PR #5, opened 2016-03-13
Create loadbalancers for all namespaces

munnerz:
This allows the ingress controller to create nginx resources for all namespaces, not just default. At the moment it polls the apiserver for a list of namespaces every `resyncPeriod` (default set to 30s).

I haven't implemented this same change for the plus controller, but will do if it's necessary to get this merged in :)

### DIFF
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/net/context"
+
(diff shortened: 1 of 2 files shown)


jdirausquin:
This commit is going to be merged into master?

pleshakov:
I took a different approach here to add support for different namespaces in https://github.com/nginxinc/kubernetes-ingress/pull/13 . This is consistent with glbc and other nginx implementations

PR #14, opened 2016-04-11
Handle missing Path in HTTPIngressPath

oyvindio:
I had an Ingress with no Path set, which caused the ingress controller to go into a crash/restart loop because the relevant nginx config file ended up invalid (`location` without a parameter). The documentation says that we should send all traffic to the backend when Path is empty, so use "/" as the default.

I also added some logging to more easily debug issues like this, which can be enabled with `-V=2`.

### DIFF
--- a/nginx-controller/Makefile
+++ b/nginx-controller/Makefile
@@ -6,7 +6,10 @@ PREFIX = nginxdemos/nginx-ingress
 nginx-ingress:
 	CGO_ENABLED=0 GOOS=linux godep go build -a -installsuffix cgo -ldflags '-w' -o nginx-ingress *.go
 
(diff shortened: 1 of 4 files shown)


pleshakov:
Looks good!
About logging stderr from the failed commands, could you put the code outside of the  ```if glog.V(2)``` statements? For now, we don't have any granular control on logging level in other places, and by default log everything. Just to be consistent

oyvindio:
> About logging stderr from the failed commands, could you put the code outside of the if glog.V(2) statements? For now, we don't have any granular control on logging level in other places, and by default log everything. Just to be consistent

Sure, I can do that.

PR #15, opened 2016-04-13
Updated NGINX version to 1.9.14

pleshakov:


munnerz:
May I ask what the reasoning behind not using the official nginx docker
image is? Seems like the easiest way to ensure we have a working nginx
installation, and reduces the complexity of this image, unless I'm missing
something.

pleshakov:
@munnerz 

We can do it, PRs are welcomed. 

When started developing the controller, I encountered an issue when using the official image -- controller hanged when starting NGINX. A quick short term solution was to not use the official image. Now it's good time to fix it

PR #16, opened 2016-04-14
adds headers for X-Real-IP and X-Forwarded-For…

goblain:
…to correctly pass orig in IP address to backend svc


pleshakov:
Could you also add setting the following headers? :
```
proxy_set_header X­-Forwarded­-Host $host
proxy_set_header X­-Forwarded­-Port $server_port
```

goblain:
added also proto so the schema can be read on the backend

PR #23, opened 2016-05-02
access_log to stdout and error_log to stderr

oyvindio:
To easily get query logs into the [kubernetes fluentd-elasticsearch setup](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/fluentd-elasticsearch), log access_log to stdout and error_log to stderr.

* Move nginx.conf from the base image into the source tree so that we can modify it.
* When we start nginx, attach the controller's stdout and stderr to the nginx process. For this to work we have to run nginx in the foreground (`daemon off`). Since we're running nginx in the foreground, use a goroutine to avoid blocking the controller process.
* Set access_log to /dev/stdout and error_log to stderr.

I don't know if you want to merge this as it is a significant change to how logging happens, but we're  using this and I think it's useful, so I thought I'd share.

Since we can now edit nginx.conf, I've also added `$host`, so we can more easily tell which backend we're hitting when using host based Ingress, and `$request_time`, since that is generally useful, to the main access_log format.

### DIFF
--- a/nginx-controller/Dockerfile
+++ b/nginx-controller/Dockerfile
@@ -18,10 +18,4 @@ COPY nginx/nginx.conf /etc/nginx/nginx.conf
 
 RUN rm /etc/nginx/conf.d/*
 
-# This works only if the nginx master process is started from the process which is the Docker entrypoint, and has its
-# stdout and stderr set to the stdout and stderr of its parent (entrypoint). nginx must also run with `daemon off`,
-# otherwise it will detach its stdout.
-RUN ln -sf /dev/stdout /var/log/nginx/access.log
-RUN ln -sf /dev/stderr /var/log/nginx/error.log
-
 CMD ["/nginx-ingress"]
--- a/nginx-controller/Dockerfile
+++ b/nginx-controller/Dockerfile
@@ -14,6 +14,7 @@ EXPOSE 80 443
 
 COPY nginx-ingress /
 COPY nginx/ingress.tmpl /
+COPY nginx/nginx.conf /etc/nginx/nginx.conf
 
 RUN rm /etc/nginx/conf.d/*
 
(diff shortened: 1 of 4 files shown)


pleshakov:
Hi, 

Thanks for submitting a pull request. I think it's a good idea to make redirecting nginx logs to stdout and stderr by default. 

Another way to do it is to insert the following lines in the Dockerfile:
```Dockerfile
RUN ln -sf /proc/1/fd/1 /var/log/nginx/access.log \
```
What we do here is redirecting logs to the stdout and stderr of the process with PID 1, which is the nginx-ingress process. Its stdout and stderr are seen in docker logs.  Changes to nginx conf files and ingress controller code are not required in this case.

oyvindio:
> RUN ln -sf /proc/1/fd/1 /var/log/nginx/access.log \
>    && ln -sf /proc/1/fd/2 /var/log/nginx/error.log

That's a nice solution! 👍  

Do you want to merge the access_log format changes, or would you rather keep the defaults? We'll still need nginx.conf in the source tree to do anything with log format I guess

pleshakov:
Do you already make or need any other customizations of NGINX configuration?

We want to add support for customization of NGINX configuration via Kubernetes resources--ConfigMap and Annotations in an Ingress Resource. So that the user can be able to set  `proxy_read_timeout`, `proxy_connect_timeout`, etc. Log format is a good candidate for adding the ability to customize.  Will it work for you if you configure log format via [ConfigMap](http://kubernetes.io/docs/user-guide/configmap/)?

oyvindio:
We don't have any other customizations right now, but I think we likely will need to do some sort of customization at some point.

ConfigMap would probably be suitable, yeah. That's what the contrib/kubernetes ingress controller is supporting also: https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx#custom-nginx-configuration

Issue #34, opened 2016-06-20
could not build server_names_hash, you should increase server_names_hash_bucket_size: 64


rawlingsj:
I'm deploying nginxdemos/nginx-ingress:0.3 as an RC but the pod goes into ERROR status. 

 From a bit of googling a suggested workaround is to edit the [http{} nginx configuration](http://nginx.org/en/docs/http/server_names.html)..

```
http {
    server_names_hash_bucket_size 64;
    ...
}
```

but I can't figure out if I can do that.  Even if the config isn't able to be [customised yet](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/customization#customization-of-nginx-configuration) yet I still don't know how I can build my own image adding in the http config. Any pointers?

Pod logs...

```
kubectl logs ingress-nginx-bq7ym
2016/06/20 11:00:31 [notice] 18#18: signal process started
2016/06/20 11:00:31 [notice] 23#23: signal process started
2016/06/20 11:00:31 [notice] 25#25: signal process started
2016/06/20 11:00:31 [notice] 27#27: signal process started
2016/06/20 11:00:31 [notice] 29#29: signal process started
2016/06/20 11:00:31 [notice] 31#31: signal process started
2016/06/20 11:00:31 [notice] 33#33: signal process started
2016/06/20 11:00:31 [emerg] 35#35: could not build server_names_hash, you should increase server_names_hash_bucket_size: 64
E0620 11:00:31.397974       1 nginx.go:207] Command nginx -s reload stdout: ""
E0620 11:00:31.398051       1 nginx.go:208] Command nginx -s reload stderr: "nginx: [emerg] could not build server_names_hash, you should increase server_names_hash_bucket_size: 64\n"
F0620 11:00:31.398064       1 nginx.go:209] Command nginx -s reload finished with error: exit status 1
```

pleshakov:
Thanks for the feedback. Yes, we don't have a mechanism to customize this parameter yet. I'll add it to our list of things.

To customize the server_names_hash_bucket_size parameter, you need to change the default nginx.conf file and add this file to the image, replacing the /etc/nginx/nginx.conf.  I'll add this file to the repo and include adding it in the Dockerfile, so that people can easily customize if they need it.

rawlingsj:
Ok great - I'll give that a go, many thanks @pleshakov

pleshakov:
[...]

PR #40, opened 2016-07-25
set a default server_names_hash_bucket_size so that controller logic …

rawlingsj:
…applies ConfigMap overrides fixes nginxinc/kubernetes-ingress#34


rawlingsj:
yeah the configmap works great - thanks!

PR #42, opened 2016-07-28
Avoid reloading if configtest fails

tfheen:
If the generated configuration is invalid, it's better to not reload
than to end up not serving traffic.

A trivial way to generate an invalid configuration is having an
ingress rule like:

```
spec:
  rules:
    - http:
        paths:
          - path: '/ { /testpath'
            backend:
              serviceName: echoheaders
              servicePort: 80
```

### DIFF
--- a/nginx-controller/nginx/nginx.go
+++ b/nginx-controller/nginx/nginx.go
@@ -2,6 +2,7 @@ package nginx
 
 import (
 	"bytes"
+	"fmt"
 	"html/template"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


nginx-controller/nginx/nginx.go
```diff
@@ -175,7 +175,11 @@ func (nginx *NginxController) templateIt(config IngressNginxConfig, filename str
 // Reload reloads NGINX
 func (nginx *NginxController) Reload() {
 	if !nginx.local {
-		shellOut("nginx -s reload")
+		if shellOut("nginx -t") {
+			shellOut("nginx -s reload")
```
pleshakov: `nginx -s reload` can also finish with error code, so it's better to check also

pleshakov:
Thanks! Agree. 

A couple of suggestions:
* It's better to propagate the error to the `Configurator` (configurator.go) and then log the error there (methods `AddOrUpdateIngress` and `DeleteIngress`). In the error message we can print the ingress resource name that caused the problem. This will point the user to the cause of the problem.

tfheen:
@pleshakov, thanks for the review, I've reworked it a bit now. I don't know how the error reporting from the configurator is supposed to work, but is this more along the lines you wanted?  The output is pretty ugly still, IMO:

```
W0802 06:53:20.178338       1 configurator.go:97] Host field of ingress rule in default/test is empty
2016/08/02 06:53:20 [emerg] 20#20: unexpected "{" in /etc/nginx/conf.d/default-test.conf:29
E0802 06:53:20.187652       1 configurator.go:40] Error when adding or updating ingress "default-test": "Invalid nginx configuration detected, not reloading%!(EXTRA *errors.errorString=Command nginx -t stdout: \"\"\nstderr: \"nginx: [emerg] unexpected \\\"{\\\" in /etc/nginx/conf.d/default-test.conf:29\\nnginx: configuration file /etc/nginx/nginx.conf test failed\\n\"\nfinished with error: exit status 1)"
```

pleshakov:
@tfheen thx. the `Configurator` looks good. The output is informative. I made a comment in the `nginx.go` file.

nginx-controller/nginx/nginx.go
```diff
@@ -206,17 +215,17 @@ func shellOut(cmd string) {
 	command.Stdout = &stdout
 	command.Stderr = &stderr
 
-	err := command.Start()
+	err = command.Start()
 	if err != nil {
-		glog.Fatalf("Failed to execute %v, err: %v", cmd, err)
+		fmt.Errorf("Failed to execute %v, err: %v", cmd, err)
```
pleshakov: it's better to return to do `return fmt.Errorf(...)` here

tfheen:
@pleshakov, thanks for the review! Fixed up now.

PR #46, opened 2016-09-10
Fix log messages.

abourget:
Colleagues had a bad impression of this projects when the logs filled with typos !


pleshakov:
thx!

PR #60, opened 2016-10-26
Redirect proxy pass subpaths

jmastr:
Otherwise more complex examples like e.g. RabbitMQ will not work,
because they try to load e.g. 'src=js/foo.js'.

### DIFF
--- a/nginx-controller/nginx/ingress.tmpl
+++ b/nginx-controller/nginx/ingress.tmpl
@@ -24,7 +24,7 @@ server {
 	{{end}}
 
 	{{range $location := $server.Locations}}
-	location {{$location.Path}} {
+	location ~ ^{{$location.Path}}(.*)$ {
 		proxy_http_version 1.1;
 		{{if $location.Websocket}}
 		proxy_set_header Upgrade $http_upgrade;
@@ -39,6 +39,7 @@ server {
 		proxy_set_header X-Forwarded-Host $host;
 		proxy_set_header X-Forwarded-Port $server_port;
 		proxy_set_header X-Forwarded-Proto $scheme;
+		rewrite ^{{$location.Path}}/(.*) /$1 break;
 		{{if $location.SSL}}
 		proxy_pass https://{{$location.Upstream.Name}};
 		{{else}}
--- a/nginx-plus-controller/nginx/ingress.tmpl
+++ b/nginx-plus-controller/nginx/ingress.tmpl
@@ -29,7 +29,7 @@ server {
 	{{end}}
 
 	{{range $location := $server.Locations}}
-	location {{$location.Path}} {
+	location ~ ^{{$location.Path}}(.*)$ {
 		proxy_http_version 1.1;
 		{{if $location.Websocket}}
 		proxy_set_header Upgrade $http_upgrade;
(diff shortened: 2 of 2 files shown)


pleshakov:
hi @jmastr 
I'm not sure if I completely understand what you're trying to accomplish. Could you provide a detailed example?

jmastr:
Hi @pleshakov 

Thank you very much for your response. I try to get a more complex example up and running namely RabbitMQ. I put these to files into the `examples/complete-example` directory:
````

**Comment by pleshakov** [comment]
Thanks for providing the details.

When you expose rabitmq via  the `/rabitmq` path, the Ingress controller assumes that rabbitmq service will respond to requests that start with `/rabitmq`. Unfortunately, it expects requests that start with `/`.  

To fix it, path rewriting is required, which is what you did by modifying the template. Often applications return web pages with absolute links so that the rewriting of response bodies is also necessary. 

Path rewriting is not supported with Ingress. What I suggest is to expose rabitmq like this:
```yaml
kind: Ingress
metadata:
  name: cafe-ingress
spec:
  rules:
  - host: rabitmq.example.com
    http:
      paths:
      - path: /
        backend:
          serviceName: rabbitmq-svc
          servicePort: 15672
```

Another option is to configure the rabbitmq to respond to requests that start with `/rabbitmq` if it is possible.

jmastr:
Hi @pleshakov 

Thanks for your answer. In my special setup I want to do something like this:
```yaml
kind: Ingress
metadata:
  name: cafe-ingress
spec:
  rules:
  - host: ingress.example.com
    http:
      paths:
      - path: /rabbitmq
        backend:
          serviceName: rabbitmq-svc
          servicePort: 15672
      - path: /another
        backend:
          serviceName: another-svc
          servicePort: 443
      - path: /another-another
        backend:
          serviceName: another-another-svc
          servicePort: 80
```
to get rid of AWS load balancer configuration overhead. So your proposal is not an option for me unfortunately.

Does my patch break anything? I tested it with the `tea-svc` and the `coffee-svc` and they still worked. In my understanding the patch just adds a feature and is backward compatible.

pleshakov:
Unfortunately, the patch breaks things. For example, a request to `/coffee/abc` to NGINX becomes a request to `/abc`to the coffee service.

One of the other options is to have another NGINX between NGINX Ingress controller and rabbitmq that does the necessary rewriting and exposes rabitmq via `/rabitmq` path.

jmastr:
But isn't that exactly what you want, that the coffee service deals with `/abc` if the request is post to `/coffee/abc` with Ingress?

I am sorry, maybe it is very obvious, but could you please explain why one does not want that behavior and what NGINX would do with the `/abc` request, if it is not handled by the coffee service.

pleshakov:
It might look counterintuitive, but for HTTP load balancing APIs, such as Ingress, it is required that paths exposed on the load balancer (`/rabitmq` in your case) must exist on the backend services.

If the default behavior was like this -- `client --/rabitmq/something--> Load balancer --/something--> rabitmq-application`  -- a lot of applications wouldn't work.  One of the problems is applications can return absolute links, which become broken. To fix that, rewriting of the response  body is necessary.  (For example `http://<your-domain>/resource` needs to be rewritten to `http://<your-domain>/rabitmq/resource` in the response body). But it is often dependent on the application and such rewriting rules are hard to generalize to work for every case.

jmastr:
Thank you very much for your detailed response. I am closing the pull request and think about another solution for my special case.

PR #62, opened 2016-11-02
Add SSL services annotation

### LINKED ISSUE #61: proxy_pass hard-coded to HTTP
I have a setup in which I use SSL for inter-container communtication. To have the nginx-ingress-rc working I need to be able to set:
```
proxy_pass https://...
```
(not `http://`)

Since Kubernetes does not hold that information, my suggestion is: If a container opens port 443 (instead of 80 or whatever) we assume, that SSL is spoken.

For this suggestion I have already a patch. Let me know if you are interested in having it.

jmastr:
If an application in a container requires HTTPS, one needs to set the
'proxy_path https://...' instead of 'proxy_pass http://...'. With this
patch one can use the annotation:

  'nginx.org/ssl-services: "service1[,service2,...]"'

to specify, which services require HTTPS.

Fixes #61

### DIFF
--- a/examples/ssl-services/README.md
+++ b/examples/ssl-services/README.md
@@ -1,0 +1,34 @@
+# SSL Services Support
+
+To load balance an application that requires HTTPS with NGINX Ingress controllers, you need to add the **nginx.org/ssl-services** annotation to your Ingress resource definition. The annotation specifies which services are SSL services. The annotation syntax is as follows:
+```
+nginx.org/ssl-services: "service1[,service2,...]"
+```
+
+In the following example we load balance three applications, one of which requires HTTPS:
+```yaml
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: cafe-ingress
+  annotations:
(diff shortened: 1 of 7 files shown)


pleshakov:
Hi @jmastr 
Thanks for the pull request! Looks good!

I have a few small suggestions:

1. Could you rename the annotation from `nginx.org/ssl-backends` to `nginx.org/ssl-services`? This will make the naming consistent: we already have `nginx.org/websocket-services` and `nginx.com/sticky-cookie-services`. My bad suggesting the wrong name in https://github.com/nginxinc/kubernetes-ingress/issues/61 

1. As a consequence of renaming, please rename the example folder to `ssl-services`, the title of the README.md to `SSL Services Support`, the name of the service from `ssl-backend-svc` to `ssl-svc`.

1. Could you also replace `an application speaking HTTPS` with `an application that requires HTTPS` , ` one of which is speaking SSL` with `one of which requires HTTPS` and `is a service for the SSL backend application` with `is a service for an HTTPS application`

jmastr:
@pleshakov done

PR #64, opened 2016-11-07
Add rewrites annotation

### LINKED ISSUE #63: Introduce annotation for rewrite rule
See #60

Instead of hard-coding the rewrite rule one could introduce a further annotation like in #61.

jmastr:
Closes #63

### DIFF
--- a/examples/rewrites/README.md
+++ b/examples/rewrites/README.md
@@ -1,0 +1,41 @@
+# Rewrites Support
+
+To load balance an application that requires rewrites with NGINX Ingress controllers, you need to add the **nginx.org/rewrites** annotation to your Ingress resource definition. The annotation specifies which services need rewrites. The annotation syntax is as follows:
+```
+nginx.org/rewrites: "serviceName=service1 rewrite=/rewrite1/[;serviceName=service2 rewrite=/rewrite2/;...]"
+```
+
+In the following example we load balance two applications, which require rewrites:
+```yaml
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: cafe-ingress
+  annotations:
+    nginx.org/rewrites: "serviceName=tea-svc rewrite=/;serviceName=coffee-svc rewrite=/beans/"
+spec:
+  rules:
+  - host: cafe.example.com
+    http:
+      paths:
+      - path: /tea/
+        backend:
+          serviceName: tea-svc
+          servicePort: 80
+      - path: /coffee/
+        backend:
+          serviceName: coffee-svc
+          servicePort: 80
+```
+
+Requests to the tea service are rewritten as follows:
+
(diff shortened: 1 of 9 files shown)


jmastr:
@pleshakov please review

pleshakov:
@jmastr 
Thx for another pull request!

A few suggestions:

- Remove `{{if $location.Rewrite}}/{{end}}` from `location {{$location.Path}}{{if $location.Rewrite}}/{{end}}` in the `ingress.tmpl`.  I think it's better to stay transperent. I'll add notes about the necessity to put `/` at the end of Ingress paths it to the README file.

- Don't validate the service name and the URL in the annotation.
```go
r, _ := regexp.Compile("serviceName=([a-z\\-]+) +rewrite=([/a-z0-9]+)")
```
A service name can also contain numbers.  A URL can contain characters like `_`. I suggest to not perform validation at all. If a service name is not valid, the controller will simply igoner the service later. If a URL is not correct, NGINX will report error in the logs when the controller does the configuration check before reloading the configuration.

Please look at https://github.com/nginxinc/kubernetes-ingress/blob/master/nginx-plus-controller/nginx/configurator.go#L232 , where we're parsing a similar annotation.

- Add logging. Please add error reporting when the annotation doesn't include all the required parts, similar to https://github.com/nginxinc/kubernetes-ingress/blob/master/nginx-plus-controller/nginx/configurator.go#L232

- It might be helpful to write a unit test similar to this one https://github.com/nginxinc/kubernetes-ingress/blob/master/nginx-plus-controller/nginx/configurator_test.go

jmastr:
Very nice! Will do!

jmastr:
Hi @pleshakov,

I worked in your comments about parsing, logging and testing.

Note that I removed the `/` at the end of `location {{$location.Path}}` from my patch, but I added it to `examples/rewrites/README.md` already to be consistent.

pleshakov:
@jmastr 
Awesome!

Could you replace `nginx.com` with `nginx.org`  in the  following line in `configurator.go` : `glog.Errorf("In %v nginx.com/rewrites contains invalid declaration: %v, ignoring", ingEx.Ingress.Name, err)`  ?

jmastr:
@pleshakov done

pleshakov:
Great. Thx

pleshakov:
We'll prepare a new release -- https://github.com/nginxinc/kubernetes-ingress/releases --  that will have this and other features. A new tag will be published to the https://hub.docker.com/r/nginxdemos/nginx-ingress/ repo.

jmastr:
That's awesome! Thank you very, very much!

jmastr:
I am going to give a talk about this at the next Kubernetes meetup in Berlin, Germany.

pleshakov:
That's awesome!!

Issue #11, opened 2016-03-29
Lighter Docker image


cescoferraro:
Alpine linux supports nginx, got a great package manager and a full container with nginx weights 36MB instead of 255MB just from the ubuntu base image we are using here.
Something like (not tested btw)

```
FROM gliderlabs/alpine:3.3
# Install nginx
RUN apk add --update nginx=1.8.0-r1 && \
    rm -rf /var/cache/apk/* && \
    chown -R nginx:www-data /var/lib/nginx

# Add the files
ADD root /

# Expose the ports for nginx
EXPOSE 80 443

COPY nginx-ingress /
COPY nginx/ingress.tmpl /

RUN rm /etc/nginx/conf.d/*

CMD ["/nginx-ingress"]

```

munnerz:
I'd be inclined to suggest we base off the official nginx image, and defer
these kind of efficiency improvements to the upstream repo. This way we can
fix nginx version in the image.

cescoferraro:
Just realized there is an official alpine base nginx image on dockerhub. We could use that

naphthalene:
There is no reason both an :alpine and a :latest set of images can't be maintained, aside from some additional code. There are certainly benefits and drawbacks to using either.

PR #65, opened 2016-11-11
Add HTTP2 as option in configmap and annotations

TheTechNick:
Hi Nginx Team,

I have added default http2 support to the ingress template.
Is there a reason, that the nginx ingress controller does not enable http2 for ssl connections by default?
The only obstacle I had, was that the environment in the debian container (openssl 1.0.1f) does not support http2. That is why I changed the base container to alpine. (wich should also solve #11)

The changed image is uploaded to: (Tag: 0.5.0)
https://quay.io/repository/nico_schieder/nginxinc-kubernetes-ingress

Please keep in mind that I could not test the nginx plus config.

### DIFF
--- a/examples/customization/nginx-config.yaml
+++ b/examples/customization/nginx-config.yaml
@@ -8,3 +8,4 @@ data:
   client-max-body-size: "2m" # default is "1m". See http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size
   server-names-hash-bucket-size: "64" # default value depends on the size of the processor’s cache line. See http://nginx.org/en/docs/http/ngx_http_core_module.html#server_names_hash_bucket_size
   server-names-hash-max-size: "1024" # default is "512". See http://nginx.org/en/docs/http/ngx_http_core_module.html#server_names_hash_max_size
+  http2: 'True' # default is "False". Enables http2 in servers with SSL enabled. See https://nginx.org/en/docs/http/ngx_http_v2_module.html
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -320,6 +321,13 @@ func (lbc *LoadBalancerController) syncCfgm(key string) {
 		}
 		if serverNamesHashMaxSize, exists := cfgm.Data["server-names-hash-max-size"]; exists {
 			cfg.MainServerNamesHashMaxSize = serverNamesHashMaxSize
+		}
+		if HTTP2Str, exists := cfgm.Data["http2"]; exists {
+			if HTTP2, err := strconv.ParseBool(HTTP2Str); err == nil {
+				cfg.HTTP2 = HTTP2
+			} else {
+				glog.Errorf("In configmap %v/%v 'http2' contains invalid declaration: %v, ignoring", cfgm.Namespace, cfgm.Name, err)
+			}
 		}
 	}
 	lbc.cnf.UpdateConfig(cfg)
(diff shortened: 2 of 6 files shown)


pleshakov:
I don't recommend enabling http/2 by default. It is a new protocol and some browsers still have problems fully implementing it, which leads to problems in some cases.

What we can do is to introduce an option to enable http/2. We can put such an option into [ConfigMap and annotation](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/customization).

Regarding the base image, we can switch from debian to ubuntu:16.04, which has openssl with support for ALPN required for [Chrome users](https://www.nginx.com/blog/supporting-http2-google-chrome-users/). 
I don't know how many people are using alpine linux so that we can switch to it by default.

We can add support for http/2. Let me know if you're interested in helping.

TheTechNick:
Hi,
sure I am interested in helping!
I will make http2 configurable in the configmap and individually on each ingress object via annotations.

Does this look good to you, in terms of annotation naming?
```yaml
annotations:
  nginx.org/enable-http2: 'True'

# configmap
enable-http2: 'True'
```

pleshakov:
@TheTechNick thx for the pull request!

The naming of ` nginx.org/http2: 'True'` looks good to me.

> About alpine linux vs. debian/ubuntu:

The arguments you provided for alpine make a good case for alpine. However, I would stick to debian as a default and as an option provide the alpine based image, similar to what we're providing on DockerHub https://hub.docker.com/_/nginx/

> We also have some more pull requests incoming for:
ipv6 binding
log_format
forward secrecy
auth_request
proxy_set_headers Do you prefer small pull requests per option/feature or do you prefer the pull requests to be larger?

That's awesome! Pull requests per option are preferable. If you have an idea how to implement a particular feature, I suggest we discuss it via a GitHub issue

Regarding this pull request:

- Since you're parsing the value of the annotation -- https://github.com/TheTechNick/kubernetes-ingress/blob/914fa992db02e5f353331a14bc287f225a5d1da7/nginx-controller/controller/controller.go#L326 -- it makes sense to report an error in case of a parsing error, similar to https://github.com/nginxinc/kubernetes-ingress/blob/master/nginx-controller/nginx/configurator.go#L198
- Another place where a server is created is here -- https://github.com/TheTechNick/kubernetes-ingress/blob/914fa992db02e5f353331a14bc287f225a5d1da7/nginx-controller/nginx/configurator.go#L147 Http2 must be added
- Let's stick to the debian base image, we will add an option to create an alpine base image. 
- Please remove http2 from the nginx-plus ingress.tmpl.

TheTechNick:
> The arguments you provided for alpine make a good case for alpine. However, I would stick to debian as a default and as an option provide the alpine based image, similar to what we're providing on DockerHub https://hub.docker.com/_/nginx/

Sounds great!

> That's awesome! Pull requests per option are preferable. If you have an idea how to implement a particular feature, I suggest we discuss it via a GitHub issue

Watch out for some issues :)
Do I somehow interrupt your plans if I add unittests for some of the config mapping functions?
I would like to be sure new config options do work without having to test them on a  kubernetes cluster.

pleshakov:
Thx! Looks good to me!
Before I merge, could you squash all commits into a single commit?

> Do I somehow interrupt your plans if I add unittests for some of the config mapping functions?
I would like to be sure new config options do work without having to test them on a kubernetes cluster.

That would be great if you add unittests for that!

TheTechNick:
I have squashed the commits into one and I have also added an entry to examples/customization/nginx-config.yaml, so the configmap entry is documented. Should I also create a examples/http2 folder with the annotations like ssl-services?

pleshakov:
Thx! No need for the examples/http2 folder. 

We'll make a new release (0.5.1) this Friday that will include this feature. We'll also create a new tag on our DockerHub to support the alpine distro.

pleshakov:
We'll make a new release next week.

PR #66, opened 2016-11-15
Support for custom log-format

df-mreidel:
I noticed that the log-format could not be customized and created a small patch that makes exactly this possible so everyone can add fields to his or her liking or switch to a totally different format. Use like this to switch to JSON-logging for example:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: nginx-ingress-conf
  namespace: kube-system
data:
  log-format: '{ "@timestamp": "$time_iso8601", "@version": "1", "clientip": "$remote_addr", "tag": "ingress", "remote_user": "$remote_user", "bytes": $bytes_sent, "duration": $request_time, "status": $status, "request": "$request_uri", "urlpath": "$uri", "urlquery": "$args", "method": "$request_method", "referer": "$http_referer", "useragent": "$http_user_agent", "software": "nginx", "version": "$nginx_version", "host": "$host", "upstream": "$upstream_addr", "upstream-status": "$upstream_status" }'
```
I also noticed that the tempalte engine was `html/template`,  which caused all text inside single quotes to be escaped like it was an HTML attribute. I therefore switched to `text/template`, I think this was meant to be used anyway.

### DIFF
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -329,6 +329,9 @@ func (lbc *LoadBalancerController) syncCfgm(key string) {
 				glog.Errorf("In configmap %v/%v 'http2' contains invalid declaration: %v, ignoring", cfgm.Namespace, cfgm.Name, err)
 			}
 		}
+		if logFormat, exists := cfgm.Data["log-format"]; exists {
+			cfg.MainLogFormat = logFormat
+		}
 	}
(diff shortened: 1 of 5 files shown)


pleshakov:
@df-mreidel 
Thx! That's very useful. 
`html/template` was used by mistake. Thanks for fixing it.

PR #68, opened 2016-11-15
Add support for configuring pxoxy-buffer-settings

df-mreidel:
A small patch to enable configuring `proxy_buffers` and `proxy_buffer_size` which is necessary especially when your backends send a lot of headers and you want to increase the default value.

### DIFF
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -332,6 +332,22 @@ func (lbc *LoadBalancerController) syncCfgm(key string) {
 		if logFormat, exists := cfgm.Data["log-format"]; exists {
 			cfg.MainLogFormat = logFormat
 		}
+		if proxyBufferingStr, exists := cfgm.Data["proxy-buffering"]; exists {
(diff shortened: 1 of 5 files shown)


pleshakov:
@df-mreidel 
Thanks for the pull request!

I have few suggestions:

- Since buffering requirements can differ from an application to an application, it's better to make them configurable per Ingress resource, similar to what we have for the `proxy_read_timeout` directive.
- There are a couple of others important buffering-releated NGINX directives:
  - [http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffering](proxy_buffering), which turns buffering on/off.
  - [http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_max_temp_file_size](proxy_max_temp_file_size), which can disable buffering responses to temp files.

df-mreidel:
I can make these settings configurable per resource, no problem. I'll take care of the other 2 settings as well as soon as I'm back to work. Thanks for looking over it!

df-mreidel:
Okay, made it location-aware and added your suggestions. Already tested and working.

pleshakov:
@df-mreidel 
Awesome! thx!

PR #71, opened 2016-11-16
Added unittests for generateNginxCfg

thetechnick:
I have added a (quite big) unittest for the generateNginxCfg method.
My goal was to test the annotation mapping without having to deploy the ingress controller on a kubernetes cluster.
The assertion library gets the job done, but I am not attached to it, if there is a better lib out there I have no problem to change it.

TODO before merge:
- "github.com/stretchr/testify/assert" needs to be added to Godeps (I was not able to do this, somehow...)

### DIFF
--- a/nginx-controller/nginx/configurator_test.go
+++ b/nginx-controller/nginx/configurator_test.go
@@ -2,6 +2,12 @@ package nginx
 
 import (
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/apis/extensions"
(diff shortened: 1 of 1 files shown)


pleshakov:
@TheTechNick 
Thx for important work!

I have no problems with using the `github.com/stretchr/testify/assert` library. 

I see that you're setting several annotations in `testIngressMetadata`, however, in the test you only check a few of them. Are you planning to cover the rest of them?

I won't merge it until we update the existing dependencies, which we'll do soon. We can add the assert library after that.

TheTechNick:
@pleshakov 
You are welcome :)
I would like to cover all  annotations, just ping me after the dependencies are updated and I will revisit the unittests.

thetechnick:

PR #74, opened 2016-11-18
Fixed the issue of ingress.class annotation handling

tangle329:


pleshakov:
@tangle329 
Thanks!

The bug appears if we do the following:
1) Create an Ingress resource that is not supposed to be handled by NGINX. For example, with the annotation `kubernetes.io/ingress.class: "gce"`
2) Change the Configmap with the NGINX configuration or Change one of the services referenced in the Ingress
3) Ingress resources gets handled by NGINX --> Bug

The bug appears if the endpoints of one of the services change, for example, if we scale the corresponding replication controller. Could you fix that too?

pleshakov:
@tangle329 thx!

Issue #52, opened 2016-09-28
default route is wrong


jacek-marchwicki:
``` yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: cafe-ingress
spec:
  tls:
  - hosts:
    - cafe.minikube
    - tea.minikube
    - coffee.minikube
    secretName: cafe-secret
  rules:
  - host: tea.minikube
    http:
      paths:
      - path: /
        backend:
          serviceName: tea-svc
          servicePort: 80
  - host: coffee.minikube
    http:
      paths:
      - path: /
        backend:
          serviceName: coffee-svc
          servicePort: 80
  - host: cafe.minikube
    http:
      paths:
      - path: /tea
        backend:
          serviceName: tea-svc
          servicePort: 80
      - path: /coffee
        backend:
          serviceName: coffee-svc
          servicePort: 80
```

in this example:
- `https://cofee.minikube/` gives `coffee-rc-XYZ` (correct)
- `https://cafe.minikube/coffee` gives `coffee-rc-XYZ` (correct)
- `https://cafe.minikube/tea` gives `tea-rc-XYZ` (correct)
- `https://wrong.minikube/` gives `coffee-rc-XYZ` (WRONG) it should return 404

pleshakov:
Currently, the controller doesn't include the default server, which catches all HTTP/HTTPS requests with a host header that doesn't match any of the Ingress hosts. Thx for pointing that out.  We'll add it.

pleshakov:
@jacek-marchwicki 
I can suggest the following solution. We define an Ingress resource that exposes a service that handles all requests with a value of the host header that doesn't match any of the existing Ingress hosts. We also setup SSL termination to handle HTTPS requests.
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: default-ingress
annotations:
  nginx.org/default-server: "true"
spec:
  tls:
    - secretName: default-secret
  backend:
    serviceName: default-service
    servicePort: 80
```
Note that we don't define any Ingress rules, just a default backend without a hostname. The ` nginx.org/default-server` annotation tells NGINX to create a default server.

This way you can create a custom service that returns 404s or other things if you want.

Let me know you thoughts.

jacek-marchwicki:
@pleshakov pretty hacky but if it works it's good quick fix ;) 

Thank you!

pleshakov:
@jacek-marchwicki 
[...]

PR #79, opened 2016-11-21
Fix default server

thetechnick:
If I create a default server using a ingress config like this:
```yaml
kind: Ingress
metadata:
  name: default-backend
  namespace: default
spec:
  backend:
    serviceName: default-backend
    servicePort: 80
```
the configuration is created, but because of a missing default_server option in the servers listen directive it is not actually been used as the default backend.

### DIFF
--- a/nginx-controller/nginx/ingress.tmpl
+++ b/nginx-controller/nginx/ingress.tmpl
@@ -6,9 +6,9 @@ upstream {{$upstream.Name}} {
 
 {{range $server := .Servers}}
 server {
-	listen 80;
+	listen 80{{if not $server.Name}} default_server{{end}};
 	{{if $server.SSL}}
-	listen 443 ssl{{if $server.HTTP2}} http2{{end}};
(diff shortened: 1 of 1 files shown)


pleshakov:
@thetechnick 
Thx! 

In this case only one default_server can be created, which is enforced by NGINX. This also entails that all the Ingress rules with empty hostnames will be rejected: NGINX will fail to reload, you'll get something like: 
```
nginx: [emerg] a duplicate default server for 0.0.0.0:80 in /etc/nginx/conf.d/default_server.conf:19
nginx: configuration file /etc/nginx/nginx.conf test failed
```

It also solves this https://github.com/nginxinc/kubernetes-ingress/issues/52

However, since NGINX configuration stays invalid until we remove all Ingress rules with the empty hostnames until there is only one, I think we should not merge it until we have a handling of cases when Ingress rules have the same hostname in the controller.

thetechnick:

Issue #73, opened 2016-11-17
Support proxy protocol


thetechnick:
If the nginx ingress controller is scheduled using mechanisms like nodePort, the client ip address is not preserved, because the kube-proxy is setting up NAT to connect the backend pods to the host network.

The client ip address is needed in several use cases:
- Enforcing rate limits
- Detecting the country of the user
- ...

Until https://github.com/kubernetes/features/issues/27 is completed, there are only a few workarounds to preserve the client ip address.

1. Directly binding to the host network
  Using `hostNetwork: true` in combination with a daemon-set nginx can bind directly to the host network, so there is no NAT.
  In this deployment scenario it is necessary to deploy a external TCP loadbalancer with a alive checking mechanism to enable high availability, if a host goes down unexpected.

2. Enable `proxy protocol` so a external TCP LB can forward the requests without loosing the client ip.
  A major advantage of this approach is, that the nginx ingress deployment can be scaled independently of the host systems.

To enable the proxy protocol feature i would suggest to add a new entry in the nginx configmap:

```yaml
data:
  proxy-protocol: 'True'
```

This entry will reconfigure the generated nginx configs like described in the official blog post:
https://www.nginx.com/resources/admin-guide/proxy-protocol/

I am not sure nginx is able to mix server entries with and without proxy-protocol in the listen directive. But because this is more of a global deployment choice, I do not see the need to support to annotations for the ingress object.

pleshakov:
@TheTechNick 

NGINX must know the address or CIDR of the proxy for the http://nginx.org/en/docs/http/ngx_http_realip_module.html#set_real_ip_from directive. We can configure it via a separate config map entry:
```yaml
data:
  proxy-protocol: 'True'
  proxy-protocol-from: '192.168.0.1/24'
```

In the NGINX configuration file, we must do the following:
```nginx
listen 80 proxy_protocol;
listen 443 ssl proxy_protocol;
...
set_real_ip_from 192.168.0.1/24;
real_ip_header   proxy_protocol;
...
```
This way, the client IP address is accessible via `$remote_addr` variable.

> I am not sure nginx is able to mix server entries with and without proxy-protocol in the listen directive. But because this is more of a global deployment choice, I do not see the need to support to annotations for the ingress object.

Agree, makes sense to make it global.

thetechnick:
@pleshakov 
I think we should stick to the nginx directive names for the configmap option names:
```yaml
data:
  proxy-protocol: 'True'
[...]

PR #81, opened 2016-11-24
Added configmap option to turn on proxy protocol and set_real_ip_from

thetechnick:
As discussed in #73

### DIFF
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -364,6 +364,33 @@ func (lbc *LoadBalancerController) syncCfgm(key string) {
 			}
 		}
 
+		if proxyProtocol, exists, err := nginx.GetMapKeyAsBool(cfgm.Data, "proxy-protocol", cfgm); exists {
+			if err != nil {
+				glog.Error(err)
+			} else {
+				cfg.ProxyProtocol = proxyProtocol
+			}
+		}
+
+		// ngx_http_realip_module
+		if realIPHeader, exists := cfgm.Data["real-ip-header"]; exists {
+			cfg.RealIPHeader = realIPHeader
+		}
+		if setRealIPFrom, exists, err := nginx.GetMapKeyAsStringSlice(cfgm.Data, "set-real-ip-from", cfgm); exists {
+			if err != nil {
+				glog.Error(err)
+			} else {
+				cfg.SetRealIPFrom = setRealIPFrom
+			}
+		}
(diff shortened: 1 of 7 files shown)


pleshakov:
@thetechnick thx for another pull request!

Couple of suggestions:

1. The reason I suggested to use the `proxy-protocol` and `proxy-protocol-from` keys is that setting those keys results in setting several (`set_real_ip_from`, `real_ip_header`) NGINX directives and changing the `listen` directive. So that using fewer keys makes configuration a little easier.
If you'd like to have the `set-real-ip-from` key, I suggest to also add  the `real-ip-header` key. This way to enable proxy protocol the user has to set `proxy-protocol: "True"` and if the users wants to see the IP from the proxy protocol in the logs as well as in the headers to backends, he will need to set the `set-real-ip-from` and `real_ip_header` keys.
(Also, it makes sense to add the 	real-ip-recursive key as it is the last remaining directive of the real ip module http://nginx.org/en/docs/http/ngx_http_realip_module.html)

2. If the the `real_ip_header` is set to `proxy_protocol` and the `set_real_ip_from` is set to the correct address, the `$remote_addr` variable will contain the IP obtained via proxy protocol. Thus, it's not necessary  to use `$proxy_protocol_addr` variable in the `proxy_set_header` directives.

3. In the `TestGetMapKeyAsStringSlice` test, I suggest to have:
```go
expected := []string{"1.String","2.String","3.String"}
fmt.Println(expected)
if !reflect.DeepEqual(expected, slice) {
	t.Errorf("Unexpected return value:\nGot: %v\nExpected: %v", slice, expected)
}
```

4. I created a pool request https://github.com/nginxinc/kubernetes-ingress/pull/82 that changes GetMapKeyAs* functions and fixes a few bugs in parsing when values were set to the config regardless of the presence of keys or annotations.

thetechnick:
@pleshakov 
Sounds good

pleshakov:
@thetechnick thx. Looks good! I added few comments about the unit test.

thetechnick:
@pleshakov 
Where did you add your comments?
I have not found them, or am I too impatient :)

pleshakov:
@thetechnick
> Where did you add your comments?
I have not found them, or am I too impatient :) 

My apologies. I forget to publish the comments

thetechnick:
@pleshakov 
Ok the unittest was quite broken for this one :)
fixed in latest commit

pleshakov:
@thetechnick Great! Could you squash the two commits?

thetechnick:
@pleshakov done

pleshakov:
@thetechnick thx!

PR #82, opened 2016-11-24
Fix configmaps and annotation parsing

pleshakov:
Fixes parsing of annotations and configmaps.


pleshakov:
@thetechnick 
I added a third return value to the GetMapKeyAs* functions. This value shows if the key exists. I think it's better to treat a missing key as not an error.  
There were few related bugs. For example, in Configurator:
```go
HTTP2, err := GetMapKeyAsBool(ingEx.Ingress.Annotations, "nginx.org/http2", ingEx.Ingress)
if err != nil && err != ErrorKeyNotFound {
	glog.Error(err)
} else {
	ingCfg.HTTP2 = HTTP2
}
```
the line `ingCfg.HTTP2 = HTTP2` was executed even when the annotation wasn't present in the Ingress resource. This led to overriding values set via ConfigMap.

thetechnick:
@pleshakov 
I like the additonal return value.
Sorry for the error. I will test my next PRs more thoroughly.

PR #84, opened 2016-11-26
Secret retry

### LINKED ISSUE #78: Watch for secret after ingress creation
The nginx ingress controller will ignore the TLS configuration of ingress objects, if the specified secret does not yet exist. If the secret is created afterwards, the nginx ingress controller does not update the rendered ingress configuration and the ingress will still be served without TLS.

I would expect the nginx ingress controller to wait for the secret to be created, or at least to update the generated configuration.
This feature, aside from #76, would be needed to support https://github.com/jetstack/kube-lego

thetechnick:
If a certificate secret referenced in an ingress object is not found,
the ingress controller will reject the ingress object. but retries every 5s.
This fixes #78.

I have tested this with success on my private cluster merged together with #77 and #81 and kube-lego.

This is a quite simple approach to the issue, next steps to improve this feature are:
- add a error event to the rejected ingress object
- only reject ingress rules referencing a host, that should be secured by the secret, when it can't be found

I think we can cover point 1 when we implement the event logging system also improving #77,
point 2 is more difficult, but I think most users will use only one host per ingress object, so this is not an issue in most cases.

### DIFF
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -285,7 +285,11 @@ func (lbc *LoadBalancerController) syncEndp(key string) {
 		ings := lbc.getIngressForEndpoints(obj)
 
 		for _, ing := range ings {
-			ingEx := lbc.createIngress(&ing)
+			ingEx, err := lbc.createIngress(&ing)
+			if err != nil {
+				lbc.ingQueue.requeueAfter(key, err, 5*time.Second)
+				return
+			}
 			glog.V(3).Infof("Updating Endpoints for %v/%v", ing.Name, ing.Namespace)
 			name := ing.Namespace + "-" + ing.Name
 			lbc.cnf.UpdateEndpoints(name, &ingEx)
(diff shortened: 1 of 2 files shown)


pleshakov:
@thetechnick thx! I've added few comments.

It doesn't fix #78 though, but it's an important step towards it.

> add a error event to the rejected ingress object

I'll create a separate issue for the task of adding error events.

> only reject ingress rules referencing a host, that should be secured by the secret, when it can't be found... point 2 is more difficult, but I think most users will use only one host per ingress object, so this is not an issue in most cases.

 Not sure if it's necessary, since if there is a missing secret, its  gets caught by this PR.

thetechnick:
@pleshakov 
What do you think is still needed for #78?
At least for my use case (kube-lego) it works as expected, if the secrets are manually added afterwards the nginx controller will also pick up the change.

pleshakov:
@thetechnick 
> What do you think is still needed for #78?
At least for my use case (kube-lego) it works as expected, if the secrets are manually added afterwards the nginx controller will also pick up the change.

Agree, It covers #78. I misunderstood the issue a little bit. There is a separate problem of updating a certificate: if a secret changes, the controller won't update the certificate.

nginx-controller/controller/controller.go
```diff
@@ -287,10 +287,14 @@ func (lbc *LoadBalancerController) syncEndp(key string) {
 			if !isNginxIngress(&ing) {
 				continue
 			}
-			ingEx := lbc.createIngress(&ing)
+			ingEx, err := lbc.createIngress(&ing)
+			if err != nil {
+				glog.Warningf("Error updating endpoints for %v/%v: %v, skipping", ing.Namespace, ing.Name, err)
+				return
```
pleshakov: a small bug. must be `continue`  instead of `return`. This leads to Ignoring the rest of the Ingresses.

pleshakov:
@thetechnick thx! 

Please make changes related to a review in a separate commit, instead of changing the last commit. This way it's easier to review it and also get notified about it. I will squash the commits before the merging (forgot to do it for this pull request)

PR #85, opened 2016-11-29
Added configmap settings to support perfect forward secrecy

### LINKED ISSUE #69: How would I implement customer DH parameters for DHE ciphers
I'm asking upfront before creating a pull-request. We need to pass in custom DH parameters by supplying `ssl_dhparam <file>` to the nginx config. I could create a patch that simply supports setting `ssl_dhparam` and it would then require the user to make sure he mounts `<file>` into the container at the given location.

On the other hand, you might prefer that the value to `ssl-dhparam`/`nginx.org/ssl-dhparam` is not a filename, but rather the name of a namespace/secret and we look for the key dhparam.pem - very much like certificates are handled right now. So let's explain by examples.

# Variant 1: filename
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: nginx-ingress
  namespace: kube-system
data:
  ssl-dhparam: ciphers/dh4096.pem
```
And then mount that file to `/etc/nginx/ciphers/dh4096/pem` in the DaemonSet or Deployment

# Variant 2: reference to a secret
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: nginx-ingress
  namespace: kube-system
data:
  ssl-dhparam: kube-system/dh-params
```

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: dh-params
  namespace: kube-system
type: Opaque
data:
  dhparam.pem: ABC==
```

The second variant would require the controller to automagically create the file from the reference, but it would also allow for a seamless update of these parameters - for whatever reason.

Of course variant 1 is easier to implement, so I would like to know, which version you prefer?

thetechnick:
I have tested this PR on my environment using the https://www.ssllabs.com checker.
Fixes #69

### DIFF
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -383,8 +383,9 @@ func (lbc *LoadBalancerController) syncCfgm(key string) {
 			fileName, err := lbc.cnf.AddOrUpdateDHParam(sslDHParamFile)
 			if err != nil {
 				glog.Errorf("Configmap %s/%s: Could not update dhparams: %v", cfgm.GetNamespace(), cfgm.GetName(), err)
+			} else {
+				cfg.MainServerSSLDHParam = fileName
 			}
-			cfg.MainServerSSLDHParam = fileName
 		}
 
 		if logFormat, exists := cfgm.Data["log-format"]; exists {
@@ -490,7 +491,7 @@ func (lbc *LoadBalancerController) createIngress(ing *extensions.Ingress) nginx.
 			glog.Warningf("Error retrieving secret %v for Ingress %v: %v", secretName, ing.Name, err)
 			continue
 		}
-		ingEx.Secrets[fmt.Sprintf("%s-%s", ing.GetNamespace(), secretName)] = secret
+		ingEx.Secrets[secretName] = secret
 	}
 
 	ingEx.Endpoints = make(map[string][]string)
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -364,6 +364,29 @@ func (lbc *LoadBalancerController) syncCfgm(key string) {
 			}
 		}
 
+		// SSL block
+		if sslProtocols, exists := cfgm.Data["ssl-protocols"]; exists {
+			cfg.MainServerSSLProtocols = sslProtocols
+		}
+		if sslPreferServerCiphers, exists, err := nginx.GetMapKeyAsBool(cfgm.Data, "ssl-prefer-server-ciphers", cfgm); exists {
+			if err != nil {
+				glog.Error(err)
+			} else {
+				cfg.MainServerSSLPreferServerCiphers = sslPreferServerCiphers
+			}
+		}
+		if sslCiphers, exists := cfgm.Data["ssl-ciphers"]; exists {
+			cfg.MainServerSSLCiphers = strings.Trim(sslCiphers, "\n")
+		}
+		if sslDHParamFile, exists := cfgm.Data["ssl-dhparam-file"]; exists {
+			sslDHParamFile = strings.Trim(sslDHParamFile, "\n")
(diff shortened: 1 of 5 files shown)


nginx-controller/nginx/nginx.go
```diff
@@ -85,13 +93,15 @@ func NewUpstreamWithDefaultServer(name string) Upstream {
 // NewNginxController creates a NGINX controller
 func NewNginxController(nginxConfPath string, local bool) (*NginxController, error) {
 	ngxc := NginxController{
+		nginxEtcPath:   path.Join(nginxConfPath, "etc"),
```
pleshakov: Can we use the `nginxCertsPath`folder to store the pem file?

nginx-controller/nginx/nginx.go
```diff
@@ -211,9 +239,9 @@ func (nginx *NginxController) Start() {
 	}
 }
 
-func (nginx *NginxController) createCertsDir() {
-	if err := os.Mkdir(nginx.nginxCertsPath, os.ModeDir); err != nil {
-		glog.Fatalf("Couldn't create directory %v: %v", nginx.nginxCertsPath, err)
+func (nginx *NginxController) createDir(path string) {
```
pleshakov: should be a function, not a method, since it doesn't depend on NginxController anymore

nginx-controller/nginx/nginx.go
thetechnick: I was worried that someone uses `dhparam` as a secret name which would override the certificate secret. (I know this is a edge case, but I do not trust users)
But if we add the namespace to the filename of the certificates we can also use the cert folder for the `dhparam.pem`, this will also solve a not yet raised issue, if a secret with the same name is used in two namespaces.

pleshakov: > But if we add the namespace to the filename of the certificates we can also use the cert folder for the dhparam.pem, this will also solve a not yet raised issue, if a secret with the same name is used in two namespaces.

Thanks for catching that. I've created a separate issue for that -- https://github.com/nginxinc/kubernetes-ingress/issues/89

nginx-controller/controller/controller.go
```diff
@@ -467,7 +490,7 @@ func (lbc *LoadBalancerController) createIngress(ing *extensions.Ingress) nginx.
 			glog.Warningf("Error retrieving secret %v for Ingress %v: %v", secretName, ing.Name, err)
 			continue
 		}
-		ingEx.Secrets[secretName] = secret
+		ingEx.Secrets[fmt.Sprintf("%s-%s", ing.GetNamespace(), secretName)] = secret
```
pleshakov: This creates a bug. The code in the `configurator` still uses the `secretName` as a key.

Better to solve the problem https://github.com/nginxinc/kubernetes-ingress/issues/89 in a separate pull request.

nginx-controller/controller/controller.go
```diff
@@ -364,6 +364,29 @@ func (lbc *LoadBalancerController) syncCfgm(key string) {
...
+		if sslCiphers, exists := cfgm.Data["ssl-ciphers"]; exists {
+			cfg.MainServerSSLCiphers = strings.Trim(sslCiphers, "\n")
+		}
+		if sslDHParamFile, exists := cfgm.Data["ssl-dhparam-file"]; exists {
+			sslDHParamFile = strings.Trim(sslDHParamFile, "\n")
+			fileName, err := lbc.cnf.AddOrUpdateDHParam(sslDHParamFile)
+			if err != nil {
+				glog.Errorf("Configmap %s/%s: Could not update dhparams: %v", cfgm.GetNamespace(), cfgm.GetName(), err)
+			}
+			cfg.MainServerSSLDHParam = fileName
```
pleshakov: In case of an error when writing the dh param file, setting `cfg.MainServerSSLDHParam = fileName` will lead to an invalid NGINX configuration, when the dh param file doesn't exist

PR #88, opened 2016-11-30
Add proxy_hide_header directive

### LINKED ISSUE #86: HSTS header is sent multiple times
The current implementation of hsts may sent the header multiple times, if the backend application is already adding it to the http response.
This **should** not break clients as stated in #67, which is true for browsers.
But if the site is tested using https://www.ssllabs.com for example a error message is shown:
"Server sent invalid HSTS policy. See below for further information."
"Strict Transport Security (HSTS)	Invalid   Server provided more than one HSTS header "

Maybe we can find a way to add the header only if it does not exist already?

thetechnick:
Fixes #86 

```yaml
#configmap
data:
  proxy-hide-headers: X-My-Header,Strict-Transport-Security

#ingress
annotations:
  nginx.org/proxy-hide-headers: X-My-Header,Strict-Transport-Security
```

will render:
```
proxy_hide_header Strict-Transport-Security;
proxy_hide_header X-My-Header;
```

### DIFF
--- a/nginx-controller/nginx/configurator.go
+++ b/nginx-controller/nginx/configurator.go
@@ -113,6 +113,7 @@ func (cnf *Configurator) generateNginxCfg(ingEx *IngressEx, pems map[string]stri
 			SetRealIPFrom:         ingCfg.SetRealIPFrom,
 			RealIPRecursive:       ingCfg.RealIPRecursive,
 			ProxyHideHeaders:      ingCfg.ProxyHideHeaders,
+			ProxyPassHeaders:      ingCfg.ProxyPassHeaders,
 		}
 
 		if pemFile, ok := pems[serverName]; ok {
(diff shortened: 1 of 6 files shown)


nginx-controller/nginx/configurator.go
```diff
@@ -190,6 +192,17 @@ func (cnf *Configurator) createConfig(ingEx *IngressEx) Config {
...
 		ingCfg.ProxyReadTimeout = proxyReadTimeout
 	}
+	if proxyHideHeaders, exists, err := GetMapKeyAsStringSlice(ingEx.Ingress.Annotations, "nginx.org/proxy-hide-headers", ingEx.Ingress); exists {
+		if err != nil {
+			glog.Error(err)
+		} else {
+			if ingCfg.ProxyHideHeaders == nil || len(ingCfg.ProxyHideHeaders) == 0 {
+				ingCfg.ProxyHideHeaders = proxyHideHeaders
+			} else {
+				ingCfg.ProxyHideHeaders = append(ingCfg.ProxyHideHeaders, proxyHideHeaders...)
```
pleshakov: If the `nginx.org/proxy-hide-headers` annotations exists, it's better to overwrite the value provided by the configmap. Such a behavior is consistent with the behaviour of the `	proxy_hide_header` NGINX directive.

pleshakov:
thx!

PR #91, opened 2016-12-01
Fix issue when operate Single Service Ingress backend

tangle329:
This issue happend when we use ingress yaml like below, and then add, delete or update the pod in service.
kind: Ingress
metadata:
  name: test-ingress
spec:
  backend:
    serviceName: testsvc
    servicePort: 80


pleshakov:
@tangle329 thx!

PR #95, opened 2016-12-06
Fix runtime.TypeAssertionError issue when delete resources

tangle329:
This issue occurred when I deleted 200 ingresses use a shell script.
The log shows:
Recovered from panic: &runtime.TypeAssertionError{interfaceString:"interface {}", concreteString:"cache.DeletedFinalStateUnknown", assertedString:"*extensions.Ingress", missingMethod:""}

This should be the same issue as https://github.com/prometheus/prometheus/pull/2180.


pleshakov:
@tangle329 thx!

The similar issue `Recovered from panic: &runtime.TypeAssertionError{interfaceString:"interface {}", concreteString:"cache.DeletedFinalStateUnknown", assertedString:"*api.Service", missingMethod:""} (interface conversion: interface {} is cache.DeletedFinalStateUnknown, not *api.Service)` still sometimes occur when removing services: I was able to reproduce it by constantly adding/removing coffee and tea services.

PR #97, opened 2016-12-09
fixed README doc for proxy_connect_timeout

rothgar:
It's correct in the example but the README was wrong. :book:


pleshakov:
@rothgar Thanks!

PR #100, opened 2016-12-13
Add health check location

### LINKED ISSUE #90: Add something like /status or /health to NGINX config
Hi,

I have a setup, where an external LoadBalancer pointing to the NGINX Ingress Controller . The NGINX Ingress Controller serves multiple Kubernetes Namespaces. For the external LoadBalancer's health checker I need a location that is available in all NGINX config files.

`ngx_http_stub_status_module` comes to mind, which is already available in the shipped NGINX:
```
location /status {
    stub_status;
    allow all;
}
```
Something like this maybe? Is there a better way? How about a new annotation like `nginx.org/status`?

jmastr:
Ideally, this feature should be a part of a bigger feature (the default
server) that solves #52.

For now, we add the "-health-status" parameter to the controller. If
it is present, the default server listening on port 80 with the health
check location "/nginx-health" gets added to the main nginx
configuration.

Closes #90

### DIFF
--- a/nginx-controller/main.go
+++ b/nginx-controller/main.go
@@ -16,6 +16,10 @@ var (
 	// Set during build
 	version string
 
+	healthStatus = flag.Bool("health-status", false,
+		`If present, the default server listening on port 80 with the health check
+		location "/nginx-health" gets added to the main nginx configuration.`)
+
 	proxyURL = flag.String("proxy", "",
 		`If specified, the controller assumes a kubctl proxy server is running on the
 		given url and creates a proxy client. Regenerated NGINX configuration files
@@ -52,7 +56,7 @@ func main() {
 		}
 	}
 
-	ngxc, _ := nginx.NewNginxController("/etc/nginx/", local)
+	ngxc, _ := nginx.NewNginxController("/etc/nginx/", local, *healthStatus)
 	ngxc.Start()
 	config := nginx.NewDefaultConfig()
 	cnf := nginx.NewConfigurator(ngxc, config)
--- a/nginx-controller/nginx/nginx.conf.tmpl
+++ b/nginx-controller/nginx/nginx.conf.tmpl
@@ -43,5 +43,18 @@ http {
     {{if .SSLPreferServerCiphers}}ssl_prefer_server_ciphers on;{{end}}
     {{if .SSLDHParam}}ssl_dhparam {{.SSLDHParam}};{{end}}
 
+    {{if .HealthStatus}}
+    server {
+        listen 80 default_server;
+        server_name _;
+
+        location /nginx-health {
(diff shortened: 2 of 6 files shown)


pleshakov:
@jmastr Thanks!

Any reason you chose the default type as `default_type text/html;`? For the response that NGINX is sending, `text/plain` suits better.

jmastr:
@pleshakov It was for debugging only. I wanted to stop my browser downloading "healthy\n" all the time. I removed those lines.

pleshakov:
@jmastr Could you add it again as the `text/plain`?

jmastr:
@pleshakov Sure. Pushed it. But it is the default anyways, isn't it?

pleshakov:
@jmastr Great! The default is `default_type  application/octet-stream;`

jmastr:
@pleshakov I am just curious, am I looking in the wrong place: http://nginx.org/en/docs/http/ngx_http_core_module.html#default_type?

pleshakov:
@jmastr That's correct. Sorry for the confusion. However, this default is overwritten by this line https://github.com/nginxinc/kubernetes-ingress/blob/master/nginx-controller/nginx/nginx.conf.tmpl#L16, which becomes the default :)

jmastr:
@pleshakov Thanks for the explanation. Everything looks good to me now.

jmastr:
@pleshakov btw. I am just preparing my talk about Ingress and NGINX Ingress Controller for tonight's Kubernetes meetup. Told you about it: https://www.meetup.com/Berlin-Kubernetes-Meetup/events/235968196/

pleshakov:
@jmastr Awesome! Good luck! :)

pleshakov:
@jmastr How did your talk go?

jmastr:
@pleshakov It went very well. The people were pretty interested. I could have talked a lot longer about that topic. There was one guy already, who manages 100+ vhosts with the NGINX Ingress Controller. Keep up the good work!

pleshakov:
@jmastr Glad it went very well! thx!

georgecrawford:
@jmastr I know this isn't yet ready for public consumption, but I did try it. I've tried adding `-health-status` as a command param to the container, and I see that the expected block is added to /etc/nginx/nginx.conf. But at this point, I don't see my custom settings as defined in `-nginx-configmaps`. However, moments, later, I see that the main config file is re-written, _with_ my custom settings from `-nginx-configmaps`, but _without_ the `-health-status` block. It only therefore persists if I don't set `-nginx-configmaps` at all. Do you agree this is a bug? Want a new ticket?

jmastr:
@georgecrawford I did not test it with configmaps indeed, so the probability that it is a bug is quite high.

PR #108, opened 2017-01-17
disable emitting nginx version

dwradcliffe:
It's generally a best practice to disable server_tokens to prevent the nginx version from being emitted.

### DIFF
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -361,6 +361,14 @@ func (lbc *LoadBalancerController) syncCfgm(key string) {
 	if cfgmExists {
 		cfgm := obj.(*api.ConfigMap)
 
+		if serverTokens, exists, err := nginx.GetMapKeyAsBool(cfgm.Data, "server-tokens", cfgm); exists {
+			if err != nil {
+				glog.Error(err)
+			} else {
+				cfg.ServerTokens = serverTokens
+			}
+		}
+
 		if proxyConnectTimeout, exists := cfgm.Data["proxy-connect-timeout"]; exists {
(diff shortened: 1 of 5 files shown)


pleshakov:
@dwradcliffe 
I don't think it's necessary to disable server_tokens. We keep the default NGINX configuration of the Ingress controller very close to the default NGINX configuration. However, it's important to have the ability to change the defaults easily for the most common use-cases.  The server_token directive is important and we should definitely add the ConfigMaps key to configure this directive.
Let me know if you're interested in implementing that.

dwradcliffe:
Yeah I can take a look at adding that.

dwradcliffe:
@pleshakov PR updated with config option for server_tokens

dwradcliffe:
I must have missed something here... getting this error:
`nginx.go:213] Failed to write template template: ingress.tmpl:20:13: executing "ingress.tmpl" at <$server.DisableServe...>: DisableServerTokens is not a field of struct type nginx.Server`

pleshakov:
@dwradcliffe 
looks like the compiled version of the controller doesn't have `DisableServerTokens` field. Suggest to recompile it.

I tried it and it works! I only suggest to rename the key: `disable-server-tokens` -> `server-tokens`  to make it more consistent with the NGINX directive.

dwradcliffe:
`server-tokens` with a default value of true?

pleshakov:
@dwradcliffe yes

dwradcliffe:
@pleshakov ok, done. Tested on my cluster and seems to be working properly.

pleshakov:
@dwradcliffe Looks good! could you squash your commits before I merge?

dwradcliffe:
squashed!

PR #110, opened 2017-01-19
add support for proxy_ignore_headers via config

dwradcliffe:
Allow setting `proxy_ignore_header` values via config, similar to `proxy_hide_header` and `proxy_pass_header`.

### DIFF
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -367,6 +367,13 @@ func (lbc *LoadBalancerController) syncCfgm(key string) {
 		if proxyReadTimeout, exists := cfgm.Data["proxy-read-timeout"]; exists {
 			cfg.ProxyReadTimeout = proxyReadTimeout
 		}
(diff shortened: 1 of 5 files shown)


pleshakov:
@dwradcliffe thanks for the PR.

The [proxy_ignore_headers](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ignore_headers) directive works differently than the [proxy_hide_header](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_hide_header): it expects a list as an argument. The list is separated by the whitespace. Thus, we need to use only one directive. 

Additionally, the values in the list are limited to the specific headers field. If there is an invalid value, NGINX will fail to reload.  Thus, it might be a good idea to check and log an error in case there is an invalid value. 

Also, `ProxyIgnoreHeaders:    ingCfg.ProxyIgnoreHeaders,` must be added here -- https://github.com/dwradcliffe/kubernetes-ingress/blob/1f7fe8ff274d3eb6f3a54b5b59a3ce522ec2b6ae/nginx-controller/nginx/configurator.go#L172 -- too. This code handles the case when there is only the default backend present in an Ingress resource.

dwradcliffe:
Closing this since we're not using this project anymore.

PR #117, opened 2017-02-19
Add template options

rchicoli:
Allow using a custom snippet for advanced configurations in different contexts:

```yaml
    nginx.org/location-snippets: |
        if ($ssl_client_verify = SUCCESS) {
            set $auth_basic off;
        }
        if ($ssl_client_verify != SUCCESS) {
            set $auth_basic "Restricted";
        }
```

Allow setting `auth_basic` and `auth-basic-user-file` values via config.

### DIFF
--- a/examples/customization/README.md
+++ b/examples/customization/README.md
@@ -31,6 +31,9 @@ The table below summarizes some of the options. More options (extensions) are av
 | N/A | `real-ip-header` | Sets the value of the [real_ip_header](http://nginx.org/en/docs/http/ngx_http_realip_module.html#real_ip_header) directive. | `X-Real-IP`|
 | N/A | `real-ip-recursive` | Enables or disables the [real_ip_recursive](http://nginx.org/en/docs/http/ngx_http_realip_module.html#real_ip_recursive) directive. | `False`|
 | `nginx.org/server-tokens` | `server-tokens` | Enables or disables the [server_tokens](http://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens) directive. Additionally, with the NGINX Plus controller, you can specify a custom string value. The empty string value disables the emission of the “Server” field. | `True`|
+| N/A | `http-snippets` | Sets a custom snippet in http context. | N/A |
+| `nginx.org/location-snippets` | `location-snippets` | Sets a custom snippet in location context. | N/A |
+| `nginx.org/server-snippets` | `server-snippets` | Sets a custom snippet in server context. | N/A |
 
 ## Using ConfigMaps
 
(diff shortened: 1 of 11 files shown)


rchicoli:
I've tested manually this changes and everything seems to be working as expected. Let me know if I forgot something.

pleshakov:
@rchicoli 
Thanks! Now it's possible to further customize NGINX configuration without modifying the templates.

However, I'm not sure if I understand why using `map[int]string` is required for storing snippets. Why not to use the `[]string` type?

Also, it seems that having  the `auth_basic` option is redundant since snippets will allow you to easily configure it.

rchicoli:
Hi @pleshakov, sorry for the late answer, for some reasons I didn't get any notification. Oh well, actually there is no special reason of using `map[int]string`, before this implementation I was playing with `map[string]string` (`key: value`), then I realize that mapping everything would be easier. Perhaps it is a good point to replace this with `[]string`.
Glad that you like the changes, this can be very handy in some cases.
Cool, I will delete the `auth_basic` stuffs as well and just leave `location-snippets`, `server-snippets` and `http-snippets`. Please let me know if I need to change these variable names too.
I will push another MR till EOD.

pleshakov:
@rchicoli 
Awesome! The only suggestion I have is related to documentation: If you could group the snippet rows in the table in the examples/customization/README.md file, that'd be great.

rchicoli:
I've updated the code and rebased against the master. The only difference now is that I had to add an extra parameter to the `GetMapKeyAsStringSlice` function, in order to avoid duplications. I hope it is already though.
Glad to contribute to this awesome project. Keep it up the good work.

rchicoli:
Not forgetting to mention that I've tested all the changes with `minikube` and everything seems to work like a charm.

pleshakov:
@rchicoli Thanks!

alirezaDavid:
Hi :)
I got error i thinks it's related to this MR.

```
executing "nginx.conf.tmpl" at <.HTTPSnippets>: can't evaluate field HTTPSnippets in type *nginx.NginxMainConfig
```

pleshakov:
@alirezaDavid 
Hi. It looks like you're using the recent template file, but an outdated controller. Please rebuild the image https://github.com/nginxinc/kubernetes-ingress/tree/master/nginx-controller  using the latest files from the repo.

alirezaDavid:
@pleshakov 
Thanks for your attention.
I clone fresh clone of repo.
And build like this `make PREFIX=boo/nginx-ingress TAG=0.4` 
build successfully complete, but pod can't start because of mentioned error.

pleshakov:
@alirezaDavid that is strange. Are you sure that the boo/nginx-ingress:0.4 image gets deployed?

rchicoli:
@alirezaDavid could you try with this image [`rchicoli/kubernetes-nginx-ingress:0.7.0.6`.](https://hub.docker.com/r/rchicoli/kubernetes-nginx-ingress/tags/) This was the one I was using for my tests, after the implementation. Could you also post your yaml file as well? So I can test it here with the same configuration.

PR #140, opened 2017-05-10
Error message is mixed around

woopstar:
Just outputting the right message for the right error :)


pleshakov:
@woopstar Good catch! :) Thanks!

PR #141, opened 2017-05-10
Auto scale workers

woopstar:
I dont see any reason not to use auto for the worker processes  too, like the plus edition uses. Why ever limit  this to 1 in an ingress case?


pleshakov:
@woopstar Thanks! The value `1` comes from the default NGINX configuration file, while the default NGINX Plus configuration file has the value `auto`. It is a good idea to make this settings the same across the controllers, and there is no need to limit the default value to `1`.

PR #168, opened 2017-08-16
Implement lb-method fixes #94

### LINKED ISSUE #94: Allow ip_hash for backends
The controller should accept some annotation to indicate client affinity, if so then set `ip_hash` in the upstream block.

sajal:
Image built from my fork : `turbobytes/nginx-ingress:0.9.0`

This adds a configmap key `lb-method` and annotation `nginx.org/lb-method` as suggested by @pleshakov  in #94 .

If this setting is not blank, then the value is pasted into the upstream block in the nginx config. This is configurable per Ingress, and not per service.

### DIFF
--- a/examples/customization/README.md
+++ b/examples/customization/README.md
@@ -34,6 +34,7 @@ The table below summarizes some of the options. More options (extensions) are av
 | N/A | `http-snippets` | Sets a custom snippet in http context. | N/A |
 | `nginx.org/location-snippets` | `location-snippets` | Sets a custom snippet in location context. | N/A |
 | `nginx.org/server-snippets` | `server-snippets` | Sets a custom snippet in server context. | N/A |
+| `nginx.org/lb-method` | `lb-method` | Sets the [load balancing method](https://www.nginx.com/resources/admin-guide/load-balancer/#method). The default `""` specifies the round-robin method. | `""` |
 
 ## Using ConfigMaps
 
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -377,6 +377,10 @@ func (lbc *LoadBalancerController) syncCfgm(task Task) {
 			}
 		}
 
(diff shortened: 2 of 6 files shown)


pleshakov:
@sajal 
thanks for the PR!

Here is a couple of suggestions:
1. In the documentation, could you change the default value from `N/A` to `""` and also add the following statement after "Sets the load balancing method.": The `""` specifies the round-robin method.
1. I think it is better to pass the `ingCfg` to the `createUpstream` and extract the method from it, rather than from `ingEx` as in https://github.com/turbobytes/kubernetes-ingress/blob/e4a43d4e700969fd9d71a90fd879350ba2c3f486/nginx-controller/nginx/configurator.go#L449 Thus,  we don't need to have `LBMethod  string` as  part of `IngressEx`. It is sufficient to have `LBMethod` only in `Config` and `Upstream`

sajal:
@pleshakov Good ideas. I will look into them now.

sajal:
@pleshakov Now im passing LBMethod as argument instead of using `IngressEx`

pleshakov:
@sajal great! thx

I have two more suggestions:

Could you move this logic https://github.com/turbobytes/kubernetes-ingress/blob/b819d7e0acdb119541b7d91228167af99c6e5b5f/nginx-controller/nginx/configurator.go#L86-L88 to the [`createConfig`](https://github.com/turbobytes/kubernetes-ingress/blob/b819d7e0acdb119541b7d91228167af99c6e5b5f/nginx-controller/nginx/configurator.go#L206) method for consistency. This way all annotation parsing happens only in one place.

Could you also squash you commits into a single one?

sajal:
Ah so I will overwrite the annotation value id present at `ingCfg` in `createConfig`.

sajal:
@pleshakov : moved annotation checking into `createConfig` and squash'd

PR #171, opened 2017-08-22
allow non-standard listen ports

### LINKED ISSUE #98: Is it possible to route on the basis of port?
I have two services (S1, S2). When a request comes on port 2775 or 8443 I have to forward the request to S1 and when request comes on port 443 I have to forward it to S2. 

I was trying to do this using  the example given but it seems that ingress doesn't allow routing on the basis of port but only on the basis of path.
https://github.com/nginxinc/kubernetes-ingress/blob/master/examples/complete-example/cafe-ingress.yaml

How do I achieve the above mentioned goal. It is possible that there is a better way of doing this. Please suggest if one exists.

seletskiy:
PR introduces two new annotations:

* `nginx.org/listen-ports`
* `nginx.org/listen-ports-ssl`

When using ingress controller with `hostNetwork: true` it is possible to allow
to have non-standard ports (not 80 and 443) served by nginx. It's useful
for dynamic ports in ingress/port-based routing.

Fixes #98.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ osx-nginx-ingress
 nginx-ingress
 osx-nginx-plus-ingress
 nginx-plus-ingress
+nginx-controller/nginx-controller
 
 # NGINX Plus license files
 *.crt
--- a/nginx-controller/nginx/configurator.go
+++ b/nginx-controller/nginx/configurator.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/golang/glog"
@@ -132,6 +133,8 @@ func (cnf *Configurator) generateNginxCfg(ingEx *IngressEx, pems map[string]stri
 			ProxyHideHeaders:      ingCfg.ProxyHideHeaders,
 			ProxyPassHeaders:      ingCfg.ProxyPassHeaders,
 			ServerSnippets:        ingCfg.ServerSnippets,
+			Ports:                 ingCfg.Ports,
+			SSLPorts:              ingCfg.SSLPorts,
 		}
 
 		if pemFile, ok := pems[serverName]; ok {
@@ -311,6 +314,15 @@ func (cnf *Configurator) createConfig(ingEx *IngressEx) Config {
 		}
 	}
 
+	ports, sslPorts := getServicesPorts(ingEx)
+	if len(ports) > 0 {
+		ingCfg.Ports = ports
+	}
+
+	if len(sslPorts) > 0 {
+		ingCfg.SSLPorts = sslPorts
+	}
+
 	return ingCfg
 }
 
@@ -403,6 +415,53 @@ func parseStickyService(service string) (serviceName string, stickyCookie string
 	}
 
 	return svcNameParts[1], parts[1], nil
+}
+
+func getServicesPorts(ingEx *IngressEx) ([]int, []int) {
+	ports := map[string][]int{}
+
+	annotations := []string{
+		"nginx.org/listen-ports",
+		"nginx.org/listen-ports-ssl",
+	}
+
+	for _, annotation := range annotations {
+		if values, exists := ingEx.Ingress.Annotations[annotation]; exists {
+			for _, value := range strings.Split(values, ",") {
+				if port, err := parsePort(value); err != nil {
+					glog.Errorf(
+						"In %v %s contains invalid declaration: %v, ignoring",
+						ingEx.Ingress.Name,
+						annotation,
+						err,
+					)
+				} else {
+					ports[annotation] = append(ports[annotation], port)
+				}
+			}
+		}
+	}
+
+	return ports[annotations[0]], ports[annotations[1]]
+}
+
+func parsePort(value string) (int, error) {
+	port, err := strconv.ParseInt(value, 10, 16)
+	if err != nil {
+		return 0, fmt.Errorf(
+			"Unable to parse port as integer: %s\n",
+			err,
+		)
+	}
+
+	if port <= 0 {
+		return 0, fmt.Errorf(
+			"Port number should be greater than zero: %q",
+			port,
+		)
+	}
+
+	return int(port), nil
 }
 
 func createLocation(path string, upstream Upstream, cfg *Config, websocket bool, rewrite string, ssl bool) Location {
--- a/nginx-controller/nginx/templates/nginx.ingress.tmpl
+++ b/nginx-controller/nginx/templates/nginx.ingress.tmpl
@@ -7,9 +7,13 @@ upstream {{$upstream.Name}} {
 
 {{range $server := .Servers}}
 server {
-	listen 80{{if $server.ProxyProtocol}} proxy_protocol{{end}};
+	{{range $port := $server.Ports}}
+	listen {{$port}}{{if $server.ProxyProtocol}} proxy_protocol{{end}};
+	{{- end}}
 	{{if $server.SSL}}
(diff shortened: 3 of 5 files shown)


pleshakov:
@seletskiy Thanks!

A few suggestions:
* Documentation: I think it is better to document those annotation in the table [here](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/customization), than having an example with yaml files. Please remove those yaml files. I can add documentation later after the PR is merged.
* It is important to redirect to a different SSL port [here](https://github.com/seletskiy/kubernetes-ingress/blob/63b924ee518222cde994a3737718616a05b14108/nginx-controller/nginx/templates/nginx.ingress.tmpl#L35), while for [this case](https://github.com/seletskiy/kubernetes-ingress/blob/63b924ee518222cde994a3737718616a05b14108/nginx-controller/nginx/templates/nginx.ingress.tmpl#L43) it is ok -- that one is for the case when NGINX is sitting behind a load balancer and SSL termination is done by that load balancer.
* It is better to add Ports and SSLPorts to [Config](https://github.com/seletskiy/kubernetes-ingress/blob/63b924ee518222cde994a3737718616a05b14108/nginx-controller/nginx/config.go#L4) and parse annotations in [createConfig](https://github.com/seletskiy/kubernetes-ingress/blob/63b924ee518222cde994a3737718616a05b14108/nginx-controller/nginx/configurator.go#L162)

seletskiy:
@pleshakov:

> Documentation: I think it is better to document those annotation in the table here, than having an example with yaml files. Please remove those yaml files. I can add documentation later after the PR is merged.

Done.

> It is important to redirect to a different SSL port here, while for this case it is ok -- that one is for the case when NGINX is sitting behind a load balancer and SSL termination is done by that load balancer.

Done: https://github.com/nginxinc/kubernetes-ingress/pull/171/commits/f1b118bf2e49b5d118baf4a5d93d092fe851b021#diff-394b0af448cb4ab4c28989f3aecae208L31

> It is better to add Ports and SSLPorts to Config and parse annotations in createConfig

This will not work, because `Config` is not passed to the `nginx.ingress.tmpl`, so we can't extract ports from it here:

nginx-controller/nginx/configurator.go
```diff
@@ -115,12 +117,20 @@ func (cnf *Configurator) generateNginxCfg(ingEx *IngressEx, pems map[string]stri
...
 			ServerSnippets:        ingCfg.ServerSnippets,
+			Ports:                 ports,
+			SSLPorts:              sslPorts,
 		}
 
 		if pemFile, ok := pems[serverName]; ok {
-			server.SSL = true
-			server.SSLCertificate = pemFile
-			server.SSLCertificateKey = pemFile
+			if len(sslPorts) > 1 {
```
pleshakov: 1. `len(sslPorts) > 1` leads to a bug, as it only enables SSL termination when we have at least 2 SSL ports.
1. this check is also unnecessary: we should always have at least 1 SSL port if SSL is enabled (`if  pemFile, ok := pems[serverName];`)

nginx-controller/nginx/templates/nginx.ingress.tmpl
```diff
@@ -28,15 +32,15 @@ server {
...
+		return 301 https://$host:{{index $server.SSLPorts 0}}$request_uri;
 	}
 	{{- if $server.HSTS}}
 	proxy_hide_header Strict-Transport-Security;
 	add_header Strict-Transport-Security "max-age={{$server.HSTSMaxAge}}; {{if $server.HSTSIncludeSubdomains}}includeSubDomains; {{end}}preload" always;{{end}}
 	{{- end}}
 	{{- if $server.RedirectToHTTPS}}
 	if ($http_x_forwarded_proto = 'http') {
-		return 301 https://$host$request_uri;
+		return 301 https://$host:{{index $server.SSLPorts 0}}$request_uri;
```
pleshakov: This redirect is for the case when SSL termination is done at a load balancer in front of NGINX. It can be enabled for Ingress resources without SSL termination. Thus, it is a bug and the config generation will fail.

.gitignore
```diff
@@ -26,6 +26,7 @@ osx-nginx-ingress
 nginx-ingress
 osx-nginx-plus-ingress
 nginx-plus-ingress
+nginx-controller/nginx-controller
```
pleshakov: what is this for?

nginx-controller/nginx/configurator.go
seletskiy: Yes, you're right, sorry.

seletskiy: That check is simple a safe guard for case when annotation specifies empty value.

nginx-controller/nginx/templates/nginx.ingress.tmpl
seletskiy: Reverted.

.gitignore
seletskiy: It's to ignore `nginx-controller` binary which appears when you run `go build` inside `nginx-controller` directory.

seletskiy:
> It be great if we could have ports and ssl ports defined in the Config, with default values set to [80] and [443]. When parsing the annotations in createConfig , we could overwrite those default values based on the annotations values, but always making sure that there is at least one port for regular and ssl ports.

OK, I've moved default values to `Config` struct. Check it out now.

pleshakov:
@seletskiy Thanks so much!
There is one thing left.  [This check](https://github.com/seletskiy/kubernetes-ingress/blob/1f6eff6cf2963edb185981b91851a07c2dce59ff/nginx-controller/nginx/configurator.go#L124) is unnecessary, as we always make sure that there is at least one element in `ingCfg.SSLPorts`.

seletskiy:
@pleshakov: I've removed excessive `if`, rebased and resolved conflicts against new commits in upstream. Check it out again.

pleshakov:
@seletskiy Thanks!!

seletskiy:
@pleshakov: Thanks for assistance. Do you plan to release new version of image on the Docker Hub?

pleshakov:
@seletskiy yes, we will release a new version today

ChristopherLClark:
can anyone provide a working example?

pleshakov:
@ChristopherLClark here is an example. instead of 80 and 443, NGINX will listen on 9080 and 9443
```yaml
kind: Ingress
metadata:
  name: cafe-ingress
  annotations:
     nginx.org/listen-ports: "9080"
     nginx.org/listen-ports-ssl: "9443"
spec:
. . .
```

Also see https://github.com/nginxinc/kubernetes-ingress/blob/master/docs/configmap-and-annotations.md

ChristopherLClark:
@pleshakov Hmm, this still doesnt quite meet my needs. I need to have kubernetes, in the spec, rules, portion be forwarding routes with ports.

So for example:

```
kind: Ingress
metadata:
  name: ingress-iot-admin
  annotations: 
    kubernetes.io/ingress.class: nginx
    nginx.org/listen-ports: "30000"
  labels:
    app: iot-admin
spec:
  rules:
  - host: localhost:30000
    http:
      paths:
      - path: /
        backend:
          serviceName: service-iot-admin
          servicePort: 8080
```

In local development we can't use dns entries. So I would like to mimic the production cluster with localhost:30000 as the domain and then all requests go through that.

pleshakov:
@ChristopherLClark that should work OK, if you change `localhost:30000` to `localhost` in your Ingress

ChristopherLClark:
@pleshakov yes I ultimately ended up running kubernetes under port 80, and killed my other processes, but I was looking for a way where I could run kubernetes under a different port. This ended up being easier just switching to 80.

boxer3379:
Couldn't tell you I just learned to mess with this stuff not to long ago
and I love it

WangHexie:
use the  example, but always bind on 80 port, how to solve?

PR #186, opened 2017-09-08
Fix hardcoded controller class

SiriusRed:
Add flag `nginx-class`, which can be set specific class for running multiple nginx ingress controllers.


pleshakov:
@SiriusRed thanks!

A couple of suggestions:

- This PR breaks the current behavior, which is  described [here](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/multiple-ingress-controllers) . Currently, Ingress resources with the annotation `kubernetes.io/ingress.class: "nginx"` or no annotation are handled by NGINX Ingress controller. It is better not to change this behavior. If you would like to make the Ingress controller process only Ingress resources with  the annotation  `kubernetes.io/ingress.class: "nginx"`, I suggest to have an additional Boolean command-line flag, something like `ingress-class-only`.
- It is better to rename the flag from `nginx-class` to `ingress-class`

SiriusRed:
Thanks for remark, mr. @pleshakov!

- Renamed the flag `nginx-class` to `ingress-class`, it is more clearly.
- Add flag 'use-ingress-class-only' for handled ingresses only with matching value of 'ingress-class'.

pleshakov:
Thanks @SiriusRed 

I have a few additional suggestions:
* Could you change `VERSION` in the Makefile from `1.0.1` to `1.0.0.` We only change VERSION, when we make a new release?
* Could you run all modified go files through `go fmt` tool to fix formatting?
* Could you squash all your commits into one commit? 

Also, there is a bug. When `use-ingress-class-only` option is enabled, if an Ingress resource doesn't have the `ingress.class` annotation, the Ingress controller processes this Ingress resource. It should not process such an Ingress resource, but only Ingress resources for the configured Ingress class.

SiriusRed:
Done all of thing which you said. Added comment.

pleshakov:
@SiriusRed 
Thanks. Works as expected now!

Could you make one last change - rename the variables  `useIngressClass` to `useIngressClassOnly`? This will help make the code more readable.

SiriusRed:
@pleshakov 
Done!

pleshakov:
@SiriusRed Thanks!

Issue #164, opened 2017-08-06
TCP Support?


hapnermw:
The current ingress resource doesn't support TCP services. It's likely to support this later; but, for now there needs to be some lower-level/controller-specific way of doing this.

Is there an undocumented way to provide ingress to TCP services.

If not, is there any ongoing development on a work-around for this. Ideally it would be a controller specific annotation and/or configmap that will allow dynamic deployment of TCP services via an ingress resource.

Currently, the kubernetes nginx-ingress-controller provides a work-around for staticly specifying TCP services; however, this isn't very useful for dynamic deployment of services (which is the whole purpose of the ingress facility).

pleshakov:
@hapnermw 
there is no undocumented way to provide Ingress for TCP services and there are no plans to add it in the nearest future.

Why doesn't the [config map approach](https://github.com/kubernetes/ingress/tree/master/controllers/nginx#exposing-tcp-services) of the nginx-ingress-controller  work you? Not sure If completely understand the distinction between static and dynamic deployments in your example.

hapnermw:
The reason k8s added the ingress abstraction is to support the dynamic registration of virtual hosts (virtual servers) for services (via  HTTP  hostname or TLS SNI).

NginX's ingress controller supports this by dynamically discovering k8s ingress resources and configuring HTTP virtual servers for them.

NginX has supported TLS/TCP virtual servers for some time. They do TLS/TCP session load balancing and require different configuration. The NginX k8s ingress controller does not support this NginX feature because currently k8s ingress does not support it.

This is unfortunate because it wouldn't have been difficult to include syntax for TCP ingress and simply note that support for this was optional. By not including this, k8s has blocked ingress controllers from supporting it.

julienvincent:
@pleshakov I am trying to use the ConfigMap approach but with little success. I am using the following yml (on GKE):

```yml
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: ingress-nginx
  name: tcp-services
data:
  2000: "lb-test/lb-tcp-headless:2000"
```

This does not seem to have any effect and I am still unable to connect to the server via port 2000. Not sure if I am miss-understanding something here?

pleshakov:
@hapnermw load balancing TCP/UDP is available through https://github.com/kubernetes/ingress-nginx/ Ingress controller

totemcaf:
[...]

PR #195, opened 2017-10-03
RFC: additional variables to log pod attributes

seletskiy:
As we know, we can log upstream(s) which served request by using
`$upstream_addr` variable in `log_format` definition. However, pod IPs in k8s
are intrinsically volatile and can be changed at pod restart, so if we will use
`$upstream_addr` in logs it will be difficult to understand which pods really
processed request.

I propose a PR which adds a way to use `$pod_name`, `$namespace_name` (sic!),
`$container_name` and `$pod_labels` to be available for use in `log_format`.

Their values are inferred from latest address in the `$upstream_addr` variable
by using `map` directive. Since mapped variables are evaluated at expansion
time, it works and will not introduce additional overhead for users who will
not use mentioned variables in their log format.

Current implementation is quite rough and I open for comments to improve it.

### DIFF
--- a/nginx-controller/Dockerfile
+++ b/nginx-controller/Dockerfile
@@ -5,7 +5,11 @@ FROM nginx:1.13.3
 RUN ln -sf /proc/1/fd/1 /var/log/nginx/access.log \
 	&& ln -sf /proc/1/fd/2 /var/log/nginx/error.log
 
-COPY nginx-ingress nginx/templates/nginx.ingress.tmpl nginx/templates/nginx.tmpl /
+COPY nginx-ingress \
+    nginx/templates/nginx.ingress.tmpl \
+    nginx/templates/nginx.tmpl \
+    nginx/templates/nginx.maps.tmpl \
+    /
 
 RUN rm /etc/nginx/conf.d/*
 
--- a/nginx-controller/Dockerfile
+++ b/nginx-controller/Dockerfile
@@ -15,6 +15,9 @@ RUN rm /etc/nginx/conf.d/*
 
 RUN mkdir -p /etc/nginx/secrets
 
+ADD https://github.com/seletskiy/killinit/releases/download/1.0/killinit /
+RUN chmod +x /killinit
+
 # Uncomment the line below if you would like to add the default.pem to the image
 # and use it as a certificate and key for the default server
 # ADD default.pem /etc/nginx/secrets/default
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -907,9 +907,9 @@ func (lbc *LoadBalancerController) createIngress(ing *extensions.Ingress) (*ngin
 		validRules++
 	}
 
-	if validRules == 0 {
-		return nil, fmt.Errorf("Ingress contains no valid rules")
-	}
+	//if validRules == 0 {
+	//    return nil, fmt.Errorf("Ingress contains no valid rules")
+	//}
 
 	return ingEx, nil
 }
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -867,13 +867,17 @@ func (lbc *LoadBalancerController) createIngress(ing *extensions.Ingress) (*ngin
 	}
 
 	ingEx.Endpoints = make(map[string][]string)
+	ingEx.EndpointsInfo = make(map[string][]nginx.EndpointInfo)
 	if ing.Spec.Backend != nil {
-		endps, err := lbc.getEndpointsForIngressBackend(ing.Spec.Backend, ing.Namespace)
+		key := ing.Spec.Backend.ServiceName + ing.Spec.Backend.ServicePort.String()
+
+		endps, info, err := lbc.getEndpointsForIngressBackend(ing.Spec.Backend, ing.Namespace)
 		if err != nil {
 			glog.Warningf("Error retrieving endpoints for the service %v: %v", ing.Spec.Backend.ServiceName, err)
-			ingEx.Endpoints[ing.Spec.Backend.ServiceName+ing.Spec.Backend.ServicePort.String()] = []string{}
+			ingEx.Endpoints[key] = []string{}
 		} else {
-			ingEx.Endpoints[ing.Spec.Backend.ServiceName+ing.Spec.Backend.ServicePort.String()] = endps
+			ingEx.Endpoints[key] = endps
+			ingEx.EndpointsInfo[key] = info
 		}
 	}
 
@@ -889,12 +893,14 @@ func (lbc *LoadBalancerController) createIngress(ing *extensions.Ingress) (*ngin
 		}
 
 		for _, path := range rule.HTTP.Paths {
-			endps, err := lbc.getEndpointsForIngressBackend(&path.Backend, ing.Namespace)
+			key := path.Backend.ServiceName + path.Backend.ServicePort.String()
+			endps, info, err := lbc.getEndpointsForIngressBackend(&path.Backend, ing.Namespace)
 			if err != nil {
 				glog.Warningf("Error retrieving endpoints for the service %v: %v", path.Backend.ServiceName, err)
-				ingEx.Endpoints[path.Backend.ServiceName+path.Backend.ServicePort.String()] = []string{}
+				ingEx.Endpoints[key] = []string{}
 			} else {
-				ingEx.Endpoints[path.Backend.ServiceName+path.Backend.ServicePort.String()] = endps
+				ingEx.Endpoints[key] = endps
+				ingEx.EndpointsInfo[key] = info
 			}
 		}
 
@@ -908,25 +914,35 @@ func (lbc *LoadBalancerController) createIngress(ing *extensions.Ingress) (*ngin
 	return ingEx, nil
 }
 
-func (lbc *LoadBalancerController) getEndpointsForIngressBackend(backend *extensions.IngressBackend, namespace string) ([]string, error) {
+func (lbc *LoadBalancerController) getEndpointsForIngressBackend(
+	backend *extensions.IngressBackend,
+	namespace string,
+) ([]string, []nginx.EndpointInfo, error) {
 	svc, err := lbc.getServiceForIngressBackend(backend, namespace)
 	if err != nil {
 		glog.V(3).Infof("Error getting service %v: %v", backend.ServiceName, err)
-		return nil, err
+		return nil, nil, err
 	}
 
 	endps, err := lbc.endpLister.GetServiceEndpoints(svc)
 	if err != nil {
 		glog.V(3).Infof("Error getting endpoints for service %s from the cache: %v", svc.Name, err)
-		return nil, err
+		return nil, nil, err
 	}
 
 	result, err := lbc.getEndpointsForPort(endps, backend.ServicePort, svc)
 	if err != nil {
 		glog.V(3).Infof("Error getting endpoints for service %s port %v: %v", svc.Name, backend.ServicePort, err)
-		return nil, err
+		return nil, nil, err
 	}
-	return result, nil
+
+	info, err := lbc.getEndpointsInfo(svc)
+	if err != nil {
+		glog.V(3).Infof("Error getting endpoints info for service %s port %v: %v", svc.Name, backend.ServicePort, err)
+		return nil, nil, err
+	}
+
+	return result, info, nil
 }
 
 func (lbc *LoadBalancerController) getEndpointsForPort(endps api_v1.Endpoints, ingSvcPort intstr.IntOrString, svc *api_v1.Service) ([]string, error) {
@@ -991,6 +1007,45 @@ func (lbc *LoadBalancerController) getTargetPort(svcPort *api_v1.ServicePort, sv
 	}
 
 	return portNum, nil
+}
+
+func (lbc *LoadBalancerController) getEndpointsInfo(
+	svc *api_v1.Service,
+) ([]nginx.EndpointInfo, error) {
+	pods, err := lbc.client.Core().Pods(svc.Namespace).List(
+		meta_v1.ListOptions{
+			LabelSelector: labels.Set(svc.Spec.Selector).String(),
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting pod information: %v", err)
+	}
+
+	if len(pods.Items) == 0 {
+		return nil, fmt.Errorf("No pods of service %s", svc.Name)
+	}
+
+	info := []nginx.EndpointInfo{}
+
+	for _, pod := range pods.Items {
+		ip := pod.Status.PodIP
+
+		for _, container := range pod.Spec.Containers {
+			for _, port := range container.Ports {
+				info = append(info, nginx.EndpointInfo{
+					Address: fmt.Sprintf("%s:%d", ip, port.ContainerPort),
+					Info: map[string]interface{}{
+						"namespace":  pod.Namespace,
+						"pod":        pod.Name,
+						"container":  container.Name,
+						"pod_labels": nginx.Labels(pod.Labels),
+					},
+				})
+			}
+		}
+	}
+
+	return info, nil
 }
 
 func (lbc *LoadBalancerController) getServiceForIngressBackend(backend *extensions.IngressBackend, namespace string) (*api_v1.Service, error) {
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -397,6 +397,9 @@ func (lbc *LoadBalancerController) syncCfgm(task Task) {
 			cfg.LBMethod = lbMethod
 		}
 
+		if proxyTimeout, exists := cfgm.Data["proxy-timeout"]; exists {
+			cfg.ProxyTimeout = proxyTimeout
+		}
 		if proxyConnectTimeout, exists := cfgm.Data["proxy-connect-timeout"]; exists {
 			cfg.ProxyConnectTimeout = proxyConnectTimeout
 		}
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -1009,9 +1009,7 @@ func (lbc *LoadBalancerController) getTargetPort(svcPort *api_v1.ServicePort, sv
 	return portNum, nil
 }
 
-func (lbc *LoadBalancerController) getEndpointsInfo(
-	svc *api_v1.Service,
-) ([]nginx.EndpointInfo, error) {
+func (lbc *LoadBalancerController) getEndpointsInfo(svc *api_v1.Service) ([]nginx.EndpointInfo, error) {
 	pods, err := lbc.client.Core().Pods(svc.Namespace).List(
 		meta_v1.ListOptions{
 			LabelSelector: labels.Set(svc.Spec.Selector).String(),
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -579,6 +579,13 @@ func (lbc *LoadBalancerController) syncCfgm(task Task) {
 				cfg.MainWorkerProcesses = cfgm.Data["worker-processes"]
 			}
 		}
+		if workerConnections, exists, err := nginx.GetMapKeyAsInt(cfgm.Data, "worker-connections", cfgm); exists {
+			if err != nil {
+				glog.Errorf("Configmap %s/%s: Invalid value for worker-connections key: must be an integer, got %q", cfgm.GetNamespace(), cfgm.GetName(), cfgm.Data["worker-connections"])
+			} else {
+				cfg.MainWorkerConnections = workerConnections
+			}
+		}
 		if workerCPUAffinity, exists := cfgm.Data["worker-cpu-affinity"]; exists {
 			cfg.MainWorkerCPUAffinity = workerCPUAffinity
 		}
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -1033,13 +1033,17 @@ func (lbc *LoadBalancerController) getEndpointsInfo(
 		for _, container := range pod.Spec.Containers {
 			for _, port := range container.Ports {
 				info = append(info, nginx.EndpointInfo{
-					Address: fmt.Sprintf("%s:%d", ip, port.ContainerPort),
-					Info: map[string]interface{}{
-						"namespace":  pod.Namespace,
-						"pod":        pod.Name,
-						"container":  container.Name,
-						"pod_labels": nginx.Labels(pod.Labels),
-					},
+					Address:   fmt.Sprintf("%s:%d", ip, port.ContainerPort),
+					Namespace: pod.Namespace,
+					Pod:       pod.Name,
+					Container: container.Name,
+					PodLabels: nginx.Labels(pod.Labels),
+					//Info: map[string]interface{}{
+					//    "namespace":  pod.Namespace,
+					//    "pod":        pod.Name,
+					//    "container":  container.Name,
+					//    "pod_labels": nginx.Labels(pod.Labels),
+					//},
 				})
 			}
 		}
--- a/nginx-controller/nginx/config.go
+++ b/nginx-controller/nginx/config.go
@@ -50,6 +50,9 @@ type Config struct {
 	SSLPorts []int
 
 	Stream bool
+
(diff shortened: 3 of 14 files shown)


seletskiy:
@pleshakov: Can you take a look at this?

pleshakov:
@seletskiy thx. this is really useful! I'll provide comments tomorrow.

nginx-controller/nginx/templates/nginx.ingress.tmpl
```diff
@@ -5,6 +5,32 @@ upstream {{$upstream.Name}} {
 	server {{$server.Address}}:{{$server.Port}};{{end}}
 }{{end}}
 
+{{with .EndpointsInfo}}
+map $upstream_addr $namespace_name {
```
pleshakov: This won't work if you have more than one Ingress resource deployed. In that case you will get several maps generated, which will collide with each other.
The way to solve that problem is to prefix the variable name with the namespace and the name of the corresponding Ingress resource:
```nginx
map $upstream_addr $IngressNameSpace_IngressName_namespace_name {
...
}
```
However, this will require to use `set`:
```nginx
 set $namespace_name $IngressNameSpace_IngressName_namespace_name;
```

Also, I suggest using the following variable names:
```
$k8s_upstream_pod
$k8s_upstream_container
$k8s_upstream_namespace
$k8s_upstream_pod_labels
```
Note the prefix `k8s` for the k8s-related variables.

nginx-controller/nginx/nginx.go
```diff
@@ -30,8 +30,9 @@ type NginxController struct {
 
 // IngressNginxConfig describes an NGINX configuration
 type IngressNginxConfig struct {
-	Upstreams []Upstream
-	Servers   []Server
+	Upstreams     []Upstream
+	Servers       []Server
+	EndpointsInfo map[string]EndpointInfo
```
pleshakov: I don't think `EndpointsInfo` should be a part of the `IngressNginxConfig`. It is better to have a special type for a map. Something like this:
```go
type Map struct {
  Source string
  Variable string
  Params map[string]string
}

File: nginx-controller/nginx/ingress.go [hunk 3]
```diff
@@ -1,15 +1,39 @@
...
 type IngressEx struct {
-	Ingress    *extensions.Ingress
-	TLSSecrets map[string]*api_v1.Secret
-	JWTKey     *api_v1.Secret
-	Endpoints  map[string][]string
+	Ingress       *extensions.Ingress
+	TLSSecrets    map[string]*api_v1.Secret
+	JWTKey        *api_v1.Secret
+	Endpoints     map[string][]string
+	EndpointsInfo map[string]map[string]EndpointInfo
```
pleshakov: I don't think a nested map is required. `map[string]string` should be enough

nginx-controller/nginx/ingress.go
```diff
@@ -1,15 +1,39 @@
...
+}
+
+// TODO
+type EndpointInfo struct {
+	ContainerName string
```
pleshakov: the suffix `Name` is not required in the variable names

nginx-controller/nginx/configurator.go
```diff
@@ -323,6 +328,12 @@ func (cnf *Configurator) createConfig(ingEx *IngressEx) Config {
 		ingCfg.SSLPorts = sslPorts
 	}
 
+	for _, bucket := range ingEx.EndpointsInfo {
```
pleshakov: Maps should be generated in the `generateNginxCfg` method.

nginx-controller/nginx/config.go
```diff
@@ -46,6 +46,8 @@ type Config struct {
 
 	Ports    []int
 	SSLPorts []int
+
+	EndpointsInfo map[string]EndpointInfo
```
pleshakov: We don't really need to have this field in the `Config` type. We pass `EndpointsInfo` in `IngressEx` to the `generateNginxCfg` method of `Configurator` and generate maps in that method. There is no need to store Endpoints info in the `Config`.

seletskiy:
@pleshakov: Thanks for comments. Most of them fixed.

However, there's one overlooked problem which I don't know how to resolve: usage of `set` will cause map to be evaluated BEFORE `$upstream_addr` is known, which cause target variable to be always empty. Previous solution worked because variable was evaluated at logging stage where `$upstream_addr` is known. The only solution I can imagine right now is to use single maps (one per variable) for all upstreams in all ingresses, which should be declared in main `nginx.conf`. Controller should be reworked to regenerate `nginx.conf` at every ingress update, then.

pleshakov:
@seletskiy 
 you're right. setting the `$upstream_addr` doesn't work.

Also the implementation uses regular expression for mapping the IP address of a pod to the pod name, and there is no better way to do it, considering the fact that the `$upstream_addr` variable can have multiple values. Using regular expressions affects performance.  

As an alternative solution, I can suggest the following solution:
1. The backend application adds additional headers into its response that contain the pod name, namespace and any additional information.
2. NGINX log format reference that information via [$upstream_http_](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#var_upstream_http_) variable
3. Additionally, NGINX can clear those headers, so that information doesn't leak outside of the k8s cluster.

seletskiy:
@pleshakov:

> Also the implementation uses regular expression for mapping the IP address of a pod to the pod name, and there is no better way to do it, considering the fact that the $upstream_addr variable can have multiple values. Using regular expressions affects performance.

We can make that option configurable, so who doesn't want performance impact can just switch it off. I by myself doubt that in the age of complex k8s clusters someone still care about regexp performance.

> The backend application adds additional headers into its response that contain the pod name, namespace and any additional information.

Unfortunately, it is not a solution, since we're running lots of third-party apps in k8s clusters which we can't modify to include our custom headers in the output. Also, bear in mind future TCP ingresses, which will not have any HTTP headers.

The only solution which I can see right now is following:

* include global state about endpoints info in the `LoadBalancerController` struct;
* modify endpoints listener, so when event is received, retrieve service with name from that event, then retrieve all pods from that service and store them into global state;
* if endpoint is removed, delete corresponding item from global state;
* create new configuration file located in the `/etc/nginx/maps.conf`, add `include /etc/nginx/maps.conf` in the `/etc/nginx/nginx.conf` at `http` level and update `/etc/nginx/maps.conf` whenever endpoint event is received.

pleshakov:
@seletskiy 
> Unfortunately, it is not a solution, since we're running lots of third-party apps in k8s clusters which we can't modify to include our custom headers in the output. Also, bear in mind future TCP ingresses, which will not have any HTTP headers.

That makes sense.

It makes sense to do some performance testing with and without that feature enabled, to see how it will affect your deployment. It is already possible with your existing application.

With the global map approach, I think the following configuration is reasonable:
a global map for each variable:
```nginx
map $k8s_namespace/$k8s_ingress $k8s_upstream_pod {
    ns1/ing1 $k8s_upstream_pod_ns1_ing1;
    ns2/ing2 $k8s_upstream_pod_ns2_ing2;
}
```
Configuration for the Ingress resource *ing1*:
```nginx
server {
  set $k8s_namespace "ns1";
  set $k8s_ingress "ing1";
  ...
}

map $upstream_addr $k8s_upstream_pod_ns1_ing1 {
    127.0.0.1:8081 ing1-pod;
}
```

Configuration for the Ingress resource *ing2*:
```nginx
server {
  set $k8s_namespace "ns2";
  set $k8s_ingress "ing2";
  ...
}

map $upstream_addr $k8s_upstream_pod_ns2_ing2 {
    127.0.0.1:9081 ing2-pod;
}
```

Note: 
* The case of multiple addresses in the `$upstream_addr` variable is not handled here
* `$k8s_namespace` and `$k8s_ingress` are additional variables that are ready to be included in the log format 

Let me know what you think. However, again, it is important to understand how it affects performance.

Currently, endpoints changes are already handled. When the endpoints are changed, the Ingress controller regenerates the configuration for the affected Ingress resources. Thus, we can extend that code to support endpoints info, as you already did.  The only missing piece is the global maps generation.

This week I'm planning to add the logic of tracking added/removed Ingress resources in `Configurator` to fix a minor Issue, related to adding endpoints for Ingress resources, for which the config was not generated yet. After that logic is added, it makes sense to add the generation of global maps next to it. Basically, when an Ingress resource is added, a new entry will be added to the global maps, when an Ingress resource is deleted, the corresponding entry will be removed from the global maps.

seletskiy:
@pleshakov: I like your solution. I need that functionality right now, so if you don't mind, I will add `ingresses` map into `Configurator` struct and add logic with adding/removing items in that map in `AddOrUpdateIngress` & `DeleteIngress` methods, then I will add `maps.conf` file and it's management.

pleshakov:
@seletskiy Please note that there is also `DeleteSecret` method, in which an the Ingress resource can be removed

seletskiy:
@pleshakov: Check it out again.

nginx-controller/Dockerfile
```diff
@@ -5,7 +5,12 @@ FROM nginx:1.13.3
 RUN ln -sf /proc/1/fd/1 /var/log/nginx/access.log \
 	&& ln -sf /proc/1/fd/2 /var/log/nginx/error.log
 
-COPY nginx-ingress nginx/templates/nginx.ingress.tmpl nginx/templates/nginx.tmpl /
+COPY \
+    nginx-ingress \
```
pleshakov: Could you change the first line to `COPY nginx-ingress  \`?

nginx-controller/controller/controller.go
```diff
@@ -979,6 +995,45 @@ func (lbc *LoadBalancerController) getTargetPort(svcPort *api_v1.ServicePort, sv
 	return portNum, nil
 }
 
+func (lbc *LoadBalancerController) getEndpointsInfo(
+	svc *api_v1.Service,
+) ([]nginx.EndpointInfo, error) {
```
pleshakov: This function declaration is not long enough to wrap it into multiple lines.

nginx-controller/nginx/configurator.go
```diff
@@ -63,6 +67,48 @@ func (cnf *Configurator) addOrUpdateIngress(ingEx *IngressEx) {
 	nginxCfg := cnf.generateNginxCfg(ingEx, pems, jwtKeyFileName)
 	name := objectMetaToFileName(&ingEx.Ingress.ObjectMeta)
 	cnf.nginx.AddOrUpdateIngress(name, nginxCfg)
+
+	key := fmt.Sprintf(
```
pleshakov: key generation happens in multiple places. better to create a function for it

nginx-controller/nginx/configurator.go
```diff
@@ -63,6 +67,48 @@ func (cnf *Configurator) addOrUpdateIngress(ingEx *IngressEx) {
...
+func (cnf *Configurator) updateMaps() {
+	maps := []Map{}
+
+	for _, variable := range []string{
+		"namespace",
+		"pod",
+		"container",
+		"pod_labels",
+	} {
+		values := map[string]string{}
```
pleshakov: it is better to create another function that takes an Ingress and generates maps for it.

nginx-controller/nginx/configurator.go
```diff
@@ -169,17 +217,69 @@ func (cnf *Configurator) generateNginxCfg(ingEx *IngressEx, pems map[string]stri
...
+	return IngressNginxConfig{
+		Upstreams: upstreamMapToSlice(upstreams),
+		Servers:   servers,
+	}
+}
+
+func getVariableMaps(ingEx *IngressEx, rule extensions.IngressRule) []Map {
+	maps := map[string]Map{}
+
+	for _, path := range rule.HTTP.Paths {
```
pleshakov: there is no need to iterate rule.HTTP.Paths . Iterating over `ingEx.Endpoints` should be enough.
Also, this function is complicated. It is better to split into multiple functions.

nginx-controller/nginx/configurator.go
```diff
@@ -529,6 +629,23 @@ func upstreamMapToSlice(upstreams map[string]Upstream) []Upstream {
 	return result
 }
 
+func getMappedVariableName(
```
pleshakov: There is no need for a multi-line function declaration.

nginx-controller/nginx/configurator.go
```diff
@@ -587,10 +704,14 @@ func GenerateCertAndKeyFileContent(secret *api_v1.Secret) []byte {
 func (cnf *Configurator) DeleteSecret(key string, ings []extensions.Ingress) error {
 	for _, ing := range ings {
 		cnf.nginx.DeleteIngress(objectMetaToFileName(&ing.ObjectMeta))
+
+		delete(cnf.ingresses, fmt.Sprintf("%s/%s", ing.Namespace, ing.Name))
 	}
 
 	cnf.nginx.DeleteSecretFile(keyToFileName(key))
 
+	cnf.updateMaps()
```
pleshakov: Updating maps only make sense when `len(ings) > 0`

nginx-controller/nginx/ingress.go
```diff
@@ -1,15 +1,37 @@
...
+	TLSSecrets    map[string]*api_v1.Secret
+	JWTKey        *api_v1.Secret
+	Endpoints     map[string][]string
+	EndpointsInfo map[string][]EndpointInfo
+}
+
+// TODO
+type EndpointInfo struct {
+	Address string
+	Info    map[string]interface{}
```
pleshakov: It is better to have `Pod`, `Container`, `Labels` fields instead of  `map[string]interface{}`

nginx-controller/nginx/nginx.go
```diff
@@ -48,8 +49,19 @@ type UpstreamServer struct {
...
+type Map struct {
+	Source   string
+	Target   string
+	Variable string
+	Values   map[string]string
+}
+
 // Server describes an NGINX server
 type Server struct {
+	Namespace   string
```
pleshakov: Namespace, IngressName as well as Maps should be part of IngressConfig.

nginx-controller/nginx/nginx.go
```diff
@@ -373,3 +395,35 @@ func (nginx *NginxController) UpdateMainConfigFile(cfg *NginxMainConfig) {
 
 	glog.V(3).Infof("The main NGINX configuration file had been updated")
 }
+
+func (nginx *NginxController) UpdateMapsConfigFile(maps []Map) {
+	tmpl, err := template.New(
```
pleshakov: Again, I don't see why we need to have a multi-line declaration here :)

nginx-controller/nginx/templates/nginx.tmpl
```diff
@@ -75,5 +75,6 @@ http {
         }
     }
 
+    include /etc/nginx/maps*.conf;
```
pleshakov: `/etc/nginx/maps*.conf` not sure why we have `*`

nginx-controller/Dockerfile
seletskiy: OK. But I did it intentionally to indicate, that all lines have same semantic.

nginx-controller/controller/controller.go
seletskiy: Line is long enough to not fit into 80 symbols per line. It's studied, that optimal line length should be <80 for balance of comprehension and reading speed — https://en.wikipedia.org/wiki/Line_length

Telling you as contributor to that code, it's very difficult to read with that long lines which even do not fit my editor, so I need to use horizontal scrolling.

However, I can join lines if you disagree.

nginx-controller/nginx/configurator.go
seletskiy: +

nginx-controller/nginx/configurator.go
seletskiy: There's already such function, it's called `getVariableMaps`, it generates maps for single ingress.

This function generates global maps.

nginx-controller/nginx/configurator.go
seletskiy: First part is fixed, regarding second part — I doubt that there's good way to split it into two functions without loosing cohesion. Method is only 40 lines long.

nginx-controller/nginx/configurator.go
seletskiy: That condition is already present inside method, not need to duplicate it.

nginx-controller/nginx/templates/nginx.tmpl
seletskiy: Because maps.conf can be not created at the moment nginx started for the first time.

nginx-controller/nginx/nginx.go
seletskiy: done

seletskiy:
@pleshakov: Please, check it out again.

nginx-controller/nginx/ingress.go
seletskiy: It will lead to ugly copy-paste code in nginx/configurator.go where variable maps are created, because I will not be able to iterate over variable names.

nginx-controller/nginx/templates/nginx.tmpl
pleshakov: @seletskiy it is possible to create it [here](https://github.com/nginxinc/kubernetes-ingress/blob/master/nginx-controller/nginx/nginx.go#L148) along with the main config.

nginx-controller/controller/controller.go
pleshakov: Consistency is the key. The other function declarations of similar length are one-line.

nginx-controller/nginx/configurator.go
pleshakov: This is slightly different.  

DeleteSecret method handles an event, when a secret is removed from k8s. If that secret is referenced in some Ingress resources, those Ingress resources will be temporary removed as well and they are passed in `ings` slice.  So, if that slice is empty, we don't need to regenerate the maps, as there were no changes to Ingress resources.

nginx-controller/nginx/ingress.go
pleshakov: I see your point. However, this approach is not desirable. In Configurator, you have the code like this:
```
		for _, variable := range []string{
			"namespace",
			"pod",
			"container",
			"pod_labels",
		} 
```
Basically, with such code, we're losing the benefits of static typing. This makes the code brittle and less maintainable. For example,  this is how EndpointInfos are generated in Controller :
```
Info: map[string]interface{}{
 +						"namespace":  pod.Namespace,
 +						"pod":        pod.Name,
 +						"container":  container.Name,
 +						"pod_labels": nginx.Labels(pod.Labels),
 +					},
```
If we make a change to Controller code, change `container` to `container_name` or remove `namespace` and compile the code,  but do not change the code Configurator, we will see no errors at the compilation time. If we had  those map  entries as fields of  EndpointInfo, the change that we made would lead to a compilation error.

pleshakov:
@seletskiy thanks for changes. I replied  to your comments. Again, some additional things:
- It makes sense to put all maps-related logic into a separate file like endpointinfo_maps.go
- Additionally, it is better to add some unit testing for that logic
- Also, could you extend the tests in templates_test.go to test how config is generated for those maps?
- It makes sense to make this feature optional. I think having a command-line argument that enables this feature is reasonable. What do you think?

nginx-controller/nginx/templates/nginx.tmpl
seletskiy: +

seletskiy:
@pleshakov: I've walked entire codebase and code under `nginx-controller` is now consistently fit 80 symbols limit (with exception of two lines which slightly longer without sane way of splitting them).

Endpoints info variable names mapping are moved into `endpoints_info.go`, but actual methods of building nginx maps are remain in the `configurator.go`, because one method (`updateMaps`) is declared on the `Configurator` and other (`getVariableNames`) have similar logic should not be moved far away of it's usage.

> Also, could you extend the tests in templates_test.go to test how config is generated for those maps?

Unfortunately, `templates_test.go` lacks any actual validation (it just compiles template and checks err), so adding another test-case with same logic is no-op.

> It makes sense to make this feature optional. I think having a command-line argument that enables this feature is reasonable. What do you think?

OK, I've added `-map-endpoints-info` command line key, which simply enables `maps.conf` inclusion.

Listen, I can't afford spending more time on that PR, so if you dislike something, can you, please, fix it by yourself?

pleshakov:
@seletskiy thanks for addressing my previous comments. 

Unfortunately, https://github.com/nginxinc/kubernetes-ingress/pull/195/commits/5b1a1fcb9f3037204a539f1add36fdca647f31e5 makes it harder to review the PR, as the changes in the commit are not related to the features related to maps. Could you possible remove this commit from the PR?

kovetskiy:
@pleshakov any progress of this pr? You can make review in a commit-by-commit manner, if it is so hard to review all changes at a time.

pleshakov:
@kovetskiy @seletskiy  this PR request has a lot of style-related changes. The added code doesn't confirm to the style of the existing code. Moreover, formatting was changed across all source files. 

I appreciate that you introduced those style changes, however, (1) such changes should be done in a separate PR and (2) it is always better to discuss it before doing anything like that.

I don't see that we need to introduce a major style change.  I won't accept a pull request that doesn't confirm to the existing style.

Regarding the main feature of this PR. The feature is useful and I hope we can merge it soon. However, the style changes make it harder to review and test the PR. I suggest removing those changes.

seletskiy:
@pleshakov: Changed style back. My only hope is that you understand, that current code style is very unwelcoming for contributors.

pleshakov:
@seletskiy thx! 

It works as expected! However, the implementation can be simplified:
* `EndpointsInfo map[string][]EndpointInfo` can be transformed into `EndpointsInfo map[string]EndpointInfo`. The key will be IP:Port.
* `EndpointInfo` no longer needs to have the `Address` field.
* The map with endpoints and a namespace is redundant, since the namespace of the endpoints of an Ingress resource will be the same namespace as the one of the Ingress resource.  
* In `getMappedVariableName`, I don't see why a namespace or a ingress must `regexp.MustCompile(`\W`)`

Ideally, we should have a style guide for contributors. However, it is generally expected that a PR doesn't introduce any style changes. Inconsistent style makes the code hard to read and maintain.

Additionally, there are still a few problems related to the style:
* There is a lot of unnecessary line wrapping. For example:
```
return fmt.Sprintf(
 +		"%s/%s",
 +		ingress.Namespace,
 +		ingress.Name,
 +	)
```
* Also, when a line gets too long either in a function declaration or invocation,  it is enough to split a line into two lines,  instead of wrapping each argument. For example, `NewNginxController` in `nginx.go`.

seletskiy:
> In getMappedVariableName, I don't see why a namespace or a ingress must regexp.MustCompile(\W)

It's used in `escape` function to prevent non-word characters to be used as variable name in the nginx config, because namespace/ingress name is used as part of variable name.

@pleshakov: I appreciate your feedback, but as I told before I can't spent more time on pushing changes to this PR and if you will not merge it, I will continue to use my own fork. So, please, if you find this PR useful and want to give your users it's functionality, but something is undesirable for you, just edit it by yourself. You can checkout this PR as branch locally: https://help.github.com/articles/checking-out-pull-requests-locally/, add new commits, merge your local branch, then this PR will be marked as merged automatically.

pleshakov:
@seletskiy thanks! Please keep the PR open.

seletskiy:
@pleshakov: Any plans to merge this?

Soon I will be working on implementing another useful feature — https://github.com/nginxinc/kubernetes-ingress/issues/164 and most probably I will do it on top of my fork since I don't want to resolve conflicts between two versions of code.

pleshakov:
@seletskiy 
We're not planning to merge this PR any time soon as items with a higher priority are planned for Jan and February. The PR requires some amount of additional work on top of your work to be done before merging it. 

However, looking into the current conflicts, it doesn't look like there are many conflicts for now and the conflicts are only in configurator.go.

It'd be interesting to see your implementation of TCP as well as your use case. That feature is a popular request.

nginx-controller/nginx/templates/nginx.tmpl
```diff
@@ -9,7 +9,7 @@ error_log  /var/log/nginx/error.log warn;
 pid        /var/run/nginx.pid;
 
 events {
-    worker_connections  1024;
+    worker_connections  {{.WorkerConnections}};
```
kovetskiy: 1024 is enough for everybody!

pleshakov:
closing this PR, as there are no plans to merge it. thx.

Issue #198, opened 2017-10-06
Allow HTTP access when TLS is defined.


lseelenbinder:
I have a service I'd like to expose over HTTP & HTTPS, but the only way to do that currently with this ingress controller is by defining a second ingress that doesn't have TLS defined on it. Is there a way to add a variable/annotation to prevent a universal HTTP -> HTTPS redirect when TLS is enabled?

I'm open to adding the patch, but I'd need some direction to get me started.

pleshakov:
@lseelenbinder We don't have an annotation for that.

 Having the annotation makes perfect sense. There is a standard annotation `ingress.kubernetes.io/ssl-redirect` that takes a boolean argument, which is common among Ingress controllers. The default is `true`.

An example of how to add an annotation can be found here -- https://github.com/nginxinc/kubernetes-ingress/commit/714fd9945b1ff3a62c00951aa7bb9112da77ed2c There is a function [here](https://github.com/nginxinc/kubernetes-ingress/blob/master/nginx-controller/nginx/convert.go#L20) that parses boolean annotations/configmap keys.

lseelenbinder:
Thank you, @pleshakov. I will see what I can do!

PR #199, opened 2017-10-07
Add support for Kubernetes ssl-redirect annotation.

lseelenbinder:
As requested in #198.

I was not certain the acceptable way to test this, so please advise; `make test` does succeed.


pleshakov:
@lseelenbinder 
Thanks!

There is a small problem: Using `$server.sslRedirect` in the templates leads to a failure. The correct field  is `$server.SSLRedirect`.

It is possible to test it, if you add the following lines into the Server [here](https://github.com/nginxinc/kubernetes-ingress/blob/master/nginx-controller/nginx/templates/templates_test.go#L27):
```go
			SSL:               true,
			SSLCertificate:    "secret.pem",
			SSLCertificateKey: "secret.pem",
			SSLPorts:          []int{443},
			SSLRedirect:       true,
```
Without changing to `$server.SSLRedirect`, the test will fail.

Other than that, it works as expected!

Could you fix the problem, change the test and commit the all changes? Please squash all commits into a single one.

lseelenbinder:
Done, @pleshakov. Let me know if there's anything else!

pleshakov:
@lseelenbinder thx!

PR #208, opened 2017-11-01
Add worker-shutdown-timeout configmap key

Sarga:


pleshakov:
@Sarga thx! Looks good!

* Could you set the `WorkerShutdownTimeout` with some value [here](https://github.com/nginxinc/kubernetes-ingress/blob/master/nginx-controller/nginx/templates/templates_test.go#L54) in the `templates_test.go`? The tests in that file make sure that the templates are executed without any errors.
* Could you squash your commits into a single one?

Sarga:
@pleshakov done. Thanks for the advice!

PR #223, opened 2018-01-02
Add worker-connections configmap key

Sarga:


pleshakov:
@Sarga thanks! we'll release version 1.1.1 with this feature as well as with updating NGINX  to 1.13.8 this week.

PR #236, opened 2018-02-09
Always respect `-ingress-class` option

nicknovitski:
If I understand this code correctly, when `-ingress-class` is set to `foo` and `-use-ingress-class-only` is false, the controller only controls ingresses with the class `nginx`, not foo.


pleshakov:
@nicknovitski 
thanks for finding and fixing the bug and simplifying the code!

pleshakov:
Added a unit test for that particular function https://github.com/nginxinc/kubernetes-ingress/pull/237

PR #245, opened 2018-02-15
RBAC clarification

rayfoss:
It is rarely optional anymore.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -31,7 +31,7 @@ The installation manifests are located in the [install](../install) folder. In t
 
 ## 2. Configure RBAC
 
-If RBAC is enabled in your cluster, create a cluster role and bind it to the service account, created in Step 1:
+RBAC is enabled by default since v1.6. If it is enabled on your cluster, create a cluster role and bind it to the service account, created in Step 1:
 ```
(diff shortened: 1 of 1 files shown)


pleshakov:
@rayfoss 
thanks for the PR!

While RBAC was introduced in 1.6 (in beta), it wasn't enabled by default in many Kubernetes platforms. For example, on GKE, RBAC is enabled since 1.8 The existing wording in line 34 is OK for now.

Regarding the GKE, it is better to have a separate note below the note about being a cluster admin:
**Note about GKE**: To grant yourself the cluster admin permissions, run `kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user=$(gcloud config get-value core/account)`

In line 99, I think it is better to move `randomly` to the first sentence: Kubernetes will *randomly*  allocate two ports

rayfoss:
i take it you meant erase, alright...

Dean-Coakley:
@rayfoss I meant raise, as in open a new PR with requested changes.

Feel free to reopen and apply requested changes. Otherwise I'll take this over tomorrow.
Apologies for the misunderstanding.

PR #248, opened 2018-02-22
[readme] use right aws yaml to create loadbalancer

zariye:
just a little mistake in the readme, the file loadbalancer-aws.yaml does not exist.


pleshakov:
@zariye thank you

Issue #250, opened 2018-02-28
Slow performance with many site mirrors


damour:
We have many Ingresses with site mirrors (all refers to same service, only different ssl domains):
- site1.com
- site1.biz
- site1.net
etc...

Nginx configs applies very slow (4-5 minutes for 200 mirrors) with increased memory usage. Also websocket connections often drops. 

Possible solution: reload nginx once when configs for all endpoints generated.

pleshakov:
@damour thanks for the PR. that logic is not optimized for this case. We'll review and test it shortly.

However, as an interim or an alternative solution, is it possible in your case to have one Ingress resource that includes all those hostnames? In such case, a change of the endpoints will only affect one Ingress resource.

damour:
Of course. If there is a way to specify mirrors with https in one ingress resource, this would be ideal.

Sarga:
We have ssl wildcard certificates *.domain.com for every domain
To specify all the domains in the server_name of a single server {} block we need SAN SSL (multi domain ssl), which is not possible in our case.
Also some of the domains might have different redirects with different conditions. Therefore such is easier to be done in a separate server {} blocks

pleshakov:
@damour 
Does the example below address your use case?

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: cafe-ingress
  annotations:
     kubernetes.io/ingress.class: "nginx"
spec:
  tls:
  - hosts:
    - cafe.example.org
    secretName: cafe-secret-org
  - hosts:
    - cafe.example.com
    secretName: cafe-secret-com
  rules:
  - host: "cafe.example.com"
    http:
      paths:
      - path: /
        backend:
          serviceName: cafe-svc
          servicePort: 80
  - host: "cafe.example.org"
    http:
      paths:
      - path: /
        backend:
          serviceName: cafe-svc
          servicePort: 80
```

Sarga:
@pleshakov 

We’ve followed your advice, but there is a problem. 
Our service has 476 domains and the same number of ssl certificates
After we have redeploy our service, the ingress controller updates the configuration of upstreams and registers the new IP addresses of our PODs in at least 1 minute and 36 seconds. As a result, the traffic for a long time goes to the IP on which our PODs no longer exist.
[...]

PR #251, opened 2018-02-28
Reload nginx after all endpoints update

damour:
Fix https://github.com/nginxinc/kubernetes-ingress/issues/250


pleshakov:
@damour thanks the PR! 

I've prepared a PR https://github.com/nginxinc/kubernetes-ingress/pull/254/ that solves this issue in slightly different way and makes sure it covers both NGINX and NGINX Plus.  I will close this PR and merge the other one.

PR #252, opened 2018-03-01
Remove Broken Links

cpretzer:
There are links to files which have been removed and this change removes those links.


pleshakov:
@cpretzer thanks!

PR #258, opened 2018-03-23
Handle annotations and conflicting paths for MergeableTypes

### LINKED ISSUE #264: Mergeable Ingress Types ignore root "/" paths
The new Mergeable Ingress Types feature works beautifully, however while testing it out we encountered one gotchya. It seems to explicitly silently ignore root paths and we are curious if that can be safely changed.

# Background

We are utilizing an Ingress with multiple paths defined with one path as only the root (`/`) and others as subpaths from there. Within a single Ingress it works well so that other defined paths go to their specific services and any other paths fall back to that service with the root path. Acting like a transparent fallback if you will.

With Mergeable Ingress Types in effect, the defined paths all come through ok in the generated nginx configuration file but the root path's Location block was missing and attempts to navigate to it resulted in 404s and in the nginx Ingress controller's logs it appears to actually try to browse to the URL through its local `/etc/nginx/html` directory.

```
2018/03/30 00:41:57 [error] 327#327: *594 open() "/etc/nginx/html/testing-api/v1/schema" failed (2: No such file or directory), client: 10.22.122.237, server: xxx.branch-testing.lab.k8s.xxx.com, request: "GET /testing-api/v1 HTTP/1.1", host: "xxx.branch-testing.lab.k8s.xxx.com"
10.22.122.237 - - [30/Mar/2018:00:41:57 +0000] "GET /testing-api/v1/schema HTTP/1.1" 404 564 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/65.0.3325.181 Safari/537.36" "-"
```

# Expected Behavior

We expected the root route to have a corresponding `Location` block in the generated nginx configuration file.

# Proposal

We propose altering the conditional in [configurator.go#L112-L115](https://github.com/nginxinc/kubernetes-ingress/blob/master/nginx-controller/nginx/configurator.go#L112-L115):

https://github.com/nginxinc/kubernetes-ingress/blob/6a8bff81f7bf7a74ccfd1e9e044fcd537792d91c/nginx-controller/nginx/configurator.go#L112-L115

We were able to compile with it commented out and things are working as we expected.

Understandably the conditional was added for a reason though, so we are curious to learn more and figure out the best way to move forward.

We tried to think through why it might have been added and the best we could think of is that it is to prevent two minion Ingresses from both using `/`, but that same problem still exists if two minion Ingress try to both claim other paths such as `/anything`. Perhaps we could allow usage of `/` but add a warning log line? Or only let one in and ignore the others like we do now but with a warning? Or perhaps there is a danger in using the root path we aren't aware of?

We can open an MR with those lines removed but felt opening up dialogue first would be better.

diazjf:
Adds new rules for the MergeableTypes. Minions will not be able to have conflicting locations, and can only have service level annotations. Masters will only be able to have host level annotations.

Fixes #264

### DIFF
--- a/examples/mergable-ingress-types/README.md
+++ b/examples/mergable-ingress-types/README.md
@@ -10,9 +10,32 @@ host level, which includes the TLS configuration, and any annotations which will
 can only be one ingress resource on a unique host that contains the master value. Paths cannot be part of the
 ingress resource.
 
+Masters cannot contain the following annotations:
+* nginx.org/rewrites
+* nginx.org/ssl-services
+* nginx.org/websocket-services
+* nginx.com/sticky-cookie-services
+
 A Minion is declared using `nginx.org/mergible-ingress-type: minion`. A Minion will be used to append different
-locations to an ingress resource with the Master value. The annotations of minions are replaced with the annotations of
-their master. TLS configurations are not allowed. There can be multiple minions which must have the same host as the master.
+locations to an ingress resource with the Master value. TLS configurations are not allowed. Multiple minions can be
+applied per master as long as they do not have conflicting paths. If a conflicting path is present then the path defined
+on the oldest minion will be used.
+
+Minions cannot contain the following annotations:
+* nginx.org/proxy-hide-headers
+* nginx.org/proxy-pass-headers
+* nginx.org/redirect-to-https
+* ingress.kubernetes.io/ssl-redirect
+* nginx.org/hsts
+* nginx.org/hsts-max-age
+* nginx.org/hsts-include-subdomains
+* nginx.org/server-tokens
+* nginx.org/listen-ports
+* nginx.org/listen-ports-ssl
+* nginx.com/jwt-key
+* nginx.com/jwt-realm
+* nginx.com/jwt-token
+* nginx.com/jwt-login-url
 
 Note: Ingress Resources with more than one host cannot be used.
 
--- a/nginx-controller/controller/controller.go
+++ b/nginx-controller/controller/controller.go
@@ -37,6 +37,7 @@ import (
 	api_v1 "k8s.io/api/core/v1"
 	extensions "k8s.io/api/extensions/v1beta1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sort"
 )
 
 const (
@@ -1203,7 +1204,14 @@ func (lbc *LoadBalancerController) getMinionsForMaster(master *nginx.IngressEx) 
 		return []*nginx.IngressEx{}, err
 	}
 
+	// ingresses are sorted by creation time
+	sort.Slice(ings.Items[:], func(i, j int) bool {
+		return ings.Items[i].CreationTimestamp.Time.UnixNano() < ings.Items[j].CreationTimestamp.Time.UnixNano()
+	})
+
 	var minions []*nginx.IngressEx
+	var minionPaths = make(map[string]*extensions.Ingress)
+
 	for i, _ := range ings.Items {
 		if !lbc.isNginxIngress(&ings.Items[i]) {
 			continue
@@ -1222,6 +1230,20 @@ func (lbc *LoadBalancerController) getMinionsForMaster(master *nginx.IngressEx) 
 			glog.Errorf("Ingress Resource %v/%v with the 'nginx.org/mergible-ingress-type' annotation set to 'minion' must contain a Path", ings.Items[i].Namespace, ings.Items[i].Name)
 			continue
 		}
+
+		uniquePaths := []extensions.HTTPIngressPath{}
+		for _, path := range ings.Items[i].Spec.Rules[0].HTTP.Paths {
+			if val, ok := minionPaths[path.Path]; ok {
+				glog.Errorf("Ingress Resource %v/%v with the 'nginx.org/mergible-ingress-type' annotation set to 'minion' cannot contain the same path as another ingress resource, %v/%v.",
+					ings.Items[i].Namespace, ings.Items[i].Name, val.Namespace, val.Name)
+				glog.Errorf("Path %s for Ingress Resource %v/%v will be ignored", path.Path, val.Namespace, val.Name)
+			} else {
+				minionPaths[path.Path] = &ings.Items[i]
+				uniquePaths = append(uniquePaths, path)
+			}
+		}
+		ings.Items[i].Spec.Rules[0].HTTP.Paths = uniquePaths
+
 		ingEx, err := lbc.createIngress(&ings.Items[i])
 		if err != nil {
 			glog.Errorf("Error creating ingress resource %v/%v: %v", ingEx.Ingress.Namespace, ingEx.Ingress.Name, err)
@@ -1233,6 +1255,7 @@ func (lbc *LoadBalancerController) getMinionsForMaster(master *nginx.IngressEx) 
 		}
 		minions = append(minions, ingEx)
 	}
+
 	return minions, nil
 }
 
--- a/nginx-controller/nginx/configurator.go
+++ b/nginx-controller/nginx/configurator.go
@@ -83,6 +83,13 @@ func (cnf *Configurator) addOrUpdateMergableIngress(mergeableIngs *MergeableIngr
 	var locations []Location
 	var upstreams []Upstream
 	var keepalive string
+	var removedAnnotations []string
+
+	removedAnnotations = filterMasterAnnotations(mergeableIngs.Master.Ingress.Annotations)
+	if len(removedAnnotations) != 0 {
+		glog.Errorf("Ingress Resource %v/%v with the annotation 'nginx.com/mergeable-ingress-type' set to 'master' cannot contain the '%v' annotation(s). They will be ignored",
+			mergeableIngs.Master.Ingress.Namespace, mergeableIngs.Master.Ingress.Name, strings.Join(removedAnnotations, ","))
+	}
 
 	pems, jwtKeyFileName := cnf.updateSecrets(mergeableIngs.Master)
 	masterNginxCfg := cnf.generateNginxCfg(mergeableIngs.Master, pems, jwtKeyFileName)
@@ -101,20 +108,28 @@ func (cnf *Configurator) addOrUpdateMergableIngress(mergeableIngs *MergeableIngr
 
 	minions := mergeableIngs.Minions
 	for _, minion := range minions {
+		// Remove the default backend so that "/" will not be generated
+		minion.Ingress.Spec.Backend = nil
+
+		// Add acceptable master annotations to minion
+		mergeMasterAnnotationsIntoMinion(minion.Ingress.Annotations, mergeableIngs.Master.Ingress.Annotations)
+
+		removedAnnotations = filterMinionAnnotations(minion.Ingress.Annotations)
+		if len(removedAnnotations) != 0 {
+			glog.Errorf("Ingress Resource %v/%v with the annotation 'nginx.com/mergeable-ingress-type' set to 'minion' cannot contain the %v annotation(s). They will be ignored",
+				minion.Ingress.Namespace, minion.Ingress.Name, strings.Join(removedAnnotations, ","))
+		}
+
 		pems, jwtKeyFileName := cnf.updateSecrets(minion)
 		nginxCfg := cnf.generateNginxCfg(minion, pems, jwtKeyFileName)
 
-		// Replace all minion annotations with master annotations
-		minion.Ingress.Annotations = mergeableIngs.Master.Ingress.Annotations
-
 		for _, server := range nginxCfg.Servers {
 			for _, loc := range server.Locations {
-				if loc.Path != "/" {
-					loc.IngressResource = objectMetaToFileName(&minion.Ingress.ObjectMeta)
-					locations = append(locations, loc)
-				}
+				loc.IngressResource = objectMetaToFileName(&minion.Ingress.ObjectMeta)
+				locations = append(locations, loc)
 			}
 		}
+
 		for _, val := range nginxCfg.Upstreams {
 			upstreams = append(upstreams, val)
 		}
@@ -764,6 +779,48 @@ func (cnf *Configurator) updatePlusEndpoints(ingEx *IngressEx) error {
 	}
 
 	return nil
+}
+
+func filterMasterAnnotations(annotations map[string]string) []string {
+	var removedAnnotations []string
+
+	for _, blacklistAnn := range masterBlacklist {
+		if _, ok := annotations[blacklistAnn]; ok {
+			removedAnnotations = append(removedAnnotations, blacklistAnn)
+			delete(annotations, blacklistAnn)
+		}
+	}
+
+	return removedAnnotations
+}
+
+func filterMinionAnnotations(annotations map[string]string) []string {
+	var removedAnnotations []string
+
+	for _, blacklistAnn := range minionBlacklist {
+		if _, ok := annotations[blacklistAnn]; ok {
+			removedAnnotations = append(removedAnnotations, blacklistAnn)
+			delete(annotations, blacklistAnn)
+		}
+	}
+
+	return removedAnnotations
+}
+
+func mergeMasterAnnotationsIntoMinion(minionAnnotations map[string]string, masterAnnotations map[string]string) {
+	for key, val := range masterAnnotations {
+		isBlacklisted := false
+		if _, ok := minionAnnotations[key]; !ok {
+			for _, blacklistAnn := range minionBlacklist {
+				if blacklistAnn == key {
+					isBlacklisted = true
+				}
+			}
+			if !isBlacklisted {
+				minionAnnotations[key] = val
+			}
+		}
+	}
 }
 
 // UpdateConfig updates NGINX Configuration parameters
--- a/nginx-controller/nginx/configurator_test.go
+++ b/nginx-controller/nginx/configurator_test.go
@@ -1,6 +1,7 @@
 package nginx
 
 import (
+	"reflect"
 	"testing"
 )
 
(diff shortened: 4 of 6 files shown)


diazjf:
@pleshakov just a WIP of the annotation and conflicting path rules. We can talk sometime next week. Have a good weekend!

diazjf:
Note: I still need to add the logic to make sure there are not other ingress resources(non-master or minion) which conflict with a master or minion.

nginx-controller/controller/controller.go
```diff
@@ -1204,6 +1204,7 @@ func (lbc *LoadBalancerController) getMinionsForMaster(master *nginx.IngressEx)
 	}
 
 	var minions []*nginx.IngressEx
+IngressLoop:
```
diazjf: Instead of using the label we can also set a boolean flag to continue if a certain annotation is found or if there are multiple paths, but using a label is quicker since we won't have to wait for the loop to end.

nginx-controller/controller/controller.go
```diff
@@ -1222,6 +1223,30 @@ func (lbc *LoadBalancerController) getMinionsForMaster(master *nginx.IngressEx)
 			glog.Errorf("Ingress Resource %v/%v with the 'nginx.org/mergible-ingress-type' annotation set to 'minion' must contain a Path", ings.Items[i].Namespace, ings.Items[i].Name)
 			continue
 		}
+
+		for key, val := range ings.Items[i].Annotations {
+			if key == "nginx.org/mergible-ingress-type" || !strings.Contains(key, "nginx.org/") {
+				continue
+			}
+			if !strings.Contains(val, "serviceName=") {
```
diazjf: I was thinking minion annotations should only contain annotations where serviceName is applied.

We can make "serviceName=" a global since it is used in multiple places.

nginx-controller/controller/controller.go
```diff
@@ -1222,6 +1223,30 @@ func (lbc *LoadBalancerController) getMinionsForMaster(master *nginx.IngressEx)
...
+				glog.Errorf("Ingress Resource %v/%v with the 'nginx.org/mergible-ingress-type' annotation set to 'minion' cannot contain any host level annotations",
+					ings.Items[i].Namespace, ings.Items[i].Name)
+				continue IngressLoop
+			}
+		}
+
+		for _, path := range ings.Items[i].Spec.Rules[0].HTTP.Paths {
```
diazjf: Here we will skip any minion with the same path as another minion. Therefore only 1 of the 2 minions will be applied.

Is there any type of order used when listing ingresses using `lbc.ingLister.List()`? Should I apply some type of sorting after the list is generated, so that we know it will always be the first successful minion that will be active?

nginx-controller/controller/controller.go
```diff
@@ -1271,6 +1297,17 @@ func (lbc *LoadBalancerController) createMergableIngresses(master *extensions.In
 		return &mergeableIngresses, err
 	}
 
+	for key, val := range master.Annotations {
+		if key == "nginx.org/mergible-ingress-type" || !strings.Contains(key, "nginx.org/") {
+			continue
+		}
+		if strings.Contains(val, "serviceName=") {
```
diazjf: Master configuration should be able to take any annotation that is not only for a specific location.

nginx-controller/controller/controller.go
pleshakov: I think we should handle most of the annotations that could be applied per a location block in the NGINX config or  per an upstream (service). Such a handling enables new use cases.

For example, consider an Ingress resource with multiple paths and multiple services. If we want to choose a load balancing method, other than the default round robin, we need to use `nginx.org/lb-method` annotation. However, this annotation applies to all services in the Ingress resource. What if we want to change the lb method only for one service. To overcome this limitation, we could potentially change the `nginx.org/lb-method` to support specifying a service, similar to what we did for [session persistence](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/session-persistence). However, with mergible Ingresses, we can split the Ingress resource into multiple ones and in each minion Ingress resource have only one path/service and configure the lb method using the `nginx.org/lb-method` annotation for each service.

Another example is `nginx.org/location-snippets` annotation.

diazjf: That works for me. So for example:

Minion_1 contains `/tea` and `/coffee`. If we want to change the load-balancing for both then we apply `nginx.org/lb-method` which configures both. Now if we wanted it only for `/coffee` then we split it up into 2 minions and apply it only to the 1 with `/coffee`.

I can remove the `serviceName` check and then apply all the master's annotations to the minions and then from each minion apply only their specific annotations overwriting the masters. All within the functions which generates the Nginx Configuration.

nginx-controller/nginx/ingress.go
```diff
@@ -15,6 +15,32 @@ type IngressEx struct {
 }
 
 type MergeableIngresses struct {
-	Master *IngressEx
+	Master  *IngressEx
 	Minions []*IngressEx
 }
+
+var mergeableIngressAnnotationBlacklist = map[string][]string{
```
pleshakov: I think it is better to have two maps of map[string]bool for two black lists -- the master one and the minion one.

nginx-controller/controller/controller.go
```diff
@@ -1243,6 +1264,11 @@ func (lbc *LoadBalancerController) findMasterForMinion(minion *extensions.Ingres
 		return &extensions.Ingress{}, err
 	}
 
+	// ingresses are sorted by creation time
```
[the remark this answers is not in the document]
pleshakov: Let's skip this check for now. Currently, in the Ingress controller we don't handle the case when there are multiple regular Ingresses for the same host. I think once implemented, the logic for handling that conflict would also work for multiple masters for the same host.

nginx-controller/controller/controller.go
```diff
@@ -1222,6 +1229,19 @@ func (lbc *LoadBalancerController) getMinionsForMaster(master *nginx.IngressEx)
 			glog.Errorf("Ingress Resource %v/%v with the 'nginx.org/mergible-ingress-type' annotation set to 'minion' must contain a Path", ings.Items[i].Namespace, ings.Items[i].Name)
 			continue
 		}
+
+		for _, path := range ings.Items[i].Spec.Rules[0].HTTP.Paths {
+			for _, minion := range minions {
```
pleshakov: to check for map conflicts, it is better to use a map[string]*nginx.IngressEx, where the map key is a path, rather iterating over an array of minions.

nginx-controller/controller/controller.go
```diff
@@ -1222,6 +1229,19 @@ func (lbc *LoadBalancerController) getMinionsForMaster(master *nginx.IngressEx)
...
+				for _, minionPath := range minion.Ingress.Spec.Rules[0].HTTP.Paths {
+					if path.Path == minionPath.Path {
+						glog.Errorf("Ingress Resource %v/%v with the 'nginx.org/mergible-ingress-type' annotation set to 'minion' cannot contain the same path as another ingress resource, %v/%v",
+							ings.Items[i].Namespace, ings.Items[i].Name, minion.Ingress.Namespace, minion.Ingress.Name)
+						continue IngressLoop
+					}
```
pleshakov: In this case, if an Ingress resource has a conflicting path and is not the oldest Ingress resource, the Ingress resource is completely Ignored. Maybe it is better to not ignore it, but only ignore that conflicting path?

nginx-controller/nginx/ingress.go
diazjf: To keep it simple, can we just add a `[]string`

nginx-controller/controller/controller.go
diazjf: I agree, we can just log an error and not include the newest path. The configuration will also contain which Ingress Resource generates which path. This can also aid the user in debugging. I can also add more instructions on this feature.

nginx-controller/controller/controller.go
diazjf: Using map[sting]*extensions.Ingress, where key is the path, and value is the minion Ingress.

examples/mergable-ingress-types/README.md
```diff
@@ -10,9 +10,33 @@ host level, which includes the TLS configuration, and any annotations which will
...
+* nginx.org/rewrites
+* nginx.org/ssl-services
+* nginx.org/websocket-services
+* nginx.com/sticky-cookie-services
+
 A Minion is declared using `nginx.org/mergible-ingress-type: minion`. A Minion will be used to append different
-locations to an ingress resource with the Master value. The annotations of minions are replaced with the annotations of
-their master. TLS configurations are not allowed. There can be multiple minions which must have the same host as the master.
+locations to an ingress resource with the Master value. TLS configurations are not allowed. Multiple minions can be
+applied per master as long as they do not have conflicting paths. Any conflicting paths will be ignored. Also note that the
```
pleshakov: It is worth mentioning that in case of several minions having the same path, the path defined in the oldest minion will be used.

nginx-controller/controller/controller.go
```diff
@@ -1243,6 +1266,11 @@ func (lbc *LoadBalancerController) findMasterForMinion(minion *extensions.Ingres
 		return &extensions.Ingress{}, err
 	}
 
+	// ingresses are sorted by creation time
+	sort.Slice(ings.Items[:], func(i, j int) bool {
```
pleshakov: sorting ingresses is unnecessary here, as we will not check for duplicated masters for now

nginx-controller/nginx/configurator.go
```diff
@@ -84,6 +84,16 @@ func (cnf *Configurator) addOrUpdateMergableIngress(mergeableIngs *MergeableIngr
 	var upstreams []Upstream
 	var keepalive string
 
+	for _, blacklistAnn := range masterBlacklist {
```
pleshakov: could we have a function for filtering the master annotations?

nginx-controller/nginx/configurator.go
```diff
@@ -101,20 +111,36 @@ func (cnf *Configurator) addOrUpdateMergableIngress(mergeableIngs *MergeableIngr
 
 	minions := mergeableIngs.Minions
 	for _, minion := range minions {
+
+		for _, blacklistAnn := range minionBlacklist {
```
pleshakov: could we have a function for filtering minion annotations?

nginx-controller/nginx/configurator.go
```diff
@@ -84,6 +84,16 @@ func (cnf *Configurator) addOrUpdateMergableIngress(mergeableIngs *MergeableIngr
 	var upstreams []Upstream
 	var keepalive string
 
+	for _, blacklistAnn := range masterBlacklist {
+		for _, masterAnn := range mergeableIngs.Master.Ingress.Annotations {
```
pleshakov: would it be more efficient if we iterate over all master annotations and look up if a particular annotation is in the black list using a map?

nginx-controller/nginx/configurator.go
pleshakov: would it be more efficient if we iterate over all minion annotations and look up if a particular annotation is in the black list using a map?

pleshakov:
@diazjf could we also remove the default backend in minions if it is present? Otherwise, if a minion has the default backend, the configurator will generate the "/" for it. Consider the example below:
```yaml
kind: Ingress
metadata:
  name: minon-ingress
  annotations:
   . . .
spec:
  backend:
    serviceName: testsvc
    servicePort: 80
. . .
```

examples/mergable-ingress-types/README.md
diazjf: Done

nginx-controller/controller/controller.go
diazjf: Done

nginx-controller/nginx/configurator.go
diazjf: Done

nginx-controller/nginx/configurator.go
diazjf: I optimized it, since annotations are already in a map[string]string, I just look over all the blacklist items and if any are in the master they are removed. Looking through the blacklist since its small and there could be many annotations in an Ingress Resource.

nginx-controller/nginx/configurator.go
diazjf: I optimized it, since annotations are already in a map[string]string, I just look over all the blacklist items and if any are in the master they are removed. Looking through the blacklist since its small and there could be many annotations in an Ingress Resource.

pleshakov:
Could you also remove `"kubernetes.io/ingress.class"` from the minion black list?

nginx-controller/nginx/configurator.go
```diff
@@ -766,6 +786,45 @@ func (cnf *Configurator) updatePlusEndpoints(ingEx *IngressEx) error {
...
+	var minionOverwrites []string
+
+	for key, val := range masterAnnotations {
+		if _, ok := minionAnnotations[key]; !ok {
+			minionAnnotations[key] = val
+			minionOverwrites = append(minionOverwrites, key)
+		}
+	}
+
+	return minionAnnotations, minionOverwrites
```
pleshakov: I don't think we need to return minionOverwrites.
Moreover, this function should not return anything, because the map minionAnnotations is passed by reference and will be changed by the function.

nginx-controller/nginx/configurator.go
```diff
@@ -766,6 +786,45 @@ func (cnf *Configurator) updatePlusEndpoints(ingEx *IngressEx) error {
...
+		if _, ok := annotations[blacklistAnn]; ok {
+			removedAnnotations = append(removedAnnotations, blacklistAnn)
+			delete(annotations, blacklistAnn)
+		}
+	}
+
+	return annotations, removedAnnotations
+}
+
+func filterMinionAnnotations(annotations map[string]string) (map[string]string, []string) {
```
pleshakov: we don't need to return a map, only the removedAnnotations.

nginx-controller/nginx/configurator.go
```diff
@@ -101,20 +109,32 @@ func (cnf *Configurator) addOrUpdateMergableIngress(mergeableIngs *MergeableIngr
 
 	minions := mergeableIngs.Minions
 	for _, minion := range minions {
+		// Remove the default backend so that "/" will not be generated
+		minion.Ingress.Spec.Backend = nil
+
+		// Add acceptable master annotations to minion
+		minion.Ingress.Annotations, masterOverwrites = mergeMasterAnnotationsIntoMinion(minion.Ingress.Annotations, mergeableIngs.Master.Ingress.Annotations)
+		if len(masterOverwrites) != 0 {
```
pleshakov: no need to report overwritten annotations

nginx-controller/nginx/configurator.go
```diff
@@ -766,6 +786,45 @@ func (cnf *Configurator) updatePlusEndpoints(ingEx *IngressEx) error {
 	return nil
 }
 
+func filterMasterAnnotations(annotations map[string]string) (map[string]string, []string) {
```
pleshakov: we don't need to return a map, only the removedAnnotations.

diazjf:
@pleshakov Done! And updated the unit tests. Thanks again.

nginx-controller/nginx/configurator.go
```diff
@@ -766,6 +781,43 @@ func (cnf *Configurator) updatePlusEndpoints(ingEx *IngressEx) error {
...
+	}
+
+	return removedAnnotations
+}
+
+func mergeMasterAnnotationsIntoMinion(minionAnnotations map[string]string, masterAnnotations map[string]string) {
+	var minionOverwrites []string
+
+	for key, val := range masterAnnotations {
+		if _, ok := minionAnnotations[key]; !ok {
```
pleshakov: we also need to check that an annotation is not in the minion black list as well. If we don't do that, then an non-supported annotation will be merged into a minion. That will lead to the filterMinionAnnotations function returning that annotation in its removedAnnotations list. That will lead to an error message in the log, which should not happen!

nginx-controller/nginx/configurator.go
```diff
@@ -766,6 +781,43 @@ func (cnf *Configurator) updatePlusEndpoints(ingEx *IngressEx) error {
...
+			removedAnnotations = append(removedAnnotations, blacklistAnn)
+			delete(annotations, blacklistAnn)
+		}
+	}
+
+	return removedAnnotations
+}
+
+func mergeMasterAnnotationsIntoMinion(minionAnnotations map[string]string, masterAnnotations map[string]string) {
+	var minionOverwrites []string
```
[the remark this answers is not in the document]
diazjf: Good catch

nginx-controller/nginx/configurator.go
diazjf: Good catch as well! Thanks.

nginx-controller/nginx/configurator_test.go
```diff
@@ -61,3 +62,86 @@ func TestParseStickyServiceInvalidFormat(t *testing.T) {
...
+
+	if !reflect.DeepEqual(expectedfilteredMinionAnnotations, minionAnnotations) {
+		t.Errorf("filterMinionAnnotations returned %v, but expected %v", minionAnnotations, expectedfilteredMinionAnnotations)
+	}
+	if !reflect.DeepEqual(expectedRemovedAnnotations, removedAnnotations) {
+		t.Errorf("filterMinionAnnotations returned %v, but expected %v", removedAnnotations, expectedRemovedAnnotations)
+	}
+}
+
+func TestMergeMasterAnnotationsIntoMinion(t *testing.T) {
```
pleshakov: Since we also check if a master annotation is in the minion black list before merging, could you test thing logic in the unit test here?

nginx-controller/nginx/configurator.go
```diff
@@ -766,6 +781,49 @@ func (cnf *Configurator) updatePlusEndpoints(ingEx *IngressEx) error {
...
+
+	return removedAnnotations
+}
+
+func mergeMasterAnnotationsIntoMinion(minionAnnotations map[string]string, masterAnnotations map[string]string) {
+	var isBlacklisted bool
+
+	for key, val := range masterAnnotations {
+		if _, ok := minionAnnotations[key]; !ok {
+			for _, blacklistAnn := range minionBlacklist {
```
pleshakov: the check if an annotation is in the black list has a bug -- `isBlacklisted` must be initialized to `false` prior to each iteration of the loop.

Instead of iterating over the minionBlacklist, I suggest to change the minionBlacklist from `[]string` to `map[string]bool` for performance reason. The code will look like this:
```golang
	for key, val := range masterAnnotations {
		if _, ok := minionAnnotations[key]; !ok && !minionBlacklist[key] {
			minionAnnotations[key] = val
		}
	}
```

nginx-controller/nginx/configurator_test.go
diazjf: Done!

nginx-controller/nginx/configurator.go
diazjf: I can change to maps in another PR if that ok.

pleshakov: sounds good!

pleshakov:
@diazjf thank you!!

PR #284, opened 2018-05-17
Correct Logs for Mergeable Types with Duplicate Location

diazjf:
The logs displayed, showed the wrong ingress resource as the ignored ingress resource when a duplicate location was setup


pleshakov:
@diazjf thanks!!

PR #285, opened 2018-05-22
Add support for active health checks for Plus

isaachawley:
NGINX Plus supports active health checks.

To use, set the annotation `nginx.com/health-checks: "true"`. The Ingress Controller will pull health check (Readiness Probe) information from the Kubernetes API and configure matching active health checks in NGINX Plus


isaachawley:
Merge problem, closing for now.

PR #311, opened 2018-07-13
Support JWT auth in mergeable minions

pleshakov:


nginx-controller/controller/controller.go
```diff
@@ -661,50 +722,93 @@ func (lbc *LoadBalancerController) syncSecret(task Task) {
...
+					lbc.recorder.Eventf(&minion, api_v1.EventTypeNormal, "Updated", "Configuration for %v/%v was updated", minion.Namespace, minion.Name)
+				}
 			}
 		}
 	}
 }
 
-func (lbc *LoadBalancerController) findIngressesForSecret(secret string) ([]extensions.Ingress, error) {
-	res := []extensions.Ingress{}
+func (lbc *LoadBalancerController) findIngressesForSecret(secret string) ([]extensions.Ingress, []extensions.Ingress, error) {
```
Dean-Coakley: I think we should use named return types so that nonMinions and minions are not mixed up in future

nginx-controller/controller/controller.go
```diff
@@ -661,50 +722,93 @@ func (lbc *LoadBalancerController) syncSecret(task Task) {
...
+					if jwtKey == secret {
+						nonMinions = append(nonMinions, ing)
+					}
+				}
 			}
+			continue
 		}
+
+		// we're dealing with a minion
+		// only JWT secrets are allowed in a minion
```
Dean-Coakley: only JWT secrets are allowed in a minion->JWT secrets are only allowed in a minion

pleshakov: nope, JWT secrets are allowed in both minions and non-minions
however, minions can only have JWT secrets.

PR #331, opened 2018-08-03
Stop modifying minions

isaachawley:
### We need to DeepCopy Ingresses before modifying them

When `-report-ingress-status` is turned on, it makes a k8s API call to [UpdateStatus](https://godoc.org/k8s.io/client-go/kubernetes/typed/extensions/v1beta1#IngressInterface) but unfortunately that call _doesn't_ just update the Status. So we were changing Ingress Annotations.

It's safer in general to DeepCopy items from the store before modifying them.

- [x ] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/master/CONTRIBUTING.md) doc
- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] I have checked that all unit tests pass after adding my changes
- [x ] I have updated necessary documentation
- [ x] I have rebased my branch onto master
- [ x] I will ensure my PR is targeting the master branch and pulling from my branch from my own fork

### DIFF
--- a/nginx-controller/controller/status.go
+++ b/nginx-controller/controller/status.go
@@ -70,15 +70,16 @@ func (su *StatusUpdater) updateIngressWithStatus(ing v1beta1.Ingress, status []a
 		return nil
 	}
 
-	// get a pristine ing with no annotation changes
+	// Get a pristine Ingress from the Store. Required because annotations can be modified
+	// for mergable Ingress objects and the update status API call will update annotations, not just status.
 	key, err := su.keyFunc(&ing)
 	if err != nil {
 		glog.V(3).Infof("error getting key for ing: %v", err)
 		return err
 	}
-	ingCopy, exists, err := su.ingLister.GetByKey(key)
+	ingCopy, exists, err := su.ingLister.GetByKeySafe(key)
 	if err != nil {
-		glog.V(3).Infof("error getting ing by key: %v", err)
+		glog.V(3).Infof("error getting ing from Store by key: %v", err)
 		return err
 	}
 	if !exists {
--- a/nginx-controller/controller/status.go
+++ b/nginx-controller/controller/status.go
@@ -24,6 +24,8 @@ type StatusUpdater struct {
 	externalStatusAddress    string
 	externalServiceAddresses []string
 	status                   []api_v1.LoadBalancerIngress
+	keyFunc                  func(obj interface{}) (string, error)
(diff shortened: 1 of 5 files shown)


nginx-controller/controller/utils.go
```diff
@@ -175,7 +175,7 @@ func (s *StoreToIngressLister) List() (ing extensions.IngressList, err error) {
 // Note that this ignores services without the right nodePorts.
 func (s *StoreToIngressLister) GetServiceIngress(svc *api_v1.Service) (ings []extensions.Ingress, err error) {
 	for _, m := range s.Store.List() {
-		ing := *m.(*extensions.Ingress)
+		ing := *m.(*extensions.Ingress).DeepCopy()
```
pleshakov: why?

isaachawley: When endpoints change we fetch ingresses using this method (via `lbc.getIngressForEndpoints`) in `syncEndp` which calls the same `UpdateEndpointsMergeableIngress` that messes with the annotations.

nginx-controller/controller/status.go
```diff
@@ -67,12 +69,26 @@ func (su *StatusUpdater) updateIngressWithStatus(ing v1beta1.Ingress, status []a
 	if reflect.DeepEqual(ing.Status.LoadBalancer.Ingress, status) {
 		return nil
 	}
-	// Objects from the LoadBalancerController.ingLister.Store are retrieved by reference
-	// and are not safe to modify.
-	ingCopy := ing.DeepCopy()
+
+	// get a pristine ing with no annotation changes
```
pleshakov: Please explain why need to do that. For example. `get a pristine ing. required because annotations can be modified for mergeable Ingress resources and the update status API call updates annotations`

nginx-controller/controller/utils.go
```diff
@@ -163,10 +163,21 @@ type StoreToIngressLister struct {
 	cache.Store
 }
 
+// GetByKey calls Store.GetByKey and returns a copy of the ingress so it is
+// safe to modify.
+func (s *StoreToIngressLister) GetByKey(key string) (ing *extensions.Ingress, exists bool, err error) {
```
pleshakov: perhaps it makes sense to use a different name for GetByKey? Consider that in controller.go we call GetByKey of the other Listers, which will continue to return an unsafe to modify resource

PR #336, opened 2018-08-14
Stub status

r4j4h:
### Proposed changes
Not too long ago a `health-status` command line argument was added to add a base location that offers a 200 "healthy" endpoint.

Recently, [nginx-prometheus-exporter](https://github.com/nginxinc/nginx-prometheus-exporter) appeared and supports both Nginx and Nginx Plus!

For Nginx basic, it needs access to a stub_status endpoint. I was unable to use `location-snippets` to accomplish this as the accompanying `proxypass` directives seem to over-ride the `stub_status` directive.

So I decided to follow suit with the health-status and add a `stub-status` command line argument to add another base `server` location that offers a *stub_status*-powered endpoint.

This change worked for my needs and seemed good. I then felt that leaving that status page wide open was a scarier proposition than leaving open a 200 "Healthy" endpoint, so I also added a `status-allow-ip` command line argument that defaults to Off but allows defining an IP to use for allowing and denying everything else in the status routes.

I'm not sure if command line argument is still the best home for these, but since it is is server-wide it very well may be. I also am not sure if we would want to customize things such as the endpoint's path, so I just went with the default one that the nginx-prometheus-exporter uses by default.

For Nginx Plus, instead of stub_status we are to use api as described by nginx-prometheus-exporter but the Plus config already generates an /api route. So I changed my "stub-status" flag to no-op on Nginx Plus. So for Nginx Plus this change just allows locking down the health endpoint via `status-allow-ip`.

### Checklist

### DIFF
--- a/docs/cli-arguments.md
+++ b/docs/cli-arguments.md
@@ -17,6 +17,10 @@ Usage of ./nginx-ingress:
   -health-status
     	Add a location "/nginx-health" to the default server. The location responds with the 200 status code for any request.
 	Useful for external health-checking of the Ingress controller
+  -stub-status
+        Add a location "/stub_status" to the default server. The location responds with the stats from stub status
+    module for any request.
+    Useful in tandem with nginx-prometheus-exporter for monitoring of the Ingress controller
   -ingress-class string
     	A class of the Ingress controller. The Ingress controller only processes Ingress resources that belong to its class
 	- i.e. have the annotation "kubernetes.io/ingress.class" equal to the class. Additionally,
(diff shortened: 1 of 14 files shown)


r4j4h:
Hi @isaachawley ,

Thanks for the detailed response and consideration of my work. The approach described in that issue works great for me, but I don't have any suggestions to add. I think it's a great direction though, being able to configure the port will be handy for many I am sure.

I will await your notice and once the new feature is ready I will add `-status-allow-ip` onto that code  and prepare a new PR. That sounds like a great plan to me! 👍

PR #345, opened 2018-08-21
Allow the http2 directive for non-SSL ports

paigr:
### Proposed changes
Currently the `http2` directive only gets added to SSL ports; this change will also add the directive to non-SSL ports if the ConfigMap value has been set.

I saw that this feature was added for SSL ports in #65, but didn't see a reason why it was not enabled for normal ports as well. Seems to me like a trivial change, but, if for some reason this was excluded deliberately, feel free to close.

### DIFF
--- a/nginx-controller/nginx/templates/nginx-plus.ingress.tmpl
+++ b/nginx-controller/nginx/templates/nginx-plus.ingress.tmpl
@@ -22,7 +22,7 @@ upstream {{$upstream.Name}} {
 server {
 	{{if not $server.GRPCOnly}}
 	{{range $port := $server.Ports}}
-	listen {{$port}}{{if $server.ProxyProtocol}} proxy_protocol{{end}};
+	listen {{$port}}{{if $server.HTTP2}} http2{{end}}{{if $server.ProxyProtocol}} proxy_protocol{{end}};
 	{{- end}}
 	{{end}}
 	{{if $server.SSL}}
--- a/nginx-controller/nginx/templates/nginx-plus.tmpl
+++ b/nginx-controller/nginx/templates/nginx-plus.tmpl
@@ -63,7 +63,7 @@ http {
 
  
     server {
-        listen 80 default_server{{if .ProxyProtocol}} proxy_protocol{{end}};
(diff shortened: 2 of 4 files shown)


pleshakov:
@paigr thanks for the PR. unfortunately, this PR will break web browsers.

NGINX doesn’t support both http/1.1 and http2 on a non-TLS socket. At the same time, web browsers when making a request to a non-TLS port, use http/1.1. As a result, a browser will fail to get a response from NGINX.

Do you have some clients other than web browsers that require http2 on a non-TLS socket?

paigr:
Hi @pleshakov, thanks for clarifying this for me. Yes, we currently have multiple Kubernetes clusters running Thanos, and would like to connect them all together. Unfortunately, they can currently only communicate over unencrypted gRPC (improbable-eng/thanos#468 is tracking this). Until this is implemented, we would like some kind of work-around.

I see that servers can have the `GRPCOnly` property; for these servers are we still concerned about non-TLS HTTP/1.1 connections? If not, could we then do something along the lines of:
```
{{range $port := $server.Ports}}
listen {{$port}}{{if $server.HTTP2 and $server.GRPCOnly}} http2{{end}}{{if $server.ProxyProtocol}} proxy_protocol{{end}};
{{- end}}
```

Alternatively, do you know if we could instead use some sort of forward proxy to broker gRPC TLS connections between Thanos clients and Nginx?

pleshakov:
@paigr for your case, I suggest customizing the template to use `http2` and then attaching to the Ingress Controller. Please see https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/custom-templates

> I see that servers can have the GRPCOnly property; for these servers are we still concerned about non-TLS HTTP/1.1 connections? If not, could we then do something along the lines of:

GRPCOnly property is per Ingress resource. Other Ingress resources might still for HTTP apps. In that case we can't make port 80 http2 only. Essentially, port 80 is shared among all the Ingress resources. Enabling http2 in on config with your suggestion will enable http2 in all other configs.

paigr:
@pleshakov OK, great, I'll give custom templates a try. Thanks again.

PR #380, opened 2018-09-27
Verify reloads & cache secrets

isaachawley:
When calling `nginx -s reload` block until we verify that the new config has been loaded.

- Cache secrets to improve performance.
- Adds a new config file to hold config version.
- Adds a verify client that fetches the version.
- Protects the Plus API from making changes on a worker process with old config.
- Skips `nginx -t`, since it's done by `nginx -s reload`.

Reload performance is worse after this change since we block after reloading. However with the secrets cache fix overall performance is improved.

### Checklist

### DIFF
--- a/internal/nginx/nginx.go
+++ b/internal/nginx/nginx.go
@@ -9,6 +9,7 @@ import (
 	"path"
 
 	"github.com/golang/glog"
+	"github.com/nginxinc/kubernetes-ingress/internal/nginx/verify"
 )
 
 const dhparamFilename = "dhparam.pem"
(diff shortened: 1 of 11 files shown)


internal/nginx/nginx.go
```diff
@@ -399,3 +417,37 @@ func (nginx *Controller) UpdateIngressConfigFile(name string, cfg []byte) {
...
+	if !nginx.local {
+		w, err := os.Create(tempname)
+		if err != nil {
+			glog.Fatalf("Failed to open %v: %v", filename, err)
+		}
+		_, err = w.Write(cfg)
+		if err != nil {
+			glog.Fatalf("Failed to write to %v: %v", filename, err)
+		}
+		w.Close()
```
peterkellyonline: @isaachawley should you not `defer` right after `os.Create` above?

isaachawley: Hmm. We don't normally do that. See [writing secrets here](https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/nginx/nginx.go#L250). I'm open to changing it but I'd like to change them all at once.

PR #399, opened 2018-10-16
Improve secret handling

pleshakov:
### Proposed changes

- Simplify the secret handling logic.
- Change the secret handling logic. See below.

Changes:

1. An Ingress includes TLS termination, but the referenced
TLS Secret is missing in Kubernetes/invalid or goes missing/
becomes invalid. An Ingress can be a regular Ingress or a
mergeable Master.

Before: the Ingress resource was rejected.

Now: the Ingress resource is not rejected. Instead, the generated
config for that Ingress resource now includes the ssl_ciphers
directive set to "NULL", which makes NGINX break any attempt to
establish a TLS connection with the corresponding Ingress host.

2. An Ingress includes JWT auth, but the referenced JWK is
missing in Kubernetes/invalid or goes missing/becomes invalid.
An Ingress can be a regular Ingress, a mergeable Master or
a mergeable Minion.

Before: the Ingress resource was rejected.

Now. the Ingress resource is not rejected. However, the generated
config for that Ingress still references the JWK on the file system,
which does not exist. This makes NGINX Plus return a 500 response
for a request to the corresponding Ingress host (or the path of the
host for mergeable minions).

### Checklist

### DIFF
--- a/internal/nginx/configurator.go
+++ b/internal/nginx/configurator.go
@@ -17,11 +17,13 @@ import (
 
 const emptyHost = ""
 
+const pemFileNameForMissingTLSSecret = "/etc/nginx/secrets/default"
+
 // DefaultServerSecretName is the filename of the Secret with a TLS cert and a key for the default server
 const DefaultServerSecretName = "default"
 
-// JWTKey is the key of the data field of a Secret where the JWK must be stored.
-const JWTKey = "jwk"
+// JWTKeyKey is the key of the data field of a Secret where the JWK must be stored.
+const JWTKeyKey = "jwk"
 
 // JWTKeyAnnotation is the annotation where the Secret with a JWK is specified.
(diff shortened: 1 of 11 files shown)


internal/nginx/configurator.go
```diff
@@ -17,11 +17,13 @@ import (
...
 
+const pemFileNameForMissingTLSSecret = "/etc/nginx/secrets/default"
+
 // DefaultServerSecretName is the filename of the Secret with a TLS cert and a key for the default server
 const DefaultServerSecretName = "default"
 
-// JWTKey is the key of the data field of a Secret where the JWK must be stored.
-const JWTKey = "jwk"
+// JWTKeyKey is the key of the data field of a Secret where the JWK must be stored.
+const JWTKeyKey = "jwk"
```
peterkellyonline: KeyKey!
```suggestion
const JWTKey = "jwk"
```

Issue #262, opened 2018-03-27
Support for ExternalName's used for cross-namespace Services


r4j4h:
Hi there,

# Summary

ExternalName services do not get endpoints created, and as far as I can tell the Ingress controller depends exclusively on the Endpoints created from Services, so ExternalName services cannot currently be used to reach a Service across namespaces using the Nginx Ingress Controller.

# Background

Today we tried including a surrogate Service in the local namespace referencing a service in another namespace via an ExternalName pointed to its in-cluster DNS:

```
kind: Service
apiVersion: v1
metadata:
  name: other-service
  namespace: current-namespace
spec:
  type: ExternalName
  externalName: other-service.other-namespace.svc.cluster.local
  ports:
  - port: 443
```

When resolving that service within a pod, the appropriate Pod IP from the other namespace is returned.

When nginx comes to handle the service, however, it logs "no endpoints found" for the service and `NewUpstreamWithDefaultServer()` marks the upstream as `127.0.0.1:8181`. When accessing the path through nginx, an error is given.

This is to be expected, as externalName Services do not get endpoints created, but the Ingress controller depends exclusively on the Endpoints created from Services. This latter part was surprising to me.

# Proposal

This issue is to bring up this point and see what thoughts are around the idea of having the nginx Ingress Controller behave differently when no Endpoints can be found to support this case.

My thought is that there are two locations we could try to resolve the name using DNS before falling back to the current behavior:

* https://github.com/nginxinc/kubernetes-ingress/blob/8aca003ee5f94b1d874cd786b8be5d6d87136fd0/nginx-controller/nginx/configurator.go#L580-L602
* https://github.com/nginxinc/kubernetes-ingress/blob/8aca003ee5f94b1d874cd786b8be5d6d87136fd0/nginx-controller/nginx/nginx.go#L135-L140

A [similar discussion on the contrib nginx Ingress controller](https://github.com/kubernetes/ingress-nginx/issues/1600#issuecomment-340019652) has led me to think modifying the proxy_pass directive directly is more correct by convention for pointing to DNS. If so, then the changes needed would affect here (and in the respective Nginx Plus-variant template)
https://github.com/nginxinc/kubernetes-ingress/blob/8aca003ee5f94b1d874cd786b8be5d6d87136fd0/nginx-controller/nginx/templates/nginx.ingress.tmpl#L92-L96

For reference on ExternalName Services, see https://github.com/kubernetes/website/issues/5822.

[...]

PR #480, opened 2019-01-15
Add support for Kubernetes ExternalName service

sombralibre:
### Proposed changes

I've tried to use ExternalName service with this ingress-controller but doesn't work, so I perform some changes to the `controller.go` to use `backend.ServiceName:backend.ServicePort` as a default endpoint instead of look for them in kubernetes objects.

These changes are based in the issue [#262](https://github.com/nginxinc/kubernetes-ingress/issues/262) and this comment [ExternalName](https://github.com/nginxinc/kubernetes-ingress/issues/262#issuecomment-376918311)

I've just modify `getEndpointsForIngressBackend` method to add these lines:

```go
	if svc.Spec.Type == api_v1.ServiceTypeExternalName {
		var endpoints []string
		endpoint := fmt.Sprintf("%s:%d", backend.ServiceName, int32(backend.ServicePort.IntValue()) )
		endpoints = append(endpoints, endpoint)
		return endpoints, nil
	}
```

Hope this help :smiley: 

### Checklist

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -1072,9 +1072,17 @@ func compareContainerPortAndServicePort(containerPort api_v1.ContainerPort, svcP
 
 func (lbc *LoadBalancerController) getEndpointsForIngressBackend(backend *extensions.IngressBackend, namespace string) ([]string, error) {
 	svc, err := lbc.getServiceForIngressBackend(backend, namespace)
+
 	if err != nil {
 		glog.V(3).Infof("Error getting service %v: %v", backend.ServiceName, err)
 		return nil, err
+	}
+
+	if svc.Spec.Type == api_v1.ServiceTypeExternalName {
+		var endpoints []string
+		endpoint := fmt.Sprintf("%s:%d", backend.ServiceName, int32(backend.ServicePort.IntValue()) )
+		endpoints = append(endpoints, endpoint)
+		return endpoints, nil
 	}
 
 	endps, err := lbc.endpointLister.GetServiceEndpoints(svc)


pleshakov:
@sombralibre thx for the PR.

Unfortunately, the suggested solution has limitations:
- *Stale DNS records*. Once a DNS name of an ExternalName service is resolved by NGINX during a configuration reload, NGINX will not try to re-resolve it, even when the TTL of the resolved records expires. NGINX will only be able to re-resolve the name during any subsequent configuration reload. 
- *Potential reload failures*. If a DNS name cannot be resolved, NGINX will fail to reload. Please note that if the Ingress Controller tries to make any further changes to NGINX configuration -- for example, as a result of the changes that affect other Ingress resources -- those changes will not be applied by NGINX until the DNS name can be resolved again and therefore NGINX can be successfully reloaded.

Those limitations come from the fact that NGINX doesn't support dynamic DNS resolution. More info about this topic is written in  https://www.nginx.com/blog/dns-service-discovery-nginx-plus/

Additionally, the suggested solution assumes that the Ingress Controller and the service in question is deployed in the same namespace. This is an example of the generated config: `server my-service:80 max_fails=1 fail_timeout=10s;`. If the my-service service is deployed in a different namespace, NGINX will not be able to resolve it.

As a workaround for the externalname services, I suggest to try to use a service without a selector, described here https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors In this case, in addition to a service resource you will have to create a corresponding endpoints resources that contains the list of IPs of the service.
Can this approach work for you?

Note that we are implementing support of type ExternalName services for NGINX Plus, because it supports dynamic DNS resolving.

sombralibre:
It's ok, that make sense.

So there isn't reason to keep this PR open.

Thanks for you help and accurate explanation.

Issue #404, opened 2018-10-24
Enabling HSTS not working


achandak123:
**Describe the bug**
Using 1.2.0 version of the ingress controller, hsts is not working. Dont see strict-transport-security header

**To Reproduce**
Here are the config files

**deployment file**
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: nginx-ingress
  namespace: nginx-ingress
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx-ingress
  template:
    metadata:
      labels:
        app: nginx-ingress
    spec:
      serviceAccountName: nginx-ingress
      containers:
      - image: nginx/nginx-ingress:1.2.0
        name: nginx-ingress
        ports:
        - name: http
          containerPort: 80
        - name: https
          containerPort: 443
        env:
        - name: POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        args:
          - -nginx-configmaps=$(POD_NAMESPACE)/nginx-ingress-config
          - -default-server-tls-secret=$(POD_NAMESPACE)/default-server-secret
         #- -v=3 # Enables extensive logging. Useful for trooublshooting.
        volumeMounts:
        - name: config-volume
          mountPath: /etc/nginx/custom-snippets/
      volumes:
      - name: config-volume
        configMap:
          name: cors-config
```

Service file

```
apiVersion: v1
kind: Service
metadata:
  name: nginx-ingress
  namespace: nginx-ingress
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: "http"
    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: "*"
    service.beta.kubernetes.io/aws-load-balancer-ssl-ports: https
    service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "3600"
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: <aws ssl arn>
    #    Unfortunately, we cant use ACM certs from another account:(https://forums.aws.amazon.com/thread.jspa?threadID=231116)
spec:
  type: LoadBalancer
  ports:
  - port: 80
    targetPort: 80
    name: http
  - port: 443
    targetPort: 80
    name: https
  selector:
    app: nginx-ingress

```

ConfigMap

```
kind: ConfigMap
apiVersion: v1
metadata:
  name: nginx-ingress-config
  namespace: nginx-ingress
data:
[...]

PR #481, opened 2019-01-23
Fix HSTS header over non tls ingresses

Dean-Coakley:
Fixes: https://github.com/nginxinc/kubernetes-ingress/issues/404

### Proposed changes
* HSTS headers are no longer added to responses over HTTP
* Added HSTSBehindProxy annotation and configmap to allow HSTS headers to be sent over HTTP. This is useful for the use case of TLS termination being done in a proxy deployed in front of the ingress controller

### Checklist

### DIFF
--- a/internal/nginx/templates/nginx-plus.ingress.tmpl
+++ b/internal/nginx/templates/nginx-plus.ingress.tmpl
@@ -75,12 +75,12 @@ server {
 		return 301 https://$host:{{index $server.SSLPorts 0}}$request_uri;
 	}
 	{{- end}}
+	{{- end}}
 
 	{{- if $server.RedirectToHTTPS}}
 	if ($http_x_forwarded_proto = 'http') {
 		return 301 https://$host$request_uri;
 	}
-	{{- end}}
 	{{- end}}
 
 	{{with $jwt := $server.JWTAuth}}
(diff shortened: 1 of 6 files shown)


internal/nginx/templates/nginx-plus.ingress.tmpl
```diff
@@ -55,21 +55,34 @@ server {
...
 	{{end}}
 	{{- end}}
+
 	{{- if $server.RedirectToHTTPS}}
 	if ($http_x_forwarded_proto = 'http') {
 		return 301 https://$host$request_uri;
 	}
 	{{- end}}
 
+	{{- if $server.SSLRedirect}}
```
pleshakov: the SSL redirect part here must go into the following block
```
	{{if $server.SSL}}
	{{if not $server.GRPCOnly}}
          <here, after hsts stuff>
        {{end}}
        {{end}}
```
Otherwise it is a BUG, because the redirect will be enabled for Ingress resources without any TLS termination.

Please double check that for an Ingress  resource without TLS termination, no SSL redirect is generated.

pleshakov: same for nginx oss template

Dean-Coakley: 👍 nice catch

PR #485, opened 2019-02-05
Add type ExternalName service support for NGINX Plus

### LINKED ISSUE #262: Support for ExternalName's used for cross-namespace Services

### LINKED ISSUE #446: External Name Support
I would like to have a way to route traffic using ExternalName services and the ingress controller. The kubernetes/ingress-nginx project has solved this issue but I would like to use NGINX Plus to proxy this traffic. 

I would like to see this project match the feature set for ExternalName services on the kubernetes/ingress-nginx project.

Thank you!

Rulox:
### Proposed changes
Closes #262
Closes #446

This PR adds support for type [ExternalName](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) services for NGINX Plus only.

### Checklist

### DIFF
--- a/docs/configmap-and-annotations.md
+++ b/docs/configmap-and-annotations.md
@@ -101,6 +101,11 @@ spec:
 | N/A | `worker-shutdown-timeout` | Sets the value of the [worker_shutdown_timeout](http://nginx.org/en/docs/ngx_core_module.html#worker_shutdown_timeout) directive. | N/A | |
 | N/A | `server-names-hash-bucket-size` | Sets the value of the [server_names_hash_bucket_size](http://nginx.org/en/docs/http/ngx_http_core_module.html#server_names_hash_bucket_size) directive. | Depends on the size of the processor’s cache line. | |
 | N/A | `server-names-hash-max-size` | Sets the value of the [server_names_hash_max_size](http://nginx.org/en/docs/http/ngx_http_core_module.html#server_names_hash_max_size) directive. | `512` | |
+| N/A | `resolver-addresses` | Sets the value of the [resolver](http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver) addresses. Note: If you use a DNS name (ex., `kube-dns.kube-system.svc.cluster.local`) as a resolver address, NGINX Plus will resolve it using the system resolver during the start and on every configuration reload. As a consequence, If the name cannot be resolved or the DNS server doesn't respond, NGINX Plus will fail to start or reload. To avoid this, consider using only IP addresses as resolver addresses. Supported in NGINX Plus only. | N/A | [Support for Type ExternalName Services](../examples/externalname-services). |
+| N/A | `resolver-ipv6` | Enables IPv6 resolution in the resolver. Supported in NGINX Plus only. | `True` | [Support for Type ExternalName Services](../examples/externalname-services). |
+| N/A | `resolver-valid` | Sets the time NGINX caches the resolved DNS records. Supported in NGINX Plus only. | TTL value of a DNS record | [Support for Type ExternalName Services](../examples/externalname-services). |
+| N/A | `resolver-timeout` | Sets the [resolver_timeout](http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver_timeout) for name resolution. Supported in NGINX Plus only.  | `30s` | [Support for Type ExternalName Services](../examples/externalname-services). |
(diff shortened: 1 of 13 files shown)


examples/externalname-services/README.md
```diff
@@ -0,0 +1,61 @@
...
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: nginx-config
+  namespace: nginx-ingress
+data:
+  resolver-addresses: "10.0.0.10"
+```
+
+Additional resolver parameters, including the caching of DNS records, are available. Check the corresponding [ConfigMap and Annotations](../docs/configmap-and-annotations.md) section.
```
pleshakov:
```suggestion
Additional resolver parameters, including the caching of DNS records, are available. Check the corresponding [ConfigMap and Annotations](../../docs/configmap-and-annotations.md) section.
```

docs/configmap-and-annotations.md
```diff
@@ -101,6 +101,11 @@ spec:
...
+| N/A | `resolver-ipv6` | Enables IPv6 resolution in the resolver. Supported in NGINX Plus only. | True | [Support for Type ExternalName Services](../examples/externalname-services). |
+| N/A | `resolver-valid` | Sets the time NGINX caches the resolved DNS records. Supported in NGINX Plus only. | TTL value of a DNS record | [Support for Type ExternalName Services](../examples/externalname-services). |
+| N/A | `resolver-timeout` | Sets the [resolver_timeout](http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver_timeout) for name resolution. Supported in NGINX Plus only.  | 30s | [Support for Type ExternalName Services](../examples/externalname-services). |
```
pleshakov:
```suggestion
| N/A | `resolver-timeout` | Sets the [resolver_timeout](http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver_timeout) for name resolution. Supported in NGINX Plus only.  | `30s` | [Support for Type ExternalName Services](../examples/externalname-services). |
```

Issue #502, opened 2019-03-06
GCP GKE 1.12's "instance groups" disabled ability to open firewall port 443 😭


rayfoss:
**Is your feature request related to a problem? **
An update to GKE changed how nodes are created in Google Compute Engine, in a way where you can no longer open port 443 on an instance. To be specific, Nodes are now created in an a k8s managed instance group using a k8s managed instance template. Instance groups disable the option to make manual changes to individual instances, and therefore the firewall of k8s nodes.

    You can't edit this VM instance because it's managed by a group. To edit this VM instance, first remove it from group gke-my-pool-0-2312adff-grp

Before this update we were able to run a deamonset on port 443, open the port on the node and bob's your uncle.

**Describe the solution you'd like**
Documentation on how to open the port? or modifying k8s node config to have it create nodes with an open port 443?

**Describe alternatives you've considered**
Screaming at Google on twitter

**Additional context**
![screen shot 2019-03-06 at 9 13 44 am](https://user-images.githubusercontent.com/876076/53891650-73434480-3ff0-11e9-8d18-f034af16c881.png)

rayfoss:
This works, even though the GUI disabled the ability to do so

`gcloud compute instances add-tags gke-clusty-pool-0-7696af58-52nf --zone=us-central1-b --tags https-server`

boxer3379:
Yes it did

PR #515, opened 2019-03-21
Highlight DaemonSet port firewall issues

rayfoss:
### Proposed changes
This was a surprise when I upgraded to k8 1.12

https://github.com/nginxinc/kubernetes-ingress/issues/502

### Checklist

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -85,7 +85,7 @@ $ kubectl get pods --namespace=nginx-ingress
 
 ## 4. Get Access to the Ingress Controller
 
-**If you created a daemonset**, ports 80 and 443 of the Ingress controller container are mapped to the same ports of the node where the container is running. To access the Ingress controller, use those ports and an IP address of any node of the cluster where the Ingress controller is running. Make sure the firewall ports of your nodes are open by creating the cluster with the correct network tags, editing the node template or manually opening them your self.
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -85,7 +85,7 @@ $ kubectl get pods --namespace=nginx-ingress
 
 ## 4. Get Access to the Ingress Controller
 
-**If you created a daemonset**, ports 80 and 443 of the Ingress controller container are mapped to the same ports of the node where the container is running. To access the Ingress controller, use those ports and an IP address of any node of the cluster where the Ingress controller is running.
+**If you created a daemonset**, ports 80 and 443 of the Ingress controller container are mapped to the same ports of the node where the container is running. To access the Ingress controller, use those ports and an IP address of any node of the cluster where the Ingress controller is running. Opening HTTP ports node on the GKE firewall may require changing the node template, see https://github.com/nginxinc/kubernetes-ingress/issues/502
```
pleshakov: Using port mapping on GKE is uncommon - typically people would use a service with type LoadBalancer. That’s why GKE specific advice here doesn’t make sense to me.

We can mention something like “Make sure that the firewall allows traffic on those ports”.

What do you think?

pleshakov:
@rayfoss thanks for making the changes. However, the new instructions are still too specific - they make sense for GKE and Google Cloud, not Kubernetes in general.

I suggest either having a very generic message (please also see my previous comment https://github.com/nginxinc/kubernetes-ingress/pull/515#discussion_r268145780), or do not provide any instructions at all.

PR #534, opened 2019-04-07
Add ConfigMap name to values.yaml and allow the option to specify the leader election lock name

### LINKED ISSUE #528: Multiple HA deployments in the namespace share the same leader, preventing ingress status updates
**Describe the bug**
I've created a namespace `networking` and deployed two HA deployments (public and private)
The private deployment has the following arguments

          - -ingress-class=nginx-private
          - -use-ingress-class-only=true
          - -report-ingress-status
          - -external-service=nginx-ingress-private
          - -enable-leader-election

The public deployment has the following arguments

          - -ingress-class=nginx-public
          - -use-ingress-class-only=true
          - -report-ingress-status
          - -external-service=nginx-ingress-public
          - -enable-leader-election

When deployed, the following configmaps are created

$ kubectl  --namespace=networking get configmap
NAME                      DATA   AGE
leader-election           0      6d
nginx-config              4      7d

These mean that only one ingress-class gets a leader, and that the nginx-config is shared between them despite differences.

**To Reproduce**
Deploy using helm, and add the above options.

**Expected behavior**
Each configmap include the ingress-class so that they aren't shared between deployments

**Your environment**
* Version of the Ingress Controller - controller.image.tag": "1.4.3-alpine"
* Version of Kubernetes 1.10
* Kubernetes platform (e.g. Mini-kube or GCP) KOPS
* Using NGINX 

**Additional context**
Add any other context about the problem here. Any log files you want to share.

paigr:
### Proposed changes
Allow the customisation of the ConfigMap name within the helm template. The value is used for `.metadata.name` in the ConfigMap yaml, as well as with the `-nginx-configmaps` flag for the container.

Also allows specifying the name used for the lock with leader election enabled. This was implemented as a new CLI flag, `-leader-election-lock-name`, and an accompanying entry in the Helm template's `values.yaml`.

Resolves  #528 

### Checklist

### DIFF
--- a/cmd/nginx-ingress/main.go
+++ b/cmd/nginx-ingress/main.go
@@ -84,6 +84,9 @@ The external address of the service is used when reporting the status of Ingress
 	leaderElectionEnabled = flag.Bool("enable-leader-election", false,
 		"Enable Leader election to avoid multiple replicas of the controller reporting the status of Ingress resources -- only one replica will report status. See -report-ingress-status flag.")
 
+	leaderElectionLockName = flag.String("leader-election-lock-name", "nginx-ingress-leader-election",
+		`Specifies the name of the ConfigMap, within the same namespace as the controller, used as the lock for leader election. Requires -enable-leader-election.`)
+
 	nginxStatusAllowCIDRs = flag.String("nginx-status-allow-cidrs", "127.0.0.1", `Whitelist IPv4 IP/CIDR blocks to allow access to NGINX stub_status or the NGINX Plus API. Separate multiple IP/CIDR by commas.`)
 
 	nginxStatusPort = flag.Int("nginx-status-port", 8080,
@@ -297,6 +300,7 @@ func main() {
 		ControllerNamespace:     controllerNamespace,
 		ReportIngressStatus:     *reportIngressStatus,
 		IsLeaderElectionEnabled: *leaderElectionEnabled,
+		LeaderElectionLockName:  *leaderElectionLockName,
 		WildcardTLSSecret:       *wildcardTLSSecret,
 		ConfigMaps:              *nginxConfigMaps,
 	}
--- a/deployments/helm-chart/Chart.yaml
+++ b/deployments/helm-chart/Chart.yaml
@@ -1,5 +1,5 @@
 name: nginx-ingress
-version: 0.3.5
+version: 0.3.6
 appVersion: edge
 description: NGINX Ingress Controller
 icon: https://raw.githubusercontent.com/nginxinc/kubernetes-ingress/master/deployments/helm-chart/chart-icon.png
--- a/deployments/helm-chart/README.md
+++ b/deployments/helm-chart/README.md
@@ -65,7 +65,7 @@ Parameter | Description | Default
 `controller.image.repository` | The image repository of the Ingress controller. | nginx/nginx-ingress
 `controller.image.tag` | The tag of the Ingress controller image. | edge
 `controller.image.pullPolicy` | The pull policy for the Ingress controller image. | IfNotPresent
-`controller.config.name` | The name of the ConfigMap used by the Ingress controller container.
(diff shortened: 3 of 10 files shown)


pleshakov:
@paigr thanks for the PR!

The part about the leader election looks good to me! Thanks for changing all the necessary docs. The only thing I would change is the default for the leader-election-lock-name
 -- from `leader-election` to something more specific like `nginx-ingress-leader-election`. I understand that that default was already present. But I think it is a good opportunity to change it.

Regarding the part about the `controller.config.name`parameter. I am not sure if it solves the problem completely. Her is how I see the problem:
```
$ kubectl create ns nginx-ingress-1
$ helm install --namespace nginx-ingress-1 --name my-r-1 .
. . .
NOTES:
The NGINX Ingress Controller has been installed.
$ kubectl create ns nginx-ingress-2
$ helm install --namespace nginx-ingress-2 --name my-r-2 .
Error: release my-r-2 failed: clusterroles.rbac.authorization.k8s.io "nginx-ingress" already exists
```
Because some of the resources in the helm chart are created with the same names, the second `helm install` command fails.
Is it the problem that you're trying to solve? Or you have a different problem?

paigr:
Thanks for your comments @pleshakov.

For `controller.config.name`, the main use-case for this was to be able to deploy multiple ICs to the same namespace. We typically keep all of our ingress related resources within a single namespace, including two copies of the IC; one for internal traffic, and one for external. When using the Helm chart, there is no option of templating a different value for the ConfigMap name, so both ICs are required to share a config.

I agree with your point that the default value for the leader lock should be `nginx-ingress-leader-election`. However, have you thought about prefixing your resources with `{{ .Values.controller.name }}`? It would set sensible defaults, and also guarantee unique names fo cluster-level objects, solving the issue of installing via Helm twice on a single cluster, as you mentioned above.

pleshakov:
@paigr 
Thanks for the updates. Just a small suggestion: could you change the defaults in the helm chart README and the values file for `leaderElectionLockName`?

If this solves your problem, we're are happy to merge. However, please address the suggestion above and squash the leaderElectionLockName-related commits into a single one. 

> However, have you thought about prefixing your resources with {{ .Values.controller.name }}

Yes, we're looking into that. We might use the release name as part of the name as well. 
When we introduce that change, it will make sense to change the default for `controller.config.name`  from `nginx-config` to a generated name as well as to use a generated name for `controller.reportIngressStatus.leaderElectionLockName`.  Could this potentially break your installation?

paigr:
@pleshakov during my tests I identified another potential bug. The template helper that defines `nginx-ingress.name` was using the name of the chart instead of the value for `controller.name`, thus `controller.name` is not actually used. Was this intentional?

I think generating the ConfigMap name will not break anything on our end. If we retain the option to set the name in the Helm template, as well as via CLI flag, there should be no problems.

pleshakov:
@paigr 
> thus controller.name is not actually used. Was this intentional?

this was not intentional. good catch! However, I suggest leaving that fix out the scope of this PR. We're gonna change the logic of how the name is generated for the resources anyway -- as I mentioned we will use the release name as part of the name as well.

thanks

paigr:
@pleshakov will this be done soon? Happy to remove it from this PR, but we do need that change as it gets in the way of deploying multiple ICs into the same namespace (which was the motivation for this PR).

pleshakov:
@paigr yes, this should be done in the next two weeks (by April 26th or sooner). would you be ok to use a modified copy for the time being?

paigr:
> @paigr yes, this should be done in the next two weeks (by April 26th or sooner). would you be ok to use a modified copy for the time being?

Sure, no problem, I have removed the commit.

pleshakov:
@paigr great! we'll merge shortly

Dean-Coakley:
@paigr thanks a lot!

PR #546, opened 2019-04-23
Allow Deploying Multiple Ingress Controllers per Cluster

Dean-Coakley:
### Proposed changes
Re: https://github.com/nginxinc/kubernetes-ingress/issues/528
Fixes: https://github.com/nginxinc/kubernetes-ingress/issues/549

All names of resources created by a deployment must be unique.

* By default helm chart now prepends release name to all resources it creates to ensure no conflicts occur when deploying 2 or more ingress controllers in the same namespace.

* `leaderLockName` now set on daemonset
* Add missing docs about defaultTLS secret to values.yaml

### Checklist


deployments/helm-chart/templates/controller-deployment.yaml
```diff
@@ -80,12 +80,12 @@ spec:
           - -report-ingress-status
```
dlapcenko: ` - -external-service={{ .Values.controller.reportIngressStatus.externalService }}` will result in a bug.
This has to be pointing to the new Service name, otherwise no ingresses would have the `ADDRESS`.
`- -external-service={{ include "nginx-ingress.name" . }}`

Dean-Coakley: Excellent catch. 

I changed the generation of the service name as we want that independently configurable to the name of the nginx-ingress. 

Thanks

paigr: This will not allow customization of the service name via `controller.reportIngressStatus.externalService`. Is the plan to deprecate that?

dlapcenko: @Dean-Coakley know only cause I banged my head against this for couple hours :D (had to copy to chart locally to do same namespace change)

dlapcenko:
@Dean-Coakley could `metadata.namespace: {{ .Release.Namespace }}` be added as part of this PR please? is necessary for certain deploy processes (spinnaker helm).

deployments/helm-chart/templates/controller-deployment.yaml
Dean-Coakley: @paigr Yeah, I'll change it back to what I had 😆 
https://github.com/nginxinc/kubernetes-ingress/pull/546/commits/a4d7ffd58b229f657eb8158d1616d071cf908c17

Dean-Coakley:
@dlapcenko That change is reasonable. We'll land that change in a separate PR shortly - as it's not related to the target issue

deployments/helm-chart/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: nginx-ingress
-version: 0.3.6
+version: 0.3.7
```
pleshakov: Please change it to 0.4.0

deployments/helm-chart/templates/rbac.yaml
```diff
@@ -68,15 +68,15 @@ rules:
...
     {{- include "nginx-ingress.labels" . | nindent 4 }}
 subjects:
 - kind: ServiceAccount
-  name: {{ .Values.controller.serviceAccount.name }}
+  name: {{ include "nginx-ingress.serviceAccountName" . }}
   namespace: {{ .Release.Namespace }}
 roleRef:
   kind: ClusterRole
-  name: nginx-ingress
+  name: {{ include "nginx-ingress.serviceAccountName" . }}
```
pleshakov: this is a bug. should be nginx-ingress.name

PR #552, opened 2019-04-29
Add an example about wildcard cert

pleshakov:
### Proposed changes

Add an example about wildcard cert


Rulox:
Great! example was missing 👍

PR #561, opened 2019-05-13
Make Service name configurable in charts

Rulox:
### Proposed changes
Makes the Service.name configurable in Helm Charts to avoid name/ip change in the Service when upgrading from a older version of the Chart.


deployments/helm-chart/templates/controller-daemonset.yaml
```diff
@@ -89,7 +89,7 @@ spec:
 {{- if .Values.controller.reportIngressStatus.externalService }}
           - -external-service={{ .Values.controller.reportIngressStatus.externalService }}
 {{- else if and (.Values.controller.service.create) (eq .Values.controller.service.type "LoadBalancer") }}
-          - -external-service={{ include "nginx-ingress.name" . }}
+          - -external-service={{ default (include "nginx-ingress.name" .) .Values.controller.service.name }}
```
pleshakov: Since the name of the service is used in multiple manifests,
I would move the logic of creating that name to `_helpers.tpl`, similar to other parameters:
```
{{- define "nginx-ingress.serviceAccountName" -}}
{{- default (include "nginx-ingress.name" .) .Values.controller.serviceAccount.name -}}
{{- end -}}
```

Rulox:
I implemented your suggestion @pleshakov can you have a look? thanks

PR #571, opened 2019-05-16
Improve flaky tests

tellet:
Add an additional step into pre-configuration

### DIFF
--- a/tests/suite/fixtures.py
+++ b/tests/suite/fixtures.py
@@ -13,7 +13,8 @@ from suite.custom_resources_utils import create_crd_from_yaml, delete_crd, creat
     delete_virtual_server
 from suite.kube_config_utils import ensure_context_in_config, get_current_context_name
 from suite.resources_utils import create_namespace_with_name_from_yaml, delete_namespace, create_ns_and_sa_from_yaml, \
-    patch_rbac, create_example_app, wait_until_all_pods_are_ready, delete_common_app
+    patch_rbac, create_example_app, wait_until_all_pods_are_ready, delete_common_app, \
+    ensure_connection_to_public_endpoint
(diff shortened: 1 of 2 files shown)


tests/suite/test_virtual_server_tls.py
```diff
@@ -34,9 +34,9 @@ def assert_ssl_error(virtual_server_setup):
         get_server_certificate_subject(virtual_server_setup.public_endpoint.public_ip,
                                        virtual_server_setup.vs_host,
                                        virtual_server_setup.public_endpoint.port_ssl)
-        pytest.fail("We expected an exception here, but didn't get it. Exiting...")
+        pytest.fail("We expected an SSLError here, but didn't get it or got another error. Exiting...")
```
pleshakov: if we get another one, will it get printed so that we understand what happened?

tests/suite/fixtures.py
```diff
@@ -269,6 +273,9 @@ def crd_ingress_controller(cli_arguments, kube_apis, ingress_controller_prerequi
     print("------------------------- Create IC -----------------------------------")
     name = create_ingress_controller(kube_apis.v1, kube_apis.extensions_v1_beta1, cli_arguments, namespace,
                                      request.param.get('extra_args', None))
+    ensure_connection_to_public_endpoint(ingress_controller_endpoint.public_ip,
```
pleshakov: will it make sense to add ` ensure_connection_to_public_endpoind` to `ingress_controller`?

tellet: there are no flaky tests among 'ingress' tests, so I don't see any reason for that.

tests/suite/test_virtual_server_tls.py
tellet: yes, pytest will post the exception in the stdout.

PR #575, opened 2019-05-24
Add delay as some tests are flaky in aws cloud

tellet:
Couple of tests tend to be flaky in aws cluster. Tests got 502 instead of 200. Existing condition ('wait_until_all_pods_are_ready') for some reason is not enough. Add a workaround for such flaky tests.

### DIFF
--- a/tests/suite/resources_utils.py
+++ b/tests/suite/resources_utils.py
@@ -850,7 +850,7 @@ def ensure_connection(request_url) -> None:
     :param request_url: url to request
     :return:
     """
-    for _ in range(0, 4):
+    for _ in range(4):
         try:
             resp = requests.get(request_url, verify=False)
             if resp.status_code == 404:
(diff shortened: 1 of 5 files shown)


tests/suite/resources_utils.py
```diff
@@ -914,3 +914,19 @@ def get_events(v1: CoreV1Api, namespace) -> []:
...
+
+def ensure_response_from_backend(req_url, host) -> None:
+    """
+    Wait for 502 to disappear.
+
+    :param req_url: url to request
+    :param host:
+    :return:
+    """
+    for _ in range(0, 5):
```
Rulox:
```suggestion
    for _ in range(5):
```
The `from` in range is 0 by default I think

tellet: Right, I'll update

PR #589, opened 2019-06-11
Print out IC logs on test failure

tellet:
Print out IC logs on test failure and change Pod condition

### DIFF
--- a/tests/suite/resources_utils.py
+++ b/tests/suite/resources_utils.py
@@ -157,21 +157,21 @@ def create_daemon_set(extensions_v1_beta1: ExtensionsV1beta1Api, namespace, body
 
 def wait_until_all_pods_are_ready(v1: CoreV1Api, namespace) -> None:
     """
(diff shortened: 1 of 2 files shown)


tests/suite/resources_utils.py
```diff
@@ -385,12 +385,12 @@ def ensure_item_removal(get_item, *args, **kwargs) -> None:
     """
     try:
         counter = 0
-        while counter < 20:
+        while counter < 30:
```
Rulox: Do you think it makes sense to make this as a variable/parameter? eg if this is going to be changed many times because as tests get slower/more complex?

tellet: This shouldn't really depend on a complexity of the tests. I plan to keep an eye on this and if we need to increase this again then we'll have to investigate the cause deeper.

Rulox: Ok thanks!

PR #596, opened 2019-06-21
Add lb-method support in vs and vsr

Rulox:
### Proposed changes
Allow the use of `lb-method` in Upstreams in VirtualServer and VirtualServerRoutes CRDs.

### DIFF
--- a/pkg/apis/configuration/validation/validation_test.go
+++ b/pkg/apis/configuration/validation/validation_test.go
@@ -23,9 +23,10 @@ func TestValidateVirtualServer(t *testing.T) {
 			},
 			Upstreams: []v1alpha1.Upstream{
 				{
-					Name:    "first",
-					Service: "service-1",
-					Port:    80,
+					Name:     "first",
+					Service:  "service-1",
+					LBMethod: "random",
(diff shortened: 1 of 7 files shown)


pkg/apis/configuration/validation/validation_test.go
```diff
@@ -1378,3 +1381,59 @@ func TestValidateVirtualServerRouteSubroutesFails(t *testing.T) {
...
+			method: "",
+			isPlus: true,
+		},
+	}
+
+	for _, test := range tests {
+		allErrs := validateUpstreamLBMethod(test.method, field.NewPath("lb-method"), test.isPlus)
+
+		if len(allErrs) != 0 {
+			t.Errorf("validateUpstreamLBMethod() returned errors for method %s", test.method)
```
pleshakov: Could you possibly also print out the value of isPlus variable. Because the signature of the function is simple, I suggest something like this:
validateUpstreamLBMethod(%q, %v) returned ...
Note that I used %q to make sure when the method is empty, the empty string is printed like “”. 
This comments also applies for line 1436.

PR #602, opened 2019-06-25
Add upstream object to the generation methods

Rulox:
### Proposed changes
Refactor the generation of Upstreams for Virtual Server and Virtual Server Routes in preparation for all the new features we will add soon.

Instead of passing all the parameters one by one, it is better to pass the whole Upstream, and work from there.

### DIFF
--- a/internal/configs/virtualserver_test.go
+++ b/internal/configs/virtualserver_test.go
@@ -732,6 +732,7 @@ func TestGenerateVirtualServerConfigForVirtualServerWithRules(t *testing.T) {
 
 func TestGenerateUpstream(t *testing.T) {
 	name := "test-upstream"
+	upstream := conf_v1alpha1.Upstream{}
 	endpoints := []string{
 		"192.168.10.10:8080",
 	}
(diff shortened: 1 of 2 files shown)


internal/configs/virtualserver_test.go
```diff
@@ -754,7 +754,11 @@ func TestGenerateUpstream(t *testing.T) {
 		LBMethod: "random",
 	}
 
-	result := generateUpstream(name, "", endpoints, isPlus, &cfgParams)
+	upstream := conf_v1alpha1.Upstream{
```
pleshakov: the Name field is not used in the generateUpstream. I suggest not setting it all:
```
upstream := conf_v1alpha1.Upstream{}
```
Also, please place that line between lines 734 and 735, so that the input parameters are initialized in the order they are present in the generateUpstream signature. This is for consistency:
```
	name := "test-upstream"
	upstream := conf_v1alpha1.Upstream{}
	endpoints := []string{
		"192.168.10.10:8080",
	}
        . . .
```
This comment also applies for `TestGenerateUpstreamForZeroEndpoints` function.

PR #607, opened 2019-06-27
Add fail-timeout and max-fails support to vs/vsr

Rulox:
### Proposed changes
`fail-timeout` and `max-fails` can be used in Upstreams in VirtualServer and VirtualServerRoutes resources.

Also changed the `parseSlowStart` function to make it more generic as `ParseTime` exportable function to validate all the times and timeouts of configuration.


pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -25,10 +25,12 @@ type VirtualServerSpec struct {
...
-	Name     string `json:"name"`
-	Service  string `json:"service"`
-	Port     uint16 `json:"port"`
-	LBMethod string `json:"lb-method"`
+	Name        string `json:"name"`
+	Service     string `json:"service"`
+	Port        uint16 `json:"port"`
+	LBMethod    string `json:"lb-method"`
+	FailTimeout string `json:"fail-timeout"`
+	MaxFails    int    `json:"max-fails"`
```
pleshakov: max_fails can't be negative. we need to validate that.

internal/configs/virtualserver.go
```diff
@@ -235,6 +235,20 @@ func generateLBMethod(method string, defaultMethod string) string {
...
 
+func generateTime(time string, defaultTime string) string {
+	if time == "" {
+		return defaultTime
+	}
+	return time
+}
+
+func generateInt(n int, defaultN int) int {
+	if n == 0 {
```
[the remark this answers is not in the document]
Rulox: You are right, I missed that in the doc, good catch!

pkg/apis/configuration/validation/validation.go
```diff
@@ -57,6 +57,20 @@ func validateTLS(tls *v1alpha1.TLS, fieldPath *field.Path) field.ErrorList {
 	return validateSecretName(tls.Secret, fieldPath.Child("secret"))
 }
 
+func validateTime(time string, fieldPath *field.Path) field.ErrorList {
```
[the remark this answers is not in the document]
Rulox: I wasn't sure, ParseTime has a lot of tests and this `validTime` function pretty much only calls that, I can add a test for consistency with other validation functions, however I think it will be repetition of tests.

komapa:
Hey folks, while you are all here, can we also add `max_conns` as a parameter to the upstreams? We were about to ask/open a PR about it but here seems like a very small lift to have it added. Thoughts?

pleshakov:
@komapa `max_conns` makes sense. We'll add it shortly. thx.

pkg/apis/configuration/validation/validation.go
```diff
@@ -57,6 +57,33 @@ func validateTLS(tls *v1alpha1.TLS, fieldPath *field.Path) field.ErrorList {
 	return validateSecretName(tls.Secret, fieldPath.Child("secret"))
 }
 
+func validatePositiveInt(n *int, fieldPath *field.Path) field.ErrorList {
```
pleshakov: positive integers do not include 0. that's why it's better to change the name as well. For example, `validatePositiveIntOrZero`

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -1437,3 +1437,73 @@ func TestValidateUpstreamLBMethodFails(t *testing.T) {
...
+		allErrs := validatePositiveInt(test.number, field.NewPath("int-field"))
+
+		if len(allErrs) != 0 {
+			t.Errorf("validatePositiveInt returned errors for case: %v", test.msg)
+		}
+	}
+}
+
+func TestValidatePositiveIntFails(t *testing.T) {
+	tests := []struct {
```
pleshakov: since there is only one set of input arguments here, having an array of tests as well as the loop `for _, test := range tests` seems redundant.

pkg/apis/configuration/validation/validation.go
```diff
@@ -57,6 +57,33 @@ func validateTLS(tls *v1alpha1.TLS, fieldPath *field.Path) field.ErrorList {
...
+	allErrs := field.ErrorList{}
+	if n == nil {
+		return allErrs
+	}
+
+	if *n < 0 {
+		return append(allErrs, field.Invalid(fieldPath, n, "must be positive"))
```
pleshakov:
```suggestion
		return append(allErrs, field.Invalid(fieldPath, n, "must be positive or zero"))
```

pleshakov:
@Rulox lgtm

Dean-Coakley:
@komapa `max_conns` support was just added in https://github.com/nginxinc/kubernetes-ingress/pull/634 

It is available now via the `:edge` tag.

PR #612, opened 2019-07-03
Add timeouts support to vs/vsr

Dean-Coakley:
### Proposed changes
Support configuring timeouts in upstreams of VirtualServer/VirtualServerRoute

Add support for the following when using to vs/vsrs:
* connect-timeout
* read-timeout 
* send-timeout

### Checklist


Makefile
```diff
@@ -29,9 +29,9 @@ lint:
 
 test:
 ifeq ($(BUILD_IN_CONTAINER),1)
-	$(DOCKER_RUN) $(GOLANG_CONTAINER) go test ./...
+	$(DOCKER_RUN) $(GOLANG_CONTAINER) go test -race ./...
```
pleshakov: this change and the one in line 34 have nothing to do with the PR. please remove.

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -189,6 +192,9 @@ max-fails: 1
...
 | `max-fails` | The number of unsuccessful attempts to communicate with an upstream server that should happen in the duration set by the `fail-timeout` to consider the server unavailable. See the [max_fails](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails) parameter of the server directive. The default is set in the `max-fails` ConfgMap key. | `int` | No |
+`connect-timeout` | The timeout for establishing a connection with an upstream server. See the [proxy_connect_timeout](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_connect_timeout) directive. The default is specified in the `proxy-connect-timeout` ConfigMap key. | `string` | No
+`read-timeout` | The timeout for reading a response from an upstream server. See the [proxy_read_timeout](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout) directive.  The default is specified in the `proxy-read-timeout` ConfigMap key. | `string` | No
+`send-timeout` | The timeout for transmitting a request to an upstream server. See the [proxy_send_timeout](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_send_timeout) directive. The default is 60s. | `string` | No
```
```suggestion
`send-timeout` | The timeout for transmitting a request to an upstream server. See the [proxy_send_timeout](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_send_timeout) directive. The default is `60s`. | `string` | No
```

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -189,6 +192,9 @@ max-fails: 1
...
 | `fail-timeout` | The time during which the specified number of unsuccessful attempts to communicate with an upstream server should happen to consider the server unavailable. See the [fail_timeout](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#fail_timeout) parameter of the server directive. The default is set in the `fail-timeout` ConfigMap key. | `string` | No |
 | `max-fails` | The number of unsuccessful attempts to communicate with an upstream server that should happen in the duration set by the `fail-timeout` to consider the server unavailable. See the [max_fails](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails) parameter of the server directive. The default is set in the `max-fails` ConfgMap key. | `int` | No |
+`connect-timeout` | The timeout for establishing a connection with an upstream server. See the [proxy_connect_timeout](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_connect_timeout) directive. The default is specified in the `proxy-connect-timeout` ConfigMap key. | `string` | No
```
pleshakov: For the added links, use `https://nginx.org` for consistency with the existing links

internal/configs/version1/config.go
```diff
@@ -105,6 +105,7 @@ type Location struct {
 	Upstream             Upstream
 	ProxyConnectTimeout  string
 	ProxyReadTimeout     string
+	ProxySendTimeout     string
```
pleshakov: this is not required. the scope of the PR is for VS/VSR, not Ingress resources.

internal/configs/ingress_test.go
```diff
@@ -161,6 +161,7 @@ func createExpectedConfigForCafeIngressEx() version1.IngressNginxConfig {
 						Upstream:            coffeeUpstream,
 						ProxyConnectTimeout: "60s",
 						ProxyReadTimeout:    "60s",
+						ProxySendTimeout:    "60s",
```
pleshakov: all the changes in this file are not required. See the comment in the `internal/configs/version1/config.go` file.

internal/configs/version2/config.go
```diff
@@ -11,9 +11,12 @@ type VirtualServerConfig struct {
 
 // Upstream defines an upstream.
 type Upstream struct {
-	Name     string
-	Servers  []UpstreamServer
-	LBMethod string
+	Name                string
+	Servers             []UpstreamServer
+	LBMethod            string
+	ProxyConnectTimeout string
```
pleshakov: lines 17-19 are not necessary

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -3,10 +3,9 @@ package validation
 import (
```
pleshakov: no need to change this file

internal/configs/virtualserver.go
```diff
@@ -401,14 +419,14 @@ func generateRulesRouteConfig(route conf_v1alpha1.Route, upstreamNamer *upstream
...
 
-var specialMapParameters = map[string]bool{
-	"default":   true,
-	"hostnames": true,
-	"include":   true,
-	"volatile":  true,
-}
-
 func generateValueForRulesRouteMap(matchedValue string) (value string, isNegative bool) {
+	var specialMapParameters = map[string]bool{
```
pleshakov: this map is a constant. there is no need to put in the `generateValueForRulesRouteMap` func

internal/configs/ingress.go
```diff
@@ -235,6 +235,7 @@ func createLocation(path string, upstream version1.Upstream, cfg *ConfigParams,
 		Upstream:             upstream,
 		ProxyConnectTimeout:  cfg.ProxyConnectTimeout,
 		ProxyReadTimeout:     cfg.ProxyReadTimeout,
+		ProxySendTimeout:     cfg.ProxySendTimeout,
```
pleshakov: this is not required. see the comment in the `internal/configs/version1/config.go` file

internal/configs/virtualserver.go
Rulox: What is the reason to change this to inside the function? I am curious 🤔

internal/configs/virtualserver.go
```diff
@@ -81,9 +81,22 @@ func (namer *variableNamer) GetNameForVariableForRulesRouteMainMap(rulesIndex in
...
+	}
+	return conf_v1alpha1.Upstream{}
+}
+
 func generateVirtualServerConfig(virtualServerEx *VirtualServerEx, tlsPemFileName string, baseCfgParams *ConfigParams, isPlus bool) version2.VirtualServerConfig {
 	ssl := generateSSLConfig(virtualServerEx.VirtualServer.Spec.TLS, tlsPemFileName, baseCfgParams)
 
+	// oUpstreams keeps track of conf_v1alpha1.Upstreams as they are created
+	// necessary to know what Location references which Upstream
+	var oUpstreams []conf_v1alpha1.Upstream
```
pleshakov: use a map to store upstreams. this way you won't need to have `getUpstream` function.

the key of the map should be the generated upstream name. That name is unique among all upstream. Right now there is a bug in your code: consider the case when you have an upstream with the same name  in VS and VSR.

regarding the variable name, I suggest calling it `crUpstreams`, where cr stands fo custom resources.

internal/configs/virtualserver.go
Dean-Coakley: I simply don't like to have global variables unless there's a good reason for them. This seems only used within this func. I think it's also slightly easier to read.

Would you like me to revert this? - Why?

pleshakov: it is a variable but it is a constant. we never change it. there is no need to initialize it every time that function is called.

PR #614, opened 2019-07-03
Added max-conns annotation

vrrs:
### Proposed changes
These changes enable the annotation max-conns which set the parameter max_conns on upstream server in nginx config. For instance, 
`nginx.org/max-conns: 6` on ingress.yaml should add to ngnix config
```
upstream myService-ingress-prod-myService.production.ingress.net-myService-prod-80 {
   server 125.255.255.255:8888 max_fails=3 fail_timeout=10 max_conns=6;
}
```
Here it's the documentation for max_conn https://nginx.org/en/docs/http/ngx_http_upstream_module.html

> max_conns=number
> limits the maximum number of simultaneous active connections to the proxied server (1.11.5). Default value is zero, meaning there is no limit. If the server group does not reside in the shared memory, the limitation works per each worker process.

## Use case
This parameter is very useful on load balancing long lived connections and limiting load on the destination servers. 

### Checklist

### DIFF
--- a/docs/configmap-and-annotations.md
+++ b/docs/configmap-and-annotations.md
@@ -164,7 +164,7 @@ spec:
 | `nginx.org/grpc-services` | N/A | Enables gRPC for services. Note: requires HTTP/2 (see `http2` ConfigMap key); only works for Ingresses with TLS termination enabled. | N/A | [GRPC Services Support](../examples/grpc-services).|
 | `nginx.org/websocket-services` | N/A | Enables WebSocket for services. | N/A | [WebSocket support](../examples/websocket). |
 | `nginx.org/max-fails` | `max-fails` | Sets the value of the [max_fails](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails) parameter of the `server` directive. | `1` | |
-| `nginx.org/max-conns` | `max-conns` | Sets the value of the [max_conns](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_conns) parameter of the `server` directive. | `0` | |
+| `nginx.org/max-conns` | N\A | Sets the value of the [max_conns](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_conns) parameter of the `server` directive. | `0` | |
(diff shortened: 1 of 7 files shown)


internal/configs/annotations.go
```diff
@@ -285,6 +286,14 @@ func parseAnnotations(ingEx *IngressEx, baseCfgParams *ConfigParams, isPlus bool
 		}
 	}
 
+	if maxConns, exists, err := GetMapKeyAsInt(ingEx.Ingress.Annotations, "nginx.org/max-fails", ingEx.Ingress); exists {
```
Dean-Coakley:
```suggestion
	if maxConns, exists, err := GetMapKeyAsInt(ingEx.Ingress.Annotations, "nginx.org/max-conns", ingEx.Ingress); exists {
```

vrrs:
@Dean-Coakley Made the changes suggested.

internal/configs/annotations.go
Dean-Coakley: @vrrs Can you please apply this change too?

vrrs: @Dean-Coakley done

docs/configmap-and-annotations.md
```diff
@@ -164,6 +164,7 @@ spec:
 | `nginx.org/grpc-services` | N/A | Enables gRPC for services. Note: requires HTTP/2 (see `http2` ConfigMap key); only works for Ingresses with TLS termination enabled. | N/A | [GRPC Services Support](../examples/grpc-services).|
 | `nginx.org/websocket-services` | N/A | Enables WebSocket for services. | N/A | [WebSocket support](../examples/websocket). |
 | `nginx.org/max-fails` | `max-fails` | Sets the value of the [max_fails](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails) parameter of the `server` directive. | `1` | |
+| `nginx.org/max-conns` | `max-conns` | Sets the value of the [max_conns](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_conns) parameter of the `server` directive. | `0` | |
```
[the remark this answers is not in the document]
vrrs: @pleshakov done

PR #617, opened 2019-07-05
Add keepalive support to vs/vsr

Dean-Coakley:
### Proposed changes
Support configuring keepalive connections in upstreams of VirtualServer/VirtualServerRoute.

### Checklist


internal/configs/version2/nginx.virtualserver.tmpl
```diff
@@ -8,8 +8,10 @@ upstream {{ $u.Name }} {
     server {{ $s.Address }} max_fails={{ $s.MaxFails }} fail_timeout={{ $s.FailTimeout }};
     {{ end }}
 
-    {{ if $.Keepalive }}
-    keepalive {{ $.Keepalive }};
+    {{ if $u.Keepalive }}
+    keepalive {{ $u.Keepalive }};
+    {{ else }}
+    {{ if $.Keepalive }}keepalive {{ $.Keepalive }};{{ end }}
```
Rulox: There is no need to have `$.Keepalive` / `version2.VirtualServerConfig.Keepalive` now, that was done because it wasn't possible to specify a keepalive for each upstream, but I think that can be replaced now for the `Upstream.Keepalive` now that you have implemented it 🤔 

PS: I think this directive also needs to add the `proxy_set_header Connection "";` in the location blocks of the upstream (ref https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) you can also have a look at the template for the Ingress Resource to see this: https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/version1/nginx.ingress.tmpl#L127

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -31,6 +31,7 @@ type Upstream struct {
 	LBMethod            string `json:"lb-method"`
 	FailTimeout         string `json:"fail-timeout"`
 	MaxFails            *int   `json:"max-fails"`
+	Keepalive           *int64 `json:"keepalive"`
```
Rulox: I think there's no need of using a pointer here. I would say `0` means keepalive is disabled (by default is disabled in the ConfigMap Key, so cfgm will be 0) Check https://github.com/nginxinc/kubernetes-ingress/blob/master/docs/configmap-and-annotations.md the `keepalive` ConfigMap Key

What do you think?

pkg/apis/configuration/validation/validation.go
```diff
@@ -105,6 +118,14 @@ func validateUpstreamLBMethod(lBMethod string, fieldPath *field.Path, isPlus boo
 	return allErrs
 }
 
+func validatePositiveInt64(num int64, fieldPath *field.Path) field.ErrorList {
```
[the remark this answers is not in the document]
Dean-Coakley: Ah yes, rebase fail

pkg/apis/configuration/v1alpha1/types.go
Dean-Coakley: yeah that makes sense

pleshakov: For simplicity, I suggest using`*int`. This way you don't need to have `validatePositiveInt64OrZero` and related functions.

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -192,6 +193,7 @@ send-timeout: 30s
...
 | `fail-timeout` | The time during which the specified number of unsuccessful attempts to communicate with an upstream server should happen to consider the server unavailable. See the [fail_timeout](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#fail_timeout) parameter of the server directive. The default is set in the `fail-timeout` ConfigMap key. | `string` | No |
 | `max-fails` | The number of unsuccessful attempts to communicate with an upstream server that should happen in the duration set by the `fail-timeout` to consider the server unavailable. See the [max_fails](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails) parameter of the server directive. The default is set in the `max-fails` ConfgMap key. | `int` | No |
+| `keepalive` | Configures the cache for connections to upstream servers. See the [keepalive](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) directive. The default is set in the `keepalive` ConfigMap key. | `int` | No
```
```suggestion
| `keepalive` | Configures the cache for connections to upstream servers. The value `0` disables the cache. See the [keepalive](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) directive. The default is set in the `keepalive` ConfigMap key. | `int` | No
```

internal/configs/version2/nginx.virtualserver.tmpl
```diff
@@ -8,8 +8,10 @@ upstream {{ $u.Name }} {
     server {{ $s.Address }} max_fails={{ $s.MaxFails }} fail_timeout={{ $s.FailTimeout }};
     {{ end }}
 
-    {{ if $.Keepalive }}
-    keepalive {{ $.Keepalive }};
+    {{ if $u.Keepalive }}
```
pleshakov: We need to set the keeaplive directive if u.Keepalive > 0. Note that `0` is invalid value here. 
As mentioned https://github.com/nginxinc/kubernetes-ingress/pull/617/files#r301005434, you need to set the connection header to `""` if keepalive is enabled.

pkg/apis/configuration/v1alpha1/types.go
pleshakov: I think we need to use a pointer here. Consider the case if the keeaplive is set in the ConfigMap, and the user wants to disable it. In that case, they can use:
```yaml
name: tea
service: tea-svc
port: 80
keepalive: 0
```
However, if you don't have keepalive as a pointer, the following yaml will also disable keepalives:
```yaml
name: tea
service: tea-svc
port: 80
```

Rulox: If we allow the user to disable it by setting it to 0 then yes a pointer is required @Dean-Coakley

internal/configs/virtualserver_test.go
```diff
@@ -759,6 +764,74 @@ func TestGenerateUpstream(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		result := generateUpstream(name, test.upstream, endpoints, isPlus, test.cfgParams)
+		if !reflect.DeepEqual(result, test.expected) {
+			t.Errorf("generateUpstream() returned %v but expected %v", result, test.expected)
```
pleshakov: please see the comment for the test `TestUpstreamHasKeepalive` about understanding which input causes a failure.

internal/configs/virtualserver.go
```diff
@@ -248,13 +243,22 @@ func generateTime(time string, defaultTime string) string {
...
-func generatePositiveIntFromPointer(n *int, defaultN int) int {
+func generateIntFromPointer(n *int, defaultN int) int {
 	if n == nil {
 		return defaultN
 	}
 	return *n
 }
 
+func upstreamHasKeepalive(upstream conf_v1alpha1.Upstream, cfgParams *ConfigParams) bool {
+	if upstream.Keepalive != nil && *upstream.Keepalive != 0 {
```
pleshakov: I suggest simplifying the logic here:
```
	if upstream.Keepalive != nil {
		return *upstream.Keepalive != 0
	}
	return cfgParams.Keepalive != 0
```

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -23,10 +23,11 @@ func TestValidateVirtualServer(t *testing.T) {
...
 				{
-					Name:     "first",
-					Service:  "service-1",
-					LBMethod: "random",
-					Port:     80,
+					Name:      "first",
+					Service:   "service-1",
+					LBMethod:  "random",
+					Port:      80,
+					Keepalive: &keepalive,
```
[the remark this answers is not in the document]
Dean-Coakley: ignored due to import cycle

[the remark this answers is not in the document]
pleshakov: my bad

PR #628, opened 2019-07-22
Add retries support for vs/vsr

LorcanMcVeigh:
### Proposed changes
Added support and documentation for next-upstream-timeout and next-upstream-retries

### Checklist


docs/virtualserver-and-virtualserverroute.md
```diff
@@ -197,9 +200,12 @@ tls:
...
+| `connect-timeout` | The timeout for establishing a connection with an upstream server. See the [proxy_connect_timeout](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_connect_timeout) directive. The default is specified in the `proxy-connect-timeout` ConfigMap key. | `string` | No
+| `read-timeout` | The timeout for reading a response from an upstream server. See the [proxy_read_timeout](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout) directive.  The default is specified in the `proxy-read-timeout` ConfigMap key. | `string` | No
+| `send-timeout` | The timeout for transmitting a request to an upstream server. See the [proxy_send_timeout](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_send_timeout) directive. The default is specified in the `proxy-send-timeout` ConfigMap key. | `string` | No
+| `next-upstream` | Specifies in which cases a request should be passed to the next upstream server. See the proxy_next_upstream directive. The default is `error timeout` | `string` | No
```
pleshakov: couple of consistency issues:
- each row must end with `|`
- each description must end with `.`

Additionally, please add links to the mentioned directives. See the row for the `send-timeout` field as an example.

examples-of-custom-resources/basic-configuration/cafe-virtual-server.yaml
```diff
@@ -10,9 +10,15 @@ spec:
   - name: tea
```
pleshakov: these changes are not required. please remove.

internal/configs/virtualserver.go
```diff
@@ -264,20 +264,30 @@ func generateProxyPassProtocol(upstream conf_v1alpha1.Upstream) string {
 	return "http"
 }
 
+func generateStringOrDefault(s string, defaultS string) string {
```
pleshakov: this function needs a unit test

pleshakov: Additionally, other similar functions don't have "OrDefault" suffix. Please remove it.

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -25,17 +25,20 @@ type VirtualServerSpec struct {
...
-	Port                uint16      `json:"port"`
-	LBMethod            string      `json:"lb-method"`
-	FailTimeout         string      `json:"fail-timeout"`
-	MaxFails            *int        `json:"max-fails"`
-	Keepalive           *int        `json:"keepalive"`
-	ProxyConnectTimeout string      `json:"connect-timeout"`
-	ProxyReadTimeout    string      `json:"read-timeout"`
-	ProxySendTimeout    string      `json:"send-timeout"`
-	TLS                 UpstreamTLS `json:"tls"`
+	Name                     string      `json:"name"`
```
pleshakov: Please run `./hack/update-codegen.sh` to regenerate the autogenerated code and commit that code.

internal/configs/virtualserver.go
```diff
@@ -264,20 +264,30 @@ func generateProxyPassProtocol(upstream conf_v1alpha1.Upstream) string {
...
+		ProxyReadTimeout:         generateTime(upstream.ProxyReadTimeout, cfgParams.ProxyReadTimeout),
+		ProxySendTimeout:         generateTime(upstream.ProxySendTimeout, cfgParams.ProxySendTimeout),
+		ClientMaxBodySize:        cfgParams.ClientMaxBodySize,
+		ProxyMaxTempFileSize:     cfgParams.ProxyMaxTempFileSize,
+		ProxyBuffering:           cfgParams.ProxyBuffering,
+		ProxyBuffers:             cfgParams.ProxyBuffers,
+		ProxyBufferSize:          cfgParams.ProxyBufferSize,
+		ProxyPass:                fmt.Sprintf("%v://%v", generateProxyPassProtocol(upstream), upstreamName),
+		ProxyNextUpstream:        generateStringOrDefault(upstream.ProxyNextUpstream, "error timeout"),
+		ProxyNextUpstreamTimeout: generateStringOrDefault(upstream.ProxyNextUpstreamTimeout, "0s"),
```
Rulox: There is a function called `generateTime` that does pretty much the same. I think it might be a good time to remove `generateTime` and use `generateString` for times and strings, what do you think? They do pretty much the same but with a different name.

pleshakov: @Rulox good suggestion 👍

internal/configs/virtualserver_test.go
```diff
@@ -895,6 +925,33 @@ func TestGenerateProxyPassProtocol(t *testing.T) {
 	}
 }
 
+func TestGenerateString(t *testing.T) {
+	tests := []struct {
+		nextUpstream string
```
pleshakov: the signature of the function under test is `func generateString(s string, defaultS string) string {`
however, in the test you use `nextUpstream` variable. Also, because the values are related to a specific NGINX directive, this relationship can confuse the reader.

Consider:
- renaming `nextUpstream` to something like `inputString` or other thing related to the function signature 
- using non NGINX directives values like `foo` and `abc`. 

Also, the third test case look very similar to the first one. I think it is redundant. Although if removing one of them, in the remaining please make sure that inputS and default has different values.

pkg/apis/configuration/validation/validation.go
```diff
@@ -156,6 +159,32 @@ func validateUpstreams(upstreams []v1alpha1.Upstream, fieldPath *field.Path, isP
 	return allErrs, upstreamNames
 }
 
+// ValidateNextUpstream checks the values given for passing queries to a upstream
+func validateNextUpstream(nextUpstream string, fieldPath *field.Path) field.ErrorList {
```
pleshakov: this function validates a complicated directive. Because of that, I suggest having a separate unit test for that function with additional cases of valid/invalid values.

For example, consider the following input "error error".

Also, now need to create a map every time this function is invoked. You can put in a global variable like this one https://github.com/nginxinc/kubernetes-ingress/blob/e22d7974018929304473e92b40780cd237e4d935/pkg/apis/configuration/validation/validation.go#L441

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -25,17 +25,20 @@ type VirtualServerSpec struct {
...
+	LBMethod                 string      `json:"lb-method"`
+	FailTimeout              string      `json:"fail-timeout"`
+	MaxFails                 *int        `json:"max-fails"`
+	Keepalive                *int        `json:"keepalive"`
+	ProxyConnectTimeout      string      `json:"connect-timeout"`
+	ProxyReadTimeout         string      `json:"read-timeout"`
+	ProxySendTimeout         string      `json:"send-timeout"`
+	ProxyNextUpstream        string      `json:"next-upstream"`
+	ProxyNextUpstreamTimeout string      `json:"next-upstream-timeout"`
+	ProxyNextUpstreamTries   *int        `json:"next-upstream-tries"`
```
Rulox: Does this need to be a pointer?

pkg/apis/configuration/validation/validation.go
```diff
@@ -156,6 +159,41 @@ func validateUpstreams(upstreams []v1alpha1.Upstream, fieldPath *field.Path, isP
...
+	"http_503":       true,
+	"http_504":       true,
+	"http_403":       true,
+	"http_404":       true,
+	"http_429":       true,
+	"non_idempotent": true,
+	"off":            true,
+}
+
+// ValidateNextUpstream checks the values given for passing queries to a upstream
```
Rulox: You don't need to comment private functions. However, if you do it, I think is better to change the capital `V` for `v` for consistency.

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -237,6 +306,56 @@ func TestValidateUpstreamsFails(t *testing.T) {
...
+		if len(allErrs) > 0 {
+			t.Errorf("validateNextUpstream() returned errors %v for valid input for the case of %s", allErrs, test.inputS)
+		}
+	}
+}
+
+func TestValidateNextUpstreamFails(t *testing.T) {
+	tests := []struct {
+		inputS   string
+		expected bool
```
Rulox: I think `expected` is not used.

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -237,6 +306,56 @@ func TestValidateUpstreamsFails(t *testing.T) {
 	}
 }
 
+func TestValidateNextUpstream(t *testing.T) {
+	tests := []struct {
+		inputS   string
+		expected bool
```
Rulox: I think `expected` is not used.

internal/configs/virtualserver_test.go
```diff
@@ -895,6 +925,33 @@ func TestGenerateProxyPassProtocol(t *testing.T) {
 	}
 }
 
+func TestGenerateString(t *testing.T) {
+	tests := []struct {
+		inputS   string
+		expected string
+	}{
+		{
```
pleshakov: this comment wasn't fully addresses -- https://github.com/nginxinc/kubernetes-ingress/pull/628#discussion_r306830297

pkg/apis/configuration/validation/validation.go
```diff
@@ -156,6 +159,41 @@ func validateUpstreams(upstreams []v1alpha1.Upstream, fieldPath *field.Path, isP
 	return allErrs, upstreamNames
 }
 
+var validVariableParams = map[string]bool{
```
pleshakov: it is a global variable, so the name must be more specific. Consider changing to `validNextUpstreamParams`

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -237,6 +306,48 @@ func TestValidateUpstreamsFails(t *testing.T) {
...
+
+func TestValidateNextUpstreamFails(t *testing.T) {
+	tests := []struct {
+		inputS string
+	}{
+		{
+			inputS: "error error",
+		},
+		{
+			inputS: "",
```
pleshakov: `""` should be valid input, no?

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -237,6 +306,48 @@ func TestValidateUpstreamsFails(t *testing.T) {
...
+		},
+		{
+
+			inputS: "http_404 timeout",
+		},
+	}
+	for _, test := range tests {
+		allErrs := validateNextUpstream(test.inputS, field.NewPath("next-upstreams"))
+		if len(allErrs) > 0 {
+			t.Errorf("validateNextUpstream() returned errors %v for valid input for the case of %s", allErrs, test.inputS)
```
pleshakov: consider printing `validateNextUpstream("error timeout")` similar to the other tests. For example, see https://github.com/nginxinc/kubernetes-ingress/blob/e6d5cf47761f31d1218c688314b02880ed3c7e3b/pkg/apis/configuration/validation/validation_test.go#L67

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -237,6 +306,48 @@ func TestValidateUpstreamsFails(t *testing.T) {
...
+			inputS: "",
+		},
+		{
+			inputS: "https_404",
+		},
+	}
+	for _, test := range tests {
+		allErrs := validateNextUpstream(test.inputS, field.NewPath("next-upstreams"))
+		if len(allErrs) < 0 {
+			t.Errorf("validateNextUpstream() didn't return errors %v for invalid input for the case of %s", allErrs, test.inputS)
```
pleshakov: same comment as for the line `324`

pkg/apis/configuration/validation/validation.go
```diff
@@ -156,6 +159,41 @@ func validateUpstreams(upstreams []v1alpha1.Upstream, fieldPath *field.Path, isP
...
+// validateNextUpstream checks the values given for passing queries to a upstream
+func validateNextUpstream(nextUpstream string, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	var occur int
+	params := strings.Fields(nextUpstream)
+	for _, para := range params {
+		if !validVariableParams[para] {
+			allErrs = append(allErrs, field.Invalid(fieldPath, para, "not a valid parameter"))
+		}
+		for _, p := range params {
```
pleshakov: this part - lines 186-190 could be greatly simplified if you use a Set. For example, see https://github.com/nginxinc/kubernetes-ingress/blob/e6d5cf47761f31d1218c688314b02880ed3c7e3b/pkg/apis/configuration/validation/validation.go#L228

pkg/apis/configuration/validation/validation.go
```diff
@@ -156,6 +159,41 @@ func validateUpstreams(upstreams []v1alpha1.Upstream, fieldPath *field.Path, isP
...
+			if p == para {
+				occur++
+			} else if occur > 1 {
+				allErrs = append(allErrs, field.Invalid(fieldPath, para, "can not have recurring parameters"))
```
pleshakov: Recurring is not as commonly used as `Duplicated`, which we use in other places. Please consider changing that.

pkg/apis/configuration/validation/validation_test.go
LorcanMcVeigh: will do, note there are a good few tests in the code that don't follow this practice

LorcanMcVeigh: will do, note many other tests don't follow this practice

pkg/apis/configuration/validation/validation.go
LorcanMcVeigh: but this is to check if theres are duplicates what would make it a bad/wrong input, making it a set would removed the duplicates that the code is checking for, no?

pleshakov: you can use a Set to check for duplicates. When you process a parameter and check that is is valid, you add it to a Set. for the next parameter, after you check it is valid, you see if it is in the set ... if it is in the set, you have a duplicate...

pkg/apis/configuration/validation/validation_test.go
pleshakov: yep. the other tests have more complicated arguments (structs) that are not suitable for printing  like that

pkg/apis/configuration/validation/validation.go
LorcanMcVeigh: ahh okay, thanks for clearing that up 👍

pkg/apis/configuration/validation/validation.go
```diff
@@ -156,6 +159,40 @@ func validateUpstreams(upstreams []v1alpha1.Upstream, fieldPath *field.Path, isP
...
+	"http_500":       true,
+	"http_502":       true,
+	"http_503":       true,
+	"http_504":       true,
+	"http_403":       true,
+	"http_404":       true,
+	"http_429":       true,
+	"non_idempotent": true,
+	"off":            true,
+	"":               true,
```
pleshakov: this is the recent change. better to handle that case in the `validateNextUpstream`. similar to https://github.com/nginxinc/kubernetes-ingress/blob/d9f88740990e27bf12fe44e0fbf9a1e20cd64e59/pkg/apis/configuration/validation/validation.go#L76
Better to add this case to the `TestValidateNextUpstream`

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -314,14 +314,13 @@ func TestValidateNextUpstream(t *testing.T) {
...
 		{
-
 			inputS: "http_404 timeout",
 		},
 	}
 	for _, test := range tests {
 		allErrs := validateNextUpstream(test.inputS, field.NewPath("next-upstreams"))
 		if len(allErrs) > 0 {
-			t.Errorf("validateNextUpstream() returned errors %v for valid input for the case of %s", allErrs, test.inputS)
+			t.Errorf("validateNextUpstream(%q) returned errors %v for valid input for the case of %s", test.inputS, allErrs, test.inputS)
```
pleshakov: ` for the case of %s` is redundant...

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -333,17 +332,14 @@ func TestValidateNextUpstreamFails(t *testing.T) {
...
-		},
 		{
 			inputS: "https_404",
 		},
 	}
 	for _, test := range tests {
 		allErrs := validateNextUpstream(test.inputS, field.NewPath("next-upstreams"))
 		if len(allErrs) < 0 {
-			t.Errorf("validateNextUpstream() didn't return errors %v for invalid input for the case of %s", allErrs, test.inputS)
+			t.Errorf("validateNextUpstream(%q) didn't return errors %v for invalid input for the case of %s", test.inputS, allErrs, test.inputS)
```
pleshakov: please see the comment for the line 323

PR #629, opened 2019-07-22
Added upstream zone directive annotation 

vrrs:
### Proposed changes
These changes enable the zone directive on upstream block with upstream name equal to zone name and zone size equals to the annotation value. This follow the same convention as nginx-plus ingress config. For instance, `nginx.org/upstream-zone-size: "32k"` in your ingress.yaml will generate a config such as
```
upstream myService-ingress-prod-myService.production.ingress.net-myService-prod-80 {
   zone myService-ingress-prod-myService.production.ingress.net-myService-prod-80 32k;
   server 125.255.255.255:8888 max_fails=3 fail_timeout=10 max_conns=6;
}
```
If the annotation is not present then the zone directive will not be present on the config. 

[Nginx Documentation for upstream zone](https://nginx.org/en/docs/http/ngx_http_upstream_module.html?#zone)

| | |
-- | --
Syntax: | zone **name** [size]; 
Default: | N/A 
Context: | upstream 

## Motivation
In multi worker nginx setup([worker_processes](http://nginx.org/en/docs/ngx_core_module.html#worker_processes) > 1), the zone directive fixes the least_conn lb method as well as max_fails, max_conns parameters. From the nginx document on [load balancing](https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/#zone-size):
>For example, if the configuration of a group is not shared, each worker process maintains its own counter for failed attempts to pass a request to a server (set by the max_fails parameter). In this case, each request gets to only one worker process. When the worker process that is selected to process a request fails to transmit the request to a server, other worker processes don’t know anything about it. While some worker process can consider a server unavailable, others might still send requests to this server. For a server to be definitively considered unavailable, the number of failed attempts during the timeframe set by the fail_timeout parameter must equal max_fails multiplied by the number of worker processes. On the other hand, the zone directive guarantees the expected behavior.
Similarly, the Least Connections load‑balancing method might not work as expected without the zone directive, at least under low load.

## Use case
A properly working least_conn load balancer is very useful in balancing long lived connections and limiting load on the destination servers; particularly when upstream servers are streaming web apps. 

### Checklist

### DIFF
--- a/docs/configmap-and-annotations.md
+++ b/docs/configmap-and-annotations.md
@@ -166,7 +166,7 @@ spec:
 | `nginx.org/websocket-services` | N/A | Enables WebSocket for services. | N/A | [WebSocket support](../examples/websocket). |
 | `nginx.org/max-fails` | `max-fails` | Sets the value of the [max_fails](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails) parameter of the `server` directive. | `1` | |
 | `nginx.org/max-conns` | N\A | Sets the value of the [max_conns](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_conns) parameter of the `server` directive. | `0` | |
-| `nginx.org/upstream-zone-size` | N\A | Sets the size of the shared memory [zone](https://nginx.org/en/docs/http/ngx_http_upstream_module.html?#zone). If this annotation is not present then the zone directive will not be set. | `N\A` | |
+| `nginx.org/upstream-zone-size` | `upstream-zone-size` | Sets the size of the shared memory [zone](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#zone) for upstreams. For NGINX, the special value `0` disables the shared memory zones. | `256K` | |
 | `nginx.org/fail-timeout` | `fail-timeout` | Sets the value of the [fail_timeout](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#fail_timeout) parameter of the `server` directive. | `10s` | |
 | `nginx.com/sticky-cookie-services` | N/A | Configures session persistence. | N/A | [Session Persistence](../examples/session-persistence). |
 | `nginx.org/keepalive` | `keepalive` | Sets the value of the [keepalive](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) directive. Note that `proxy_set_header Connection "";` is added to the generated configuration when the value > 0. | `0` | |
--- a/docs/configmap-and-annotations.md
+++ b/docs/configmap-and-annotations.md
@@ -166,6 +166,7 @@ spec:
 | `nginx.org/websocket-services` | N/A | Enables WebSocket for services. | N/A | [WebSocket support](../examples/websocket). |
 | `nginx.org/max-fails` | `max-fails` | Sets the value of the [max_fails](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails) parameter of the `server` directive. | `1` | |
 | `nginx.org/max-conns` | N\A | Sets the value of the [max_conns](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_conns) parameter of the `server` directive. | `0` | |
+| `nginx.org/upstream-zone-size` | N\A | Sets the size of the shared memory [zone](https://nginx.org/en/docs/http/ngx_http_upstream_module.html?#zone). If this annotation is not present then the zone directive will not be set. | `N\A` | |
 | `nginx.org/fail-timeout` | `fail-timeout` | Sets the value of the [fail_timeout](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#fail_timeout) parameter of the `server` directive. | `10s` | |
 | `nginx.com/sticky-cookie-services` | N/A | Configures session persistence. | N/A | [Session Persistence](../examples/session-persistence). |
 | `nginx.org/keepalive` | `keepalive` | Sets the value of the [keepalive](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) directive. Note that `proxy_set_header Connection "";` is added to the generated configuration when the value > 0. | `0` | |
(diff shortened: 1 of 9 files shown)


pleshakov:
@vrrs thanks for the PR! We'll review it shortly

pleshakov:
@vrrs 
Thanks for bringing this PR! Having zones in the config for enabling passive health checks and the least conn algorithm to work across multiple workers for NGINX OSS makes sense. We should've had it implemented before.

I wonder though if it makes sense to have a configmap key `upstream-zone-size` instead of the suggestion annotation. It seems that it is a more common case for the zone size to be configured/enabled per all Ingress resources than an per individual Ingress resource. For example, if the Ingress Controller is configured with multiple NGINX workers, why not to enable zones for all Ingresses? Does that apply for your requirements as well? 

Additionally, having zones enabled by default also makes sense, so that the mentioned benefits are enabled by default. The default size of `256K` will work for the case of more than 100 endpoints in a single upstream. We use the same size in the upstreams of NGINX Plus. Having a special value `0` to remove the zones from upstreams is reasonable to me.

Let me know what you think.

vrrs:
@pleshakov Our use case requires individual resources to specify max_conn and zone since our ingress nginx is shared. Different set of upstream servers might require different zone sizes
or no zones at all. In addition, this change is backward compatible. We have already deployments without upstream zone and adding a default zone might have an impact on performance. 
However, We can add a configmap key with a special value(0) as you suggested and set it to 0 as default. This way it covers both use cases.

In summary, We should add a configmap key along with the annotation defaulted to 0 which will enable both use cases of having all ingresses with one zone(using configmap key) or having individual ingresses with specific zones(with annotation).
Let me know what you think.

docs/configmap-and-annotations.md
```diff
@@ -166,6 +166,7 @@ spec:
 | `nginx.org/websocket-services` | N/A | Enables WebSocket for services. | N/A | [WebSocket support](../examples/websocket). |
 | `nginx.org/max-fails` | `max-fails` | Sets the value of the [max_fails](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails) parameter of the `server` directive. | `1` | |
 | `nginx.org/max-conns` | N\A | Sets the value of the [max_conns](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_conns) parameter of the `server` directive. | `0` | |
+| `nginx.org/upstream-zone-size` | N\A | Sets the size of the shared memory [zone](https://nginx.org/en/docs/http/ngx_http_upstream_module.html?#zone). If this annotation is not present then the zone directive will not be set. | `N\A` | |
```
pleshakov:
```suggestion
| `nginx.org/upstream-zone-size` | `upstream-zone-size` | Sets the size of the shared memory [zone](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#zone) for upstreams. For NGINX, the special value `0` disables the shared memory zones. | `256K` | |
```

internal/configs/version1/nginx.ingress.tmpl
```diff
@@ -2,6 +2,7 @@
 
 {{range $upstream := .Upstreams}}
 upstream {{$upstream.Name}} {
+	{{if $upstream.UpstreamZoneSize }}zone {{$upstream.Name}} {{$upstream.UpstreamZoneSize}};{{end}}
```
pleshakov:
```suggestion
	{{if $upstream.UpstreamZoneSize}}zone {{$upstream.Name}} {{$upstream.UpstreamZoneSize}};{{end}}
```

vrrs:
@pleshakov Performed the changes

vrrs:
@Rulox @Dean-Coakley 👀 👀 --^

docs/configmap-and-annotations.md
Dean-Coakley: Looks like you missed https://github.com/nginxinc/kubernetes-ingress/pull/629#discussion_r307377239 unless I'm mistaken?

internal/configs/config_params.go
```diff
@@ -109,6 +110,7 @@ func NewDefaultConfigParams() *ConfigParams {
 		SSLPorts:                   []int{443},
 		MaxFails:                   1,
 		MaxConns:                   0,
+		UpstreamZoneSize: 			"256k",
```
Dean-Coakley: Looks like this is missing a run of `gofmt`, not gonna block the pr for it, but would be nice.

Rulox: `go fmt` should be run for consistency before merging

vrrs:
@Dean-Coakley Modified the documentation.

Dean-Coakley:
@vrrs The format of config_params.go still isn't quite right. 
Please run `$ gofmt -w internal/configs/config_params.go`

vrrs:
Ohh i see, check  now @Dean-Coakley

Rulox:
@vrrs Looks great, just 2 small things before I can merge:

1. It seems this https://github.com/nginxinc/kubernetes-ingress/pull/629#discussion_r307377917 was missed? (it's just removing the space there)

2. Make sure you follow the (style guides)[https://github.com/nginxinc/kubernetes-ingress/blob/master/CONTRIBUTING.md#git-style-guide]. You need to rebase against `master` and also squash all your commits into one, using a present tense commit message like `Add upstream zone directive annotation/configmap`.

Both changes are just to follow the rules and consistency with the rest of the code/repo.

Once that's done I will merge it, thanks again for your time!

vrrs:
@Rulox 
1. Done. FYI, the lbmethod upstream line is not following this standard which is the one i was following. 
2. Regarding squashed commits, when merging you can select squash and merge. I always rebase before submitting PRs and using one commit(From the guideline: `Keep a clean, concise and meaningful git commit history on your branch, rebasing locally and squashing before submitting a PR`). The extra commits you see were as a consequence of the review. This branch has no merge conflicts. Am i suppose to close this PR and create a new one ?

Rulox:
@vrrs 

1. Yes I noticed when reviewing, it's a mistake, I'll fix it.
2. No worries, I'll squash them

👍

PR #632, opened 2019-07-24
Fix hsts support when not using SSL

### LINKED ISSUE #627: HSTS not working using hsts-behind-proxy
Hello,

I'm trying to setup HSTS support using annotations in the ingress but I'm not having luck with it. We have the nginx-ingress in front of an AWS ALB that handles SSL termination.

This is how the ingress looks like:

```
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: nginx
    nginx.org/hsts: "True"
    nginx.org/hsts-behind-proxy: "True"
    nginx.org/hsts-include-subdomains: "True"
    nginx.org/hsts-max-age: "2592000"
    nginx.org/mergeable-ingress-type: master
    nginx.org/redirect-to-https: "True"
  creationTimestamp: 2019-07-03T15:02:57Z
  generation: 1
  name: nginx-host-master-app.dev.company.net
  namespace: default
  resourceVersion: "50956162"
  selfLink: /apis/extensions/v1beta1/namespaces/default/ingresses/nginx-host-master-app.dev.company.net
  uid: a434f4e8-9da3-11e9-9d08-1296d0bc4772
spec:
  rules:
  - host: app.dev.company.net
status:
  loadBalancer:
    ingress:
    - {}
```

I've checked the dynamic configuration that is built by the controller under `/etc/nginx/conf.d/` and the expected directives are not there. 

I verified that changing `nginx.org/proxy-connect-timeout` did actually make a change on the nginx configuration so the annotations are working to a degree.

**Expected behavior**
The HSTS relevant headers are added in the response.

**Your environment**
* Version of the Ingress Controller: 1.5.1-alpine
* Version of Kubernetes: 1.11
* Kubernetes platform: EKS
* Using NGINX or NGINX Plus: NGINX

bilby91:
Closes #627

### Proposed changes
This change will fix the HSTS support when not handling SSL at the ingress layer. Previously, the templating code would only render the HSTS related directives if the `SSL` was true which is never the case when a load balancer that sits in front of the ingress handle the TLS.  

Fixes the following issue: https://github.com/nginxinc/kubernetes-ingress/issues/627

### Checklist

### DIFF
--- a/internal/configs/version1/nginx-plus.ingress.tmpl
+++ b/internal/configs/version1/nginx-plus.ingress.tmpl
@@ -53,9 +53,7 @@ server {
 	proxy_pass_header {{$proxyPassHeader}};{{end}}
 	{{end}}
 
-	{{if $server.SSL}}
-	{{if not $server.GRPCOnly}}
-	{{- if $server.HSTS}}
+	{{- if and $server.HSTS (or $server.SSL $server.HSTSBehindProxy)}}
 	set $hsts_header_val "";
 	proxy_hide_header Strict-Transport-Security;
 	{{- if $server.HSTSBehindProxy}}
@@ -69,6 +67,8 @@ server {
 	add_header Strict-Transport-Security "$hsts_header_val" always;
 	{{end}}
 
+	{{if $server.SSL}}
+	{{if not $server.GRPCOnly}}
 	{{- if $server.SSLRedirect}}
 	if ($scheme = http) {
 		return 301 https://$host:{{index $server.SSLPorts 0}}$request_uri;
--- a/internal/configs/version1/nginx.ingress.tmpl
+++ b/internal/configs/version1/nginx.ingress.tmpl
@@ -39,9 +39,7 @@ server {
 	{{range $proxyPassHeader := $server.ProxyPassHeaders}}
 	proxy_pass_header {{$proxyPassHeader}};{{end}}
 
-	{{if $server.SSL}}
-	{{if not $server.GRPCOnly}}
-	{{- if $server.HSTS}}
+	{{- if and $server.HSTS (or $server.SSL $server.HSTSBehindProxy)}}
 	set $hsts_header_val "";
 	proxy_hide_header Strict-Transport-Security;
 	{{- if $server.HSTSBehindProxy}}
@@ -55,6 +53,8 @@ server {
 	add_header Strict-Transport-Security "$hsts_header_val" always;
 	{{end}}
 
+	{{if $server.SSL}}
+	{{if not $server.GRPCOnly}}
 	{{- if $server.SSLRedirect}}
 	if ($scheme = http) {
 		return 301 https://$host:{{index $server.SSLPorts 0}}$request_uri;


pleshakov:
@bilby91 thanks for the PR! we will review it shortly.

bilby91:
@Dean-Coakley Good catch, I totally missed that. Do you think that moving the block of code on top of `{{if $server.SSL}}` would make sense ? 

BTW, do you have a cheap way of testing this ? The way I tested this took me some time to get stuff setup.

Dean-Coakley:
@bilby91 I think this issue is more complex than we initially expected. I'd recommend waiting until  we can provide you with some guidance on how to meet all edge cases. But feel free to share commits/comments in the meantime if you feel so inclined.


There is no way to test without a cluster - if that's what you're hoping for. However you don't need to test with a cloud loadbalancer. I'm testing against a minikube cluster with the [complete-example](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/complete-example) deployed. And then use curl `curl -k http://<MINIKUBE>/tea -I -L -H "X-Forwarded-Proto: https"` Perhaps that's easier for you.


You can also run the binary locally against a running cluster. You'd set this up by
1. Compiling the binary `go build -installsuffix cgo -ldflags "-w -X main.version=my-version" -o nginx-ingress github.com/nginxinc/kubernetes-ingress/cmd/nginx-ingress`
2. Run `kubectl proxy` is a separate terminal window
3. Make sure the default server secret is deployed in your cluster
4. Run the binary `./nginx-ingress -proxy 127.0.0.1:8001 -default-server-tls-secret=nginx-ingress/default-server-secret -v=3 -main-template-path=internal/nginx/templates/nginx.tmpl -ingress-template-path=internal/nginx/templates/nginx.ingress.tmpl`

This will help you debug template compilation. However please note _**this will not substitute testing in a cluster**_. There is no nginx running here. You can use this for dev purposes before testing in a cluster.

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -55,6 +55,14 @@ server {
...
 	{{if not $server.GRPCOnly}}
+	{{- if $server.SSLRedirect}}
+	if ($scheme = http) {
+		return 301 https://$host:{{index $server.SSLPorts 0}}$request_uri;
+	}
+	{{- end}}
+	{{end}}
+	{{- end}}
+
 	{{- if $server.HSTS}}
```
pleshakov: with this suggestion HSTS will only be enabled when TLS termination is enabled and `hsts-behind-proxy` annotation/configmapkey is true.
```suggestion
	{{- if and $server.HSTS (or $server.SSL $server.HSTSBehindProxy)}}
```

pleshakov: Additionally, please move lines 66-78 (the `{{- if and $server.HSTS (or $server.SSL $server.HSTSBehindProxy)}} ... {{end}}` block) above line 56 (`{{if $server.SSL}})`. This will fix the warning message like this `*84 using uninitialized "hsts_header_val" variable`.

bilby91: @pleshakov Thanks for taking the time to review the change! The suggested changes would apply for nginx-plus and regular nginx right ? Just making sure!

pleshakov: yep. for both templates! thanks for double checking.

bilby91: Pushed suggested changes!

Dean-Coakley:
Thanks a lot @bilby91 !

PR #634, opened 2019-07-24
Add Active Connections support to vs/vsr

Dean-Coakley:
### Proposed changes
Add support to be able to configure max-conns for an upstream in my VirtualServer/VirtualServerRoute.

### Checklist

### DIFF
--- a/pkg/apis/configuration/v1alpha1/types.go
+++ b/pkg/apis/configuration/v1alpha1/types.go
@@ -31,6 +31,7 @@ type Upstream struct {
 	LBMethod                 string      `json:"lb-method"`
 	FailTimeout              string      `json:"fail-timeout"`
 	MaxFails                 *int        `json:"max-fails"`
+	MaxConns                 *int        `json:"max-conns"`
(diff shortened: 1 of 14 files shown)


pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -31,6 +31,7 @@ type Upstream struct {
 	LBMethod            string      `json:"lb-method"`
 	FailTimeout         string      `json:"fail-timeout"`
 	MaxFails            *int        `json:"max-fails"`
+	MaxConns            *int        `json:"max-conns"`
```
Rulox: Why do we need a pointer here? 🤔

Dean-Coakley: @Rulox My thinking was if you set MaxConns globally via configmap you may want to overwrite for one vs/vsr back to `0` (equivalent to unset).

Rulox: Yup ok.

PR #635, opened 2019-07-24
Add HealthChecks support for vs/vsr

Rulox:
### Proposed changes
Add HealthCheck features for VS and VSR (NGINX Plus only).


pkg/apis/configuration/validation/validation.go
```diff
@@ -57,6 +57,16 @@ func validateTLS(tls *v1alpha1.TLS, fieldPath *field.Path) field.ErrorList {
 	return validateSecretName(tls.Secret, fieldPath.Child("secret"))
 }
 
+func validatePositiveInt(n int, fieldPath *field.Path) field.ErrorList {
```
LorcanMcVeigh: technically zero isn't positive

LorcanMcVeigh: but lgtm 👍

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -82,6 +82,20 @@ server {
     }
     {{ end }}
 
+    {{ range $hc := $s.HealthChecks }}
+    location @hc-{{ $hc.Name }} {
+        {{ range $n, $v := $hc.Headers }}
+        proxy_set_header {{ $n }} {{ $v }};
```
pleshakov: Because we allow spaces in headers
```suggestion
        proxy_set_header {{ $n }} "{{ $v }}";

```

pkg/apis/configuration/validation/validation.go
```diff
@@ -105,6 +115,82 @@ func validateUpstreamLBMethod(lBMethod string, fieldPath *field.Path, isPlus boo
...
+	}
+
+	for _, msg := range isValidHeaderValue(h.Value) {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("value"), h.Value, msg))
+	}
+
+	return allErrs
+}
+
+const headerValueFmt = "[^{};'\"]*"
```
pleshakov: I think here we need to have a validation similar to machValue - a valid value must have all '"' (double quotes) escaped and must not end with an unescaped '\' (backslash). And additionally, we can disallow using variables by banning `$` sign. Note, for health check most of NGINX built in variables will be empty, because they are typically related to request/response.

pkg/apis/configuration/validation/validation.go
```diff
@@ -105,6 +115,82 @@ func validateUpstreamLBMethod(lBMethod string, fieldPath *field.Path, isPlus boo
...
+	if hc.Port > 0 {
+		for _, msg := range validation.IsValidPortNum(int(hc.Port)) {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("port"), hc.Port, msg))
+		}
+	}
+
+	return allErrs
+}
+
+const statusMatchFmt = "(\\d*|\\s|\\-|\\!\\s)*"
```
pleshakov: for status match the validation rules should be 
possible ! followed by valid (in the sense that NGINX config parser doesn't through any errors) status codes (100-999) or valid status code ranges

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -201,12 +203,68 @@ tls:
...
+  - name: Host
+    value: my.service
+  statusMatch: "! 500"
+```
+
+| Field | Description | Type | Required |
+| ----- | ----------- | ---- | -------- |
+| `enable` | Enables a health check for an upstream server. The default is `false`. | `boolean` | No |
+| `path`  | The path used for health check requests. The default is `/`. | `string` | No |
+| `interval` | The interval between two consecutive health checks. The default is `5s`. | `time` | No |
```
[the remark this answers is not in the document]
Rulox: Yes that makes sense 👍

pkg/apis/configuration/validation/validation.go
Rulox: I'll update. I think there must be a space ` ` after the ! `! 500` instead of `!500`  as well. 👍

pkg/apis/configuration/validation/validation.go
Rulox: Sounds good

Rulox:
Ready to review

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -1620,3 +1652,210 @@ func TestValidateTimeFails(t *testing.T) {
...
+				Value: "my.service",
+			},
+		},
+		StatusMatch: "! 500",
+	}
+
+	allErrs := validateUpstreamHealthCheck(hc, field.NewPath("healthCheck"))
+
+	if len(allErrs) != 0 {
+		t.Errorf("validateUpstreamHealthCheck returned errors for valid input %v", hc)
```
pleshakov: small comment about consistency for all added test functions. In existing test functions in error messages we use `function()` - i.e. with the name of the function and parentheses.

pkg/apis/configuration/validation/validation.go
```diff
@@ -105,6 +116,116 @@ func validateUpstreamLBMethod(lBMethod string, fieldPath *field.Path, isPlus boo
...
+}
+
+const headerValueFmt = `([^{};$"\\]|\\.)*`
+const headerValueFmtErrMsg string = `a valid header value must not include '$', '{', '}' or ';', must have all '"' escaped and must not end with an unescaped '\'`
+
+var headerValueFmtRegexp = regexp.MustCompile("^" + headerValueFmt + "$")
+
+func isValidHeaderValue(s string) []string {
+	if !headerValueFmtRegexp.MatchString(s) {
+		return []string{validation.RegexError(headerValueFmtErrMsg, headerValueFmt, "my.service", "\"my service\"")}
```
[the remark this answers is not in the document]
Rulox: Let me see if I can double scape the `\` otherwise I'll simplify

pkg/apis/configuration/validation/validation.go
```diff
@@ -105,6 +116,116 @@ func validateUpstreamLBMethod(lBMethod string, fieldPath *field.Path, isPlus boo
...
+	}
+
+	statuses := strings.Split(s, " ")
+	for i, value := range statuses {
+		if value == "!" {
+			if i != 0 {
+				allErrs = append(allErrs, field.Invalid(fieldPath, s, "`!` can only appear once at the beginning"))
+			}
+		} else if strings.Contains(s, "-") {
+			statusRange := strings.Split(value, "-")
```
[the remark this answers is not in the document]
Rulox: I'll have to rewrite all this block, otherwise is going to be impossible to read, this will take time.

pkg/apis/configuration/validation/validation_test.go
Rulox: They are mixed actually, there's no consistency in the whole file

pkg/apis/configuration/validation/validation.go
```diff
@@ -184,15 +188,49 @@ func validateStatusMatch(s string, fieldPath *field.Path) field.ErrorList {
...
+	numberInt, err := strconv.ParseInt(number, 10, 64)
 	if err != nil {
-		return fmt.Sprintf("%v must be a valid integer", status)
+		return 0, fmt.Sprintf("%v must be a valid integer", number)
+	}
+
+	return int(numberInt), ""
+}
+
+func validateStatusCodeInRange(status string) string {
```
pleshakov: with this addition, we now have two functions with very similar names `validateStatusCodeInRange` and `validateStatusCodeRange`. For clarity, I suggest renaming `validateStatusCodeInRange` to `validateStatusCode`. So that `validateStatusCode` validates an individual status code and `validateStatusCodeRange` validates a range.

pkg/apis/configuration/validation/validation.go
```diff
@@ -184,15 +188,49 @@ func validateStatusMatch(s string, fieldPath *field.Path) field.ErrorList {
 	return allErrs
 }
 
-func validateStatusCodeInRange(status string) string {
-	code, err := strconv.ParseInt(status, 10, 64)
+func validateStatusCodeRange(statusRange []string) string {
```
pleshakov: this function always uses  2 elements of the  statusRange slice. Because of that, I suggest changing the signature to take 2 arguments. For example, `validateStatusCodeRange(min sting, max string)`. This will make the code easier to understand.

Alternatively, it is possible to bring the following the code into this function and make it take a string argument like `validateStatusCodeRange(statusRange string)`
```
statusRange := strings.Split(value, "-")
if len(statusRange) != 2 {
				allErrs = append(allErrs, field.Invalid(fieldPath, s, "ranges must only have 2 numbers"))
			}
```
Let me know what you think
```

File: pkg/apis/configuration/validation/validation.go [hunk 11]
```diff
@@ -214,14 +252,14 @@ func validateHeader(h v1alpha1.Header, fieldPath *field.Path) field.ErrorList {
 	return allErrs
 }
 
-const headerValueFmt = `([^{};$"\\]|\\.)*`
-const headerValueFmtErrMsg string = `a valid header value must not include '$', '{', '}' or ';', must have all '"' escaped and must not end with an unescaped '\'`
+const headerValueFmt = `([^"\\]|\\.)*`
```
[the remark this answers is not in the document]
Rulox: Yes

pkg/apis/configuration/validation/validation.go
Rulox: That won't work, the check has to be done both in ranges and separate status codes. Renaming to `validateStatusCode` but checking that the code is in the range there too. It's still a validation of the status code so it will work.

PR #641, opened 2019-08-01
Add support for ExternalName Services for vs/vsr

Rulox:
### Proposed changes
Upstreams in VS and VSR can reference ExternalName services now.


internal/k8s/controller.go
```diff
@@ -1504,10 +1504,25 @@ func (lbc *LoadBalancerController) createVirtualServer(virtualServer *conf_v1alp
...
+			ServiceName: u.Service,
+			ServicePort: intstr.FromInt(int(u.Port)),
+		}
+
+		svc, err := lbc.getServiceForIngressBackend(backend, virtualServer.Namespace)
+		if err != nil {
+			glog.V(3).Infof("Error getting service %v: %v", u.Service, err)
+		} else {
+			endpointsKey := configs.GenerateEndpointsKey(virtualServer.Namespace, u.Service, u.Port)
+			endps, external, err := lbc.getEndpointsForIngressBackend(backend, virtualServer.Namespace, svc)
```
pleshakov: in case of an error here and in line 1579, could we report a warning similarly to what we do in `createIngressEx` https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/k8s/controller.go#L1408

internal/k8s/controller.go
```diff
@@ -1551,32 +1566,37 @@ func (lbc *LoadBalancerController) createVirtualServer(virtualServer *conf_v1alp
...
-		ServiceName: name,
-		ServicePort: intstr.FromInt(port),
-	}
-
-	svc, err := lbc.getServiceForIngressBackend(backend, namespace)
-	if err != nil {
-		glog.Warningf("Error getting service %v: %v", name, err)
-		return nil
-	}
+func (lbc *LoadBalancerController) getEndpointsForService(svc *api_v1.Service, backend *extensions.IngressBackend) []string {
```
pleshakov: this new function is not used anywhere. do we need it?

internal/k8s/controller.go
```diff
@@ -1551,32 +1566,37 @@ func (lbc *LoadBalancerController) createVirtualServer(virtualServer *conf_v1alp
...
 	}
 
 	virtualServerEx.Endpoints = endpoints
 	virtualServerEx.VirtualServerRoutes = virtualServerRoutes
+	virtualServerEx.ExternalNameSvcs = externalNameSvcs
 
 	return &virtualServerEx, virtualServerRouteErrors
 }
 
-func (lbc *LoadBalancerController) getEndpointsForService(namespace string, name string, port int) []string {
```
pleshakov: please note this function existed to separate any Ingress-related code (functions like `getServiceForIngressBackend` and structs like `extensions.IngressBackend` from the virtual server related code so that in `createVirtualServer` we don't use any Ingress related code directly.

internal/configs/virtualserver.go
```diff
@@ -198,15 +198,24 @@ func generateVirtualServerConfig(virtualServerEx *VirtualServerEx, tlsPemFileNam
 	}
 }
 
-func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, endpoints []string, isPlus bool, cfgParams *ConfigParams) version2.Upstream {
+func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, virtualServerEx *VirtualServerEx, isPlus bool, isResolverConfigured bool, cfgParams *ConfigParams) version2.Upstream {
```
pleshakov: I don't think ` virtualServerEx *VirtualServerEx` is neither required nor at the right level of abstractions for `generateUpstream`. I know we have something similar in the old code (like `createUpstream` in `ingress.go`). However, for the new code, I would rather not use the old code as a good example. 

As a bad consequece of this change, a lot of unnecessary test input data was created in generateUpstream tests. For example:
```
upstream := conf_v1alpha1.Upstream{Service: name, Port: 80}
	vsEx := &VirtualServerEx{
		VirtualServer: &conf_v1alpha1.VirtualServer{
			ObjectMeta: meta_v1.ObjectMeta{
				Name:      name,
				Namespace: "test-namespace",
			},
		},
		Endpoints: map[string][]string{
			"test-namespace/test-upstream:80": []string{
				"192.168.10.10:8080",
			},
		},
	}
```
Perhaps two arguments in this case - `endpoints[]string` and `externalNameSvcs map[string]bool` would be more suitable for `generateUpstream` instead of `virtualServerEx *VirtualServerEx`. While we have 2 arguments instead of one, both arguments are required by `generateUpstream` and at the right level of abstraction.

internal/configs/virtualserver_test.go
```diff
@@ -894,12 +929,109 @@ func TestGenerateUpstreamForZeroEndpoints(t *testing.T) {
...
+		upstream             conf_v1alpha1.Upstream
+		vsEx                 *VirtualServerEx
+		isPlus               bool
+		isResolverConfigured bool
+		expected             version2.Upstream
+	}{
+		{
+			upstream:             conf_v1alpha1.Upstream{},
+			vsEx:                 vsEx,
+			isPlus:               false,
```
pleshakov: for NGINX OSS, ExternalNameSvcs map will always be empty, as mandated by the following code https://github.com/nginxinc/kubernetes-ingress/blob/211b8f98bb2cbf4e1e83be040b161ff58707465d/internal/k8s/controller.go#L1521 This is a precondition for the `generateUpstream` function and if this precondition is not true -- it is a bug in the code, not invalid input. That's why I don't think `isPlus: false` cases are necessary.  To avoid confusion, perhaps it makes sense to add a comment in the `generateUpstream` function that `isExternalNameSvc` is never true for OSS?

internal/configs/virtualserver_test.go
```diff
@@ -894,12 +929,109 @@ func TestGenerateUpstreamForZeroEndpoints(t *testing.T) {
...
+			},
+		},
+		Endpoints: map[string][]string{
+			"test-namespace/test-upstream:0": []string{"example.com"},
+		},
+		ExternalNameSvcs: map[string]bool{
+			name: true,
+		},
+	}
+	cfgParams := ConfigParams{
```
pleshakov: these cfgParams values doesn't seem to be related to the external name svc at all

internal/configs/virtualserver.go
Rulox: Too many arguments, I don't see any problem with using the `virtualServerEx` Actually I jsut realized I need to pass the namespace too, cos is different for VS and VSR, so, either we have different functions for both (this could be a good use of Generics when they come to go), or I use `virtualServerEx`, otherwise, it will be too many arguments.

I understand tests are part of the code, but, if I have to choose between simplicity in code or tests I'll go with code all the time. You spend more time with actual code than tests. 

Unless you really thing this is super important, I'll go with `virtualServerEx`.

internal/configs/virtualserver_test.go
Rulox: The tests is important to check that we don't fill `ExternalNameSvcs` map in OSS. If we change the way we fill that by mistake, this test should fail and we will catch it (resolve=True for OSS fails) no?

internal/configs/virtualserver_test.go
Rulox: 👍 from copy/paste

internal/configs/virtualserver.go
Rulox: Alternatively, we can create a Struct like `UpstreamConfig` or something, and just use the important fields there.

pleshakov: Thanks for the explanation.

What if we use another approach and put the logic related to transforming the endpoints in a separate function like this (not tested, just a prototype):
```go
func prepareEndpoints(endpoints []string, isExternalNameSvc bool, isResolverConfigured bool, isPlus bool, service string) []string {
	if !isPlus && len(endpoints) == 0 {
		return []string{nginx502Server}
	}

	if isExternalNameSvc && !isResolverConfigured {
		glog.Warningf("A resolver must be configured for Type ExternalName service %s, no upstream servers will be created", service)
		return []string{}
	}

	return endpoints
}
```

In `generateVirtualServerConfig` will have:
```go
		upstreamName := virtualServerUpstreamNamer.GetNameForUpstream(u.Name)
		endpointsKey := GenerateEndpointsKey(virtualServerEx.VirtualServer.Namespace, u.Service, u.Port)
		endpoints := virtualServerEx.Endpoints[endpointsKey]
		endpoints = prepareEndpoints(endpoints, virtualServerEx.ExternalNameSvcs[endpointsKey], isResolverConfigured, isPlus, u.Service)
		upstreamNamespace := virtualServerEx.VirtualServer.Namespace
		ups := generateUpstream(upstreamName, u, endpoints, virtualServerEx.ExternalNameSvcs[endpointsKey], baseCfgParams)
```

This way we will simplify the `generateUpstream` function considerably, so its signature can look like this:
```go
func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, endpoints []string, isExternalNameSvc bool, cfgParams *ConfigParams) version2.Upstream {
```

Note, we only have 5 parameters, as before this PR.

I believe in this case both the code and the test will be simplified.

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -195,7 +195,7 @@ tls:
 | Field | Description | Type | Required |
 | ----- | ----------- | ---- | -------- |
 | `name` | The name of the upstream. Must be a valid DNS label as defined in RFC 1035. For example, `hello` and `upstream-123` are valid. The name must be unique among all upstreams of the resource. | `string` | Yes |
-| `service` | The name of a [service](https://kubernetes.io/docs/concepts/services-networking/service/). The service must belong to the same namespace as the resource. If the service doesn't exist, NGINX will assume the service has zero endpoints and return a `502` response for requests for this upstream. | `string` | Yes |
+| `service` | The name of a [service](https://kubernetes.io/docs/concepts/services-networking/service/). The service must belong to the same namespace as the resource. If the service doesn't exist, NGINX will assume the service has zero endpoints and return a `502` response for requests for this upstream. For NGINX Plus only, services of type [ExternalName](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) are also supported (check the [prerrequisites](../examples/externalname-services#prerequisites)). | `string` | Yes |
```
Dean-Coakley:
```suggestion
| `service` | The name of a [service](https://kubernetes.io/docs/concepts/services-networking/service/). The service must belong to the same namespace as the resource. If the service doesn't exist, NGINX will assume the service has zero endpoints and return a `502` response for requests for this upstream. For NGINX Plus only, services of type [ExternalName](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) are also supported (check the [prerequisites](../examples/externalname-services#prerequisites)). | `string` | Yes |
```

Dean-Coakley: `prerrequisites`->`prerequisites`

internal/k8s/controller.go
Rulox: That's what I thought at the beginning, but then I saw that `getServiceForIngressBackend` was called inside, so it's not really separated, is just a useless abstraction. Better to split in a more clever way I would say.

internal/configs/virtualserver.go
```diff
@@ -167,7 +169,23 @@ func generateUpstreamStatusMatch(upstreamName string, status string) version2.St
 	}
 }
 
-func generateVirtualServerConfig(virtualServerEx *VirtualServerEx, tlsPemFileName string, baseCfgParams *ConfigParams, isPlus bool) version2.VirtualServerConfig {
+func getEndpointsForVirtualServerEx(namespace string, upstream conf_v1alpha1.Upstream, virtualServerEx *VirtualServerEx, isResolverConfigured bool, isPlus bool) []string {
```
pleshakov: name is slightly misleading. the function gets endpoints for an upstream, not a virtualserverEx. Perhaps `getEndpointsForUpstream` or `generateEndpointsForUpstream`?

internal/configs/virtualserver.go
```diff
@@ -302,25 +322,19 @@ func generateVirtualServerConfig(virtualServerEx *VirtualServerEx, tlsPemFileNam
 	}
 }
 
-func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, endpoints []string, isPlus bool, cfgParams *ConfigParams) version2.Upstream {
+func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, externalNameSvcs map[string]bool, endpoints []string, cfgParams *ConfigParams) version2.Upstream {
 	var upsServers []version2.UpstreamServer
 
+	// isExternalNameSvc is always false for OSS
+	_, isExternalNameSvc := externalNameSvcs[upstream.Service]
```
pleshakov: Looks like there is a bug in the `externalNameSvcs` map because the keys contain names without namespaces. 

Consider a (corner) case if we have 2 VirtualServerRoutes in different namespaces and they both refer to a service with the same name. If in one namespace the service is of the externalname type, the IC will think that the service in the second namespace is also of the externalname type.

internal/configs/virtualserver_test.go
```diff
@@ -866,13 +867,12 @@ func TestGenerateUpstreamWithKeepalive(t *testing.T) {
 func TestGenerateUpstreamForZeroEndpoints(t *testing.T) {
```
pleshakov: looks like we need this test func anymore, because the logic for handling 0 endpoints for Plus/OSS was moved to a different function

internal/configs/virtualserver_test.go
```diff
@@ -1829,3 +1854,130 @@ func TestGenerateHealthCheck(t *testing.T) {
...
+					ObjectMeta: meta_v1.ObjectMeta{
+						Name:      name,
+						Namespace: namespace,
+					},
+				},
+				Endpoints: map[string][]string{},
+			},
+			isPlus:               true,
+			isResolverConfigured: false,
+			expected:             nil,
```
pleshakov: msg is missing

internal/configs/virtualserver.go
```diff
@@ -630,18 +644,26 @@ func createUpstreamServersForPlus(virtualServerEx *VirtualServerEx) map[string][
 	virtualServerUpstreamNamer := newUpstreamNamerForVirtualServer(virtualServerEx.VirtualServer)
 
 	for _, u := range virtualServerEx.VirtualServer.Spec.Upstreams {
+		if _, isExternalNameSvc := virtualServerEx.ExternalNameSvcs[u.Service]; isExternalNameSvc {
```
pleshakov: i think it makes sense to add a case to `TestCreateUpstreamServersForPlus` or new test function for VS/VSR and externalname svc.

internal/configs/virtualserver_test.go
```diff
@@ -1829,3 +1854,130 @@ func TestGenerateHealthCheck(t *testing.T) {
 		}
 	}
 }
+
+func TestGetEndpointsForVirtualServerEx(t *testing.T) {
+	name := "test-upstream"
```
pleshakov: the name is used both for VirtualServer and the service, but not the upstream. That's why `test-upstream` might be confusing. Consider using a more general name like `test`.

internal/k8s/controller.go
```diff
@@ -1551,36 +1567,41 @@ func (lbc *LoadBalancerController) createVirtualServer(virtualServer *conf_v1alp
...
-	svc, err := lbc.getServiceForIngressBackend(backend, namespace)
-	if err != nil {
-		glog.Warningf("Error getting service %v: %v", name, err)
-		return nil
+		ServicePort: intstr.FromInt(int(port)),
 	}
 
-	endps, _, err := lbc.getEndpointsForIngressBackend(backend, namespace, svc)
+	endps, isExternal, err := lbc.getEndpointsForIngressBackend(backend, namespace, svc)
 	if err != nil {
```
pleshakov: this function always returns err as `nil`. consider removing that return argument or returning the err when it is not nil.

internal/k8s/controller.go
```diff
@@ -1504,10 +1504,26 @@ func (lbc *LoadBalancerController) createVirtualServer(virtualServer *conf_v1alp
 	}
 
 	endpoints := make(map[string][]string)
+	externalNameSvcs := make(map[string]bool)
 
 	for _, u := range virtualServer.Spec.Upstreams {
-		endpointsKey := configs.GenerateEndpointsKey(virtualServer.Namespace, u.Service, u.Port)
-		endpoints[endpointsKey] = lbc.getEndpointsForService(virtualServer.Namespace, u.Service, int(u.Port))
+		backend := &extensions.IngressBackend{
```
pleshakov: seems like you forgot to update this part or vice versa.  the logic in the  loop `for _, u := range vsr.Spec.Upstreams` in line 1569 looks different to this one.

internal/k8s/controller.go
```diff
@@ -1551,36 +1567,41 @@ func (lbc *LoadBalancerController) createVirtualServer(virtualServer *conf_v1alp
...
 
 	virtualServerEx.Endpoints = endpoints
 	virtualServerEx.VirtualServerRoutes = virtualServerRoutes
+	virtualServerEx.ExternalNameSvcs = externalNameSvcs
 
 	return &virtualServerEx, virtualServerRouteErrors
 }
 
-func (lbc *LoadBalancerController) getEndpointsForService(namespace string, name string, port int) []string {
+func (lbc *LoadBalancerController) getEndpointsForService(namespace string, name string, port uint16, svc *api_v1.Service) (result []string, isExternal bool, err error) {
```
pleshakov: I think if you call `lbc.getServiceForUpstream()` from this function, rather from than `createVirtualServer`, it will simplify the logic. For example, `getEndpointsForService` will only need `(namespace string, name string, port uint16)` as arguments. Let me know what you think.

internal/configs/virtualserver.go
Rulox: Yeah I had the namespaces in the first implementation, then I removed them cos I saw we didn't use them for externalname services for Ingress Resources. Adding them back.

internal/k8s/controller.go
Rulox: I'll do it, but then the meaning will change, cos we need the upstream to get the service.

PR #643, opened 2019-08-02
Add flag to control accumulating IC logs in stdout

tellet:
Add flag `--show-ic-logs`, default is 'no' (do Not show the logs)


tellet:
@Rulox I wasn't sure about that. I'll update README for consistency then

PR #650, opened 2019-08-07
Add checks for `health check` upstream options

tellet:
The cases are similar to the other upstream options but only run for Nginx Plus.

### DIFF
--- a/tests/data/virtual-server-route-upstream-options/plus-route-m-invalid-keys.yaml
+++ b/tests/data/virtual-server-route-upstream-options/plus-route-m-invalid-keys.yaml
@@ -1,0 +1,48 @@
+apiVersion: k8s.nginx.org/v1alpha1
+kind: VirtualServerRoute
+metadata:
+  name: backends
+spec:
+  host: virtual-server-route.example.com
+  upstreams:
+  - name: backend1
+    service: backend1-svc
+    port: 80
+    healthCheck:
+      enable: True
+      path: "invalid"
+      interval: 1.5d
+      jitter: invalid
+      fails: -5
+      passes: -1
+      port: -1
+      connect-timeout: invalid
+      read-timeout: m
+      send-timeout: d
+      headers:
+        - name: ""
+          value: "$one"
+      statusMatch: "invalid"
+  - name: backend3
+    service: backend3-svc
+    port: 80
+    healthCheck:
+      enable: True
+      path: "invalid"
+      interval: 1.5d
+      jitter: invalid
+      fails: -5
+      passes: -1
+      port: -1
+      connect-timeout: invalid
+      read-timeout: m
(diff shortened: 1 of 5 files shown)


tests/data/virtual-server-upstream-options/plus-virtual-server-with-invalid-keys.yaml
```diff
@@ -0,0 +1,47 @@
...
+      jitter: invalid
+      fails: -5
+      passes: -1
+      port: -1
+      tls:
+        enable: true
+      connect-timeout: invalid
+      read-timeout: m
+      send-timeout: d
+      headers:
```
Rulox: I think this is a valid header no?

pleshakov: 👍

tellet: Thanks for noticing!

tellet:
Updated the manifests as requested and squashed the commits.

tests/data/virtual-server-route-upstream-options/plus-route-m-invalid-keys.yaml
```diff
@@ -0,0 +1,46 @@
...
+      interval: 1.5d
+      jitter: invalid
+      fails: -5
+      passes: -1
+      port: -1
+      connect-timeout: invalid
+      read-timeout: m
+      send-timeout: d
+      headers:
+        - name: "`=1^&*"
```
pleshakov: this is a valid value :) 
please see https://github.com/nginxinc/kubernetes-ingress/blob/master/pkg/apis/configuration/validation/validation.go#L255 additionally, `$` are disallowed (we didn't mention that in the description, we'll fix soon)

pleshakov: my mistake - it is an invalid value. 

it is an invalid value for the name of the header. I got confused and thought it was for the value of the header. Btw, that check is missing. For the value of the header, that value is valid.

tellet: Well, there is no check for header.value, I'll add it here too, but I really think that integration test is more about common flow (what system does in case if a value is invalid). All the checks for validation rules should be a part of unit tests, what do you think? @pleshakov @Rulox

Rulox: There are unit tests for those validations for sure. About integration tests for this project specifically, my personal opinion is that 1 full yaml with **all** valid values and another one with **all** invalid values is enough.

PR #653, opened 2019-08-09
Add websockets support for vs/vsr upstreams

Dean-Coakley:
### Proposed changes
This PR adds support for Websocket connections for upstreams without any additional configuration when configuring via VirtualServer or VirtualServerRoute.

### Checklist

### DIFF
--- a/internal/configs/version1/nginx-plus.tmpl
+++ b/internal/configs/version1/nginx-plus.tmpl
@@ -67,7 +67,10 @@ http {
         default upgrade;
         ''      close;
     }
-
+    map $http_upgrade $vs_connection_header {
+        default upgrade;
+        ''      $default_connection_header;
+    }
     {{if .SSLProtocols}}ssl_protocols {{.SSLProtocols}};{{end}}
     {{if .SSLCiphers}}ssl_ciphers "{{.SSLCiphers}}";{{end}}
     {{if .SSLPreferServerCiphers}}ssl_prefer_server_ciphers on;{{end}}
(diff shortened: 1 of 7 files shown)


internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -1,3 +1,8 @@
+map $http_upgrade $connection_header {
```
pleshakov: this map is defined in the http context, which is shared among the generated config for all Ingress and VS resources. For example, if you have multiple VS resources, the generated config for each resource will include this map. As a result, you will get a conflict and NGINX will fail to reload.

You need to move this map into the main NGINX configuration file.

internal/configs/version1/nginx.ingress.tmpl
```diff
@@ -123,12 +123,10 @@ server {
 		{{end}}
 		{{else}}
 		proxy_http_version 1.1;
-		{{if $location.Websocket}}
+
+		set $default_connection_header {{if $.Keepalive}}''{{else}}close{{end}};
```
pleshakov: there is no need to change the implementation for the Ingress resource.

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -65,7 +65,7 @@ http {
 
     map $http_upgrade $connection_upgrade {
```
pleshakov: Please configure a different map so that your change doesn't interfere with the implementation for the Ingress resource. To configure a different map, choose a different variable. For example:
```yaml
map $http_upgrade $vs_connection_upgrade {
```

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -67,7 +67,10 @@ http {
         default upgrade;
         ''      close;
     }
-
+    map $http_upgrade $vs_connection_upgrade {
```
pleshakov: Please rename before merging `$vs_connection_upgrade` to `$vs_connection_header` here and in the other places. `$vs_connection_header` better reflects the meaning of that variable. Note that `$connection_upgrade ` from `map $http_upgrade $connection_upgrade` is fine.

PR #654, opened 2019-08-09
Use vendored deps in code-generation script

pleshakov:
### Proposed changes
Use vendored deps in code-generation script.


pleshakov:
To use the code generation script, the repo must be inside the `GOPATH`. We need to be aware of that. Couldn't figure out how to make it work without GOPATH.

PR #655, opened 2019-08-13
Add slow-start support in vs/vsr

B-V-R:
### Proposed changes

### Checklist

### DIFF
--- a/docs/virtualserver-and-virtualserverroute.md
+++ b/docs/virtualserver-and-virtualserverroute.md
@@ -217,6 +217,7 @@ tls:
 | `client-max-body-size` | Sets the maximum allowed size of the client request body. See the [client_max_body_size](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) directive. The default is set in the `client-max-body-size` ConfigMap key. | `string` | No |
 | `tls` | The TLS configuration for the Upstream. | [`tls`](#UpstreamTLS) | No |
 | `healthCheck` | The health check configuration for the Upstream. See the [health_check](http://nginx.org/en/docs/http/ngx_http_upstream_hc_module.html#health_check) directive. Note: this feature is supported only in NGINX Plus. | [`healthcheck`](#UpstreamHealthcheck) | No |
+| `slow-start` | The slow start allows an upstream server to gradually recover its weight from 0 to its nominal value after it has been recovered or became available or when the server becomes available after a period of time it was considered unavailable. By default, the slow start is disabled. See the [slow_start](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#slow_start) parameter of the server directive. Note: The parameter cannot be used along with the `random`, `hash` or `ip_hash` load balancing methods and will be ignored. | `string` | No |
 
 ### Upstream.TLS
 
--- a/internal/configs/virtualserver.go
+++ b/internal/configs/virtualserver.go
@@ -14,6 +14,15 @@ import (
 
 const nginx502Server = "unix:/var/run/nginx-502-server.sock"
 
+var incompatibleLBMethodsForSlowStart = map[string]bool{
+	"random":                          true,
+	"ip_hash":                         true,
+	"random two":                      true,
+	"random two least_conn":           true,
+	"random two least_time=header":    true,
+	"random two least_time=last_byte": true,
+}
+
 // VirtualServerEx holds a VirtualServer along with the resources that are referenced in this VirtualServer.
 type VirtualServerEx struct {
 	VirtualServer       *conf_v1alpha1.VirtualServer
@@ -211,7 +220,7 @@ func generateVirtualServerConfig(virtualServerEx *VirtualServerEx, tlsPemFileNam
 		endpoints := generateEndpointsForUpstream(upstreamNamespace, u, virtualServerEx, isResolverConfigured, isPlus)
 		// isExternalNameSvc is always false for OSS
 		_, isExternalNameSvc := virtualServerEx.ExternalNameSvcs[GenerateExternalNameSvcKey(upstreamNamespace, u.Service)]
-		ups := generateUpstream(upstreamName, u, isExternalNameSvc, endpoints, baseCfgParams)
+		ups := generateUpstream(upstreamName, u, isExternalNameSvc, endpoints, baseCfgParams, isPlus)
 		upstreams = append(upstreams, ups)
 		crUpstreams[upstreamName] = u
 
@@ -231,7 +240,7 @@ func generateVirtualServerConfig(virtualServerEx *VirtualServerEx, tlsPemFileNam
 			endpoints := generateEndpointsForUpstream(upstreamNamespace, u, virtualServerEx, isResolverConfigured, isPlus)
 			// isExternalNameSvc is always false for OSS
 			_, isExternalNameSvc := virtualServerEx.ExternalNameSvcs[GenerateExternalNameSvcKey(upstreamNamespace, u.Service)]
-			ups := generateUpstream(upstreamName, u, isExternalNameSvc, endpoints, baseCfgParams)
+			ups := generateUpstream(upstreamName, u, isExternalNameSvc, endpoints, baseCfgParams, isPlus)
 			upstreams = append(upstreams, ups)
 			crUpstreams[upstreamName] = u
 
@@ -332,8 +341,9 @@ func generateVirtualServerConfig(virtualServerEx *VirtualServerEx, tlsPemFileNam
 	}
 }
 
-func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isExternalNameSvc bool, endpoints []string, cfgParams *ConfigParams) version2.Upstream {
+func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isExternalNameSvc bool, endpoints []string, cfgParams *ConfigParams, isPlus bool) version2.Upstream {
 	var upsServers []version2.UpstreamServer
+	lbMethod := generateLBMethod(upstream.LBMethod, cfgParams.LBMethod)
 
 	for _, e := range endpoints {
 		s := version2.UpstreamServer{
@@ -343,16 +353,39 @@ func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isEx
 			MaxConns:    generateIntFromPointer(upstream.MaxConns, cfgParams.MaxConns),
 			Resolve:     isExternalNameSvc,
 		}
+
+		if isPlus {
+			s.SlowStart = generateSlowStartForPlus(upstream, lbMethod)
+		}
+
 		upsServers = append(upsServers, s)
 	}
 
 	return version2.Upstream{
 		Name:             upstreamName,
 		Servers:          upsServers,
-		LBMethod:         generateLBMethod(upstream.LBMethod, cfgParams.LBMethod),
+		LBMethod:         lbMethod,
 		Keepalive:        generateIntFromPointer(upstream.Keepalive, cfgParams.Keepalive),
 		UpstreamZoneSize: cfgParams.UpstreamZoneSize,
 	}
+}
+
+func generateSlowStartForPlus(upstream conf_v1alpha1.Upstream, lbMethod string) string {
+	if upstream.SlowStart == "" {
+		return ""
+	}
+
+	if strings.HasPrefix(lbMethod, "hash") {
+		glog.Warningf("slow-start will be disabled for the Upstream %v", upstream.Name)
+		return ""
+	}
+
+	if _, exists := incompatibleLBMethodsForSlowStart[lbMethod]; exists {
+		glog.Warningf("slow-start will be disabled for the Upstream %v", upstream.Name)
+		return ""
+	}
+
+	return upstream.SlowStart
 }
 
 func generateLBMethod(method string, defaultMethod string) string {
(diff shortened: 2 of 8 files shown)


internal/configs/virtualserver_test.go
```diff
@@ -238,7 +238,7 @@ func TestGenerateVirtualServerConfig(t *testing.T) {
 				Name: "vs_default_cafe_tea",
```
pleshakov: please fix the formatting in this file using `gofmt`.

internal/configs/virtualserver.go
```diff
@@ -342,6 +347,12 @@ func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isEx
 			FailTimeout: generateString(upstream.FailTimeout, cfgParams.FailTimeout),
 			MaxConns:    generateIntFromPointer(upstream.MaxConns, cfgParams.MaxConns),
 			Resolve:     isExternalNameSvc,
+			SlowStart:   upstream.SlowStart,
+		}
+		if incompatibleLBMethods[upstream.LBMethod] {
```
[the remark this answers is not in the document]
pleshakov: I would further suggest putting the logic for checking for incompatible method in a separate function and writing a unit test for that function.

```go
	lbMethod := generateLBMethod(upstream.LBMethod, cfgParams.LBMethod)
	slowStart := 	for _, e := range endpoints {
		s := version2.UpstreamServer{
			Address:     e,
			MaxFails:    generateIntFromPointer(upstream.MaxFails, cfgParams.MaxFails),
			FailTimeout: generateString(upstream.FailTimeout, cfgParams.FailTimeout),
			MaxConns:    generateIntFromPointer(upstream.MaxConns, cfgParams.MaxConns),
			Resolve:     isExternalNameSvc,
                         SlowStart:  generateSlowStart(lbMethod, upstream.SlowStart, isPlus)

,
		}
		upsServers = append(upsServers, s)
	}


	return version2.Upstream{
		Name:      upstreamName,
		Servers:   upsServers,
		LBMethod: lbMethod,
		Keepalive: generateIntFromPointer(upstream.Keepalive, cfgParams.Keepalive),
	}
```

Note that `isPlus` will have to be added to the list of `generateUpstream` arguments. We need it because slow start only works for NGINX Plus.

Additionally, also note that `hash` method is defined by  `hash` + some varialble(s), for example `$hash $request_id`. As a result, the check `incompatibleLBMethods[lbMethod]` for hash method won't work. We need to check for the prefix, not the exact match here. Please also see the implementation of ParseLBMethod function https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/parsing_helpers.go#L92

Also, I suggest renaming `incompatibleLBMethods ` to `incompatibleLBMethodsForSlowStart `and making it a global variable similar to https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/parsing_helpers.go#L119 so that the map is only created once.

internal/configs/virtualserver.go
```diff
@@ -342,6 +347,12 @@ func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isEx
...
+			//TODO trigger warning
+			glog.Warningf("The parameter slow-start cannot be used along with %s, load balancing method", upstream.LBMethod)
```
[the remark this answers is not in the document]
pleshakov: @Rulox warningf for now works OK. we can add a PR later to support emitting events for warnings.

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -41,6 +41,7 @@ type Upstream struct {
 	ProxyNextUpstreamTries   int          `json:"next-upstream-tries"`
 	TLS                      UpstreamTLS  `json:"tls"`
 	HealthCheck              *HealthCheck `json:"healthCheck"`
+	SlowStart                string       `json:"slow-start,omitempty"`
```
pleshakov: `omitempty` is not necessary here

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -215,6 +215,7 @@ tls:
...
 | `tls` | The TLS configuration for the Upstream. | [`tls`](#UpstreamTLS) | No |
 | `healthCheck` | The health check configuration for the Upstream. See the [health_check](http://nginx.org/en/docs/http/ngx_http_upstream_hc_module.html#health_check) directive. Note: this feature is supported only in NGINX Plus. | [`healthcheck`](#UpstreamHealthcheck) | No |
+| `slow-start` | The slow start allows an upstream server to gradually recover its weight from 0 to its nominal value after it has been recovered or became available or when the server becomes available after a period of time it was considered unavailable. By default, the slow start is disabled. See the [slow-start](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#slow_start) parameter of the server directive. Note: The parameter cannot be used along with the `random`, `hash` or `ip_hash` load balancing methods and will be ignored. | `string` | No |
```
pleshakov: the previous suggestion wasn't fully implemented -- https://github.com/nginxinc/kubernetes-ingress/pull/655#discussion_r314030550
```suggestion
| `slow-start` | The slow start allows an upstream server to gradually recover its weight from 0 to its nominal value after it has been recovered or became available or when the server becomes available after a period of time it was considered unavailable. By default, the slow start is disabled. See the [slow_start](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#slow_start) parameter of the server directive. Note: The parameter cannot be used along with the `random`, `hash` or `ip_hash` load balancing methods and will be ignored. | `string` | No |
```

internal/configs/virtualserver.go
```diff
@@ -332,8 +338,10 @@ func generateVirtualServerConfig(virtualServerEx *VirtualServerEx, tlsPemFileNam
 	}
 }
 
-func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isExternalNameSvc bool, endpoints []string, cfgParams *ConfigParams) version2.Upstream {
+func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isExternalNameSvc bool, endpoints []string, cfgParams *ConfigParams, isPlus bool) version2.Upstream {
 	var upsServers []version2.UpstreamServer
+	lbMethod := generateLBMethod(upstream.LBMethod, cfgParams.LBMethod)
+	slowStart := generateSlowStartForPlus(upstream, lbMethod)
```
pleshakov: instead of 
```
		if isPlus {
			s.SlowStart = slowStart
		}
```
I suggest to do the following:
```
slowStart := ""
if isPlus {
  slowStart := generateSlowStartForPlus(upstream, lbMethod)
}
```
this way `generateSlowStartForPlus` will only be invoked for Plus.

internal/configs/virtualserver.go
```diff
@@ -354,6 +366,23 @@ func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isEx
 	}
 }
 
+func generateSlowStartForPlus(upstream conf_v1alpha1.Upstream, lbMethod string) string {
+
```
pleshakov: the empty line is not necessary

internal/configs/virtualserver.go
```diff
@@ -354,6 +366,23 @@ func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isEx
 	}
 }
 
+func generateSlowStartForPlus(upstream conf_v1alpha1.Upstream, lbMethod string) string {
```
pleshakov: this function needs to be rewritten as it doesn't implement the slow start logic completely. The logic is below:
- if slowStart is "" , return ""
- if the lbMethod is either `random`, `ip_hash` or `hash`, print a warning and return "" . Note that the hash method is defined as hash + some varialble(s), for example `$hash $request_id`. The current implementation incorrectly allows methods like  `$hash $request_id`. Also, the random method is similar to the hash, but has a finite number of possible values. Please see them here -- https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/parsing_helpers.go#L122-L126 Those should be excluded as well.
- return slowStart

Also, the corresponding unit test must be rewritten as well. For the test, consider the following test cases:

Positive cases:
- slow start is "" and the method is "least_conn",  expects  ""
- slow_start is "10s" and the method is "least_conn",  expects "10s"

Negative cases:
- slow_start is "10s" and the method is "hash $something", expects ""
- slow_start is "10s" and the method is "ip_hash", expects ""
- slow_start is "10s" and the method is "random", expects "". Note, this can be repeated for all allowed variations of the random method -- https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/parsing_helpers.go#L122-L126

Please add any additional cases if necessary.

Dean-Coakley: Normally we have a preference for using a `map` like: https://github.com/nginxinc/kubernetes-ingress/blob/master/pkg/apis/configuration/validation/validation.go#L343-L357 which I think should be possible.

If sticking with a slice is necessary, you should be able to make it `const` , right?

internal/configs/virtualserver.go
```diff
@@ -343,17 +354,41 @@ func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isEx
...
+
+	slowStart := upstream.SlowStart
+
+	if slowStart == "" {
+		return ""
+	}
+
+	for _, nLBMethod := range incompatibleLBMethodsForSlowStart {
```
pleshakov: to check for the incompatible method, I suggest using approach similar to https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/parsing_helpers.go#L141-L148 :
1) first check if the method starts with hash
2) check if the method is in the map. This will require changing `incompatibleLBMethodsForSlowStart` to a map.

internal/configs/virtualserver.go
pleshakov: this wasn't addressed. we don't put empty lines at the beginning of the function

internal/configs/virtualserver_test.go
```diff
@@ -1960,3 +1961,44 @@ func TestGenerateEndpointsForUpstream(t *testing.T) {
...
+			expected: "",
+		},
+		{
+			upstream: conf_v1alpha1.Upstream{Service: name, Port: 80, SlowStart: "10s", LBMethod: "least_conn"},
+			expected: "10s",
+		},
+	}
+
+	for _, test := range tests {
+		lbMethod := generateLBMethod(test.upstream.LBMethod, "least_conn")
```
pleshakov: there is no need to call `generateLBMethod`. I suggest making the method the third field of the struct you use in tests array.

internal/configs/virtualserver_test.go
```diff
@@ -1960,3 +1961,44 @@ func TestGenerateEndpointsForUpstream(t *testing.T) {
 		}
 	}
 }
+
+func TestGenerateSlowStartForPlusWithInCompatibleLBMethods(t *testing.T) {
+	name := "test-slowstart-with-incompatible-LBMethods"
+	upstream := conf_v1alpha1.Upstream{Service: name, Port: 80, SlowStart: "10s"}
+	expected := ""
+
+	for lbMethod, _ := range incompatibleLBMethodsForSlowStart {
```
pleshakov: it is better to have an array of the input parameters like you have in the `TestGenerateSlowStartForPlus`. this will make the tests much more readable.

internal/configs/virtualserver_test.go
```diff
@@ -1960,3 +1961,44 @@ func TestGenerateEndpointsForUpstream(t *testing.T) {
...
+		nlbMethod := generateLBMethod(upstream.LBMethod, lbMethod)
```
pleshakov: there is no need to call `generateLBMethod`. this is the same comment as for line 1998.

internal/configs/virtualserver.go
```diff
@@ -14,6 +14,16 @@ import (
 
 const nginx502Server = "unix:/var/run/nginx-502-server.sock"
 
+var incompatibleLBMethodsForSlowStart = map[string]bool{
+	"random":                          true,
+	"ip_hash":                         true,
+	"hash":                            true,
```
pleshakov: hash is not required here, considering this change https://github.com/nginxinc/kubernetes-ingress/pull/655/files#diff-d8f2204ffe0ef11d167489c09e9f4170R380

internal/configs/virtualserver_test.go
```diff
@@ -1960,3 +1961,56 @@ func TestGenerateEndpointsForUpstream(t *testing.T) {
...
+		expected string
+	}{
+		{
+			upstream: conf_v1alpha1.Upstream{Service: serviceName, Port: 80, SlowStart: "10s"},
+			expected: "",
+		},
+	}
+
+	for _, test := range tests {
+		for lbMethod, _ := range incompatibleLBMethodsForSlowStart {
```
pleshakov: this suggestion wasn't implemented https://github.com/nginxinc/kubernetes-ingress/pull/655#discussion_r316461167 please have a list of input parameters rather than taking lb methods from `incompatibleLBMethodsForSlowStart`

internal/configs/virtualserver.go
```diff
@@ -343,17 +354,42 @@ func generateUpstream(upstreamName string, upstream conf_v1alpha1.Upstream, isEx
...
 		Name:      upstreamName,
 		Servers:   upsServers,
-		LBMethod:  generateLBMethod(upstream.LBMethod, cfgParams.LBMethod),
+		LBMethod:  lbMethod,
 		Keepalive: generateIntFromPointer(upstream.Keepalive, cfgParams.Keepalive),
 	}
 }
 
+func generateSlowStartForPlus(upstream conf_v1alpha1.Upstream, lbMethod string) string {
+	slowStart := upstream.SlowStart
```
pleshakov: we don't usually create auxiliary variables like this one. Note that is only used twice. Can we just use `upstream.SlowStart` instead?

B-V-R:
@pleshakov can you please review changes

PR #657, opened 2019-08-14
Tests upstream tls support

tellet:
Refactor code a bit and add tests for tls option in upstreams for VS/VSRs.


tests/data/common/app/vsr/secure/multiple.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: apps/v1
```
pleshakov: for these deployments we don't need TLS, right?

tests/suite/test_v_s_route_upstream_tls.py
```diff
@@ -0,0 +1,116 @@
...
+                                            v_s_route_setup.vs_name,
+                                            ic_pod_name,
+                                            ingress_controller_prerequisites.namespace)
+        req_url = f"http://{v_s_route_setup.public_endpoint.public_ip}:{v_s_route_setup.public_endpoint.port}"
+        resp_1 = requests.get(f"{req_url}{v_s_route_setup.route_m.paths[0]}",
+                              headers={"host": v_s_route_setup.vs_host})
+        resp_2 = requests.get(f"{req_url}{v_s_route_setup.route_s.paths[0]}",
+                              headers={"host": v_s_route_setup.vs_host})
+
+        assert 'proxy_pass https://' in config
```
pleshakov: does it matter how many `proxy_pass https://` are in the config?

tests/suite/test_v_s_route_upstream_tls.py
```diff
@@ -0,0 +1,116 @@
...
+        assert_response_codes(resp_1, resp_2)
+
+    def test_invalid_value_rejection(self, kube_apis, ingress_controller_prerequisites,
```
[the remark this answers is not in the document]
tellet: Yes, the test implements the scenario given in the automated-examples for TLS termination via Ingresses. So, it tests a different thing, that's why it is a separate test. I can add checks to existing tests for invalid fields but in a different PR, what do you think?

tests/data/common/app/vsr/secure/multiple.yaml
tellet: Yes, we don't, we need them here because of the way VSR setup is organised (1 VS + VSR with multiple BEs in one ns and another VSR with single BE). So there are 2 paths here: leave multiple.yaml without TLS in a 'secure' directory (which is a bit misleading) or create a separate fixture for VSR test with simpler setup (1 VS, 1 VSR and 2 BEs)

tests/suite/test_v_s_route_upstream_tls.py
tellet: I'll improve assertions to `assert 'proxy_pass https://{exact_host_here}' in config`

tests/data/common/app/vsr/secure/multiple.yaml
tellet: I'll leave it as it is for now and will switch to 2nd path later.

tests/suite/test_v_s_route_upstream_tls.py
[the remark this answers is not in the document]
tellet: I'll add events checks.

PR #663, opened 2019-08-16
Add tests for VS/VSR externalName service support

tellet:
Cover test-cases for ExternalName service support for VS and VSRs

### DIFF
--- a/tests/data/virtual-server-route-externalname/externalname-svc.yaml
+++ b/tests/data/virtual-server-route-externalname/externalname-svc.yaml
@@ -1,0 +1,7 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: externalname-service
+spec:
+  type: ExternalName
+  externalName: nginx.com
--- a/tests/suite/test_virtual_server_external_name.py
+++ b/tests/suite/test_virtual_server_external_name.py
@@ -1,0 +1,108 @@
+import pytest
+import requests
+
+from settings import TEST_DATA
+from suite.custom_resources_utils import get_vs_nginx_template_conf
+from suite.resources_utils import replace_configmap_from_yaml, \
+    ensure_connection_to_public_endpoint, replace_configmap, create_service_from_yaml, get_first_pod_name, get_events, \
+    read_service, replace_service, wait_before_test
+from suite.yaml_utils import get_external_host_from_service_yaml
+
+
+class ExternalNameSetup:
+    """Encapsulate ExternalName example details.
+
+    Attributes:
+        ic_pod_name:
+        external_host: external service host
+    """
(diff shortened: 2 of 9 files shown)


tests/data/virtual-server-route-externalname/externalname-svc.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: externalname-service
+spec:
+  type: ExternalName
+  externalName: nginx.com
```
pleshakov: please note that if `nginx.com` is either (1) not available or (2) not resolved properly, the tests will see those cases as successful. Because in both cases, a request to `externalnginx.com` will return 502, which the tests expect.  Note that case (2) is hard to distinguish from the case of misconfigured resolver, which we configure using the  `resolver-addresses` configmap key.

It is possible to have a backend running inside the cluster and expose it as external service (assumung we can have coffee service in the default namespace):
```
kind: Service
apiVersion: v1
metadata:
  name: externalname-service
spec:
  type: ExternalName
  externalName: coffee.default.svc.cluster.local
```
```
kind: ConfigMap
apiVersion: v1
metadata:
  name: nginx-config
data:
  resolver-addresses: "kube-dns.kube-system.svc.cluster.local" 
```
This way we can always be sure that the backend is available and returns 200 repossess. As a result, we won't have (1) and (2).

I don't think we need to address (1) and (2) right now, but this is something we should address later.

tests/suite/test_virtual_server_external_name.py
```diff
@@ -0,0 +1,108 @@
...
+                                                 virtual_server_setup.vs_name,
+                                                 vs_externalname_setup.ic_pod_name,
+                                                 ingress_controller_prerequisites.namespace)
+        line = f"zone vs_{virtual_server_setup.namespace}_{virtual_server_setup.vs_name}_backend1 256k;"
+        assert line in result_conf
+        assert "random two least_conn;" in result_conf
+        assert f"server {vs_externalname_setup.external_host}:80 max_fails=1 fail_timeout=10s max_conns=0 resolve;"\
+               in result_conf
+
+    def test_events(self, kube_apis, crd_ingress_controller, virtual_server_setup, vs_externalname_setup):
```
pleshakov: note that in tests for VSR, `test_events_flows` is used.

tests/data/virtual-server-route-externalname/externalname-svc.yaml
tellet: I'll address this in a separate PR, cause this requires a significant changes in pre-conditions fixtures.

PR #668, opened 2019-08-21
Run golang-ci linter in travisci. Add golang-ci configuration

Dean-Coakley:
### Proposed changes
* Add `golangci-lint` config to set what linters to run
* Add linting step to travis.yml so build will fail if format is bad.

### Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,15 +4,15 @@ linters-settings:
 
 linters:
     enable:
-    - golint
     - goimports
+    - gosimple
+    - govet
     - misspell
     - gofmt
     - unparam
     - unconvert
     - interfacer
-    disable:
-    - maligned
-    - prealloc
-    - gochecknoglobals
+    - structcheck
+    - errcheck
+    disable-all: true
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,18 @@
+linters-settings:
+  misspell:
+    locale: US
+
+linters:
+    enable:
+    - golint
+    - goimports
+    - misspell
+    - gofmt
+    - unparam
+    - unconvert
+    - interfacer
+    disable:
+    - maligned
+    - prealloc
+    - gochecknoglobals
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,5 +19,5 @@ before_install:
   wget https://${fossalocation}/${fossafile} && tar xzf ${fossafile}
   && ./fossa init
   && GO111MODULE=on FOSSA_API_KEY=${fossapush} ./fossa analyze -t kubernetes-ingress -b ${TRAVIS_BRANCH}; fi
-- echo "Checking ingress controller for linting errors" && wget -O - -q ${golangci_lint} | sh -s v1.17.1 -b $(pwd)
-- make lint
+- echo "Checking ingress controller for linting errors" && wget -O - -q ${golangci_lint} | sh -s v1.17.1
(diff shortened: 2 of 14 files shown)


.travis.yml
```diff
@@ -19,3 +19,5 @@ before_install:
   wget https://${fossalocation}/${fossafile} && tar xzf ${fossafile}
   && ./fossa init
   && GO111MODULE=on FOSSA_API_KEY=${fossapush} ./fossa analyze -t kubernetes-ingress -b ${TRAVIS_BRANCH}; fi
+- echo "Checking ingress controller for linting errors" && wget -O - -q ${golangci_lint} | sh -s v1.17.1
```
pleshakov: `sh -s v1.17.1` ? 
judging by the output from here https://travis-ci.org/nginxinc/kubernetes-ingress/builds/575403783?utm_source=github_status&utm_medium=notification, `$ export golangci_lint_version=v1.17.1`,  the env variable was supposed to be used here

.golangci.yml
```diff
@@ -0,0 +1,18 @@
...
+    enable:
+    - golint
+    - goimports
+    - misspell
+    - gofmt
+    - unparam
+    - unconvert
+    - interfacer
+    disable:
+    - maligned
```
pleshakov: aren't those disabled by default https://github.com/golangci/golangci-lint#disabled-by-default-linters--e--enable ?

.golangci.yml
```diff
@@ -0,0 +1,18 @@
+linters-settings:
+  misspell:
+    locale: US
+
+linters:
+    enable:
```
pleshakov: all those linters are enabled by default and there are more enabled linters!
if you want to control which are enabled, consider the following example from the docs https://github.com/golangci/golangci-lint#config-file:
```
linters:
  enable:
    - megacheck
    - govet
  enable-all: false
  disable:
    - maligned
    - prealloc
  disable-all: false
  presets:
    - bugs
    - unused
  fast: false
```
I think we should lock down the linters that way

.travis.yml
```diff
@@ -19,3 +19,5 @@ before_install:
...
+- ./bin/golangci-lint run --config .golangci.yml
```
pleshakov: note that .golangci.yml will be used even if you don't specify `--config`

pleshakov: also, we have `make lint` target, can we use that?

.travis.yml
Dean-Coakley: I decided against using a travisci variable as you saw in that log there. 
Version of the linter should be tracked in version control.

.golangci.yml
Dean-Coakley: I found the following to not be enabled by default.
```
gofmt: Gofmt checks whether code was gofmt-ed. By default this tool runs with -s option to check for code 
goimports: Goimports does everything that gofmt does. Additionally it checks unused imports [fast: true, auto-fix: true]
interfacer: Linter that suggests narrower interface types [fast: false, auto-fix: false]
misspell: Finds commonly misspelled English words in comments [fast: true, auto-fix: true]
unconvert: Remove unnecessary type conversions [fast: true, auto-fix: false]
unparam: Reports unused function parameters [fast: false, auto-fix: false]
```

Tested with `$ golangci-lint linters` in dir with config and outside it.

.golangci.yml
Dean-Coakley: Locking down the linters like that looks nice though, I'll investigate that.

PR #670, opened 2019-08-22
Add buffering support for vs/vsr

ampant:
### Proposed changes
Added support for configuring buffering in upstream of vs/vsr.

### Checklist


docs/virtualserver-and-virtualserverroute.md
```diff
@@ -217,6 +217,9 @@ tls:
...
 | `tls` | The TLS configuration for the Upstream. | [`tls`](#UpstreamTLS) | No |
 | `healthCheck` | The health check configuration for the Upstream. See the [health_check](http://nginx.org/en/docs/http/ngx_http_upstream_hc_module.html#health_check) directive. Note: this feature is supported only in NGINX Plus. | [`healthcheck`](#UpstreamHealthcheck) | No |
+| `buffering` | Enables buffering of responses from the upstream server. See the [proxy_buffering](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffering) directive. The default is set in the proxy-buffering  ConfigMap key. | `boolean` | No |
+| `buffers` | Configures the buffers used for reading a response from the upstream server for a single connection. The buffers field configures the buffers used for reading a response from the upstream server for a single connection: number: 4 size: 8K .See the [proxy_buffers](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffers) directive for additional information.| `buffers` | No |
```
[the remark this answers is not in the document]
Dean-Coakley: I think Buffers warrents an entry into docs similar to: https://github.com/nginxinc/kubernetes-ingress/blob/master/docs/virtualserver-and-virtualserverroute.md#upstreamtls

pkg/apis/configuration/validation/validation.go
```diff
@@ -331,6 +361,8 @@ func validateUpstreams(upstreams []v1alpha1.Upstream, fieldPath *field.Path, isP
 		allErrs = append(allErrs, validatePositiveIntOrZeroFromPointer(u.MaxConns, idxPath.Child("max-conns"))...)
 		allErrs = append(allErrs, validateSize(u.ClientMaxBodySize, idxPath.Child("client-max-body-size"))...)
 		allErrs = append(allErrs, validateUpstreamHealthCheck(u.HealthCheck, idxPath.Child("healthCheck"))...)
+		allErrs = append(allErrs, validateBuffer(u.ProxyBuffers, idxPath.Child("proxy-buffers"))...)
```
Dean-Coakley: The `idxPath` needs to match the entry in the crd, not the upstream struct.

```suggestion
		allErrs = append(allErrs, validateBuffer(u.ProxyBuffers, idxPath.Child("buffers"))...)
```

pkg/apis/configuration/validation/validation.go
```diff
@@ -331,6 +361,8 @@ func validateUpstreams(upstreams []v1alpha1.Upstream, fieldPath *field.Path, isP
...
+		allErrs = append(allErrs, validateBufSize(u.ProxyBufferSize, idxPath.Child("proxy-buffer-size"))...)
```
Dean-Coakley:
```suggestion
		allErrs = append(allErrs, validateBufSize(u.ProxyBufferSize, idxPath.Child("buffer-size"))...)
```

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -39,11 +39,20 @@ type Upstream struct {
 	ProxyNextUpstream        string       `json:"next-upstream"`
 	ProxyNextUpstreamTimeout string       `json:"next-upstream-timeout"`
 	ProxyNextUpstreamTries   int          `json:"next-upstream-tries"`
+	ProxyBuffering           *bool        `json:"buffering"`
```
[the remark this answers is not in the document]
ampant: We need to check if this option is set in upstream or not. In that case, I am comparing it with nil so its type should be *bool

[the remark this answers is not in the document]
Dean-Coakley: This seems like the most reasonable way to achieve that. Looks fine.

pkg/apis/configuration/validation/validation.go
```diff
@@ -116,6 +116,36 @@ func validateSize(size string, fieldPath *field.Path) field.ErrorList {
...
+
+	if buff.Number <= 0 {
+		return append(allErrs, field.Invalid(fieldPath, buff.Number, "must be positive"))
+	}
+
+	if buff.Size == "" {
+		return append(allErrs, field.Invalid(fieldPath, buff.Size, "cannot be empty"))
+	}
+
+	if buff.Size == "4k" || buff.Size == "8k" {
```
pleshakov: buff.Size is of type size - it can take numbers, like 123 with suffixes m or M and k or K.

I suggest adding a function similar to https://github.com/nginxinc/kubernetes-ingress/blob/master/pkg/apis/configuration/validation/validation.go#L104
See also http://nginx.org/en/docs/syntax.html

Note that it is better to rename the existing function `validateSize` to `validateOffset` and create  a new function `validateSize` that validates using the rules described above.

pkg/apis/configuration/validation/validation.go
```diff
@@ -116,6 +116,36 @@ func validateSize(size string, fieldPath *field.Path) field.ErrorList {
...
+		return allErrs
+	}
+	return append(allErrs, field.Invalid(fieldPath, buff.Size, "must be 4k or 8k"))
+}
+
+func validateBufSize(size string, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if size == "" || size == "4k" || size == "8k" {
```
pleshakov: similar to proxy buffers, this field is of the type size.

internal/configs/virtualserver.go
```diff
@@ -391,6 +391,20 @@ func generateString(s string, defaultS string) string {
 	return s
 }
 
+func generateBuffers(s conf_v1alpha1.UpstreamBuffers, defaultS string) string {
```
pleshakov: I suggest adding a unit test for this function

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -1658,18 +1689,51 @@ func TestValidateTime(t *testing.T) {
...
+
+func TestValidateBuffer(t *testing.T) {
+	validbuff := &v1alpha1.UpstreamBuffers{Number: 8, Size: "8k"}
+	allErrs := validateBuffer(validbuff, field.NewPath("proxy_buffers"))
+
+	if len(allErrs) != 0 {
+		t.Errorf("validateBuffer returned errors %v valid input %v", allErrs, validbuff)
+	}
+
+	invalidbuff := &v1alpha1.UpstreamBuffers{Number: -8, Size: "15k"}
```
pleshakov: I suggest adding a test case for Size: ""

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -1658,16 +1689,61 @@ func TestValidateTime(t *testing.T) {
...
+	for _, test := range invalidInput {
+		allErrs := validateOffset(test, field.NewPath("offset-field"))
+		if len(allErrs) == 0 {
+			t.Errorf("validateOffset(%q) didn't return error for invalid input.", test)
+		}
+	}
+}
+
+func TestValidateBuffer(t *testing.T) {
+	validbuff := &v1alpha1.UpstreamBuffers{Number: 8, Size: "8k"}
```
Rulox: Here and in `TestValidateOffset` I think creating a for loop with all the cases will reduce a lot of duplicated code. For example, have a look at `TestValidateSize` you could do something like:

```golang
	var validBuffers = []v1alpha1.UpstreamBuffers{
		v1alpha1.UpstreamBuffers{
			Number: 8,
			Size:   "8k",
		},
		// etc
	}
```

And then run over the cases in a for loop like the other test cases

Also, normally in the rest of this file we separated valid and invalid test cases. For instance: `TestValidateBuffer` will have all the valid cases (if is only 1 we don't need a for loop) and `TestValidateBufferInvalid` will have the invalid cases. Same applies for `TestValidateSize` I think for consistency we can do the same here. Let me know if that makes sense.

pleshakov:
@ampant thx for the updates. the changes look good! 

I agree with @Rulox  suggestions.

Please also rebase against the master so that the conflict is resolved.

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -1658,16 +1689,64 @@ func TestValidateTime(t *testing.T) {
...
+	allErrs := validateBuffer(validbuff, field.NewPath("buffers-field"))
+
+	if len(allErrs) != 0 {
+		t.Errorf("validateBuffer returned errors %v valid input %v", allErrs, validbuff)
+	}
+
+	invalidbuff := []*v1alpha1.UpstreamBuffers{
+		{
```
pleshakov: small comment about the input for test cases - please change the input parameters so that only 2 of the fields (size or number) is invalid. For example,

			Number: -8,
			Size:   "15m",

			Number: 8,
			Size:   "15G",

                         Size: 8,
			Size:  "",

Otherwise, the test won't  fail if the validation for one of the parameter is buggy.

PR #671, opened 2019-08-22
Add a few more checks for WebSockets upstream option

tellet:
Add checks on default values in config

### DIFF
--- a/tests/suite/test_v_s_route_upstream_options.py
+++ b/tests/suite/test_v_s_route_upstream_options.py
@@ -71,6 +71,10 @@ class TestVSRouteUpstreamOptions:
 
         assert "keepalive" not in config
         assert 'set $default_connection_header "";' not in config
+        assert 'set $default_connection_header close;' in config
+        assert "proxy_set_header Upgrade $http_upgrade;" in config
+        assert "proxy_set_header Connection $vs_connection_header;" in config
+        assert "proxy_http_version 1.1;" in config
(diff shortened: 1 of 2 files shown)


tests/suite/test_v_s_route_upstream_options.py
```diff
@@ -71,6 +71,10 @@ def test_nginx_config_upstreams_defaults(self, kube_apis, ingress_controller_pre
 
         assert "keepalive" not in config
         assert 'set $default_connection_header "";' not in config
+        assert 'set $default_connection_header close;' in config
+        assert "proxy_set_header Upgrade $http_upgrade;" in config
+        assert "proxy_set_header Connection $vs_connection_header;" in config
+        assert "proxy_http_version 1.1;" in config
```
Dean-Coakley: Should this be included? It is in the config unconditionally unless I'm mistaken?

pleshakov: it is not conditional in VS/VSR https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/version2/nginx.virtualserver.tmpl#L108

tellet: As well as "Upgrade" and "Connection" lines, the test verifies defaults and doesn't check any conditions. Also as far as understand we need to set the http version to 1.1 to support websockets.

PR #672, opened 2019-08-23
Add tests for upstream-zone-size support

tellet:
Cover cases for Ingresses (OSS&Plus), ConfigMap key for VS/VSR (OSS&Plus).

### DIFF
--- a/tests/data/virtual-server-configmap-keys/configmap-validation-keys.yaml
+++ b/tests/data/virtual-server-configmap-keys/configmap-validation-keys.yaml
@@ -12,3 +12,4 @@ data:
   keepalive: "32"
   proxy-protocol: "true"
   redirect-to-https: "true"
+  upstream-zone-size: "0" # special value
--- a/tests/suite/test_annotations.py
+++ b/tests/suite/test_annotations.py
@@ -45,6 +45,23 @@ def replace_ingresses_from_yaml(extensions_v1_beta1: ExtensionsV1beta1Api, names
                 replace_ingress(extensions_v1_beta1, doc['metadata']['name'], namespace, doc)
 
 
+def get_minions_info_from_yaml(file) -> []:
+    """
+    Parse yaml file and return minions details.
+
+    :param file: an absolute path to file
+    :return: [{name, svc_name}]
+    """
+    res = []
+    with open(file) as f:
+        docs = yaml.load_all(f)
+        for dep in docs:
+            if 'minion' in dep['metadata']['name']:
+                res.append({"name": dep['metadata']['name'],
+                            "svc_name": dep['spec']['rules'][0]['http']['paths'][0]['backend']['serviceName']})
+    return res
+
+
 class AnnotationsSetup:
     """Encapsulate Annotations example details.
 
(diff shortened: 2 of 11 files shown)


tests/suite/test_annotations.py
```diff
@@ -156,15 +156,19 @@ def test_nginx_config_defaults(self, kube_apis, annotations_setup, ingress_contr
 
         assert "Strict-Transport-Security" not in result_conf
 
+        assert " 256k;" in result_conf
```
pleshakov: this line looks unclear. I suggest either adding a comment that the line asserts the zone size or assert "zone name 256k;". the name of the zone is the same as the name of the upstream.

this comment applies to the other similar asserts.

tests/suite/test_v_s_route_upstream_options.py
```diff
@@ -144,11 +147,11 @@ def test_when_option_in_v_s_r_only(self, kube_apis,
          ["max_fails=3 ", "fail_timeout=33s ", "max_conns=0;",
           "proxy_connect_timeout 44s;", "proxy_read_timeout 22s;", "proxy_send_timeout 55s;",
           "keepalive 1024;", 'set $default_connection_header "";',
-          "client_max_body_size 3m;"],
+          "client_max_body_size 3m;", " 100k;"],
```
pleshakov: zone size are not configurable in the upstream field of VS/VSR. do we need to test for them here?

tests/data/virtual-server-configmap-keys/configmap-validation-keys.yaml
```diff
@@ -11,4 +11,5 @@ data:
   max-fails: "3"
   keepalive: "32"
   proxy-protocol: "true"
-  redirect-to-https: "true"
\ No newline at end of file
+  redirect-to-https: "true"
+  upstream-zone-size: "0" # special value
```
pleshakov: It doesn't look that in the tests we're checking that for Plus we generate zones with the default size for this case

tests/suite/test_virtual_server_upstream_options.py
```diff
@@ -82,6 +82,9 @@ def test_nginx_config_defaults(self, kube_apis, ingress_controller_prerequisites
 
         assert "client_max_body_size 1m;" in config
 
+        assert "zone " in config
```
pleshakov: since we don't have a field for configuring zones in VS/VSR, do we need to check for zones in this test?

tests/suite/test_v_s_route_upstream_options.py
tellet: I wasn't sure if to add this check or not, I'll remove it then.

tests/suite/test_virtual_server_upstream_options.py
tellet: same as above

tests/suite/test_v_s_route_upstream_options.py
tellet: I agree this may be misleading (the same check exists in `test_virtual_server_configmap_keys`) as we don't support this for VS/VSR. I'll remove them then.

tests/suite/test_annotations.py
tellet: I'll add an assert "zone name 256k;". As for the assertions in the test_methods with `@pytest.mark.parametrize` annotation there is no way improve those without significant changes.

PR #686, opened 2019-09-09
Check if config map created with leader-election

LorcanMcVeigh:
### Proposed changes
If the cli argument value is not a valid value for a resource name  and/or if the IC can't create a config map, the IC exits with an error 

### Checklist

### DIFF
--- a/cmd/nginx-ingress/main.go
+++ b/cmd/nginx-ingress/main.go
@@ -138,7 +138,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	statusLockNameValidationError := validateLockName(*leaderElectionLockName)
+	statusLockNameValidationError := validateResourceName(*leaderElectionLockName)
 	if statusLockNameValidationError != nil {
 		glog.Fatalf("Invalid value for leader-election-lock-name: %v", statusLockNameValidationError)
 	}
@@ -451,12 +451,12 @@ func getSocketClient(sockPath string) *http.Client {
 	}
 }
 
-// validateLockName validates the name of a resource
+// validateResourceName validates the name of a resource
 // It returns an error if the resource name is invalid
-func validateLockName(lock string) error {
-	lst := validation.IsDNS1123Subdomain(lock)
-	if len(lst) > 0 {
-		return fmt.Errorf("invalid resource name: %v", lock)
+func validateResourceName(lock string) error {
+	allErrs := validation.IsDNS1123Subdomain(lock)
+	if len(allErrs) > 0 {
+		return fmt.Errorf("invalid resource name %v: %v", lock, allErrs)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


cmd/nginx-ingress/main.go
```diff
@@ -445,6 +451,16 @@ func getSocketClient(sockPath string) *http.Client {
 	}
 }
 
+// validateLockName validates the name of a resource
+// It returns an error if the resource name is invalid
+func validateLockName(lock string) error {
+	lst := validation.IsDNS1123Subdomain(lock)
```
Dean-Coakley: Lets use a more appropriate variable name. 
```suggestion
	allErrs := validation.IsDNS1123Subdomain(lock)
```

cmd/nginx-ingress/main.go
```diff
@@ -445,6 +451,16 @@ func getSocketClient(sockPath string) *http.Client {
...
+	if len(lst) > 0 {
+		return fmt.Errorf("invalid resource name: %v", lock)
```
Dean-Coakley: Here we lose all context of the error returned from: `validation.IsDNS1123Subdomain(lock)`

Maybe something like:
```suggestion
		return fmt.Errorf("invalid resource name %v: %v", lock, allErrs)
```

Can work? As allErrs is a slice of strings.

pkg/apis/configuration/validation/validation.go
```diff
@@ -323,8 +323,6 @@ func isValidHeaderValue(s string) []string {
 	return nil
 }
 
-// validateSecretName checks if a secret name is valid.
```
Dean-Coakley: This doesn't relate to the PR. Lets revert these changes?

cmd/nginx-ingress/main.go
```diff
@@ -445,6 +451,16 @@ func getSocketClient(sockPath string) *http.Client {
 	}
 }
 
+// validateLockName validates the name of a resource
+// It returns an error if the resource name is invalid
+func validateLockName(lock string) error {
```
Rulox: Maybe `validateResourceName` is a better function name ? That way we can re-use the validation function for other resources.

Dean-Coakley: @Rulox Do you suggest leaving `validateResourceName` in main.go and move to validate.go if/when we want to reuse it?

Rulox: Do you mean `validation.go` @Dean-Coakley ? I think main.go is better for this method.
 `validation.go` belongs to `pkg/apis/configuration/validation` This is, validation for CRDs (VS/VSR). 

Maybe we can also create another `.go` file for the validation of things that are not related with VS/VSR (or use `utils.go`). 

PS: there are more validation functions here in main.go so I don't think is a terrible idea to keep it here.

PR #691, opened 2019-09-11
Helper Functions for custom annotations

ampant:
### Proposed changes

### Checklist

### DIFF
--- a/docs/custom-annotations.md
+++ b/docs/custom-annotations.md
@@ -86,6 +86,37 @@ If you'd like to use custom annotations with Mergeable Ingress resources, please
 
 * Minions do not inherent custom annotations of the master.
 
+### Helper Functions
+
+Helper functions can be used in custom templates to transform the value of custom annotations.
+
+| Function | Description |
+| -------- | ----------- |
+| `splitinput` | Splits the arguments at specified delimiter and returns an array of strings |
+| `triminput` | Trims the trailing and leading whitespaces from the arguments |
+
+For example, the following custom annotation custom.nginx.org/allowed-ips represents a comma separated list of IP addresses as a string.
+```
+annotations:
+  custom.nginx.org/allowed-ips: "192.168.1.3, 10.0.0.13"
+```
+
+ It is possible to use helper functions such as splitinput and triminput to transform the annotation into something meaningful for NGINX. The example below shows how to use these functions in a custom template for the given annotation:
+
+```
+{{range $ip := splitinput (index $.Ingress.Annotations "custom.nginx.org/allowed-ips") ","}}
+    allow {{triminput $ip}};
+{{end}}
+deny all;
+```
+
+The template will generate the following configuration:
+```
+allow 192.168.1.3;
+allow 10.0.0.13;
+deny all;
+```
+
 ## Example
 
 See the [custom annotations example](../examples/custom-annotations).
--- a/docs/custom-annotations.md
+++ b/docs/custom-annotations.md
@@ -92,8 +92,8 @@ Helper functions can be used in the Ingress template to parse the values of cust
 
 | Function | Input Arguments| Return Arguments | Description |
 | -------- | -------------- | ---------------- | ----------- |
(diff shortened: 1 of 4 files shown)


internal/configs/version1/templates_test.go
```diff
@@ -169,3 +169,47 @@ func TestMainForNGINX(t *testing.T) {
...
+	tmpl, err := template.New("testTemplate").Funcs(helperFunctions).Parse(tpl)
+	if err != nil {
+		t.Fatalf("Failed to parse template: %v", err)
+	}
+	var buf bytes.Buffer
+
+	slice := "foo,bar"
+	expected := "foo bar "
+	err = tmpl.Execute(&buf, slice)
+	t.Log(buf.String())
```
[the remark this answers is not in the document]
pleshakov: Consider removing it here. 
The other tests do not check the generated config against the expected output -- they only check that there is no error returned.  That's mostly why `t.Log(buf.String())`, so that the output could be checked visually. In contrast, this test checks the output against the expected output, so no printing is necessary.

docs/custom-annotations.md
```diff
@@ -88,29 +88,29 @@ If you'd like to use custom annotations with Mergeable Ingress resources, please
...
-Helper functions can be used in custom templates to transform the value of custom annotations.
+Helper functions can be used in the Ingress template to parse the values of custom annotations.
 
-| Function | Description |
-| -------- | ----------- |
-| `splitinput` | Splits the arguments at specified delimiter and returns an array of strings |
-| `triminput` | Trims the trailing and leading whitespaces from the arguments |
+| Function | Input Arguments| Return Arguments | Description |
+| -------- | -------------- | ---------------- | ----------- |
+| `split` | s, sep string | []string | Splits the string s into a slice of strings separated by the sep. |
```
pleshakov:
```suggestion
| `split` | `s, sep string` | `[]string` | Splits the string `s` into a slice of strings separated by the `sep`. |
```

docs/custom-annotations.md
```diff
@@ -88,29 +88,29 @@ If you'd like to use custom annotations with Mergeable Ingress resources, please
...
+| `trim` | s string | string | Trims the trailing and leading whitespace from the string s. |
```
pleshakov:
```suggestion
| `trim` | `s string` | `string` | Trims the trailing and leading whitespace from the string `s`. |
```

internal/configs/version1/template_helper.go
```diff
@@ -5,18 +5,17 @@ import (
 	"text/template"
 )
 
-// SplitInput splits the input from "," and returns an array of strings
-func SplitInput(s string, delim string) []string {
+// splitinput splits the input from "," and returns an array of strings
+func splitinput(s string, delim string) []string {
```
pleshakov: For multiword names, the convention is to use mixed caps. Like "SplitInput". See https://golang.org/doc/effective_go.html#mixed-caps
At the same time, as mentioned here https://github.com/nginxinc/kubernetes-ingress/pull/691#discussion_r323989004, I suggest renaming to `split`. This will make the name simpler, as the suffix input seems unnecessary. 
Same applies for `triminput`.

internal/configs/version1/template_helper.go
```diff
@@ -5,18 +5,17 @@ import (
...
+// triminput trims the leading and trailing spaces in the string
+func triminput(s string) string {
 	return strings.TrimSpace(s)
 }
 
-// HelperFunctions to parse the annotations
 var helperFunctions = template.FuncMap{
-	"splitinput": SplitInput, //returns array of strings
-	"triminput":  TrimInput,  //returns string with trimmed leading and trailing spaces
+	"split": splitinput, //returns array of strings
```
pleshakov: I suggest removing these comments completely. The comment for split  s already outdated. Having the docs in the doc is enough.

internal/configs/version1/template_helper.go
pleshakov: The documentation for the function is not correct. I suggest removing it complety - the function is private and also strait-forward.
Same applies for triminput.

PR #695, opened 2019-09-13
Add tests for dynamic configuration for VS/VSR

tellet:
Also enable tests to call Nginx APIs directly (open 8080 port of IC container).


tellet:
In the end all the tests will run with API port enabled by default.

PR #707, opened 2019-09-26
Emit warning events in VS/VSR

Rulox:
### Proposed changes
To give more insights to the end user that have no access to the IC logs, now we report warnings for all VS and VSR if they are added or updated but with some errors.

### DIFF
--- a/internal/configs/virtualserver_test.go
+++ b/internal/configs/virtualserver_test.go
@@ -287,9 +287,14 @@ func TestGenerateVirtualServerConfig(t *testing.T) {
 	isPlus := false
 	isResolverConfigured := false
 	tlsPemFileName := ""
-	result := generateVirtualServerConfig(&virtualServerEx, tlsPemFileName, &baseCfgParams, isPlus, isResolverConfigured)
+	vsc := newVirtualServerConfigurator(&baseCfgParams, isPlus, isResolverConfigured)
+	result, warnings := vsc.GenerateVirtualServerConfig(&virtualServerEx, tlsPemFileName)
 	if !reflect.DeepEqual(result, expected) {
-		t.Errorf("generateVirtualServerConfig returned \n%v but expected \n%v", result, expected)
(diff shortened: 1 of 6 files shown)


internal/configs/virtualserver_test.go
```diff
@@ -287,9 +287,14 @@ func TestGenerateVirtualServerConfig(t *testing.T) {
...
-	result := generateVirtualServerConfig(&virtualServerEx, tlsPemFileName, &baseCfgParams, isPlus, isResolverConfigured)
+	vsc := newVirtualServerConfigurator(&baseCfgParams, isPlus, isResolverConfigured)
+	result, warnings := vsc.GenerateVirtualServerConfig(&virtualServerEx, tlsPemFileName)
 	if !reflect.DeepEqual(result, expected) {
-		t.Errorf("generateVirtualServerConfig returned \n%v but expected \n%v", result, expected)
+		t.Errorf("GenerateVirtualServerConfig returned \n%v but expected \n%v", result, expected)
+	}
+
+	if len(warnings) != 0 {
+		t.Errorf("GenerateVirtualServerConfig returned warnings for %v", virtualServerEx)
```
pleshakov: I suggest printing warnings instead of virtualServerEx, otherwise we don't know what warnings were generated. Also, `virtualServerEx` is a big structure and it is defined in the test. no need to print it. 

same suggestions for line 519, 781.

internal/k8s/controller_test.go
```diff
@@ -1507,3 +1507,14 @@ func TestFindVirtualServersForVirtualServerRoute(t *testing.T) {
...
 }
+
+func TestFormatWarnings(t *testing.T) {
+	warnings := []string{"Test warning", "Test warning 2"}
+
+	expected := "Test warning; Test warning 2"
+	result := formatWarningMessages(warnings)
+
+	if result != expected {
+		t.Errorf("formatWarnings(%v) returned %v but expected %v", warnings, result, expected)
```
pleshakov: formatWarnings -> formatWarningMessages
please update the name of the test function accordingly as well.

PR #709, opened 2019-09-27
Include latest tests improvements.

tellet:
Add pytest hook to update test report. Change condition back to 'ContainersReady'. Increase time of waiting for Deployment removal. Set strict versions to python libraries.


tests/suite/resources_utils.py
```diff
@@ -203,7 +203,7 @@ def are_all_pods_in_ready_state(v1: CoreV1Api, namespace) -> bool:
             return False
         for condition in pod.status.conditions:
             # wait for 'Ready' state instead of 'ContainersReady' for backwards compatibility with k8s 1.10
```
Dean-Coakley:
```suggestion
```

tellet: thanks for noticing

PR #710, opened 2019-09-27
Run IC as non-root

### LINKED ISSUE #529: Would you consider building unprivileged nginx-ingress controller or have you got this already?
We gain tls termination over tcp with this nginx-ingress controller, we appreciate a lot.
In order to follow the security compliance, we need to start pod using nonroot user.

I noticed there is unprivileged nginx image:  [https://github.com/nginxinc/docker-nginx-unprivileged/blob/master/README.md](https://github.com/nginxinc/docker-nginx-unprivileged/blob/master/README.md), and I followed the instructions [https://github.com/nginxinc/kubernetes-ingress/tree/v1.4.5/build](https://github.com/nginxinc/kubernetes-ingress/tree/v1.4.5/build) to build nginx-ingress after tweaking the dockerfiles and nginx.tmpl.
There are some privileged locations hard-coded in go source code, for example: `listen unix:/var/run/nginx-config-version.sock`, so the nginx-ingress binary built by go doesn't really want to run in unprivileged way. 
We have no expert on Go language, so we are pretty much stuck here.

Has this project ever been considered to build unprivileged nginx-ingress?
If not, what is the reason or concerns?
If yes, that's great.

Rulox:
### Proposed changes
Closes #529 

This PR makes the changes in all the Dockerfiles (and IC code) to run IC and NGINX as non root. 

Changes are:

* Containers will use `nginx` as user instead of root
* IC and NGINX will run using the `nginx` user
* Changed the `pid`, opentracing config and unix `sockets` to `/var/lib/nginx`
* Used `setcap` so we can still use ports 80 and 443 without being root
* Removed `user` directive in NGINX conf (not needed anymore as we are running NGINX as `nginx` user)
* Added security policies to IC Pod in all manifests (and helm manifests)

PS: I have arranged the Dockerfiles to keep consistency (order, number of layers, etc).

Thanks to @LorcanMcVeigh as he started working on this.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -6,14 +6,24 @@ RUN ln -sf /proc/1/fd/1 /var/log/nginx/access.log \
 	&& ln -sf /proc/1/fd/1 /var/log/nginx/stream-access.log \
 	&& ln -sf /proc/1/fd/2 /var/log/nginx/error.log
 
+RUN mkdir -p /var/lib/nginx \
+	&& mkdir -p /etc/nginx/secrets \
+	&& apt-get update \
+	&& apt-get install -y libcap2-bin \
+	&& setcap 'cap_net_bind_service=+ep' /usr/sbin/nginx \
+	&& chown -R nginx:0 /etc/nginx \
+	&& chown -R nginx:0 /var/cache/nginx \
+	&& chown -R nginx:0 /var/lib/nginx \
+	&& apt-get remove --purge -y libcap2-bin \
+	&& rm /etc/nginx/conf.d/* \
+	&& rm -rf /var/lib/apt/lists/*
+
 COPY nginx-ingress internal/configs/version1/nginx.ingress.tmpl internal/configs/version1/nginx.tmpl internal/configs/version2/nginx.virtualserver.tmpl /
-
-RUN rm /etc/nginx/conf.d/*
-
-RUN mkdir -p /etc/nginx/secrets
 
 # Uncomment the line below if you would like to add the default.pem to the image
 # and use it as a certificate and key for the default server
 # ADD default.pem /etc/nginx/secrets/default
+
+USER nginx
 
 ENTRYPOINT ["/nginx-ingress"]
--- a/build/DockerfileForAlpine
+++ b/build/DockerfileForAlpine
@@ -6,14 +6,23 @@ RUN ln -sf /proc/1/fd/1 /var/log/nginx/access.log \
 	&& ln -sf /proc/1/fd/1 /var/log/nginx/stream-access.log \
 	&& ln -sf /proc/1/fd/2 /var/log/nginx/error.log
 
+RUN mkdir -p /etc/nginx/secrets \
+	&& mkdir -p /var/lib/nginx \
+	&& apk add --no-cache libcap \
+	&& setcap 'cap_net_bind_service=+ep' /usr/sbin/nginx \
+	&& chown -R nginx:0 /etc/nginx \
+	&& chown -R nginx:0 /var/cache/nginx \
+	&& chown -R nginx:0 /var/lib/nginx \
+	&& apk del libcap \
+	&& rm /etc/nginx/conf.d/* \
+	&& rm -rf /var/cache/apk/*
+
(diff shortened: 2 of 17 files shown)


Rulox:
Checking CI :)

build/DockerfileForAlpine
```diff
@@ -6,14 +6,22 @@ RUN ln -sf /proc/1/fd/1 /var/log/nginx/access.log \
...
-RUN rm /etc/nginx/conf.d/*
+RUN mkdir -p /etc/nginx/secrets \
+	&& mkdir -p /var/lib/nginx \
+	&& apk add --no-cache libcap \
+	&& setcap 'cap_net_bind_service=ep' /usr/sbin/nginx \
+	&& chown -R nginx:0 /etc/nginx \
+	&& chown -R nginx:0 /var/cache/nginx \
+	&& chown -R nginx:0 /var/lib/nginx \
+	&& rm /etc/nginx/conf.d/* \
+	&& rm -rf /var/lib/apt/lists/*
```
pleshakov: alpine is not using apt.

build/Dockerfile
```diff
@@ -6,14 +6,23 @@ RUN ln -sf /proc/1/fd/1 /var/log/nginx/access.log \
...
 	&& ln -sf /proc/1/fd/2 /var/log/nginx/error.log
 
-COPY nginx-ingress internal/configs/version1/nginx.ingress.tmpl internal/configs/version1/nginx.tmpl internal/configs/version2/nginx.virtualserver.tmpl /
-
-RUN rm /etc/nginx/conf.d/*
+RUN mkdir -p /var/lib/nginx \
+	&& mkdir -p /etc/nginx/secrets \
+	&& apt-get update \
+	&& apt-get install -y libcap2-bin \
+	&& setcap 'cap_net_bind_service=ep' /usr/sbin/nginx \
```
pleshakov: I think it makes sense to remove this package once we used the `setcap` binary.

deployments/daemon-set/nginx-ingress.yaml
```diff
@@ -16,6 +16,14 @@ spec:
        #prometheus.io/port: "9113"
     spec:
       serviceAccountName: nginx-ingress
+      securityContext:
```
pleshakov: does it make sense to define the security context at the container level? those only make sense for the nginx container.  although we don't have any other container, what if somebody would want to add some sidecar container to the Ingress Controller pod?

build/DockerfileForAlpine
Rulox: Ooops! good catch

deployments/daemon-set/nginx-ingress.yaml
Rulox: Interesting question. Yes, I think you are right, we shouldn't set capabilities for the whole pod but just for the IC container. Will change, thanks for the heads up!

Rulox: Please note that changing this to the container will require `allowPrivilegeEscalation: true` instead of `false` (even with setcap). Otherwise, we get:

```
2019/09/30 13:59:15 [emerg] 17#17: bind() to 0.0.0.0:80 failed (13: Permission denied)
nginx: [emerg] bind() to 0.0.0.0:80 failed (13: Permission denied)
```

build/DockerfileForPlus
```diff
@@ -46,19 +47,25 @@ RUN set -x \
...
-	&& ln -sf /proc/1/fd/2 /var/log/nginx/error.log
+  && ln -sf /proc/1/fd/1 /var/log/nginx/stream-access.log \
+  && ln -sf /proc/1/fd/2 /var/log/nginx/error.log
 
+RUN  mkdir -p /var/lib/nginx \
+  && mkdir -p /etc/nginx/secrets \
+  && chown -R nginx:0 /etc/nginx \
+  && chown -R nginx:0 /var/cache/nginx \
+  && chown -R nginx:0 /var/lib/nginx/ \
+	&& apt-get remove --purge -y libcap2-bin \
```
Dean-Coakley: Indentation level looks wrong here
```suggestion
  && apt-get remove --purge -y libcap2-bin \
```

Rulox: Forgot to push :) thanks

PR #714, opened 2019-09-30
Update nginx-go-client to v0.5.0

ampant:
### Proposed changes
Updated go.mod with latest version of go-client


Rulox:
Why do we need this ? @ampant @Dean-Coakley

PR #712, opened 2019-09-27
Add service subselector support in vs/vsr

Dean-Coakley:
### Proposed changes
* When using VS/VSRs, Subselector support allows you to configure NGINX to route requests to different groups of pods of the same service based on kubernetes labels.

### Checklist
Before creating a PR, run through this checklist and mark each as complete.

- [x] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/master/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that all unit tests pass after adding my changes
- [x] I have updated necessary documentation
- [x] I have rebased my branch onto master
- [x] I will ensure my PR is targeting the master branch and pulling from my branch from my own fork

PR #719, opened 2019-10-07
Add test: Dynamic update in vsr with subselector

Dean-Coakley:
### Proposed changes
Extend unit test coverage for subselector support added in: https://github.com/nginxinc/kubernetes-ingress/pull/712

Add a test case for dynamic updates when using a subselector in a VSR.

### Checklist

### DIFF
--- a/internal/configs/virtualserver_test.go
+++ b/internal/configs/virtualserver_test.go
@@ -1206,7 +1206,7 @@ func TestGenerateSSLConfig(t *testing.T) {
 	}
 }
 
-func TestCreateUpstreamServersForPlus(t *testing.T) {
+func TestCreateUpstreamsForPlus(t *testing.T) {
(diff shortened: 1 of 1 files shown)


internal/configs/virtualserver_test.go
```diff
@@ -1247,6 +1247,10 @@ func TestCreateUpstreamServersForPlus(t *testing.T) {
 						Path:  "/coffee",
 						Route: "default/coffee",
 					},
+					{
+						Path:  "/sub",
```
pleshakov: while it works, the Route that you have here is invalid, because there is no `default/sub` VSR.

Consider the following changes to make the input valid:
1. Remove the `/sub` route completely in the VS.
2. Change the `/sub` subroute in the VSR from `/sub` to `/coffee/sub`.
Both changes will make the VS and VSR valid.

PR #722, opened 2019-10-10
Fix trailing leader election cm when using helm

Dean-Coakley:
### Proposed changes
When removing an ingress controller deployment managed by helm.
A leader election configmap was left behind because it was created by
the IC and not managed by Tiller/Helm.

This PR pre-allocates the configmap so it is now managed by Helm.
The created configmap is blank, but once the Ingress Controller starts it updates the configmap with necessary Annotations required for leader election.

### Edge cases
If the configmap somehow already exists. The helm deployment fails cleanly at the client level.
```
$ kubectl create cm dev-nginx-ingress-leader-election
configmap/dev-nginx-ingress-leader-election created
$ helm install --name dev deployments/helm-chart -f deployments/helm-chart/values.yaml
Error: release dev failed: configmaps "dev-nginx-ingress-leader-election" already exists
```

### Checklist

### DIFF
--- a/deployments/helm-chart/templates/controller-leader-election-configmap.yaml
+++ b/deployments/helm-chart/templates/controller-leader-election-configmap.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "nginx-ingress.leaderElectionName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "nginx-ingress.labels" . | nindent 4 }}
+data:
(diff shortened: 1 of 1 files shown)


deployments/helm-chart/templates/controller-leader-election-configmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "nginx-ingress.leaderElectionName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "nginx-ingress.labels" . | nindent 4 }}
+data:
```
Rulox: Do you need the `data` field if is empty?

Dean-Coakley: Nope 😄

PR #724, opened 2019-10-11
Add VS/VSR Prometheus metrics

Dean-Coakley:
### Proposed changes
The IC now exports metrics for the number of VirtualServers and VirtualServerRoutes it is currently handling.

### Checklist

### DIFF
--- a/docs/prometheus.md
+++ b/docs/prometheus.md
@@ -17,6 +17,6 @@ The Ingress Controller exports the following metrics:
   * `controller_nginx_last_reload_milliseconds`. Duration in milliseconds of the last NGINX reload.
   * `controller_ingress_resources_total`. Number of handled Ingress resources. This metric includes the label type, that groups the Ingress resources by their type (regular, [minion or master](./../examples/mergeable-ingress-types))
   * `controller_virtualserver_resources_total`. Number of handled VirtualServer resources.
-  * `controller_virtualserverroute_resources_total`. Number of handled VirtualServerRoute resources.
+  * `controller_virtualserverroute_resources_total`. Number of handled VirtualServerRoute resources. **Note**: Only counts VirtualServerRoutes that have a reference from a VirtualServer.
 
 **Note**: all metrics have the namespace nginx_ingress. For example, nginx_ingress_controller_nginx_reloads_total.
--- a/docs/prometheus.md
+++ b/docs/prometheus.md
@@ -16,5 +16,7 @@ The Ingress Controller exports the following metrics:
   * `controller_nginx_last_reload_status`. Status of the last NGINX reload, 0 meaning down and 1 up.
   * `controller_nginx_last_reload_milliseconds`. Duration in milliseconds of the last NGINX reload.
   * `controller_ingress_resources_total`. Number of handled Ingress resources. This metric includes the label type, that groups the Ingress resources by their type (regular, [minion or master](./../examples/mergeable-ingress-types))
+  * `controller_virtualserver_resources_total`. Number of handled VirtualServer resources.
+  * `controller_virtualserverroute_resources_total`. Number of handled VirtualServerRoute resources.
 
 **Note**: all metrics have the namespace nginx_ingress. For example, nginx_ingress_controller_nginx_reloads_total.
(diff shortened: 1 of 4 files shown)


docs/prometheus.md
```diff
@@ -16,5 +16,7 @@ The Ingress Controller exports the following metrics:
   * `controller_nginx_last_reload_status`. Status of the last NGINX reload, 0 meaning down and 1 up.
   * `controller_nginx_last_reload_milliseconds`. Duration in milliseconds of the last NGINX reload.
   * `controller_ingress_resources_total`. Number of handled Ingress resources. This metric includes the label type, that groups the Ingress resources by their type (regular, [minion or master](./../examples/mergeable-ingress-types))
+  * `controller_virtualserver_resources_total`. Number of handled VirtualServer resources.
+  * `controller_virtualserverroute_resources_total`. Number of handled VirtualServerRoute resources.
```
Rulox: Should we add a clarification that the VSR needs to be referenced by a VS in order to appear in the metrics? Just in case users can be confused if they deploy a VSR and metrics are not updating. What do you think @Dean-Coakley , @pleshakov ?

Dean-Coakley: I will add a note. While I think familiar users should know what `handled` means. New users may be misled. Happy to revert the change if @pleshakov disagrees.

internal/metrics/collectors/controller.go
```diff
@@ -7,43 +7,79 @@ var labelNamesController = []string{"type"}
 // ControllerCollector is an interface for the metrics of the Controller
 type ControllerCollector interface {
 	SetIngressResources(ingressType string, count int)
+	SetVsResources(count int)
```
pleshakov: For consistency with the existing method:
SetVsResources -> SetVirtualServerResources
SetVsrResources -> SetVirtualServerRouteResources

You can also remove resources in all three methods and make them plural to be less verbose:
SetVirtualServers

internal/configs/configurator.go
```diff
@@ -637,3 +643,13 @@ func (cnf *Configurator) GetIngressCounts() map[string]int {
 
 	return counters
 }
+
+// GetVSCounts returns the total count of Vs/Vsr resources that are handled by the Ingress Controller
+func (cnf *Configurator) GetVSCounts() (vsCount int, vsrCount int) {
```
pleshakov: I suggest expanding the abbreviation for clarity and consistency with GetIngressCounts:
GetVirtualServerCounts

In the comments and other places, if you abbreviate VirtualServer and VirtualServerRoute, better to use "VS" and "VSR" instead of "Vs" and "Vsr"

internal/metrics/collectors/controller.go
pleshakov: Please also update `updateVsMetrics` in Controller` to `updateVirtualServerMetrics` for the same reasons.

docs/prometheus.md
```diff
@@ -16,5 +16,7 @@ The Ingress Controller exports the following metrics:
   * `controller_nginx_last_reload_status`. Status of the last NGINX reload, 0 meaning down and 1 up.
   * `controller_nginx_last_reload_milliseconds`. Duration in milliseconds of the last NGINX reload.
   * `controller_ingress_resources_total`. Number of handled Ingress resources. This metric includes the label type, that groups the Ingress resources by their type (regular, [minion or master](./../examples/mergeable-ingress-types))
+  * `controller_virtualserver_resources_total`. Number of handled VirtualServer resources.
+  * `controller_virtualserverroute_resources_total`. Number of handled VirtualServerRoute resources. **Note**: Only counts VirtualServerRoutes that have a reference from a VirtualServer.
```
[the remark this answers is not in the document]
Dean-Coakley: Yeah, that seems better

internal/metrics/collectors/controller.go
```diff
@@ -6,17 +6,17 @@ var labelNamesController = []string{"type"}
 
 // ControllerCollector is an interface for the metrics of the Controller
 type ControllerCollector interface {
-	SetIngressResources(ingressType string, count int)
-	SetVsResources(count int)
-	SetVsrResources(count int)
+	SetIngress(ingressType string, count int)
```
pleshakov: as suggested here https://github.com/nginxinc/kubernetes-ingress/pull/724#discussion_r335249432 , please make them plural before merging

PR #726, opened 2019-10-15
Set a strict version to `cffi`

tellet:
Tests need `pyopenssl` and it has a dependency named `cffi`. Set a strict version to `cffi` cause its lates update (1.13.0) broke the tests image.


Rulox:
Lgtm

PR #728, opened 2019-10-18
Add session persistence in VS/VSR

Dean-Coakley:
### Proposed changes
Add session persistence to Upstreams in VirtualServers.

See https://nginx.org/en/docs/http/ngx_http_upstream_module.html?#sticky for more details
Requires NGINX Plus.

### Checklist


docs/virtualserver-and-virtualserverroute.md
```diff
@@ -307,7 +308,37 @@ healthCheck:
...
 | `headers` | The request headers used for health check requests. NGINX Plus always sets the `Host`, `User-Agent` and `Connection` headers for health check requests. | [`[]header`](#Header) | No |
-| `statusMatch` | The expected response status codes of a health check.  By default, the response should have status code 2xx or 3xx. Examples: `“200”`, `“! 500”`, `"301-303 307"`. See the documentation of the [match](https://nginx.org/en/docs/http/ngx_http_upstream_hc_module.html?#match) directive. | `string` | No |
+| `statusMatch` | The expected response status codes of a health check. By default, the response should have status code 2xx or 3xx. Examples: `“200”`, `“! 500”`, `"301-303 307"`. See the documentation of the [match](https://nginx.org/en/docs/http/ngx_http_upstream_hc_module.html?#match) directive. | `string` | No |
+
+### Upstream.SessionCookie
+
+The SessionCookie field configures session persistence which allows requests from the same client to be passed to the same upstream server. The information about the designated upstream server is passed in a session cookie generated by NGINX Plus.
+
+```yaml
```
pleshakov: Add a paragraph before the yaml:
In the example below, we configure session persistence with a session cookie for an upstream and configure all the available parameters:

internal/configs/version2/config.go
```diff
@@ -98,6 +99,17 @@ type HealthCheck struct {
 	Match               string
 }
 
+// SessionCookie defines a cookie for an upstream.
```
pleshakov:
```suggestion
// SessionCookie defines a session cookie for an upstream.
```

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -15,6 +15,10 @@ upstream {{ $u.Name }} {
     {{ if $u.Queue }}
     queue {{ $u.Queue.Size }} timeout={{ $u.Queue.Timeout }};
     {{ end }}
+
+    {{ with $u.SessionCookie }}
+    {{ if .Enable }}sticky cookie {{ .Name }}{{ if .Expires }} expires={{ .Expires }}{{ end }}{{ if .Domain }} domain={{ .Domain }}{{ end }}{{ if .HTTPOnly }} httponly{{ end }}{{ if .Secure }} secure{{ end }}{{ if .Path }} path={{ .Path }}{{ end }};{{ end }}
```
pleshakov: please rewrite sligihtly for better readability:
```
    {{ with $u.SessionCookie }}
        {{ if .Enable }}
    sticky cookie {{ .Name }}{{ if .Expires }} expires={{ .Expires }}{{ end }}{{ if .Domain }} domain={{ .Domain }}{{ end }}{{ if .HTTPOnly }} httponly{{ end }}{{ if .Secure }} secure{{ end }}{{ if .Path }} path={{ .Path }}{{ end }};
        {{ end }}
    {{ end }}
```

internal/configs/virtualserver_test.go
```diff
@@ -2412,3 +2412,33 @@ func TestGenerateQueueForPlus(t *testing.T) {
 	}
 
 }
+
+func TestGenerateSessionCookie(t *testing.T) {
+	tests := []struct {
+		sc   *conf_v1alpha1.SessionCookie
+		want *version2.SessionCookie
```
pleshakov: the other tests use expected/result instead want/got convention. Please stick with the former.

pkg/apis/configuration/validation/validation.go
```diff
@@ -210,6 +210,24 @@ func validateUpstreamHealthCheck(hc *v1alpha1.HealthCheck, fieldPath *field.Path
 	return allErrs
 }
 
+func validateSessionCookie(sc *v1alpha1.SessionCookie, fieldPath *field.Path) field.ErrorList {
```
pleshakov: Proper validation is missing:

* For name, use `isCookieName` function.
* For domain, use `IsDNS1123Subdomain`, although the leading `.` should be allowed.

pkg/apis/configuration/validation/validation.go
```diff
@@ -210,6 +210,24 @@ func validateUpstreamHealthCheck(hc *v1alpha1.HealthCheck, fieldPath *field.Path
...
+	allErrs := field.ErrorList{}
+
+	if sc == nil || !sc.Enable {
```
pleshakov: even if `sc.Enable` false, it is better to validate the fields.

pkg/apis/configuration/validation/validation.go
```diff
@@ -210,6 +210,24 @@ func validateUpstreamHealthCheck(hc *v1alpha1.HealthCheck, fieldPath *field.Path
...
+		return allErrs
+	}
+
+	if sc.Name == "" {
+		allErrs = append(allErrs, field.Required(fieldPath.Child("name"), ""))
+	}
+	if sc.Path != "" {
+		allErrs = append(allErrs, validatePath(sc.Path, fieldPath.Child("path"))...)
+	}
+	allErrs = append(allErrs, validateTime(sc.Expires, fieldPath.Child("expires"))...)
```
pleshakov: what about special value `max`?

pkg/apis/configuration/validation/validation.go
Rulox: Just a small thing here @pleshakov @Dean-Coakley , we do it here: https://github.com/nginxinc/kubernetes-ingress/blob/master/pkg/apis/configuration/validation/validation.go#L182

For consistency I suggest either change it in HC (different PR) or leave as it is here.

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -307,7 +308,39 @@ healthCheck:
...
+
+
+| Field | Description | Type | Required |
+| ----- | ----------- | ---- | -------- |
+| `enable` | Enables session persistence with a session cookie for an upstream server. The default is `false`. | `boolean` | No |
+| `name` | The name of the cookie. | `string` | Yes |
+| `path` | The path for which the cookie is set. | `string` | No |
+| `expires` | The time for which a browser should keep the cookie. Can be set to the special value `max`, which will cause the cookie to expire on `31 Dec 2037 23:55:55 GMT`. | `string` | No |
+| `domain` | The domain for which the cookie is set. | `string` | No |
+| `httpOnly` | Adds the `HttpOnly` attribute is added to the cookie. | `boolean` | No |
```
Rulox: This line (description) and the next one seem weird?  `Adds the X attribute is added to the cookie` , shouldn't it be `Adds the X attribute to the cookie` or `X attribute is added to the cookie` ??

pkg/apis/configuration/validation/validation.go
```diff
@@ -210,6 +210,38 @@ func validateUpstreamHealthCheck(hc *v1alpha1.HealthCheck, fieldPath *field.Path
...
+
+	if sc.Path != "" {
+		allErrs = append(allErrs, validatePath(sc.Path, fieldPath.Child("path"))...)
+	}
+	if sc.Expires != "max" {
+		allErrs = append(allErrs, validateTime(sc.Expires, fieldPath.Child("expires"))...)
+	}
+
+	if sc.Domain != "" {
+		// A Domain prefix of "." is allowed.
```
Rulox: As discussed, this could be more implicit if you think it makes sense

pkg/apis/configuration/validation/validation.go
```diff
@@ -210,6 +210,38 @@ func validateUpstreamHealthCheck(hc *v1alpha1.HealthCheck, fieldPath *field.Path
...
+	} else {
+		for _, msg := range isCookieName(sc.Name) {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("name"), sc.Name, msg))
+		}
+	}
+
+	if sc.Path != "" {
+		allErrs = append(allErrs, validatePath(sc.Path, fieldPath.Child("path"))...)
+	}
+	if sc.Expires != "max" {
```
Rulox: I would add a `\n` here for consistency between `ifs`

docs/virtualserver-and-virtualserverroute.md
Dean-Coakley: Yep, doesn't make sense. Changing both to `Adds the X attribute to the cookie`

pkg/apis/configuration/validation/validation.go
pleshakov: @Rulox good catch! 

I think it is better to validate in both cases. This way any problem could be discovered earlier before a hc/sc is enabled.

PR #753, opened 2019-11-04
Fix: Allow TLSRedirect.basedOn to be unset

Dean-Coakley:
### Proposed changes
Given the following config
```yaml
redirect:
  enable: true
  code: 301
```

The VS resource would be rejected:
```
Warning  Rejected 44s nginx-ingress-controller  VirtualServer test-namespace-1572877439202/virtual-server-tls is invalid and was rejected: spec.tls.redirect.basedOn: Invalid value: “”: accepted values are ‘scheme’ or ‘x-forwarded-proto’
```

This PR now accepts `""` (unset) as a valid input for TLSRedirect so the VS resource is accepted. The default `$scheme` value is then generated for basedOn.

New behaviour:
```
Event(v1.ObjectReference{Kind:"VirtualServer", Namespace:"default", Name:"cafe", UID:"0d840dbe-61e1-4a8c-a9ee-2b397e906a59", APIVersion:"k8s.nginx.org/v1alpha1", ResourceVersion:"354233", FieldPath:""}): type: 'Normal' reason: 'AddedOrUpdated' Configuration for default/cafe was added or updated
```
### Checklist

### DIFF
--- a/pkg/apis/configuration/validation/validation.go
+++ b/pkg/apis/configuration/validation/validation.go
@@ -76,7 +76,7 @@ func validateTLSRedirect(redirect *v1alpha1.TLSRedirect, fieldPath *field.Path) 
 	}
 
 	if redirect.BasedOn != "" && redirect.BasedOn != "scheme" && redirect.BasedOn != "x-forwarded-proto" {
-		allErrs = append(allErrs, field.Invalid(fieldPath.Child("basedOn"), redirect.BasedOn, "accepted values are '', 'scheme', 'x-forwarded-proto'"))
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("basedOn"), redirect.BasedOn, "accepted values are 'scheme', 'x-forwarded-proto'"))
 	}
 
 	return allErrs
--- a/pkg/apis/configuration/validation/validation.go
+++ b/pkg/apis/configuration/validation/validation.go
@@ -75,8 +75,8 @@ func validateTLSRedirect(redirect *v1alpha1.TLSRedirect, fieldPath *field.Path) 
 		allErrs = append(allErrs, validateTLSRedirectStatusCode(*redirect.Code, fieldPath.Child("code"))...)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/apis/configuration/validation/validation_test.go
```diff
@@ -103,6 +103,13 @@ func TestValidateTLS(t *testing.T) {
 				BasedOn: "scheme",
 			},
 		},
+		{
+			Secret: "my-secret",
```
pleshakov: Does it make sense to add cases for the following valid cases as well?
* `Redirect: &v1alpha1.TLSRedirect{}`
*  `Redirect: &v1alpha1.TLSRedirect{Enable: true}`

pkg/apis/configuration/validation/validation.go
```diff
@@ -75,8 +75,8 @@ func validateTLSRedirect(redirect *v1alpha1.TLSRedirect, fieldPath *field.Path)
 		allErrs = append(allErrs, validateTLSRedirectStatusCode(*redirect.Code, fieldPath.Child("code"))...)
 	}
 
-	if redirect.BasedOn != "scheme" && redirect.BasedOn != "x-forwarded-proto" {
-		allErrs = append(allErrs, field.Invalid(fieldPath.Child("basedOn"), redirect.BasedOn, "accepted values are 'scheme' or 'x-forwarded-proto'"))
+	if redirect.BasedOn != "" && redirect.BasedOn != "scheme" && redirect.BasedOn != "x-forwarded-proto" {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("basedOn"), redirect.BasedOn, "accepted values are '', 'scheme', 'x-forwarded-proto'"))
```
pleshakov: If the field is `''`, we assume that it is set to the default. That's why it is not necessary to report that `''` is one of the accepted value:
```suggestion
		allErrs = append(allErrs, field.Invalid(fieldPath.Child("basedOn"), redirect.BasedOn, "accepted values are 'scheme', 'x-forwarded-proto'"))
```

Dean-Coakley: sure

PR #755, opened 2019-11-06
Refactor tests code.

tellet:
Gather common assertion methods in one place. Replace 'nginx.com' as ExternalName host with an internal backend.

### DIFF
--- a/tests/suite/custom_assertions.py
+++ b/tests/suite/custom_assertions.py
@@ -5,7 +5,7 @@ from suite.custom_resources_utils import get_vs_nginx_template_conf
 
 
 def assert_no_new_events(old_list, new_list):
-    assert len(old_list) == len(new_list), "expected: lists are the same"
+    assert len(old_list) == len(new_list), "Expected: lists are of the same size"
     for i in range(len(new_list) - 1, -1, -1):
         if old_list[i].count != new_list[i].count:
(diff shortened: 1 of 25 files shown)


tests/suite/custom_assertions.py
```diff
@@ -0,0 +1,120 @@
+"""Describe the custom assertion methods"""
+import pytest
+
+from suite.custom_resources_utils import get_vs_nginx_template_conf
+
+
+def assert_no_new_events(old_list, new_list):
+    assert len(old_list) == len(new_list), "expected: lists are the same"
```
Rulox:
```suggestion
    assert len(old_list) == len(new_list), "expected: lists are the same size"
```

tests/suite/resources_utils.py
```diff
@@ -981,10 +981,10 @@ def ensure_response_from_backend(req_url, host) -> None:
     :param host:
     :return:
     """
-    for _ in range(10):
+    for _ in range(30):
         resp = requests.get(req_url, headers={"host": host}, verify=False)
         if resp.status_code != 502:
             print(f"At last after {_ * 2} seconds got 200. Continue...")
```
pleshakov: that fact that you didn't get 502 doesn't mean you get 200 :)

tellet: well, sometimes I don't expect 200, but i.e. 404 ;) I haven't ever met 504 before and still can not reproduce it in GCP or in minikube.

PR #764, opened 2019-11-15
Tests ingress class

tellet:
Add tests on IC cli arguments: `-watch-namespace`, `-ingress-class` and `-use-ingress-class-only`.

I split the changes into 2 commits to simplify the review process, I'll squash them later.

### DIFF
--- a/tests/data/ingress-class/no-class-ingress.yaml
+++ b/tests/data/ingress-class/no-class-ingress.yaml
@@ -1,0 +1,17 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: no-class-ingress
+spec:
+  rules:
+  - host: no-class.example.com
+    http:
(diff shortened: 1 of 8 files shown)


tests/data/ingress-class/no-class-ingress.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  annotations:
```
Rulox: I think you can remove `annotations` if empty?

tellet: I was thinking to keep it as a hint: "look, annotations are empty". I can remove this for sure

Rulox: Ah ok, I am ok with that too :)

PR #766, opened 2019-11-18
Add regex support to location paths

Rulox:
### Proposed changes
Add the ability to use exact matches (`=`) and case sensitive/insensitive regular expressions (`~`, `~*`) in VS/VSR route paths.

@tellet Please have a look at the Test methods and let me know if more edge cases can be added to improve validation, thank you.

### DIFF
--- a/docs/virtualserver-and-virtualserverroute.md
+++ b/docs/virtualserver-and-virtualserverroute.md
@@ -63,6 +63,12 @@ spec:
   - path: /coffee
     action:
       pass: coffee
+  - path: ~ ^/decaf/.*\\.jpg$
+    action:
+      pass: coffee
+  - path: = /green/tea
+    action:
+      pass: tea
 ```
 
 | Field | Description | Type | Required |
@@ -113,7 +119,7 @@ The route defines rules for matching client requests to actions like passing a r
 
 | Field | Description | Type | Required |
 | ----- | ----------- | ---- | -------- |
-| `path` | The path of the route. NGINX will match it against the URI of a request. The path must start with `/` and must not include any whitespace characters, `{`, `}` or `;`. For example, `/`, `/path` are valid. The path must be unique among the paths of all routes of the VirtualServer. | `string` | Yes |
+| `path` | The path of the route. NGINX will match it against the URI of a request. Possible values are: a prefix (`/`, `/path`), an exact match (`=/exact/match`), a case insensitive regular expression (`~*^/Bar.*\\.jpg`) or a case sensitive regular expression (`~^/foo.*\\.jpg`). In the case of a prefix (must start with `/`) or an exact match (must start with `=`), the path must not include any whitespace characters, `{`, `}` or `;`. In the case of the regex matches, all double quotes `"` must be escaped and the match can't end in an unescaped backslash `\`. The path must be unique among the paths of all routes of the VirtualServer. Check the [location](http://nginx.org/en/docs/http/ngx_http_core_module.html#location) directive for more information. | `string` | Yes |
 | `action` | The default action to perform for a request. | [`action`](#Action) | No* |
 | `splits` | The default splits configuration for traffic splitting. Must include at least 2 splits. | [`[]split`](#Split) | No* |
 | `matches` | The matching rules for advanced content-based routing. Requires the default `action` or `splits`.  Unmatched requests will be handled by the default `action` or `splits`. |[`matches`](#Match) | No |
@@ -193,7 +199,7 @@ action:
 
 | Field | Description | Type | Required |
 | ----- | ----------- | ---- | -------- |
-| `path` | The path of the subroute. NGINX will match it against the URI of a request. The path must start with the same path as the path of the route of the VirtualServer that references this resource. It must not include any whitespace characters, `{`, `}` or `;`. The path must be unique among the paths of all subroutes of the VirtualServerRoute. | `string` | Yes |
+| `path` | The path of the subroute. NGINX will match it against the URI of a request. Possible values are: a prefix (`/`, `/path`), an exact match (`=/exact/match`), a case insensitive regular expression (`~*^/Bar.*\\.jpg`) or a case sensitive regular expression (`~^/foo.*\\.jpg`). In the case of a prefix, the path must start with the same path as the path of the route of the VirtualServer that references this resource. In the case of an exact or regex match, the path must be the same as the path of the route of the VirtualServer that references this resource. In the case of a prefix or an exact match, the path must not include any whitespace characters, `{`, `}` or `;`.  In the case of the regex matches, all double quotes `"` must be escaped and the match can't end in an unescaped backslash `\`. The path must be unique among the paths of all subroutes of the VirtualServerRoute. | `string` | Yes |
 | `action` | The default action to perform for a request. | [`action`](#Action) | No* |
 | `splits` | The default splits configuration for traffic splitting. Must include at least 2 splits. | [`[]split`](#Split) | No* |
 | `matches` | The matching rules for advanced content-based routing. Requires the default `action` or `splits`.  Unmatched requests will be handled by the default `action` or `splits`. |[`matches`](#Match) | No |
--- a/internal/configs/virtualserver.go
+++ b/internal/configs/virtualserver.go
@@ -499,9 +499,21 @@ func generateBool(s *bool, defaultS bool) bool {
 	return defaultS
 }
 
+func generatePath(path string) string {
+	// Wrap the regular expression (if present) inside double quotes (") to avoid NGINX parsing errors
+	if strings.HasPrefix(path, "~*") {
+		return fmt.Sprintf(`~* "%v"`, strings.TrimPrefix(strings.TrimPrefix(path, "~*"), " "))
+	}
+	if strings.HasPrefix(path, "~") {
+		return fmt.Sprintf(`~ "%v"`, strings.TrimPrefix(strings.TrimPrefix(path, "~"), " "))
+	}
+
+	return path
+}
+
 func generateLocation(path string, upstreamName string, upstream conf_v1alpha1.Upstream, cfgParams *ConfigParams) version2.Location {
 	return version2.Location{
-		Path:                     path,
+		Path:                     generatePath(path),
 		Snippets:                 cfgParams.LocationSnippets,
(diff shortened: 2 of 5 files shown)


docs/virtualserver-and-virtualserverroute.md
```diff
@@ -113,7 +119,7 @@ The route defines rules for matching client requests to actions like passing a r
 
 | Field | Description | Type | Required |
 | ----- | ----------- | ---- | -------- |
-| `path` | The path of the route. NGINX will match it against the URI of a request. The path must start with `/` and must not include any whitespace characters, `{`, `}` or `;`. For example, `/`, `/path` are valid. The path must be unique among the paths of all routes of the VirtualServer. | `string` | Yes |
+| `path` | The path of the route. NGINX will match it against the URI of a request. Possible values are: uri (`/`, `/path`), exact matches (`=/exact/match`), case insensitive matches (`~*^/Bar.*\\.jpg`) or case sensitive matches (`~^/foo.*\\.jpg`). In the case of an uri, path must start with `/` and must not include any whitespace characters, `{`, `}` or `;`. In the case of the matches, all double quotes `"` must be escaped and the match can't end in an unescaped backslash `\`. The path must be unique among the paths of all routes of the VirtualServer. Check [location](http://nginx.org/en/docs/http/ngx_http_core_module.html#location) for more information. | `string` | Yes |
```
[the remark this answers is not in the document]
pleshakov: Note that the docs for the subroute in VSR are missing https://github.com/nginxinc/kubernetes-ingress/blob/master/docs/virtualserver-and-virtualserverroute.md#virtualserverroute-specification :(

internal/configs/virtualserver.go
```diff
@@ -499,9 +499,21 @@ func generateBool(s *bool, defaultS bool) bool {
 	return defaultS
 }
 
+func generatePath(path string) string {
+	// Wrap the regular expression (if present) inside double quotes (") to avoid NGINX parsing errors
+	if strings.HasPrefix(path, "~*") {
+		return fmt.Sprintf(`~* "%v"`, strings.TrimPrefix(strings.TrimPrefix(path, "~*"), " "))
```
pleshakov: is it possible to trim prefix only once, for `"~* "` ?
same for the `~`

or that whitespace is optional?

pkg/apis/configuration/validation/validation.go
```diff
@@ -660,7 +660,48 @@ func validateSplits(splits []v1alpha1.Split, fieldPath *field.Path, upstreamName
 	return allErrs
 }
 
-// For now, we only support prefix-based NGINX locations. For example, location /abc { ... }.
+// For now, we only support prefix-based NGINX locations, positive case-sensitive/insensitive regular expressions matches and exact matches.
```
pleshakov:
```suggestion
// We support prefix-based NGINX locations, positive case-sensitive/insensitive regular expressions matches and exact matches.
```

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -63,6 +63,12 @@ spec:
   - path: /coffee
     action:
       pass: coffee
+  - path: ~ ^/decaf/\\.*jpg$
```
pleshakov: it seems it should be:
```suggestion
  - path: ~ ^/decaf/.*\\.jpg$

```

pkg/apis/configuration/validation/validation.go
```diff
@@ -857,11 +898,38 @@ func validateVirtualServerRouteHost(host string, virtualServerHost string, field
 	return allErrs
 }
 
+var regexOrExactMatchPrefixes = []string{"~", "="}
+
+func isRegexOrExactMatch(path string) bool {
+	for _, p := range regexOrExactMatchPrefixes {
```
pleshakov: it seems like a bit of a overkill to have a loop here instead of:
```go
return strings.HasPrefix(path, "~") || strings.HasPrefix(path, "=")
```

pkg/apis/configuration/validation/validation.go
```diff
@@ -857,11 +898,38 @@ func validateVirtualServerRouteHost(host string, virtualServerHost string, field
...
+
+	return false
+}
+
 func validateVirtualServerRouteSubroutes(routes []v1alpha1.Route, fieldPath *field.Path, upstreamNames sets.String, pathPrefix string) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	allPaths := sets.String{}
 
+	if isRegexOrExactMatch(pathPrefix) {
```
pleshakov: it seems that this could be simplified. additionally, we can return in the if block  - no need to go into the loop:
```go
	if isRegexOrExactMatch(pathPrefix) {
		if len(routes) != 1 {
			return append(allErrs, field.Invalid(fieldPath, "subroutes", "must have only one subroute if regex match or exact match are being used"))
		}

		if routes[0].Path != pathPrefix {
			return append(allErrs, field.Invalid(fieldPath.Index(0).Child("path"), routes[0].Path, "must have the same path as the referenced VirtualServer route path"))
		}

		isRouteFieldForbidden := true
		return validateRoute(routes[0], fieldPath.Index(0).Child("path"), upstreamNames, isRouteFieldForbidden)
	}
```

pkg/apis/configuration/validation/validation.go
```diff
@@ -857,11 +898,38 @@ func validateVirtualServerRouteHost(host string, virtualServerHost string, field
 	return allErrs
 }
 
+var regexOrExactMatchPrefixes = []string{"~", "="}
+
+func isRegexOrExactMatch(path string) bool {
```
pleshakov: should we have a unit test for this one?

pkg/apis/configuration/validation/validation.go
```diff
@@ -857,11 +898,38 @@ func validateVirtualServerRouteHost(host string, virtualServerHost string, field
...
+	for _, p := range regexOrExactMatchPrefixes {
+		if strings.HasPrefix(path, p) {
+			return true
+		}
+	}
+
+	return false
+}
+
 func validateVirtualServerRouteSubroutes(routes []v1alpha1.Route, fieldPath *field.Path, upstreamNames sets.String, pathPrefix string) field.ErrorList {
```
pleshakov: should pathPrefix be renamed because it is no longer a prefix - could be regex or an exact match?
perhaps `vsPath` ?

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -63,6 +63,12 @@ spec:
...
+    action:
+      pass: coffee
+  - path: =/green/tea
```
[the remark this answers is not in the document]
Rulox: Yeah we can add a space, in terms of NGINX it doesn't matter but I agree we can add it for consistency thanks

internal/configs/virtualserver.go
Rulox: The whitespace is optional, is just there in case the user adds it, otherwise, this: `~* regex` will create: `~* " regex"` note the space before `regex` should not be part of the regex inside `"`

docs/virtualserver-and-virtualserverroute.md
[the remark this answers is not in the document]
Rulox:
```
In the case of a prefix or an exact match, the path must start with `/` and must not include...
```

This is not correct for exact match right? exact matches start with `=` ?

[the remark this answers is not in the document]
Rulox: Updated, let me know what u think

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -113,8 +119,7 @@ The route defines rules for matching client requests to actions like passing a r
...
-| `action` | The default action to perform for a request. | [`action`](#Action) | No* |
+| `path` | The path of the route. NGINX will match it against the URI of a request. Possible values are: a prefix (`/`, `/path`), an exact match (`=/exact/match`), a case insensitive regular expression (`~*^/Bar.*\\.jpg`) or a case sensitive regular expression (`~^/foo.*\\.jpg`). In the case of a prefix (must start with `/`) or an exact match (must start with `=`), the path must not include any whitespace characters, `{`, `}` or `;`. In the case of the regex matches, all double quotes `"` must be escaped and the match can't end in an unescaped backslash `\`. The path must be unique among the paths of all routes of the VirtualServer. Check the [location](http://nginx.org/en/docs/http/ngx_http_core_module.html#location) directive for more information. | `string` | Yes | `action` | The default action to perform for a request. | [`action`](#Action) | No* |
```
pleshakov: the action row is not shown anymore :(

docs/virtualserver-and-virtualserverroute.md
```diff
@@ -193,7 +198,7 @@ action:
...
+| `path` | The path of the subroute. NGINX will match it against the URI of a request. Possible values are: a prefix (`/`, `/path`), an exact match (`=/exact/match`), a case insensitive regular expression (`~*^/Bar.*\\.jpg`) or a case sensitive regular expression (`~^/foo.*\\.jpg`). In the case of a prefix, the path must start with the same path as the path of the route of the VirtualServer that references this resource. In the case of a prefix or an exact match, the path must not include any whitespace characters, `{`, `}` or `;`.  In the case of the regex matches, all double quotes `"` must be escaped and the match can't end in an unescaped backslash `\` and it must be exactly the same path as the path of the route of the VirtualServer that references this resource. The path must be unique among the paths of all subroutes of the VirtualServerRoute. | `string` | Yes |
```
```suggestion
| `path` | The path of the subroute. NGINX will match it against the URI of a request. Possible values are: a prefix (`/`, `/path`), an exact match (`=/exact/match`), a case insensitive regular expression (`~*^/Bar.*\\.jpg`) or a case sensitive regular expression (`~^/foo.*\\.jpg`). In the case of a prefix, the path must start with the same path as the path of the route of the VirtualServer that references this resource.  In the case of an exact or regex match, the path must be the sam as the path of the route of the VirtualServer that references this resource. In the case of a prefix or an exact match, the path must not include any whitespace characters, `{`, `}` or `;`.  In the case of the regex matches, all double quotes `"` must be escaped and the match can't end in an unescaped backslash `\`. The path must be unique among the paths of all subroutes of the VirtualServerRoute. | `string` | Yes |
```

docs/virtualserver-and-virtualserverroute.md
Rulox: Oops

PR #772, opened 2019-11-22
Update VS/VSR version from v1alpha1 to v1

Rulox:
### Proposed changes
Changed VirtualServer and VirtualServerRoute to stable `v1`


pkg/apis/configuration/validation/validation.go
```diff
@@ -7,21 +7,21 @@ import (
 	"strings"
 
 	"github.com/nginxinc/kubernetes-ingress/internal/configs"
+	v1 "github.com/nginxinc/kubernetes-ingress/pkg/apis/configuration/v1"
```
pleshakov: hm. is the alias `v1` required here?
same for `validation_test.go` file as well

Rulox: `goimports`

PR #773, opened 2019-11-22
Improve TLS redirect in VS

Dean-Coakley:
### Proposed changes
* Make `tls.secret` optional in the VS.

This allows a conditional redirect  based on the value of `x_forwarded_proto` to be configured which is important for the use case of deploying a proxy in front of the ingress controller to do TLS termination.

### Checklist

### DIFF
--- a/docs/virtualserver-and-virtualserverroute.md
+++ b/docs/virtualserver-and-virtualserverroute.md
@@ -84,13 +84,12 @@ The tls field defines TLS configuration for a VirtualServer. For example:
 ```yaml
 secret: cafe-secret
 redirect:
-  code: 302
-  basedOn: x-forwarded-proto
+  enable: true
 ```
 
 | Field | Description | Type | Required |
 | ----- | ----------- | ---- | -------- |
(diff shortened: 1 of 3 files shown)


docs/virtualserver-and-virtualserverroute.md
```diff
@@ -84,13 +84,12 @@ The tls field defines TLS configuration for a VirtualServer. For example:
 ```yaml
 secret: cafe-secret
 redirect:
-  code: 302
```
Rulox: I know this is based on specification but @pleshakov why do we need to remove code and basedOn from here? Seems to me adding `enable` and keep those is a more complete example?

pleshakov: @Rulox we have a complete example in `VirtualServer.TLS.Redirect` section. My thinking was that here we can show a minimal config for tls redirect.

Rulox: I see thanks for clarification

PR #778, opened 2019-11-27
Add redirect support in VS

Dean-Coakley:
### Proposed changes
* Add an action `redirect` which allow redirection of http traffic to https to be configured in the VirtualServer.

Example configuration:
```yaml
action:
  redirect:
    code: 301
    url: http://www.nginx.org 
```

### Checklist


pkg/apis/configuration/validation/validation.go
```diff
@@ -596,15 +596,140 @@ func validateRoute(route v1.Route, fieldPath *field.Path, upstreamNames sets.Str
...
+	if redirect.Code != 0 {
+		allErrs = append(allErrs, validateRedirectStatusCode(redirect.Code, fieldPath.Child("code"))...)
+	}
+
+	return allErrs
+}
+
+// Returns a slice of vars enclosed in ${}. For example ${scheme} would return scheme.
+func captureVariables(s string) [][]string {
+	regex := regexp.MustCompile(`\$\{([^}]+)\}`)
```
Rulox: I was thinking, maybe we can have this regular expression as a global var outside of the method for consistency? I'm ok with having it here too tho.

pkg/apis/configuration/validation/validation.go
```diff
@@ -596,15 +596,140 @@ func validateRoute(route v1.Route, fieldPath *field.Path, upstreamNames sets.Str
...
+		return append(allErrs, field.Required(fieldPath, "must specify a url"))
+	}
+
+	allErrs = append(allErrs, validateRedirectURLFmt(redirectURL, fieldPath)...)
+
+	if strings.Contains(redirectURL, "$") {
+		nginxVars := captureVariables(redirectURL)
+		if nginxVars != nil {
+			for _, v := range nginxVars {
+				nVar := v[1]
```
Rulox: I would use `v[1]` in the method instead of creating this variable here, I don't think it adds a lot of information:
```
allErrs = append(allErrs, validateVariable(v[1], validRedirectVariableNames, fieldPath)...)
```
Even less characters 😛

pkg/apis/configuration/validation/validation.go
```diff
@@ -596,15 +596,140 @@ func validateRoute(route v1.Route, fieldPath *field.Path, upstreamNames sets.Str
...
+	if strings.Contains(str, "${}") {
+		return append(allErrs, field.Invalid(fieldPath, str, "must not contain any empty variable ${}"))
+	}
+
+	if strings.HasSuffix(str, "$") {
+		return append(allErrs, field.Invalid(fieldPath, str, "must not end with $"))
+	}
+
+	for i, c := range str {
+		if c == '$' && i+1 < len(str) {
```
Rulox: I was thinking, `&& i+1 < len(str)` is not really needed because in line 683 we return if the last character is `$` and that's the only possible panic, when the `$` is at the end of the string. So unless I am missing something, up to you to remove it.

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -814,26 +814,197 @@ func TestValidateRouteFails(t *testing.T) {
...
+			msg:         "url with escaped backslash",
+		},
+		{
+			redirectURL: `${scheme}://${host}$`,
+			msg:         "url with ending $",
+		},
+		{
+			redirectURL: `http://${}`,
+			msg:         "url containing blank var",
+		},
```
pleshakov: I suggest adding the following case:
```
		{
			redirectURL: `http://${abca`,
			msg:         "url containing a var without ending }",
		},
```

pkg/apis/configuration/validation/validation.go
```diff
@@ -627,10 +627,18 @@ func validateActionRedirect(redirect *v1.ActionRedirect, fieldPath *field.Path)
 	return allErrs
 }
 
-// Returns a slice of vars enclosed in ${}. For example ${scheme} would return scheme.
-func captureVariables(s string) [][]string {
-	regex := regexp.MustCompile(`\$\{([^}]+)\}`)
-	return regex.FindAllStringSubmatch(s, -1)
+var regexFmtRegexp = regexp.MustCompile(`\$\{([^}]*)\}`)
```
pleshakov: `regexFmtRegexp` doesn't describe well the variable. Could you rename it to something like `nginxVariableRegexp`?

PR #780, opened 2019-12-05
Add support for canned responses to VS/VSR

Rulox:
### Proposed changes
This PR adds the "canned responses" features.

Now, it is possible to return preconfigured responses for any VS or VSR with the action `return`. 

Also refactored the `redirect` action to reuse the same code as the new `return` action, as both make usage of the "return" directive from NGINX.


pkg/apis/configuration/validation/validation.go
```diff
@@ -707,6 +738,130 @@ func validateVariable(nVar string, validVars map[string]bool, fieldPath *field.P
...
+		return []string{"a valid HTTP header must consist of alphanumeric characters or '_'"}
+	}
+	return nil
+}
+
+func validateSpecialVariable(nVar string, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	value := strings.SplitN(nVar, "_", 2)
+
+	if strings.HasPrefix(nVar, "arg") {
```
pleshakov: Let me know what you think of the following suggestion:

validate every special variable in its own function.
* isValidSpecialVariableHeader
* isValidSpecialVariableArgument
* isValidSpecialVariableCookie

Each of those function can have its own regexes - I think it would be better instead of modying the input and calling other functions, even if we have some duplication.

Also, each of those function will return a standardize special variable error message. For example: '$http_asdf-sdf must contain ... '

I think that will simplify the code and make it more readable.

Also, if a validation variable returns a slice of string, for consistency, I suggest using the `for _, msg := range` approach instead of `if len(validationMsg) >= 1 `

pkg/apis/configuration/validation/validation.go
```diff
@@ -707,6 +738,130 @@ func validateVariable(nVar string, validVars map[string]bool, fieldPath *field.P
...
+		_, err := url.ParseQuery(fmt.Sprintf("%v=%v", value[0], value[1]))
```
pleshakov: why not an approach similar to the one in `isArgumentName`?

pkg/apis/configuration/validation/validation.go
pleshakov: we can check for equality here, no need to use `HasPrefix`.  As a result, we can also use a switch statement instead of `if..else...`

internal/configs/virtualserver.go
```diff
@@ -511,9 +511,32 @@ func generatePath(path string) string {
 	return path
 }
 
+func generateReturnBlock(text string, code int, defaultCode int) *version2.Return {
```
pleshakov: I strongly suggest not using `block` with `return` in the code. The code  in this package deals with NGINX configuration. Better to be consistent with the terminology when possible.

Rulox: This doesn't make any sense. Go code != nginx config. In templates it makes sense since is kind of a mix between the two worlds, here it doesn't. But thanks for your suggestion.

pkg/apis/configuration/validation/validation.go
Rulox: This sounds reasonable, I was trying to reduce number of functions as this file is growing and growing, but I guess is fine.

Rulox: PS: about regexes, I'll give it a thought, I am not a fan of regexes all the time, it's easier to make a mistake with a regular expression and do not cover all the cases than with string manipulation in my opinion.

pkg/apis/configuration/validation/validation.go
Rulox: Because I didn't see `isArgumentName`

Rulox:
@pleshakov 

Changed the validateSpecialVar to a switch case, with the next considerations:

`isValidSpecialVariableHeader` is needed but there's a difference between using `-` and `_`. Creating a new regexp for this is against DRY and KISS, reusing the header with replacing the underscores seems a much better and elegant solution to me.

I didn't create `isValidSpecialVariableCookie` or `isValidSpecialArgument` because those would have been just wrappers to `isCookieName` and `isArgumentName` so I think that would be just a waste of time and add code for nothing. 

Let me know what you think thanks

pkg/apis/configuration/validation/validation.go
```diff
@@ -707,6 +737,124 @@ func validateVariable(nVar string, validVars map[string]bool, fieldPath *field.P
...
+
+	if (code >= 200 && code <= 299) || (code >= 400 && code <= 599) {
+		return allErrs
+	}
+
+	msg := fmt.Sprintf("must be a valid status code either 2XX, 4XX or 5XX, for example, 200 or 402.")
+	return append(allErrs, field.Invalid(fieldPath, code, msg))
+}
+
+func validateActionReturn(r *v1.ActionReturn, fieldPath *field.Path) field.ErrorList {
```
pleshakov: this function isn't tested at all. could you add a unit test or extend a test like `TestValidateAction` ?

pkg/apis/configuration/validation/validation.go
```diff
@@ -707,6 +737,124 @@ func validateVariable(nVar string, validVars map[string]bool, fieldPath *field.P
...
+
+	return allErrs
+}
+
+var actionReturnTypeFmt = `([^;\{\}"\\]|\\.)*`
+var actionReturnTypeErr = `must have all '"' (double quotes) escaped, must not contain '{', '}' or ';' and must not end with an unescaped '\' (backslash)`
+
+var actionReturnTypeRegexp = regexp.MustCompile("^" + actionReturnTypeFmt + "$")
+
+func validateActionReturnType(returnType string, fieldPath *field.Path) field.ErrorList {
```
pleshakov: this function isn't tested at all. could you add a unit test?

pkg/apis/configuration/validation/validation.go
```diff
@@ -707,6 +737,124 @@ func validateVariable(nVar string, validVars map[string]bool, fieldPath *field.P
 		msg := fmt.Sprintf("'%v' contains an invalid NGINX variable. Accepted variables are: %v", nVar, mapToPrettyString(validVars))
 		allErrs = append(allErrs, field.Invalid(fieldPath, nVar, msg))
 	}
+	return allErrs
+}
+
+func isValidSpecialVariableHeader(header string) []string {
+	errMsgs := validation.IsHTTPHeaderName(strings.Replace(header, "_", "-", -1))
```
pleshakov: modifying the input so that we can use an existing function for validation adds complexity - it makes the code harder to understand. 
I suggest adding a comment about why we need replace `_` with `-`. For example,  
```
# underscores in $http_ variable represent '-'.  
```

internal/configs/virtualserver.go
pleshakov: > Go code != nginx config.
this code is in the package config that deals with NGINX config.

generateReturnBlock - is generating something that doesn't exist in NGINX. I suggest sticking to NGINX terminology, `generateReturnBlock` -> `generateReturn`

Note that using ReturnBlock in `generateLocationForReturnBlock` would be OK, because here ReturnBlock is something we refer in YAML file, not in the NGINX config.

pkg/apis/configuration/validation/validation.go
Rulox: I don't agree, but thanks for the suggestion

pkg/apis/configuration/validation/validation.go
```diff
@@ -707,6 +737,124 @@ func validateVariable(nVar string, validVars map[string]bool, fieldPath *field.P
...
+	"server_port":         true,
+	"server_name":         true,
+	"server_protocol":     true,
+	"connections_active":  true,
+	"connections_reading": true,
+	"connections_writing": true,
+	"connections_waiting": true,
+}
+
+var returnBodySpecialVariables = []string{"arg_, http_, cookie_"}
```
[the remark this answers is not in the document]
Rulox: Yes! Good catch, not sure how this got here. Thank you!

Rulox:
`creating a new function is against DRY but not KISS. I left a comment for the corresponding line explaining why.`

Hmm not really, it's a pretty opinionated reason. To me, the solution it's DRY and KISS, keeping it like that, thanks.

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -113,8 +113,11 @@ server {
         {{ $snippet }}
         {{ end }}
 
-        {{ with $l.Redirect }}
-        return {{ .Code }} "{{ .URL }}";
+        {{ with $l.Return }}
+            {{ if $l.DefaultType }}
+        default_type {{ $l.DefaultType }};
```
pleshakov: {{ $l.DefaultType }} should be enclosed in `"..."` . otherwise, any space characters in the field in the YAML will cause problems

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -2651,3 +2651,278 @@ func TestIsRegexOrExactMatch(t *testing.T) {
...
+			Code: 301,
+			Body: "Hello World",
+		},
+		{
+			Code: 200,
+			Type: `application/"json"`,
+			Body: "Hello World",
+		},
+		{
+			Code: 200,
```
pleshakov: this test case is the same as the one above

internal/configs/virtualserver.go
pleshakov: Inconsistent terminology adds complexity. This will make code less maintainable. The code of our project is growing and it should be a priority to keep the codebase maintainable.

PR #800, opened 2019-12-26
Add tests for health-status and nginx-debug cli args

tellet:
Cover scenarios for -nginx-debug and -health-status (and -health-status-uri).

### DIFF
--- a/tests/suite/test_smoke.py
+++ b/tests/suite/test_smoke.py
@@ -51,9 +51,15 @@ def smoke_setup(request, kube_apis, ingress_controller_endpoint, ingress_control
 
 @pytest.mark.ingresses
 @pytest.mark.smoke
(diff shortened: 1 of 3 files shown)


tests/suite/test_smoke.py
```diff
@@ -51,9 +51,15 @@ def fin():
 
 @pytest.mark.ingresses
 @pytest.mark.smoke
+@pytest.mark.parametrize('ingress_controller',
+                         [
+                             pytest.param({"extra_args": None}, id="no-additional-cli-args"),
+                             pytest.param({"extra_args": ["-nginx-debug", "-health-status=true"]},
```
pleshakov: for `-nginx-debug`, is it enough just to test that the IC starts successfully?

tellet: Yes, for the smoke test.

PR #801, opened 2019-12-30
Fix tests for DaemonSet test-run

tellet:
There is a difference in the amount of events generated for VS when there are 3 pods (DaemonSet) and 1 pod (Deployment). Take that into account in the tests.

### DIFF
--- a/tests/suite/resources_utils.py
+++ b/tests/suite/resources_utils.py
@@ -227,6 +227,18 @@ def are_all_pods_in_ready_state(v1: CoreV1Api, namespace) -> bool:
     return pod_ready_amount == len(pods.items)
 
 
+def get_pods_amount(v1: CoreV1Api, namespace) -> int:
+    """
+    Get an amount of pods.
+
+    :param v1: CoreV1Api
+    :param namespace: namespace
+    :return: int
+    """
+    pods = v1.list_namespaced_pod(namespace)
+    return 0 if not pods.items else len(pods.items)
+
+
 def create_service_from_yaml(v1: CoreV1Api, namespace, yaml_manifest) -> str:
     """
     Create a service based on yaml file.
--- a/tests/suite/test_virtual_server_configmap_keys.py
+++ b/tests/suite/test_virtual_server_configmap_keys.py
@@ -2,25 +2,25 @@ import pytest
 
 from settings import TEST_DATA, DEPLOYMENTS
 from suite.resources_utils import wait_before_test, replace_configmap_from_yaml, get_events, get_first_pod_name, \
-    get_file_contents
+    get_file_contents, get_pods_amount
 from suite.custom_resources_utils import get_vs_nginx_template_conf
 from suite.yaml_utils import get_configmap_fields_from_yaml
 
 
-def assert_update_event_emitted(virtual_server_setup, new_list, previous_list):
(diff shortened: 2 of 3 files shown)


tests/suite/test_virtual_server_configmap_keys.py
```diff
@@ -7,20 +7,22 @@
 from suite.yaml_utils import get_configmap_fields_from_yaml
 
 
-def assert_update_event_emitted(virtual_server_setup, new_list, previous_list):
+def assert_update_event_emitted(virtual_server_setup, new_list, previous_list, is_deployment):
     item_name = f"{virtual_server_setup.namespace}/{virtual_server_setup.vs_name}"
     text_valid = f"Configuration for {item_name} was updated"
     text_invalid = "was updated but was not applied"
+    expected_amount = 1 if is_deployment else 3
```
pleshakov: it seems that number 3 for DS depends on the cluster - 3 means 3 nodes right? what if the number of nodes is changed? 
Should we configure that number differently instead of  hardcoding it?
Alternatively, can we get number of IC pods from the k8s API and that use that number ?

tellet: This is a testing framework, in my opinion at each point of time test should know exactly what is going on in the system.

pleshakov: @tellet 
the PR makes it is a requirement from the cluster to have 3 nodes if the tests are run with `daemon-set` deployment type. Is it possible to make it either conifgurable (so that the infrastructure can be changed without changing the source code) or at least document that in the README https://github.com/nginxinc/kubernetes-ingress/blob/master/tests/README.md ?

tellet: I'll add a method to get # of IC pods then.

tests/suite/test_virtual_server_configmap_keys.py
```diff
@@ -268,7 +271,7 @@ def test_keys_in_main_config(self, cli_arguments, kube_apis, ingress_controller_
         step_7_config = get_file_contents(kube_apis.v1,
                                           config_path, ic_pod_name, ingress_controller_prerequisites.namespace)
         step_7_events = get_events(kube_apis.v1, virtual_server_setup.namespace)
-        assert_not_applied_event_emitted(virtual_server_setup, step_7_events, step_6_events)
+        assert_not_applied_event_emitted(virtual_server_setup, step_7_events, step_6_events, is_deployment)
```
pleshakov: seems like a bug here?

tests/suite/resources_utils.py
```diff
@@ -227,6 +227,20 @@ def are_all_pods_in_ready_state(v1: CoreV1Api, namespace) -> bool:
...
+    Get an amount of pods.
+
+    :param v1: CoreV1Api
+    :param namespace: namespace
+    :return: int
+    """
+    pods = v1.list_namespaced_pod(namespace)
+    if not pods.items:
+        return 0
+    return len(pods.items)
```
Rulox: This is not a real suggestion for you to implement, just a curiosity. Tis could be rewritten as:
```python
return 0 if not pod.items else len(pods.items)
```
In a more `pythonic` way

PR #807, opened 2020-01-07
Add pod PriorityClass support

gsnegovskiy:
### Proposed changes
PriorityClass for pods is a stable feature since Kubernetes 1.14
It will be useful to have it available to configure priorities for ingress controller.


gsnegovskiy:
@Dean-Coakley @pleshakov Thank's for the review!

Hopefully I've added documentation in a right place inside values.yaml

Anything else I can do to fit standards ?

PR #814, opened 2020-01-13
Add tests for custom annotations

tellet:
Cover the flow from the documentation: custom annotations  and helper functions. Nginx OSS|Plus, standard and mergeable Ingresses.

### DIFF
--- a/tests/data/custom-annotations/mergeable/annotations-ingress.yaml
+++ b/tests/data/custom-annotations/mergeable/annotations-ingress.yaml
@@ -1,0 +1,49 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: custom-annotations-ingress-master
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
+    nginx.org/mergeable-ingress-type: "master"
+    custom.nginx.org/feature-a: "on"
+    custom.nginx.org/feature-b: "512"
+    custom.nginx.org/test-split-feature: "  master.ip,master"
+spec:
+  rules:
+  - host: custom-annotations.example.com
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: custom-annotations-ingress-backend1-minion
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
+    nginx.org/mergeable-ingress-type: "minion"
+    custom.nginx.org/test-split-feature: "  minion1"
+spec:
+  rules:
+  - host: custom-annotations.example.com
+    http:
+      paths:
+      - path: /backend1
+        backend:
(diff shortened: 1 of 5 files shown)


tests/data/custom-annotations/mergeable/annotations-ingress.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: custom-annotations-ingress-master
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
+    nginx.org/mergeable-ingress-type: "master"
+    custom.nginx.org/feature-a: "on"
+    custom.nginx.org/feature-b: "512"
+    custom.nginx.org/test-split-feature: "  master.ip--master"
```
pleshakov: when we test the split, could we use more common delimiters like `","` instead of `"--"`? otherwise, it will look a little bit more confusing

tests/suite/test_custom_annotations.py
```diff
@@ -0,0 +1,88 @@
...
+
+    return CustomAnnotationsSetup(ingress_controller_endpoint, ingress_name, test_namespace, host, ic_pod_name)
+
+
+@pytest.mark.ingresses
+@pytest.mark.parametrize('custom_annotations_setup, expected_texts',
+                         [
+                             pytest.param("standard",
+                                          ["# This is TEST configuration for custom-annotations-ingress/",
+                                           "# Insert config for test-split-feature: 192.168.1.1;",
```
pleshakov: here`# Insert config for test-split-feature: some-ip;` will also be generated, should we test for that?

tests/suite/test_custom_annotations.py
```diff
@@ -0,0 +1,88 @@
...
+    replace_configmap_from_yaml(kube_apis.v1,
+                                ingress_controller_prerequisites.config_map['metadata']['name'],
+                                ingress_controller_prerequisites.namespace,
+                                f"{TEST_DATA}/custom-annotations/{ing_type}/nginx-config.yaml")
+    print("------------------------- Deploy Custom Annotations Example -----------------------------------")
+    ing_src = f"{TEST_DATA}/custom-annotations/{ing_type}/annotations-ingress.yaml"
+    create_items_from_yaml(kube_apis, ing_src, test_namespace)
+    host = get_first_ingress_host_from_yaml(ing_src)
+    ingress_name = get_names_from_yaml(ing_src)[0]
+    create_example_app(kube_apis, "simple", test_namespace)
```
pleshakov: do we need to deploy the backends for this? it seems that we only test the generated config

tellet: It's not necessary indeed.

PR #825, opened 2020-01-16
Add multistage docker builds

Dean-Coakley:
### Proposed changes
* Converted spaces to tabs in all Dockerfiles for consistency
* Building with `BUILD_IN_CONTAINER = 1` now uses a multi stage docker build instead of using volumes.

### Checklist


Makefile
```diff
@@ -17,10 +16,10 @@ DOCKER_BUILD_OPTIONS =
 
 GIT_COMMIT=$(shell git rev-parse --short HEAD)
 
+export DOCKER_BUILDKIT = 1
+
 nginx-ingress:
-ifeq ($(BUILD_IN_CONTAINER),1)
```
pleshakov: similar thing should be done to the test stage - run it inside the builder container, not here

.travis.yml
```diff
@@ -21,3 +21,8 @@ before_install:
   && GO111MODULE=on FOSSA_API_KEY=${fossapush} ./fossa analyze -t kubernetes-ingress -b ${TRAVIS_BRANCH}; fi
 - echo "Checking ingress controller for linting errors" && wget -O - -q ${golangci_lint} | sh -s v1.17.1
 - export PATH="$PATH:./bin" && make lint
+- echo "Updating docker to latest stable release"
```
pleshakov: travis fails. needs to be fixes

Rulox:
Looks nice @Dean-Coakley ! I have mixed feelings about your proposal @pleshakov.

On the one hand the removal of the buildkit is something that I like. 
On the other hand, it seems we make things more complicated just to avoid a little bit of repetition (for instance, there are other parts of the Dockerfiles that we repeat normally, what happens with those? we keep repeating them?)

So I am not sure, let me have a deeper review and will be back to you guys (also @dboenig probably has something to say here)

lucacome:
Looks good @Dean-Coakley . I'd argue we could have even fewer Dockerfiles because of the multistage approach :)

@pleshakov not sure how I feel about your proposed solution of creating the Dockerfile on the fly instead of having a static file, I think I prefer the latter because it's more straightforward.

@pleshakov @Rulox I personally really like buildkit, because it speeds up the building process (it can build concurrently, better caching, etc) and makes the multistage approach even better. Also, it's not really new so it doesn't require the latest version of Docker but just 18.09 or higher (18.09 was released in November 2018).

Makefile
```diff
@@ -17,10 +16,10 @@ DOCKER_BUILD_OPTIONS =
 
 GIT_COMMIT=$(shell git rev-parse --short HEAD)
 
+export DOCKER_BUILDKIT = 1
+
 nginx-ingress:
```
lucacome: maybe we should rename `nginx-ingress` target to something that reflects that you're building the binary locally and remove `BUILD_IN_CONTAINER`

Makefile
```diff
@@ -49,7 +48,11 @@ endif
 
 container: test verify-codegen nginx-ingress certificate-and-key
 	cp $(DOCKERFILEPATH)/$(DOCKERFILE) ./Dockerfile
-	docker build $(DOCKER_BUILD_OPTIONS) --build-arg IC_VERSION=$(VERSION)-$(GIT_COMMIT) -f Dockerfile -t $(PREFIX):$(TAG) .
+ifeq ($(BUILD_IN_CONTAINER),1)
```
lucacome: how about you make a different make target for `local` and call the right docker target? this `container` target is also calling `nginx-ingress` when you're not building inside the container.

build/DockerfileForAlpine
```diff
@@ -1,4 +1,4 @@
-FROM nginx:1.17.7-alpine
+FROM nginx:1.17.7-alpine AS base
```
lucacome: If we pass `-alpine` as a build arg we could have just one file. And also support the other official images `-perl` and `alpine-perl`.

build/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM nginx:1.17.7
+FROM nginx:1.17.7 AS base
 
 # forward nginx access and error logs to stdout and stderr of the ingress
 # controller process
```
lucacome: I've never seen `/proc/1/fd/1` being used instead of `/dev/stdout` in a Dockerfile and looks weird to me 😄. Any chance you could change it since you're modifying the files anyway? Also, the nginx base image already has the forward for error and access log, so we really need just the one for `stream-access.log`.

Makefile
Dean-Coakley: Agreed. 

`make binary` makes sense to me

build/DockerfileForAlpine
Dean-Coakley: I assume you mean something like:

```dockerfile
ARG BASE_TAG=1.17.7

FROM nginx:$BASE_TAG AS base
```

I don't think this can work because alpine uses the apk package manager, whereas debian uses the apt package manager so the rest of the stage can't succeed.

Rulox:
I am ok with buildkit, I've never used it, so that's why I can't really have a strong opinion. If I have to choose between that and the multiple files and the `cat >>` in files I prefer buildkit, that's for sure. @lucacome @Dean-Coakley

Makefile
```diff
@@ -60,4 +53,3 @@ endif
 
 clean:
 	rm -f nginx-ingress
-	rm -f Dockerfile
```
pleshakov: why was it removed?

Dean-Coakley: Instead of copying Dockerfiles to the root of the project I point `docker build` directly to the Dockerfile required:
https://github.com/nginxinc/kubernetes-ingress/pull/825/commits/baea83e6e0e100e9c18870eea173dd76ef4a8c49#diff-b67911656ef5d18c4ae36cb6741b7965R43

So there is no trailing Dockerfile to remove/cleanup

build/DockerfileForAlpine
lucacome: Oh yeah you're right, I missed that we're installing a package. Not sure if there's a way to combine the two files to avoid the repetitions then.

Makefile
```diff
@@ -60,4 +58,3 @@ endif
...
```
pleshakov: Please update .gitignore https://github.com/nginxinc/kubernetes-ingress/blob/master/.gitignore#L42

Makefile
```diff
@@ -4,9 +4,6 @@ VERSION = edge
 TAG = $(VERSION)
 PREFIX = nginx/nginx-ingress
 
-DOCKER_TEST_RUN = docker run --rm -v $(shell pwd):/go/src/github.com/nginxinc/kubernetes-ingress -w /go/src/github.com/nginxinc/kubernetes-ingress
-DOCKER_BUILD_RUN = docker run --rm -v $(shell pwd):/go/src/github.com/nginxinc/kubernetes-ingress -w /go/src/github.com/nginxinc/kubernetes-ingress/cmd/nginx-ingress/
-GOLANG_CONTAINER = golang:1.13
```
pleshakov: I wonder if it makes sense to use this variable in the Dockerfiles? so that when we need to update it, we update it only here ?

docs-web/installation/building-ingress-controller-image.md
```diff
@@ -54,7 +54,7 @@ Next you will find the details about available Makefile targets and variables.
 
 The **Makefile** we provide has the following targets:
 * **test**: runs unit tests.
```
pleshakov:
```suggestion
* **test**: runs unit tests (ignored when `BUILD_IN_CONTAINER` is `1`).
```

docs-web/installation/building-ingress-controller-image.md
```diff
@@ -54,7 +54,7 @@ Next you will find the details about available Makefile targets and variables.
...
-* **nginx-ingress**: creates the controller binary.
+* **binary**: creates the controller binary.
```
pleshakov:
```suggestion
* **binary**: creates the controller binary using local golang environment (ignored when `BUILD_IN_CONTAINER` is `1`). 
```

build/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM nginx:1.17.7
+ARG GOLANG_CONTAINER=golang:latest
```
pleshakov: do we need to initialize this argument?

build/DockerfileWithOpentracingForPlus
```diff
@@ -2,87 +2,106 @@ FROM debian:stretch-slim AS tracer-downloader
...
-ENTRYPOINT ["/nginx-ingress"]
\ No newline at end of file
+ENTRYPOINT ["/nginx-ingress"]
+
+
+FROM base AS local
+COPY nginx-ingress /
+
+
+FROM golang:1.13 AS builder
```
pleshakov: here we're not using the argument :(

.travis.yml
```diff
@@ -6,7 +6,7 @@ go:
 - "1.13"
 script:
 - echo "Building ingress controller commit:${TRAVIS_COMMIT}"
-- make BUILD_IN_CONTAINER=0 container;
+- make container BUILD_IN_CONTAINER=0 DOCKER_BUILDKIT=0;
```
lucacome: why `DOCKER_BUILDKIT=0`?

Dean-Coakley: TravisCI comes with docker 18.06. So by default it fails due to docker version.

If you try to update docker you would inexplicably get: 
```bash
#3 [internal] load metadata for docker.io/library/nginx:1.17.7
#3 ERROR: docker.io/library/nginx:1.17.7 not found
#7 [internal] load build context
#7 transferring context: 7.59MB 0.1s done
#7 CANCELED
#4 [base 1/4] FROM docker.io/library/nginx:1.17.7
#4 resolve docker.io/library/nginx:1.17.7 0.1s done
#4 ERROR: docker.io/library/nginx:1.17.7 not found
```
(https://travis-ci.org/nginxinc/kubernetes-ingress/builds/639443840?utm_source=github_status&utm_medium=notification)

So BUILDKIT is disabled just for travis in order for the build to succeed.

PR #826, opened 2020-01-20
Add a test for the Prometheus Exporter in IC

tellet:
Include exporter port (9113) into common LB/NodePort. Automate a "smoke" test for Prometheus Exporter in IC.

### DIFF
--- a/tests/suite/test_prometheus_exporter.py
+++ b/tests/suite/test_prometheus_exporter.py
@@ -1,0 +1,57 @@
+import pytest
+import requests
+
+from kubernetes.client import V1ContainerPort
+
+from suite.resources_utils import wait_until_all_pods_are_ready, ensure_connection
+
+
+@pytest.fixture(scope="class")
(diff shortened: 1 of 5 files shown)


tests/suite/test_prometheus_exporter.py
```diff
@@ -0,0 +1,62 @@
...
+    Set annotations and containerPort for Prometheus Exporter.
+
+    :param cli_arguments: context
+    :param kube_apis: client apis
+    :param ingress_controller_prerequisites
+    :param ingress_controller: IC name
+    :return:
+    """
+    namespace = ingress_controller_prerequisites.namespace
+    annotations = {"prometheus.io/scrape": "true", "prometheus.io/port": "9113"}
```
pleshakov: those annotations are for Prometheus metrics collector running in a cluster. Our IC doesn't care about it. Considering that, do we need to add them?

tellet: No we don't, I'll remove this.

PR #827, opened 2020-01-20
Add ingress class label to all prometheus metrics

Dean-Coakley:
### Proposed changes
* Adds a label of the ingress class to all metrics to inform the user of the ingress that is being used.

Example:
  * `nginx_ingress_controller_ingress_resources_total{type="regular"} 1`
Will become:
  * `nginx_ingress_controller_ingress_resources_total{class="nginx",type="regular"} 1`

### Checklist


docs-web/logging-and-monitoring/prometheus.md
```diff
@@ -35,3 +35,5 @@ The Ingress Controller exports the following metrics:
   * `controller_virtualserverroute_resources_total`. Number of handled VirtualServerRoute resources. **Note**: The metric counts only VirtualServerRoutes that have a reference from a VirtualServer.
 
 **Note**: all metrics have the namespace nginx_ingress. For example, nginx_ingress_controller_nginx_reloads_total.
+
+**Note**: all metrics include the label `class` which is set to the `ingress-class` of the Ingress Controller deployment.
```
pleshakov:
```suggestion
**Note**: all metrics include the label `class`, which is set to the class of the Ingress Controller. The class is configured via the `-ingress-class` command-line argument.
```

cmd/nginx-ingress/main.go
```diff
@@ -242,13 +242,14 @@ func main() {
 	var registry *prometheus.Registry
 	var managerCollector collectors.ManagerCollector
 	var controllerCollector collectors.ControllerCollector
+	var ingressClassLabel = map[string]string{"class": *ingressClass}
```
pleshakov: style suggestion: for variable initialization, we use the following style:
```
ingressClassLabel := map[string]string{"class": *ingressClass}
```
Please change for consistency 

Additionally, since you're using a map, better to rename the variable to `constLabels` or something like that. If you want to use `ingressClassLabel`, you can safely change the type from map[string]string to string. String is enough to represent the value of one label.

internal/metrics/collectors/controller.go
```diff
@@ -21,12 +21,13 @@ type ControllerMetricsCollector struct {
 }
 
 // NewControllerMetricsCollector creates a new ControllerMetricsCollector
-func NewControllerMetricsCollector(crdsEnabled bool) *ControllerMetricsCollector {
+func NewControllerMetricsCollector(crdsEnabled bool, ingressClassLabel map[string]string) *ControllerMetricsCollector {
```
pleshakov: similar to the comment in `main.go`, it is not necessary to use `map` to represent just one label value. Either change the type to `string` or change to name to `constLabels`. With the former, we would be able to easily add more const labels in the future. 

Same applies to `manager.go ` and `listener.go`.

PR #847, opened 2020-02-10
Add support for error pages for VS/VSR

Rulox:
### Proposed changes
This PR introduces the [error_page](http://nginx.org/en/docs/http/ngx_http_core_module.html#error_page) directive to use with VS/VSR for circuit breaking.

### Changes
This PR introduces changes in how the NGINX configuration is generated using the templates. Prior to this PR, the internal redirections were done by using the `error_page` directive. However, this would prevent the usage of error_pages by the user, this has been changed to use the `rewrite` directive:

```
    location {{ $l.Path }} {
        rewrite ^ {{ $l.Destination }} last;
    }
```

instead of 
```
    location {{ $l.Path }} {
        error_page 418 = {{ $l.Destination }};
        return 418;
    }
```

This change affects to the locations used for internal redirect. Before, they were named locations. But because `rewrite` directive won't work with named locations, regular locations need to be used. However, to prevent these locations to be accessible, the [internal](http://nginx.org/en/docs/http/ngx_http_core_module.html#internal) directive has been added.

Additionally, in order to avoid conflicts with locations defined by the user, a prefix `internal_location_` will be used for these locations. This will make the locations look like this:
```
location /internal_location_my_location {
   internal;
   // location config
}
```

Instead of:
```
location @my_location {
   // location config
}
```

### Checklist


docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -186,6 +189,10 @@ The route defines rules for matching client requests to actions like passing a r
      - The name of a VirtualServerRoute resource that defines this route. If the VirtualServerRoute belongs to a different namespace than the VirtualServer, you need to include the namespace. For example, ``tea-namespace/tea``.
      - ``string``
      - No*
+   * - ``errorPages``
```
pleshakov: we need to add the same for the subroute

Rulox: Oops, good catch

PR #850, opened 2020-02-17
Extend redirect URI validation with protocol check

Rulox:
### Proposed changes
* Update docs with proper variables
* Make sure that validation of Redirect URLs (both in action and error pages) check for the protocol of the URL to be a fully working URL (this is to avoid NGINX to try to open local files). Now a redirect URL must contain `://`.

### DIFF
--- a/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
+++ b/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
@@ -992,7 +992,7 @@ In the example below, NGINX responds with a redirect when a response from an ups
 codes: [404]
 redirect:
   code: 301
-  url: ${scheme}://cafe.example.com/error_${status}.html
+  url: ${scheme}://cafe.example.com/error.html
 ```
 
 ```eval_rst
@@ -1008,7 +1008,7 @@ redirect:
      - ``int``
      - No
    * - ``url``
-     - The URL to redirect the request to. Supported NGINX variables: ``$scheme``\ , ``$status``\ and ``$http_x_forwarded_proto``\. Variables must be inclosed in curly braces. For example: ``${scheme}``.
+     - The URL to redirect the request to. Supported NGINX variables: ``$scheme``\ and ``$http_x_forwarded_proto``\. Variables must be inclosed in curly braces. For example: ``${scheme}``.
      - ``string``
      - Yes
 ```
@@ -1026,8 +1026,8 @@ return:
   type: application/json
   body: "{\"msg\": \"You don't have permission to do this\"}"
   headers:
-  - name: x-debug-original-status
-    value: ${status}
+  - name: x-debug-original-statuses
+    value: ${upstream_status}
 ```
 
 ```eval_rst
@@ -1047,7 +1047,7 @@ return:
      - ``string``
      - No
    * - ``body``
-     - The body of the response. Supported NGINX variable: ``$status`` \ . Variables must be inclosed in curly braces. For example: ``${status}``.
+     - The body of the response. Supported NGINX variable: ``$upstream_status`` \ . Variables must be inclosed in curly braces. For example: ``${upstream_status}``.
      - ``string``
      - Yes
    * - ``headers``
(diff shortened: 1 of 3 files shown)


pkg/apis/configuration/validation/validation.go
```diff
@@ -801,6 +801,10 @@ func validateRedirectURL(redirectURL string, fieldPath *field.Path, validVars ma
 		return append(allErrs, field.Required(fieldPath, "must specify a url"))
 	}
 
+	if !strings.Contains(redirectURL, "://") {
+		return append(allErrs, field.Required(fieldPath, "must contain the protocol with '://', for example http://, https:// or ${scheme}://"))
```
Dean-Coakley: For `field.Required`: 
```go
// Required returns a *Error indicating "value required".  This is used
// to report required values that are not provided (e.g. empty strings, null
// values, or empty arrays).
```
My interpretation of the above would be that `field.Required` should only be used if `url` was unset, empty string, null, etc. 
Considering `url` is set, but an incorrect format I would think that `field.Invalid` is more appropriate.

_Hm I noticed a similar use of `field.Required`. Perhaps I'm mistaken:_
https://github.com/nginxinc/kubernetes-ingress/blob/c2bccae7a052538d2c7fc96a6f891ad45e2d9101/pkg/apis/configuration/validation/validation.go#L1329

What do you think @Rulox ? Happy to be corrected.

Rulox: Yes it should be Invalid

pkg/apis/configuration/validation/validation_test.go
```diff
@@ -946,7 +946,7 @@ func TestValidateRedirectURL(t *testing.T) {
 			msg:         "x-forwarded-proto redirect url use case",
 		},
 		{
-			redirectURL: "${host}${request_uri}",
+			redirectURL: "http://${host}${request_uri}",
```
lucacome: should this be `${scheme}`?

pkg/apis/configuration/validation/validation.go
```diff
@@ -801,6 +801,10 @@ func validateRedirectURL(redirectURL string, fieldPath *field.Path, validVars ma
 		return append(allErrs, field.Required(fieldPath, "must specify a url"))
 	}
 
+	if !strings.Contains(redirectURL, "://") {
```
pleshakov: could you also update the maps of supported variables?

For the error page returns, both headers and body, we need to support `$upstream_status` instead of `$status`

For the error page redirect, we do not need to support the `$status` anymore.

docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -992,7 +992,7 @@ In the example below, NGINX responds with a redirect when a response from an ups
 codes: [404]
 redirect:
   code: 301
-  url: ${scheme}://cafe.example.com/error_${status}.html
+  url: ${scheme}://cafe.example.com/error.html
```
pleshakov: Please update the list of supported variables in the table

docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -1026,8 +1026,8 @@ return:
   type: application/json
   body: "{\"msg\": \"You don't have permission to do this\"}"
   headers:
-  - name: x-debug-original-status
-    value: ${status}
+  - name: x-debug-original-statuses
+    value: ${upstream_status}
```
pleshakov: Please update the example in `ErrorPage.Return.Header` and the description in the corresponding table. (upstream status instead of status)

pleshakov: Also, please update the description of the body in the table (upstream status instead of status)

pkg/apis/configuration/validation/validation_test.go
Rulox: As it says in the `msg`, this case was specific for: `"use multi variables, no scheme set",` but I am happy to add another use case with all variables if it makes sense to you.

pkg/apis/configuration/validation/validation.go
Rulox: Do we still support `upstream_status` in the error page redirect tho?

PR #851, opened 2020-02-19
Implement snippet functionality in route

kaywinpark:
### Proposed changes
Support for snippet functionality in route for arbitrary per-location configuration overrides.

We are considering of moving from Ingress to use VirtualServer and VirtualServerRoute. However, we are facing two limitations: the ability to add CORS header and rewrite functionality per location. The current functionality via configmap only allows us to make a global configuration. Thus, we think that by adding these functionalities, this would enable us to make the transition entirely from Ingress.

### Checklist


pleshakov:
Hi @kaywinpark 

Thanks for the PR!

> However, we are facing two limitations: the ability to add CORS header and rewrite functionality per location.

The VirtualServer/VirtualServerRoute resources are relatively new. The features that you mentioned are something we want to add along with some other features:
- Modifiers: modification of request/response headers, rewriting request URI 
- Policies: authentication, rate-limiting

Additionally, we want to add extension methods for VirtualServer and VirtualServerRoute resources - custom templates and things that will match the functionally available with snippets/custom annotations for the Ingress resource.

We're carefully adding features one by one, so that once we implement something, we don't have to change/remove it in the future: our goal is to avoid any backward incompatible changes.

Our next release is 1.7 and it is scheduled for the end of Match. We're planning to work on polices and modifiers for release 1.8, scheduled for the end of June.  If we have time, we will also cover some of the extensibility features.

Regarding this PR, while it solves the problem of adding a custom snippet for the route, we cannot accept it, because we would like to carefully design the mentioned extensibility features for VirtualServer and VirtualServerRoute - we might end up with something that would not look like snippets.

I wonder if an option of having your custom version of the Ingress Controller can work for you for now? Our stable releases do not include any code changes between minor versions (unless we'll have a serious bug we need to fix). So that if you base your version on a stable release, you will be able to merge your changes with any subsequent minor release.  Once we release those needed features (setting response headers, URI rewriting or extensibility features), you could migrate to that version.

kaywinpark:
Hello @pleshakov, thank you so much for your reply.

We will decide if we would transition to use the implementation that we did or wait for the proper implementation of the features. In any case, keep up the good work! We think that this pull request can be closed now.

PR #852, opened 2020-02-21
Add default-server-access-log-off to configmap

lucacome:
### Proposed changes
This PR adds a new ConfigMap `default-server-access-log-off` that enables/disables logging for the default server.
If logging is disabled at the `http` level, logging for the default server is always disabled.

### Checklist

### DIFF
--- a/docs-web/configuration/global-configuration/configmap-resource.md
+++ b/docs-web/configuration/global-configuration/configmap-resource.md
@@ -198,6 +198,10 @@ See the doc about [VirtualServer and VirtualServerRoute resources](/nginx-ingres
      - Disables the `access log <http://nginx.org/en/docs/http/ngx_http_log_module.html#access_log>`_.
      - ``False``
      - 
+   * - ``default-server-access-log-off``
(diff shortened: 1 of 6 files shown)


Rulox:
Looks good!, I just have one question.

Do you have any kind of `golint` rule that transforms the structs order into an alphabetical order or something like that? There are a lot of changes in the structs. We had a discussion around this last year. The way we normally order the structs is putting similar options together. For instance, options that are going to be together in the final NGINX Config.

I am ok with either of those methods to be honest, but I just think we should all follow the same rules instead of having some structs one way and some the other.

Let me know your thoughts (also @Dean-Coakley @pleshakov)

lucacome:
No automatic rule, I just think it's easier to find stuff when it's in alphabetical order. 🙂 
When I mentioned this to @pleshakov he didn't say anything about a convention, but I'm happy to change it back.

Rulox:
Nah it´s fine. I proposed this last year and I remember we decided not to do it (not sure why). But I am ok with it, to me is easier this way too, :shipit:

PR #859, opened 2020-02-27
Fix errorpages body yaml docs

Rulox:
### Proposed changes
The docs of the Body field for an Error Page are not correct. If you use the docs as it is, IC will fail validation because of the unescaped strings. Needs to be multiline.

### DIFF
--- a/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
+++ b/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
@@ -1024,7 +1024,8 @@ codes: [401, 403]
 return:
   code: 200
   type: application/json
-  body: "{\"msg\": \"You don't have permission to do this\"}"
+  body: |
+    {\"msg\": \"You don't have permission to do this\"}
   headers:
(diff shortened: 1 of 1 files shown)


docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -1024,7 +1024,8 @@ codes: [401, 403]
 return:
   code: 200
   type: application/json
-  body: "{\"msg\": \"You don't have permission to do this\"}"
+  body: |
+    {\"msg\": \"You don't have permission to do this\"}
```
pleshakov: because the string itself is not multi-line, why not use:
```yaml
body: '{\"msg\": \"You do not have permission to do this\"}'
```

Rulox: It's defined like that in the Ticket. Also mixing `"` with `'` looks weird to me.

pleshakov: ok. thanks

PR #883, opened 2020-03-12
Update DockerfileForPlus

creativie:
Remove unused FROM

### Proposed changes
Remove unused FROM, its crashes build in this step
`Step 13/22 : FROM base AS local
 ---> 53124caee417
Step 14/22 : COPY nginx-ingress /
COPY failed: stat /var/lib/docker/tmp/docker-builder013496253/nginx-ingress: no such file or directory`

### Checklist

- [ *] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/master/CONTRIBUTING.md) doc
- [ *] I have added tests that prove my fix is effective or that my feature works
- [ *] I have checked that all unit tests pass after adding my changes
- [ *] I have updated necessary documentation
- [ *] I have rebased my branch onto master
- [ *] I will ensure my PR is targeting the master branch and pulling from my branch from my own fork

### DIFF
--- a/build/openshift/DockerfileForPlus
+++ b/build/openshift/DockerfileForPlus
@@ -76,11 +76,6 @@ USER nginx
 
 ENTRYPOINT ["/nginx-ingress"]
 
-
-FROM base AS local
-COPY nginx-ingress /
-
-
 FROM $GOLANG_CONTAINER AS builder
 ARG VERSION
 ARG GIT_COMMIT


Dean-Coakley:
Hi @creativie 

Thanks for the PR.

It seems you tried tried to run `docker build` directly.
 
There are two targets. `local` and `build`. The docker build expects [buildkit](https://github.com/nginxinc/kubernetes-ingress/blob/945bbc7e901f9f378d3e2fe86633773f4afe0b5b/Makefile#L19) to be enabled. And a target specified as visible [here](https://github.com/nginxinc/kubernetes-ingress/blob/945bbc7e901f9f378d3e2fe86633773f4afe0b5b/Makefile#L55-L57). So it is recommended that you use the [Makefile](https://github.com/nginxinc/kubernetes-ingress/blob/master/Makefile)

`local` is faster but depends on a working local `go` installation.

If you try something like:
```sh
make container PREFIX=openshift-nginx-ingress DOCKERFILE=openshift/Dockerfile
```
it should work for you.

Let me know if that resolves your issue.
I'd prefer to keep the offerings of building locally and in a multistage docker build if possible.

creativie:
HI @Dean-Coakley 

> If you try something like

its fix. Thanks

PR #888, opened 2020-03-16
Update controller.go

creativie:
Allow to use tls secret from another namespace

### Proposed changes
Allow to set _secretName_ from another namespace
Example:

```
kind: Ingress
metadata:
  name: load-certificates
  namespace: default
spec:
  tls:
  - hosts:
    - "test.2.example.com"
    secretName: ingress-nginx/wild-second-domain-certs
  rules:
  - host: "test.2.example.com"
    http:
      paths:
      - backend:
          serviceName: my-service
          servicePort: 8080
        path: /
```

### DIFF
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -1438,6 +1438,9 @@ func (lbc *LoadBalancerController) createIngress(ing *extensions.Ingress) (*conf
 	for _, tls := range ing.Spec.TLS {
 		secretName := tls.SecretName
 		secretKey := ing.Namespace + "/" + secretName
+		if strings.Contains(secretName, "/") {
+			secretKey = secretName
+		}
 		secret, err := lbc.getAndValidateSecret(secretKey)
 		if err != nil {
 			glog.Warningf("Error trying to get the secret %v for Ingress %v: %v", secretName, ing.Name, err)


pleshakov:
Hi @creativie 

Thanks for the PR!

This restriction -- not allowing an Ingress resource referencing a secret from a different namespace -- is intentional. Basically, we don't want to allow users/accounts from one namespace to have access to the secrets from another namespace.

Note that currently, it is possible to use a secret from a different namespace for TLS termination, with the following methods:
* Use [mergeable Ingress resources](https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/cross-namespace-configuration/): you can create a master Ingress with TLS termination and then a minion(s) in a different namespace(s).
* You can configure a [wildcard TLS cert/key](https://github.com/nginxinc/kubernetes-ingress/tree/v1.6.3/examples/wildcard-tls-certificate), which can be used for TLS termination of all Ingress resources across all namespaces.
* You can use [VirtualServer (VS) and VirtualServerRoutes (VSR)](https://docs.nginx.com/nginx-ingress-controller/configuration/virtualserver-and-virtualserverroute-resources/): you can configure a VS with TLS termination in one namespace and then reference a VSR from a different namespace.

creativie:
Hi
If I have two or more wildcards?

pleshakov:
Hi @creativie 
sorry for the delay

> If I have two or more wildcards?

In that case it's not possible to use the mentioned TLS Wildcard secret feature. However, the approach will mergeable Ingresses and VS/VSRs should still work.

Dean-Coakley:
Closing as ingress referencing a secret in the same namespace is an intentional restriction.

PR #900, opened 2020-03-26
Update dependencies

Dean-Coakley:
### Proposed changes
* Update kubernetes related dependencies to `kubernetes-1.18.0`
  * `k8s.io/api@kubernetes-1.18.0`
  * `k8s.io/apimachinery@kubernetes-1.18.0`
  * `k8s.io/client-go@kubernetes-1.18.0`
  * `k8s.io/code-generator@kubernetes-1.18.0`

### Checklist


go.mod
```diff
@@ -4,32 +4,23 @@ go 1.13
...
-	gopkg.in/yaml.v2 v2.2.5 // indirect
-	k8s.io/api v0.0.0-20191016110408-35e52d86657a
-	k8s.io/apimachinery v0.0.0-20191004115801-a2eda9f80ab8
-	k8s.io/client-go v0.0.0-20191016111102-bec269661e48
-	k8s.io/code-generator v0.0.0-20191004115455-8e001e5d1894
-	k8s.io/gengo v0.0.0-20191108084044-e500ee069b5c // indirect
-	k8s.io/klog v1.0.0 // indirect
-	k8s.io/kube-openapi v0.0.0-20191107075043-30be4d16710a // indirect
-	k8s.io/utils v0.0.0-20191030222137-2b95a09bc58d // indirect
+	k8s.io/api v0.18.0
```
pleshakov: some tags are `v0.18.0`, some are `v0.18.1-beta.0`. Was it intentional? 🤔

Dean-Coakley: @pleshakov this was the result of:
```sh
go get -u "k8s.io/client-go@kubernetes-1.18.0"
go get -u "k8s.io/code-generator@kubernetes-1.18.0"
go get -u "k8s.io/apimachinery@kubernetes-1.18.0"
go get -u "k8s.io/api@kubernetes-1.18.0"
```

🤔 


Seems to just be a symbolic link/alias for the same release?:
E.g: https://github.com/kubernetes/code-generator/tags

build/openshift/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG GOLANG_CONTAINER=registry.access.redhat.com/ubi8/go-toolset:latest
+ARG GOLANG_CONTAINER=golang:latest
```
pleshakov: I would double check it with @Rulox

Rulox: Yeah, this is fine.

The main issue is that `ubi-go` image latest is Go 1.12. RedHat keeps those images outdated so there's nothing we can do. We need 1.13 for the new k8s libraries. I made some tests, both images work to compile the IC Binary, because the GOOS is `linux` in both.

The alternative was to create our own ubi based image, install Go manually and use that.

I tested with golang:latest and it works for ubi. Think of this like the `alpine` image. We use golang:latest as well there.

PR #903, opened 2020-03-30
Fix omissions for TransportServer and GlobalConfiguration

pleshakov:
### Proposed changes
* Disable preserveUnknownFields in TS and GC crds, similar to VS and VSR crds
* Add missing GlobalConfiguration manifest: required by the installation docs

### DIFF
--- a/deployments/common/global-configuration.yaml
+++ b/deployments/common/global-configuration.yaml
@@ -1,0 +1,5 @@
+apiVersion: k8s.nginx.org/v1alpha1
+kind: GlobalConfiguration
+metadata:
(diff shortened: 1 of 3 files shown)


deployments/common/global-configuration.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: k8s.nginx.org/v1alpha1
```
Rulox: Do we need helm files as well?

pleshakov: We do, but I would implement that separately

PR #909, opened 2020-04-02
Tests new crds

tellet:
Respect latest code changes in tests and increase timeouts in some places for stability.


tests/data/virtual-server/rbac-without-vs.yaml
```diff
@@ -56,4 +57,20 @@ rules:
   resources:
   - ingresses/status
   verbs:
-  - update
\ No newline at end of file
+  - update
+- apiGroups:
+  - "extensions"
+  resources:
+  - ingresses/status
```
Rulox, line 64: I think this is repeated? The whole ingresses/status

PR #921, opened 2020-04-13
Support TransportServer in Helm

pleshakov:
### Proposed changes
* Register TransportServer CRD
* Add controller.enableTLSPassthrough parameter

### DIFF
--- a/deployments/helm-chart/templates/controller-daemonset.yaml
+++ b/deployments/helm-chart/templates/controller-daemonset.yaml
@@ -128,4 +128,7 @@ spec:
           - -enable-prometheus-metrics={{ .Values.prometheus.create }}
           - -prometheus-metrics-listen-port={{ .Values.prometheus.port }}
           - -enable-custom-resources={{ .Values.controller.enableCustomResources }}
+{{- if .Values.controller.enableCustomResources  }}
+          - -enable-tls-passthrough={{ .Values.controller.enableTLSPassthrough }}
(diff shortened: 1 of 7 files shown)


deployments/helm-chart/templates/controller-daemonset.yaml
```diff
@@ -128,4 +128,7 @@ spec:
           - -enable-prometheus-metrics={{ .Values.prometheus.create }}
           - -prometheus-metrics-listen-port={{ .Values.prometheus.port }}
           - -enable-custom-resources={{ .Values.controller.enableCustomResources }}
+{{- if .Values.controller.enableCustomResources  }}
```
Rulox, line 131: Shouldn't this be `enableTLSPassthrough` ?
```suggestion
{{- if .Values.controller.enableTLSPassthrough  }}
```

 Same in the `deployment` template

pleshakov: nope. this is to prevent setting `-enable-tls-passthrough` to `true` if `enableTLSPassthrough` is `false`, since  `-enable-tls-passthrough` requires `-enable-custom-resources`. Otherwise, the IC will exit with an error.

Rulox: Ok thanks, it wasn't clear to me that this logic was in the Helm chart instead of in the IC

PR #933, opened 2020-04-21
Add Makefile for Netlify to use to build docs

jputrino:
### Proposed changes

This pull request sets up the repo to use Netlify to build and host deploy previews for the IC documentation.

Next steps:
- Netlify support will activate automatic builds for this repo once this addition is merged to master.
- Netlify will automatically build docs and create deploy previews for all pull requests into master and release branches.

### Checklist

- [x ] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/master/CONTRIBUTING.md) doc
- [ N/A] I have added tests that prove my fix is effective or that my feature works
- [ N/A] I have checked that all unit tests pass after adding my changes
- [N/A ] I have updated necessary documentation
- [ x] I have rebased my branch onto master
- [ x] I will ensure my PR is targeting the master branch and pulling from my branch from my own fork

### DIFF
--- a/docs-web/Makefile
+++ b/docs-web/Makefile
@@ -1,0 +1,29 @@
+NETLIFY_ROOT	=	/opt/build/repo
+SITE_ROOT			= ${NETLIFY_ROOT}/site
+BUILDDIR			= ${NETLIFY_ROOT}/docs-web/build
+THEME_REPO		= ${THEME_REPO_URL}
+#DOCKER_IMAGE 	=
+
+.PHONY: all build clean setup
+
+#help:
+
+all: clean setup build
+#docker-all: docker setup build
+
+clean:
+	rm -rf ${SITE_ROOT}
+
+setup:
+	mkdir ${SITE_ROOT}
+	@git clone -q --single-branch --branch master https://$(GITHUB_PAT)@${THEME_REPO} ${SITE_ROOT}/docs-nginx-com
+	pip install -r ${SITE_ROOT}/docs-nginx-com/requirements.txt
+	rsync -avOz ${NETLIFY_ROOT}/docs-web/ ${SITE_ROOT}/docs-nginx-com/source/nginx-ingress-controller/ --exclude ${NETLIFY_ROOT}/docs-web/.netlify --exclude ${NETLIFY_ROOT}/docs-web/Makefile
+
+build:
+	sphinx-build -b dirhtml -d ${BUILDDIR}/doctrees ${SITE_ROOT}/docs-nginx-com/source ${BUILDDIR}/dirhtml
+	@echo "Build finished. The HTML pages are in ${BUILDDIR}/dirhtml."
+
+#docker:
+
+#docker-serve:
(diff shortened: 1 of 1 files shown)


docs-web/Makefile
```diff
@@ -0,0 +1,28 @@
...
+#docker-all: docker setup build
+
+clean:
+	rm -rf ${SITE_ROOT}
+
+setup:
+	mkdir ${SITE_ROOT}
+	@git clone -q --single-branch --branch master https://$(GITHUB_PAT)@github.com/nginxinc/docs-nginx-com.git "${SITE_ROOT}"/docs-nginx-com
+	pip install -r ${SITE_ROOT}/docs-nginx-com/requirements.txt
+	rsync -avOz "${NETLIFY_ROOT}"/docs-web/ ${SITE_ROOT}/docs-nginx-com/source/nginx-ingress-controller/ --exclude ${NETLIFY_ROOT}/docs-web/.netlify --exclude ${NETLIFY_ROOT}/docs-web/Makefile
```
pleshakov, line 20: `--exclude ${NETLIFY_ROOT}/docs-web/.netlify` - this file doesn't exist in our repo, right?

docs-web/Makefile
```diff
@@ -0,0 +1,28 @@
+NETLIFY_ROOT	=	/opt/build/repo
```
pleshakov, line 1: could we possibly use the same assignment style as in the https://github.com/nginxinc/kubernetes-ingress/blob/master/Makefile ?
For example,
```
NETLIFY_ROOT = /opt/build/repo
```

docs-web/Makefile
```diff
@@ -0,0 +1,28 @@
...
+SITE_ROOT			= ${NETLIFY_ROOT}/site
+BUILDDIR			= ${NETLIFY_ROOT}/docs-web/build
+#DOCKER_IMAGE 	=
```
pleshakov, line 4: the Makefile includes a few commented out parts. Are those necessary to keep or could be removed?

docs-web/Makefile
jputrino: Correct! Removing.

docs-web/Makefile
```diff
@@ -0,0 +1,21 @@
...
+
+.PHONY: all build clean setup
+
+all: clean setup build
+
+clean:
+	rm -rf ${SITE_ROOT}
+
+setup:
+	mkdir ${SITE_ROOT}
```
lucacome, line 14:
```suggestion
	mkdir -p ${SITE_ROOT}
```

jputrino: @lucacome Thanks! The Netlify build environment always clones the repo into the path `/opt/build/repo`. If we're just adding another directory in that existing location, do we need to use the `-p`?

lucacome: It's just safer, if the directory already exists or if part of the path doesn't exist, it won't throw an error. But not strictly necessary, that's why I approved anyway 🙂

martialonline: I agree with @lucacome here, `-p` is considered good practice and certainly won't do any harm but it's not necessarily needed.

jputrino: Thanks guys!

Issue #863, opened 2020-02-28
Equivalent of `kubernetes.io/ingress.class` annotation for `VirtualServer`s


defect:
Hi!

We are running a couple of nginx ingress instances for isolation of different kinds of workloads. We currently do this using "normal" `Ingress` resources and annotating them with `kubernetes.io/ingress.class`

I've now started experimenting with `VirtualServer` as we're planning to move more complicated nginx configs over from bare-metal and it seems powerful. Is there an annotation or similar to make only a specific ingress class pick up a `VirtualServer`? I tried `kubernetes.io/ingress.class` but the virtual server was added to all the ingress instances.

Thanks in advance!

pleshakov:
Hi @defect 

There is nothing similar to `kubernetes.io/ingress.class` annotation for VirtualServer resource.

Adding the class annotation (or something similar to that) would not be a difficult thing to implement.

Our next release - 1.7. - is planned for end of March. I will let you now next week if we will be able to implement this feature for 1.7.

For now, the only way to use different Ingress Controllers for different VirtualServers is to isolate them on a per namespace basis.

defect:
Okay, thank you @pleshakov! That's all understandable. I might have a go at implementing that. I'd like to get familiarized with the codebase anyway :)

pleshakov:
@defect  just an update. I don't think we'll make this feature available in the upcoming 1.7, but we'll  definitely include it in 1.8.

Issue #924, opened 2020-04-17
Support for k8s 1.18 Ingress Class?


TheSithPadawan:
How will nginx support the Ingress Class for the new k8s update?

```
apiVersion: "networking.k8s.io/v1beta1"
kind: "IngressClass"
metadata:
  name: "external-lb"
spec:
  controller: "example.com/ingress-controller" --> how should we specify nginx ingress controller?
  parameters: --> what parameters are needed?
    apiGroup: "k8s.example.com/v1alpha"
    kind: "IngressParameters"
    name: "external-lb"
```
Thank you!

pleshakov:
Hi @TheSithPadawan 

Our Ingress Controller doesn’t support yet the IngressClass Resource and the new `ingressClassName` field of the Ingress resource. 

We haven’t decided yet how we’re going to support the IngressClass resource.

TheSithPadawan:
Thanks for the update! Any ETA for when you will support the IngressClass resource?

pleshakov:
@TheSithPadawan we don't have an ETA now. However, I will update this issue once we have it.

Since (1) this is a very recent feature and supported only in k8s 1.18+ and (2) the old way with the class annotation will continue to work (although it is deprecated) , I wonder if there are any reasons you're asking for this feature?

TheSithPadawan:
@pleshakov  I'm working on a networking controller for k8s, and we want to support nginx controller, so just want to see what's your design/expectation for the new updates. Plz keep me posted.

PR #937, opened 2020-04-24
VirtualServer respecting kubernetes.io/ingress.class

defect:
### Proposed changes
This is an attempt at implementing https://github.com/nginxinc/kubernetes-ingress/issues/863

I apologize for taking so long to open this PR. I actually wrote this quite a few weeks ago, deployed to our cluster, and then kinda forgot about it. The good news is that we've been having it running in production for a while now. We're still not heavy on using `VirtualServer`s but I've deployed a few, making sure they end up on the correct ingress instances.

I do need some pointers and/or help with the python test suite though. I haven't quite grok'd it enough to get a decent test for this set up. Any help is appreciated.

### Checklist

### DIFF
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -420,7 +420,7 @@ func (lbc *LoadBalancerController) syncEndpoint(task task) {
 		var mergableIngressesSlice []*configs.MergeableIngresses
 
 		for i := range ings {
-			if !lbc.IsNginxIngress(&ings[i]) {
+			if !lbc.CheckIngressClassAnnotation(ings[i].Annotations) {
 				continue
 			}
 			if isMinion(&ings[i]) {
@@ -597,7 +597,7 @@ func (lbc *LoadBalancerController) GetManagedIngresses() ([]extensions.Ingress, 
 	ings, _ := lbc.ingressLister.List()
 	for i := range ings.Items {
 		ing := ings.Items[i]
-		if !lbc.IsNginxIngress(&ing) {
+		if !lbc.CheckIngressClassAnnotation(ing.Annotations) {
 			continue
 		}
 		if isMinion(&ing) {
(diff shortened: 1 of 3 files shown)


lucacome:
Hi @defect thanks for the PR! We'll review it and get back to you.

lucacome:
Hi @defect sorry for the delay.

The code looks good and I've tested that it's working, but we were thinking about introducing a new field `ingressClassName` instead of using annotations like:

```
apiVersion: k8s.nginx.org/v1
kind: VirtualServer
metadata:
  name: cafe
spec:
  ingressClassName: prod
  host: cafe.example.com
```
Our goal with VirtualServer is to avoid using any annotations, but rather add new fields to the spec of the resources.

Would you be interested in implementing a class field? Otherwise, we will implement this feature soon :)

defect:
Yeah, that makes sense to me. I suspect it's to align with the changes in 1.18? Is there a wider plan on how to attack https://github.com/nginxinc/kubernetes-ingress/issues/924 ?

I'd be happy to work on moving to a class field, i'll try to dig around a bit this weekend.

In the meantime, I'll try to keep my fork up-to-date with master if anyone else needs to use the annotation on `VirtualServer`s.

lucacome:
We don't have a wider plan at the moment, but we will support it eventually.

I've also noticed that we probably have a bug in our current implementation: when a user changes the class of an existing Ingress resource, the configuration doesn't get deleted. I think it should be deleted and that it should be the behavior for VS/VSR as well.
What do you think? Does it make sense?

lucacome:
Hi @defect,

I just wanted to check if you had a chance to look into this or if you need any assistance. Let me know 🙂 

Cheers.

lucacome:
Closing this as it was implemented in #994 with the `ingressClassName` field

PR #954, opened 2020-05-15
Add spire-agent-address flag to command line args

kate-osborn:
### Proposed changes
In order to communicate with services in the mTLS enabled NGINX Service Mesh, the Ingress 
Controller needs to fetch certs and keys from the Spire agent running in the K8s cluster. Every location block in the `nginx.ingress` and `nginx.virtualserver` configs need to reference these certs and keys when proxying traffic to the upstream services.
- Adds a global and static configuration option,`spire-agent-address`, as an optional command line argument.
- On startup, if the Spire agent address is provided then we attempt to connect to the Spire agent and fetch the certs and keys. If successful, the certs and keys are written to disk. If the controller cannot connect to the Spire agent or the certs/keys are not written to disk after 30 seconds, the controller fails to start.
- When certs are rotated, the controller fetches new certs from the Spire agent, writes them to disk, and reloads nginx.

### Checklist


cmd/nginx-ingress/main.go
```diff
@@ -388,6 +392,7 @@ func main() {
 		NginxStatusPort:                *nginxStatusPort,
 		StubStatusOverUnixSocketForOSS: *enablePrometheusMetrics,
 		TLSPassthrough:                 *enableTLSPassthrough,
+		SpiffeCerts:                    *spireAgentAddress != "",
```
lucacome, line 395: should this be `SpireCerts`?

internal/configs/configurator.go
```diff
@@ -30,6 +37,15 @@ const WildcardSecretName = "wildcard"
 // JWTKeyKey is the key of the data field of a Secret where the JWK must be stored.
 const JWTKeyKey = "jwk"
 
+// SPIFFE filenames and modes
+const (
+	spiffeCertFileName   = "spire_cert.pem"
```
lucacome, line 42: should the naming spiffe/spire be consistent? The variables are called like `spiffeCertFileName` but the filename cert is `spire_cert.pem`

kate-osborn: I see how this is confusing, the naming should be consistent. Since these certs are generated from the SVIDs, SPIFFE Verifiable Identity Documents, I think it makes name the certs `spiffe` instead of `spire`

cmd/nginx-ingress/main.go
kate-osborn: I think `SpiffeCerts` is correct, see my comment below about `SVIDs`

PR #973, opened 2020-05-29
Add status to VirtualServer and VirtualServerRoute

Rulox:
### Proposed changes
This PR adds a status field to VirtualServer and VirtualServerRoute resources.

PR #980, opened 2020-06-04
Enable leader election as default

Rulox:
### Proposed changes
With the new feature in https://github.com/nginxinc/kubernetes-ingress/pull/973, it is required to have leader election to avoid multiple replicas to try to update the VS/VSR status.

When the KIC is deployed using Helm, `leaderElection` is enabled by default. With this change, we do the same when the users deploy the KIC manually deploying the manifests.


Rulox:
That sounds better thanks @Dean-Coakley , I'll do the changes

PR #994, opened 2020-06-15
Support Class Field in VS/VSR

lucacome:
Adding support for Class Field in VirtualServer and VirtualServerRoute resources.

### DIFF
--- a/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
+++ b/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
@@ -96,7 +96,7 @@ spec:
      - `[]route <#virtualserver-route>`_
      - No
    * - ``class``
-     - Specifies which Ingress controller must handle the Virtual Server resource. Set to `nginx` to make NGINX Ingress controller handle it.
+     - Specifies which Ingress controller must handle the VirtualServer resource. Set to `nginx` to make NGINX Ingress controller handle it.
      - ``string``
      - No
 ```
@@ -278,7 +278,7 @@ Note that each subroute must have a `path` that starts with the same prefix (her
      - `[]subroute <#virtualserverroute-subroute>`_
      - No
    * - ``class``
-     - Specifies which Ingress controller must handle the Virtual Server Route resource. Set to `nginx` to make NGINX Ingress controller handle it.
+     - Specifies which Ingress controller must handle the VirtualServerRoute resource. Set to `nginx` to make NGINX Ingress controller handle it.
      - ``string``_
      - No
 ```
--- a/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
+++ b/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
@@ -95,6 +95,10 @@ spec:
      - A list of routes.
      - `[]route <#virtualserver-route>`_
      - No
+   * - ``class``
+     - Specifies which Ingress controller must handle the Virtual Server resource. Set to `nginx` to make NGINX Ingress controller handle it.
+     - ``string``
+     - No
 ```
 
 ### VirtualServer.TLS
@@ -272,6 +276,10 @@ Note that each subroute must have a `path` that starts with the same prefix (her
    * - ``subroutes``
      - A list of subroutes.
      - `[]subroute <#virtualserverroute-subroute>`_
+     - No
+   * - ``class``
+     - Specifies which Ingress controller must handle the Virtual Server Route resource. Set to `nginx` to make NGINX Ingress controller handle it.
+     - ``string``_
      - No
 ```
 
--- a/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
+++ b/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
@@ -96,7 +96,7 @@ spec:
      - `[]route <#virtualserver-route>`_
      - No
    * - ``class``
-     - Specifies which Ingress controller must handle the VirtualServer resource. Set to `nginx` to make NGINX Ingress controller handle it.
+     - Specifies which Ingress controller must handle the VirtualServer resource.
      - ``string``
      - No
 ```
@@ -278,7 +278,7 @@ Note that each subroute must have a `path` that starts with the same prefix (her
      - `[]subroute <#virtualserverroute-subroute>`_
      - No
    * - ``class``
-     - Specifies which Ingress controller must handle the VirtualServerRoute resource. Set to `nginx` to make NGINX Ingress controller handle it.
+     - Specifies which Ingress controller must handle the VirtualServerRoute resource. Must be the same as the ``class`` of the VirtualServer that references this resource.
      - ``string``_
      - No
 ```
--- a/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
+++ b/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
@@ -95,7 +95,7 @@ spec:
      - A list of routes.
      - `[]route <#virtualserver-route>`_
      - No
-   * - ``class``
+   * - ``ingressClassName``
      - Specifies which Ingress controller must handle the VirtualServer resource.
      - ``string``
      - No
@@ -277,8 +277,8 @@ Note that each subroute must have a `path` that starts with the same prefix (her
      - A list of subroutes.
      - `[]subroute <#virtualserverroute-subroute>`_
      - No
-   * - ``class``
-     - Specifies which Ingress controller must handle the VirtualServerRoute resource. Must be the same as the ``class`` of the VirtualServer that references this resource.
+   * - ``ingressClassName``
+     - Specifies which Ingress controller must handle the VirtualServerRoute resource. Must be the same as the ``ingressClassName`` of the VirtualServer that references this resource.
      - ``string``_
      - No
 ```
--- a/docs-web/installation/running-multiple-ingress-controllers.md
+++ b/docs-web/installation/running-multiple-ingress-controllers.md
@@ -10,6 +10,7 @@ This document explains the following topics:
 The smooth coexistence of multiple Ingress Controllers in one cluster is provided by the Ingress class concept, which mandates the following:
 * Every Ingress Controller must only handle Ingress resources for its particular class. 
 * Ingress resources should be annotated with the `kubernetes.io/ingress.class` annotation set to the value, which corresponds to the class of the Ingress Controller the user wants to use. 
+* Virtual Server and Virtual Server Route resources should have the `class` field set to the value, which corresponds to the class of the Ingress Controller the user wants to use.
 
 ### Configuring Ingress Class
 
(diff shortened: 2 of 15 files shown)


docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -95,6 +95,10 @@ spec:
      - A list of routes.
      - `[]route <#virtualserver-route>`_
      - No
+   * - ``class``
+     - Specifies which Ingress controller must handle the Virtual Server resource. Set to `nginx` to make NGINX Ingress controller handle it.
```
Rulox, line 99: Just for consistency we use `VirtualServer`:
```suggestion
     - Specifies which Ingress controller must handle the VirtualServer resource. Set to `nginx` to make NGINX Ingress controller handle it.
```

docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -273,6 +277,10 @@ Note that each subroute must have a `path` that starts with the same prefix (her
      - A list of subroutes.
      - `[]subroute <#virtualserverroute-subroute>`_
      - No
+   * - ``class``
+     - Specifies which Ingress controller must handle the Virtual Server Route resource. Set to `nginx` to make NGINX Ingress controller handle it.
```
Rulox, line 281: Same as before:

```suggestion
     - Specifies which Ingress controller must handle the VirtualServerRoute resource. Set to `nginx` to make NGINX Ingress controller handle it.
```

docs-web/installation/running-multiple-ingress-controllers.md
```diff
@@ -10,6 +10,7 @@ This document explains the following topics:
 The smooth coexistence of multiple Ingress Controllers in one cluster is provided by the Ingress class concept, which mandates the following:
 * Every Ingress Controller must only handle Ingress resources for its particular class. 
 * Ingress resources should be annotated with the `kubernetes.io/ingress.class` annotation set to the value, which corresponds to the class of the Ingress Controller the user wants to use. 
+* Virtual Server and Virtual Server Route resources should have the `class` field set to the value, which corresponds to the class of the Ingress Controller the user wants to use.
```
Rulox, line 13:
```suggestion
* VirtualServer and VirtualServerRoute resources should have the `class` field set to the value, which corresponds to the class of the Ingress Controller the user wants to use.
```

internal/k8s/controller.go
```diff
@@ -2372,14 +2372,24 @@ func (lbc *LoadBalancerController) getServiceForIngressBackend(backend *extensio
 
 // IsNginxIngress checks if resource ingress class annotation (if exists) is matching with ingress controller class
```
Rulox, line 2373: I think we can also add here that is possible to pass a string now

internal/k8s/controller_test.go
```diff
@@ -85,6 +85,55 @@ func TestIsNginxIngress(t *testing.T) {
 		},
 	}
 
+	var testsWithoutIngressClassOnlyVS = []struct {
```
pleshakov, line 88: this test grew 2 times now :) maybe it makes sense split VS/VSRs and Ingress into multiple test functions?

internal/k8s/controller.go
```diff
@@ -2372,14 +2372,24 @@ func (lbc *LoadBalancerController) getServiceForIngressBackend(backend *extensio
...
 // If annotation is absent and use-ingress-class-only enabled - ingress resource would ignore
-func (lbc *LoadBalancerController) IsNginxIngress(ing *extensions.Ingress) bool {
-	if class, exists := ing.Annotations[ingressClassKey]; exists {
-		if lbc.useIngressClassOnly {
-			return class == lbc.ingressClass
+func (lbc *LoadBalancerController) IsNginxIngress(obj interface{}) bool {
```
pleshakov, line 2375: this slightly doesn't make sense to me why we take either string or `*extensions.Ingress` as argument, because they are completely different things.

 Would it make it simpler if we take  Ingress,  VirtualServer and VirtualServerRoute as arguments?

pleshakov, line 2375: because IngressClass is no longer only about Ingress resources, I suggest renaming this method to reflect that. For example, `BelongsToIngressClass`, `HasCorrectIngressClass` and `MatchesIngressClass`. I think the current name also assumes that the class of our Ingress Controller is `nginx`, but this is not always true.

docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -95,6 +95,10 @@ spec:
      - A list of routes.
      - `[]route <#virtualserver-route>`_
      - No
+   * - ``class``
+     - Specifies which Ingress controller must handle the VirtualServer resource. Set to `nginx` to make NGINX Ingress controller handle it.
```
pleshakov, line 99: The second sentence assumes that other Ingress Controllers can handle VirtualServer resources (not true) and that our Ingress Controller has the `nginx` class (not always true). That's why I suggest removing it here and in the VSR doc:
```suggestion
     - Specifies which Ingress controller must handle the VirtualServer resource.
```

docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -273,6 +277,10 @@ Note that each subroute must have a `path` that starts with the same prefix (her
      - A list of subroutes.
      - `[]subroute <#virtualserverroute-subroute>`_
      - No
+   * - ``class``
+     - Specifies which Ingress controller must handle the VirtualServerRoute resource. Set to `nginx` to make NGINX Ingress controller handle it.
```
pleshakov, line 281:
```suggestion
     - Specifies which Ingress controller must handle the VirtualServerRoute resource. Must be the same as the ``class`` of the VirtualServer that references this resource.
```

docs-web/installation/running-multiple-ingress-controllers.md
```diff
@@ -10,6 +10,7 @@ This document explains the following topics:
 The smooth coexistence of multiple Ingress Controllers in one cluster is provided by the Ingress class concept, which mandates the following:
 * Every Ingress Controller must only handle Ingress resources for its particular class. 
 * Ingress resources should be annotated with the `kubernetes.io/ingress.class` annotation set to the value, which corresponds to the class of the Ingress Controller the user wants to use. 
+* VirtualServer and VirtualServerRoute resources should have the `class` field set to the value, which corresponds to the class of the Ingress Controller the user wants to use.
```
pleshakov, line 13: I sent a few suggestions for this doc separately

internal/k8s/handlers.go
```diff
@@ -333,12 +337,20 @@ func createVirtualServerHandlers(lbc *LoadBalancerController) cache.ResourceEven
...
+				glog.Infof("Ignoring VS %v based on class %v", vs.Name, vs.Spec.IngressClass)
+				return
+			}
 			glog.V(3).Infof("Removing VirtualServer: %v", vs.Name)
 			lbc.AddSyncQueue(vs)
 		},
 		UpdateFunc: func(old, cur interface{}) {
 			curVs := cur.(*conf_v1.VirtualServer)
 			oldVs := old.(*conf_v1.VirtualServer)
+			if !lbc.IsNginxIngress(curVs.Spec.IngressClass) {
```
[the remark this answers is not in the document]
lucacome: The case you mentioned works. There's a bug when you change the class from the "right" one to a different one, where the config it's not deleted.

internal/k8s/controller.go
lucacome: Not sure it would make it simpler to pass three arguments instead of one. In the end, we're just checking the string even if it's in an annotation, right?

pleshakov: I made a mistake: used `and` instead of `or` :( so we still take the obj as an argument, but that obj can be Ingress, VS or VSR.

 What I meant is something like that:
```
	var class string
	switch obj.(type) {
	case *conf_v1.VirtualServerRoute:
		. . .
        case  *conf_v1.VirtualServer:
                . . .
	case *extensions.Ingress:
		ing := obj.(*extensions.Ingress)
		if ingClass, exists := ing.Annotations[ingressClassKey]; exists {
			class = ingClass
		}
```

Or alternatively, when you call this function, you call for VS and VRS like this: 
```
lbc.HasCorrectIngressClass(vs.Spec.IngressClass)
```
So why not do the same for Ingress resource
```
lbc.HasCorrectIngressClass(ing.Annotations[ingressClassKey])
```

internal/k8s/controller_test.go
```diff
@@ -145,29 +145,144 @@ func TestIsNginxIngress(t *testing.T) {
...
+func TestHasCorrectIngressClassVS(t *testing.T) {
+	ingressClass := "ing-ctrl"
+
+	var testsWithIngressClassOnlyVS = []struct {
+		lbc      *LoadBalancerController
+		ing      *conf_v1.VirtualServerSpec
+		expected bool
+	}{
+		{
+			&LoadBalancerController{
```
pleshakov, line 180: in seems that `LoadBalancerController` is the same in all cases for `testsWithIngressClassOnlyVS` (and similarly for `testsWithoutIngressClassOnlyVS`). Maybe it make sense to define them once as a variable, similarly to the `ingressClass` variable.

internal/k8s/controller.go
```diff
@@ -2083,6 +2073,10 @@ func newVirtualServerRouteErrorFromVSR(virtualServerRoute *conf_v1.VirtualServer
 }
 
 func (lbc *LoadBalancerController) createVirtualServer(virtualServer *conf_v1.VirtualServer) (*configs.VirtualServerEx, []virtualServerRouteError) {
+	if !lbc.HasCorrectIngressClass(virtualServer) {
```
pleshakov, line 2076: I don't think we are on the same page with the problem with `createVirtualServer`.

the problem is that in the function we do `obj, exists, err := lbc.virtualServerRouteLister.GetByKey(vsrKey)` and we don't check the class of that VSR!

So If you create the following VSR with an incorrect class:
```yaml
apiVersion: k8s.nginx.org/v1
kind: VirtualServerRoute
metadata:
  name: coffee
spec:
  ingressClassName: random
  host: cafe.example.com
  upstreams:
  - name: latte
    service: latte-svc
    port: 80
  - name: espresso
    service: espresso-svc
    port: 80
  subroutes:
  - path: "/coffee/latte"
    action:
      pass: latte
  - path: "/coffee/espresso"
    action:
      pass: espresso
```
And then create a VS with the correct class:
```yaml
apiVersion: k8s.nginx.org/v1
kind: VirtualServer
metadata:
  name: cafe
spec:
  ingressClassName: nginx
  host: cafe.example.com
  upstreams:
  - name: tea
    service: tea-svc
    port: 80
  - name: coffee
    service: coffee-svc
    port: 80
  routes:
  - path: "/tea"
    action:
      pass: tea
  - path: "/coffee"
    route: coffee
```

The Ingress Controller will handle the VSR :
```
Events:
  Type    Reason          Age   From                      Message
  ----    ------          ----  ----                      -------
  Normal  AddedOrUpdated  6m9s  nginx-ingress-controller  Configuration for default/coffee was added or updated
```

This is a bug :(

The fix in this line doesn't fix it and isn't necessary.

internal/k8s/handlers.go
```diff
@@ -316,8 +316,8 @@ func createVirtualServerHandlers(lbc *LoadBalancerController) cache.ResourceEven
 	return cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			vs := obj.(*conf_v1.VirtualServer)
-			if !lbc.HasCorrectIngressClass(vs.Spec.IngressClass) {
-				glog.Infof("Ignoring VS %v based on class %v", vs.Name, vs.Spec.IngressClass)
+			if !lbc.HasCorrectIngressClass(vs) {
+				glog.Infof("Ignoring VirtualServer %v based on class %v", vs.Name, vs.Spec.IngressClass)
```
pleshakov, line 320: could you update the other occurrences of VS/VSR in the `handlers.go` as well?

PR #995, opened 2020-06-16
Orphaned vsrs emit events

LorcanMcVeigh:
### Proposed changes
Vsr's are notified when they become orphaned by a vs, either by deleting the vs or through misconfiguration  

### Checklist


internal/k8s/controller.go
```diff
@@ -871,40 +871,52 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
...
-			err = lbc.statusUpdater.UpdateVirtualServerStatus(vs, conf_v1.StateInvalid, reason, msg)
-
-			if err != nil {
-				glog.Errorf("Error when updating the status for VirtualServer %v/%v: %v", vs.Namespace, vs.Name, err)
-			}
+		reason := "NoVirtualServerFound"
+		for _, vsr := range vsrlist {
+			msg := fmt.Sprintf("No VirtualServer references VirtualServerRoute %v/%v", vsr.Namespace, vsr.Name)
+			lbc.recorder.Eventf(vsr, api_v1.EventTypeWarning, reason, msg)
+			lbc.statusUpdater.UpdateVirtualServerRouteStatus(vsr, conf_v1.StateInvalid, reason, msg)
```
Dean-Coakley, line 917: `UpdateVirtualServerRouteStatus` returns an error. This should be checked?

* If so, this applies in the delete VirtualServer case too.

internal/k8s/controller.go
```diff
@@ -976,8 +992,49 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
...
+	key := fmt.Sprintf("%s/%s", vs.Namespace, vs.Name)
+	for _, r := range vsrs {
+		if r.Status.ReferencedBy == key {
+			result = append(result, r)
+		}
+	}
+	return result
+}
+
+func findVirtualServerRoutesForVirtualServer(virtualserver *conf_v1.VirtualServer, virtualServerRoutes []*conf_v1.VirtualServerRoute) []*conf_v1.VirtualServerRoute {
```
Dean-Coakley, line 1028: Lets keep func args short and consistent?

```suggestion
func findVirtualServerRoutesForVirtualServer(vs *conf_v1.VirtualServer, vsrs []*conf_v1.VirtualServerRoute) []*conf_v1.VirtualServerRoute {
```

internal/k8s/controller_test.go
```diff
@@ -1382,6 +1382,126 @@ func TestFindVirtualServerRoutesForService(t *testing.T) {
...
+	vsr4 := conf_v1.VirtualServerRoute{
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name:      "vsr-4",
+			Namespace: "ns-1",
+		},
+	}
+
+	virtualserverroutes := []*conf_v1.VirtualServerRoute{&vsr1, &vsr2, &vsr3, &vsr4}
+
+	vs1 := conf_v1.VirtualServer{
```
Dean-Coakley, line 1416: Call it `vs` because there is no `vs2`?
```suggestion
	vs := conf_v1.VirtualServer{
```

* If you change this, change elsewhere too.

internal/k8s/controller_test.go
```diff
@@ -1382,6 +1382,126 @@ func TestFindVirtualServerRoutesForService(t *testing.T) {
...
+	result := findVirtualServerRoutesForVirtualServer(&vs1, virtualserverroutes)
+
+	if !reflect.DeepEqual(result, expected) {
+		t.Errorf("findVirtualServerRoutesForVirtualServer returned %v but expected %v", result, expected)
+	}
+
+}
+
+func TestFindAllVirtualServerRoutesForVirtualServer(t *testing.T) {
+
```
Dean-Coakley, line 1447:
```suggestion
```

internal/k8s/controller_test.go
```diff
@@ -1382,6 +1382,126 @@ func TestFindVirtualServerRoutesForService(t *testing.T) {
...
+		},
+	}
+
+	vsr2 := conf_v1.VirtualServerRoute{
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name:      "vsr-2",
+			Namespace: "ns-1",
+		},
+		Status: conf_v1.VirtualServerRouteStatus{
+			State:        "InValid",
```
Dean-Coakley, line 1467: Pretty sure this has no effect. But, would be nice to keep consistent casing:
https://github.com/nginxinc/kubernetes-ingress/blob/7a2b51a00a642f4b432aaa5b6a1f0a904f04c94e/pkg/apis/configuration/v1/types.go#L13

* (and fix elsewhere)

internal/k8s/controller.go
```diff
@@ -976,8 +992,49 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
 				glog.Errorf("Error when updating the status for VirtualServerRoute %v/%v: %v", vsr.Namespace, vsr.Name, err)
 			}
 		}
+
+		vsrkey := fmt.Sprintf("%s/%s", vsr.Namespace, vsr.Name)
+		for i, r := range vsrlist {
```
Dean-Coakley, line 997: Optional:
Would be nice to move this to a unit tested func in my opinion.

internal/k8s/controller.go
```diff
@@ -976,8 +993,58 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
...
+		msg := fmt.Sprintf("Ignored by VirtualServer %v/%v", vs.Namespace, vs.Name)
+		lbc.recorder.Eventf(vsr, api_v1.EventTypeWarning, "Ignored", msg)
+		err := lbc.statusUpdater.UpdateVirtualServerRouteStatus(vsr, conf_v1.StateInvalid, reason, msg)
+		if err != nil {
+			glog.Errorf("Error when updating the status for VirtualServerRoute %v/%v: %v", vsr.Namespace, vsr.Name, err)
+		}
+	}
+}
+
+func checkForVirtualServerRoute(vsrkey string, vsrlist []*conf_v1.VirtualServerRoute) []*conf_v1.VirtualServerRoute {
```
Dean-Coakley, line 1011:
```suggestion
func removeVirtualServerRouteByKey(vsrkey string, vsrlist []*conf_v1.VirtualServerRoute) []*conf_v1.VirtualServerRoute {
```

internal/k8s/controller_test.go
```diff
@@ -1382,6 +1382,172 @@ func TestFindVirtualServerRoutesForService(t *testing.T) {
 	}
 }
 
+func TestCheckForVirtualServerRoute(t *testing.T) {
+
```
Dean-Coakley, line 1386:
```suggestion
```

internal/k8s/controller.go
```diff
@@ -976,8 +993,58 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
...
+	var vsrNotRef []*conf_v1.VirtualServerRoute
```
Dean-Coakley, line 1012: I suggest removing this slice and modifying the slice from args directly.

The return value is a slice of vsrs, not a slice of vsrsNotReferenced.

internal/k8s/controller.go
```diff
@@ -976,8 +993,58 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
...
+	for i, r := range vsrlist {
+		key := fmt.Sprintf("%s/%s", r.Namespace, r.Name)
+		if key == vsrkey {
+			if len(vsrlist) > 1 {
```
Rulox: Why do we need this `if len(vsrlist) > 1 / else`?

internal/k8s/controller.go
```diff
@@ -976,8 +983,56 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
...
+		}
+	}
+
+	return vsrList
+}
+
+func findAllVirtualServerRoutesForVirtualServer(key string, vsrs []*conf_v1.VirtualServerRoute) []*conf_v1.VirtualServerRoute {
+	var result []*conf_v1.VirtualServerRoute
+	for _, r := range vsrs {
+		if r.Status.ReferencedBy == key {
```
pleshakov, line 1019: The status of the resource is not a reliable source, unfortunately. for one thing, status reporting could be disabled in the IC. Additionally, there are periods when status isn't reported even if it is enabled (when a leader is being elected).

In the `Configurator`, we already track which VSs and VSRs are currently present in the configuration. Note the map `virtualServers      map[string]*VirtualServerEx`. This means you can get the list of VSRs for a VS.

Now for the case when you delete the VS configuration (either on VS deletion or misconfiguration), before deleting the VSR, you can the list of its VSRs from Configurator. For example, you can add a method `vsrs := lbc.cnf.GetVirtualServerRoutesForVirtualServer(key)`. After the deletion, you can just go through the list and report events and update statuses of those VSRs.

Now for the case when a VS stops referencing some VSR, you can do the following:
* before you call `AddOrUpdateVirtualServer`, call `previousVSRs := lbc.cnf.GetVirtualServerRoutesForVirtualServer(key)` to get the old list of VSRs.
* After the call, compare that list with the  current list (`vsEx.VirtualServerRoutes`). For the ones from `previousVSRs` that are (1) not present in `vsEx.VirtualServerRoutes` and (2) not present in `vsrErrors`, we should report `NoVirtualServersFound` event and set the status Invalid. 

Let me know what you think

internal/k8s/controller.go
```diff
@@ -976,8 +983,56 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
...
+			result = append(result, r)
+		}
+	}
+	return result
+}
+
+func findVirtualServerRoutesForVirtualServer(vs *conf_v1.VirtualServer, vsrs []*conf_v1.VirtualServerRoute) []*conf_v1.VirtualServerRoute {
```
pleshakov, line 1026: this function is not used. could you remove it?

internal/k8s/controller.go
```diff
@@ -914,6 +929,8 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
 		lbc.recorder.Eventf(vs, api_v1.EventTypeWarning, "IgnoredVirtualServerRoute", "Ignored VirtualServerRoute %v: %v", vsrError.VirtualServerRouteNsName, vsrError.Error)
 		if vsrError.VirtualServerRoute != nil {
 			lbc.recorder.Eventf(vsrError.VirtualServerRoute, api_v1.EventTypeWarning, "Ignored", "Ignored by VirtualServer %v/%v: %v", vs.Namespace, vs.Name, vsrError.Error)
+			vsrKey := fmt.Sprintf("%v/%v", vsrError.VirtualServerRoute.Namespace, vsrError.VirtualServerRoute.Name)
+			vsrs = removeVirtualServerRouteByKey(vsrKey, vsrs)
```
pleshakov, line 933: the function `removeVirtualServerRouteByKey` modifies the array that back ups the slice `vsrs`. So it will modify the array of the slices that configurator stores. That could lead to hard to catch bugs :(

perhaps instead of calling `removeVirtualServerRouteByKey` here and in line 998, you have another slice, something like 

```go
var handledVSRs []*conf_v1.VirtualServerRoute
. . .
handledVSRs = append(handledVSRs, vsrError.VirtualServerRoute)
. . .
handledVSRs = append(handledVSRs, vsr)
. . .
ignoredVSRs := filterOutVirtualServerRoutes(vsrs, handledVSRs)
for _, vsr := range ignoredVSRs 
. . .
```
In `filterOutVirtualServerRoutes`, make sure to create a new slice, instead of modifying the existing.

I would recommend renaming `vsrs` to `previosVSRs` to reflect that the slice contains a list of the VSRs that were previously added in the configuration.

internal/k8s/controller.go
```diff
@@ -969,15 +986,41 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
...
 			if err != nil {
 				glog.Errorf("Error when updating the status for VirtualServerRoute %v/%v: %v", vsr.Namespace, vsr.Name, err)
 			}
 		}
+
+		vsrkey := fmt.Sprintf("%s/%s", vsr.Namespace, vsr.Name)
+		vsrs = removeVirtualServerRouteByKey(vsrkey, vsrs)
+	}
+
+	reason := "Ignored"
```
pleshakov, line 1001: as was mentioned here https://github.com/nginxinc/kubernetes-ingress/pull/995#discussion_r442570355, this case should not have different reason and message. The VSRS that we report here are no longer referenced by VS (for example, a VS was updated). That's why its better to have the `NoVirtualServersFound` reason and `No VirtualServer references VirtualServerRoute` message

internal/k8s/controller.go
```diff
@@ -969,15 +987,49 @@ func (lbc *LoadBalancerController) syncVirtualServer(task task) {
...
+			if err != nil {
+				glog.Errorf("Error when updating the status for VirtualServerRoute %v/%v: %v", vsr.Namespace, vsr.Name, err)
+			}
+		}
+	}
+}
+
+func filterOutVirtualServerRoutes(previousVSRs []*conf_v1.VirtualServerRoute, handledVSRs []*conf_v1.VirtualServerRoute) []*conf_v1.VirtualServerRoute {
+	var orphanedVSRs []*conf_v1.VirtualServerRoute
+	for _, r := range previousVSRs {
```
pleshakov, line 1018: better to not name variables like `r1`. 
name like `r1` doesn't provide any context. it makes the code less readable.

Consider the following names for the loop:

`for _, r := range previousVSRs` -> `for _, prev := range previousVSRs`
`for _, r1 := range handledVSRs` -> `for _, handled := range handledVSRs`

`key` - > `prevKey`
`vsrKey` -> `handledKey`

internal/k8s/controller_test.go
```diff
@@ -1382,6 +1382,56 @@ func TestFindVirtualServerRoutesForService(t *testing.T) {
 	}
 }
 
+func TestCheckForVirtualServerRoute(t *testing.T) {
```
pleshakov, line 1385: the name of the unit test needs to reflect the name of the function under the test.

internal/k8s/controller_test.go
```diff
@@ -1382,6 +1382,56 @@ func TestFindVirtualServerRoutesForService(t *testing.T) {
...
+	vsrs := []*conf_v1.VirtualServerRoute{&vsr1, &vsr2, &vsr3, &vsr4, &vsr5}
+
+	handledVSRs := []*conf_v1.VirtualServerRoute{&vsr3}
+
+	expected := []*conf_v1.VirtualServerRoute{&vsr1, &vsr2, &vsr4, &vsr5}
+
+	result := filterOutVirtualServerRoutes(vsrs, handledVSRs)
+
+	if !reflect.DeepEqual(result, expected) {
+		t.Errorf("checkForVirtualServerRoute return %v but expected %v", result, expected)
```
pleshakov, line 1430: the name here needs to be the name fo the function under the test

internal/k8s/controller_test.go
```diff
@@ -1382,6 +1382,56 @@ func TestFindVirtualServerRoutesForService(t *testing.T) {
...
+	}
+
```
pleshakov, line 1432: this empty line is unnecessary

PR #996, opened 2020-06-17
Update vs/vsr status when IC becomes the leader

Rulox:
### Proposed changes
When updating VS and VSR statuses of created resources during a leading election process, those resources won't have their status updated until they're synced again and the KIC instance has become the leader.

With this change, KIC will trigger an update of all VS/VSR statuses based on their last event when the KIC becomes the leader.

### DIFF
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -1400,6 +1400,73 @@ func (lbc *LoadBalancerController) emitEventForVirtualServers(eventType string, 
 	}
 }
 
+func getStatusFromEventTitle(eventTitle string) string {
+	switch eventTitle {
+	case "AddedOrUpdatedWithError", "Rejected", "NoVirtualServersFound", "Missing Secret", "UpdatedWithError":
+		return conf_v1.StateInvalid
+	case "AddedOrUpdatedWithWarning", "UpdatedWithWarning":
+		return conf_v1.StateWarning
+	case "AddedOrUpdated", "Updated":
+		return conf_v1.StateValid
+	default:
+		return "Unknown"
+	}
+}
+
+func (lbc *LoadBalancerController) updateVirtualServerVirtualServerRouteStatusesFromEvents() error {
+	vss := lbc.getVirtualServers()
+	var allErrs []error
+	for _, vs := range vss {
+		events, err := lbc.client.CoreV1().Events(vs.Namespace).List(context.TODO(),
+			meta_v1.ListOptions{FieldSelector: fmt.Sprintf("involvedObject.name=%v,involvedObject.uid=%v", vs.Name, vs.UID)})
+		if err != nil {
+			allErrs = append(allErrs, fmt.Errorf("error trying to get events for VirtualServer %v/%v: %v", vs.Namespace, vs.Name, err))
+			break
+		}
+
+		var timestamp time.Time
+		var latestEvent api_v1.Event
+		for _, event := range events.Items {
+			if event.CreationTimestamp.After(timestamp) {
+				latestEvent = event
+			}
+		}
+
+		err = lbc.statusUpdater.UpdateVirtualServerStatus(vs, getStatusFromEventTitle(latestEvent.Reason), latestEvent.Reason, latestEvent.Message)
+		if err != nil {
+			allErrs = append(allErrs, err)
+		}
+	}
+	vsrs := lbc.getVirtualServerRoutes()
+	for _, vsr := range vsrs {
+		events, err := lbc.client.CoreV1().Events(vsr.Namespace).List(context.TODO(),
+			meta_v1.ListOptions{FieldSelector: fmt.Sprintf("involvedObject.name=%v,involvedObject.uid=%v", vsr.Name, vsr.UID)})
+		if err != nil {
+			allErrs = append(allErrs, fmt.Errorf("error trying to get events for VirtualServerRoute %v/%v: %v", vsr.Namespace, vsr.Name, err))
+			break
+		}
+
+		var timestamp time.Time
+		var latestEvent api_v1.Event
+		for _, event := range events.Items {
+			if event.CreationTimestamp.After(timestamp) {
+				latestEvent = event
+			}
+		}
+
+		err = lbc.statusUpdater.UpdateVirtualServerRouteStatus(vsr, getStatusFromEventTitle(latestEvent.Reason), latestEvent.Reason, latestEvent.Message)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


internal/k8s/controller.go
```diff
@@ -1400,6 +1400,73 @@ func (lbc *LoadBalancerController) emitEventForVirtualServers(eventType string,
...
+	case "AddedOrUpdatedWithWarning", "UpdatedWithWarning":
+		return conf_v1.StateWarning
+	case "AddedOrUpdated", "Updated":
+		return conf_v1.StateValid
+	default:
+		return "Unknown"
+	}
+}
+
+func (lbc *LoadBalancerController) updateVirtualServerVirtualServerRouteStatusesFromEvents() error {
```
LorcanMcVeigh, line 1416: optional: add and `and` between VirtualServer and VirtualServerRoutes if it doesn't make the name too long

internal/k8s/controller.go
```diff
@@ -1400,6 +1400,73 @@ func (lbc *LoadBalancerController) emitEventForVirtualServers(eventType string,
...
+		events, err := lbc.client.CoreV1().Events(vs.Namespace).List(context.TODO(),
+			meta_v1.ListOptions{FieldSelector: fmt.Sprintf("involvedObject.name=%v,involvedObject.uid=%v", vs.Name, vs.UID)})
+		if err != nil {
+			allErrs = append(allErrs, fmt.Errorf("error trying to get events for VirtualServer %v/%v: %v", vs.Namespace, vs.Name, err))
+			break
+		}
+
+		var timestamp time.Time
+		var latestEvent api_v1.Event
+		for _, event := range events.Items {
```
pleshakov, line 1429: what if the list of events is `0`. in that case latestEvent will be an empty struct and `UpdateVirtualServerStatus` will update the resource with `Unknown` state. Is it intentional? Or should we skip the update in that case?

same for VSRs

Rulox: It was intentional as we wanted to have at least something in the `status` field ASAP, but I can skip the update for sure. Either works fine

internal/k8s/controller.go
```diff
@@ -1400,6 +1400,73 @@ func (lbc *LoadBalancerController) emitEventForVirtualServers(eventType string,
...
+func getStatusFromEventTitle(eventTitle string) string {
+	switch eventTitle {
+	case "AddedOrUpdatedWithError", "Rejected", "NoVirtualServersFound", "Missing Secret", "UpdatedWithError":
+		return conf_v1.StateInvalid
+	case "AddedOrUpdatedWithWarning", "UpdatedWithWarning":
+		return conf_v1.StateWarning
+	case "AddedOrUpdated", "Updated":
+		return conf_v1.StateValid
+	default:
+		return "Unknown"
```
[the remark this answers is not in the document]
Rulox: Ok I'll skip update in this case.

internal/k8s/controller.go
```diff
@@ -1400,6 +1400,73 @@ func (lbc *LoadBalancerController) emitEventForVirtualServers(eventType string,
...
+	}
+}
+
+func (lbc *LoadBalancerController) updateVirtualServerVirtualServerRouteStatusesFromEvents() error {
+	vss := lbc.getVirtualServers()
```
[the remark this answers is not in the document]
Rulox: Good catch!

Rulox:
I've implemented your suggestions @pleshakov @LorcanMcVeigh let me know your thoughts, thanks

internal/k8s/controller.go
```diff
@@ -1408,22 +1408,25 @@ func getStatusFromEventTitle(eventTitle string) string {
...
+
 		events, err := lbc.client.CoreV1().Events(vs.Namespace).List(context.TODO(),
 			meta_v1.ListOptions{FieldSelector: fmt.Sprintf("involvedObject.name=%v,involvedObject.uid=%v", vs.Name, vs.UID)})
 		if err != nil {
 			allErrs = append(allErrs, fmt.Errorf("error trying to get events for VirtualServer %v/%v: %v", vs.Namespace, vs.Name, err))
 			break
 		}
 
+		if len(events.Items) == 0 {
+			break
```
pleshakov, line 1427: break will stop processing events for all resources if one has zero events. shall we use continue here?

internal/k8s/controller.go
```diff
@@ -1408,22 +1408,25 @@ func getStatusFromEventTitle(eventTitle string) string {
...
-		return "Unknown"
 	}
 }
 
-func (lbc *LoadBalancerController) updateVirtualServerVirtualServerRouteStatusesFromEvents() error {
-	vss := lbc.getVirtualServers()
+func (lbc *LoadBalancerController) updateVirtualServersStatusFromEvents() error {
 	var allErrs []error
-	for _, vs := range vss {
+	for _, obj := range lbc.virtualServerLister.List() {
```
pleshakov, line 1416: @lucacome the correct class will have to be handled  for this case in your PR as well https://github.com/nginxinc/kubernetes-ingress/pulls . Also for VirtualServerRoutes

internal/k8s/controller.go
```diff
@@ -1400,6 +1400,86 @@ func (lbc *LoadBalancerController) emitEventForVirtualServers(eventType string,
...
+
+		err = lbc.statusUpdater.UpdateVirtualServerStatus(vs, getStatusFromEventTitle(latestEvent.Reason), latestEvent.Reason, latestEvent.Message)
+		if err != nil {
+			allErrs = append(allErrs, err)
+		}
+	}
+	return nil
+}
+
+func (lbc *LoadBalancerController) updateVirtualServerRoutessStatusFromEvents() error {
```
pleshakov, line 1446:
```suggestion
func (lbc *LoadBalancerController) updateVirtualServerRoutesStatusFromEvents() error {
```

internal/k8s/controller.go
Rulox: Continue yeah, my mistake.

internal/k8s/controller.go
```diff
@@ -219,7 +219,7 @@ func NewLoadBalancerController(input NewLoadBalancerControllerInput) *LoadBalanc
 		}
 	}
 
-	if input.ReportIngressStatus && input.IsLeaderElectionEnabled {
+	if input.IsLeaderElectionEnabled {
```
Rulox, line 222: @pleshakov note this change, in case you think this can cause any issue

pleshakov: good catch 👍 seems like we had a bug

PR #1003, opened 2020-06-23
Add support for custom annotations in configmaps

RubyLangdon:
### Proposed changes
This allows annotations to be added to the configmaps in the helm chart.

One case where this is useful is when Spinnaker is used for deploying the chart.
Annotations are needed to control Spinnaker's behaviour when it deploys configmaps. For example, the annotation `strategy.spinnaker.io/versioned` controls Spinnaker's versioning behaviour.

### Checklist


Rulox:
Thanks @RubyLangdon we'll have a look and get back to you soon.

deployments/helm-chart/values.yaml
```diff
@@ -220,6 +223,9 @@ controller:
     ## Autogenerated if not set or set to "".
     # leaderElectionLockName: "nginx-ingress-leader-election"
 
+    ## The annotations of the leader election configmap
```
Dean-Coakley, line 226: Tiny nitpick, (for consistency with other doc strings):
```suggestion
    ## The annotations of the leader election configmap.
```

deployments/helm-chart/templates/controller-leader-election-configmap.yaml
```diff
@@ -5,3 +5,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "nginx-ingress.labels" . | nindent 4 }}
+{{- if .Values.controller.reportIngressStatus.annotations }}
```
pleshakov, line 8: Hi @RubyLangdon 

Just double checking that the spinnaker annotations are needed to be configured for the leader election configmap. This config map is used internally by the Ingress Controller for leader election to work. The IC never changes its `data` field, it only sets the annotation `control-plane.alpha.kubernetes.io/leader:`

RubyLangdon: In some situations I see spinnaker append increasing version numbers to configmaps when they haven't changed since the last deployment. I'm not sure why but I'd prefer to be able to stop versioning to be safe. Also, even if that doesn't happen the configmaps always get `v000` appended to them the first time they are deployed, which we don't necessarily want.

PR #1006, opened 2020-06-23
Add request/response modifiers to VS and VSR

Rulox:
### Proposed changes
This PR includes the ability to rewrite the URI and modify request/response headers for backend services configured using VirtualServer or VirtualServerRoute custom resources by adding a new action to the routes and subroutes `Proxy`.


internal/configs/virtualserver_test.go
```diff
@@ -3834,3 +3870,378 @@ func TestIsTLSEnabled(t *testing.T) {
...
+			path:     "/_internal_path",
+			internal: true,
+			proxy: &conf_v1.ActionProxy{
+				RewritePath: "/rewrite",
+			},
+			originalPath: "/path",
+			expected:     []string{`^ $request_uri`, `"^/path(.*)$" "/rewrite$1" break`},
+		},
+		{
+			path:     "/_internal_path",
```
pleshakov, line 3909: is it a duplicate of the one above?

internal/configs/virtualserver_test.go
```diff
@@ -3834,3 +3870,378 @@ func TestIsTLSEnabled(t *testing.T) {
...
+			originalPath: "/path",
+			expected:     []string{`^ $request_uri`, `"^/path(.*)$" "/rewrite$1" break`},
+		},
+		{
+			path:     "~/regex",
+			internal: false,
+			proxy: &conf_v1.ActionProxy{
+				RewritePath: "/rewrite",
+			},
+			expected: []string{`"^/regex$" "/rewrite" break`},
```
pleshakov, line 3932: this might be a bug in the `generateRewrites` . why are we appending `$`? 

Consider the following config:
```nginx
    location ~/test {
        rewrite ^/test$ "/rewrite" break;
        proxy_pass http://test;
    }
```
In this case:
* `/test` -> `/rewrite` - OK
* `/test/abc` -> `/test/abc` :( 

If we step back and go to CRDs, in this case it is a responsibility of the user that the path and rewrite work together. So they should be something like  `~ /tea/?(.*)` and `/$1` (from the example)

internal/configs/virtualserver_test.go
```diff
@@ -3834,3 +3870,378 @@ func TestIsTLSEnabled(t *testing.T) {
...
+		expected string
+	}{
+		{
+			proxy:    nil,
+			expected: "",
+		},
+		{
+			proxy: &conf_v1.ActionProxy{
+				ResponseHeaders: nil,
+			},
```
pleshakov, line 4168: would it be better to make `expected: ""` as in the case above?

internal/configs/virtualserver_test.go
Rulox: Yes you're right, I was adding a `$` in the regex case too by mistake, as you say, this should be responsibility of the user. I'll sync up with @vepatel to make sure we cover all the use cases in the tests. Thanks for the heads up.

Issue #967, opened 2020-05-24
Helm 3 multiple ingress controller install results in CRD conflict


nick-cromwell:
**Describe the bug**
Trying to install two Ingress Controllers via Helm 3 results in the following error:

`Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition "virtualservers.k8s.nginx.org" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "test-internal-ingress": current value is "test-ingress"`

**To Reproduce**
1. Start with a fresh cluster
2. Install an Ingress Controller and attempt to skip CRDs:
`helm install test-ingress nginx-stable/nginx-ingress --skip-crds`
3. Attempt to install a second Ingress Controller:
`helm install test-internal-ingress nginx-stable/nginx-ingress --skip-crds`
3. See error

**Expected behavior**
Two ingress controllers are installed with different release names

**Your environment**
* 1.16.8-gke.15
* GKE
* Using NGINX
* Helm 3.2.1

I believe I have worked around this by deleting the CRs between installs like so:
`kubectl delete crd transportservers.k8s.nginx.org`
`kubectl delete crd virtualservers.k8s.nginx.org`
`kubectl delete crd virtualserverroutes.k8s.nginx.org`

They were created again after installing the second ingress controller.

Dean-Coakley:
Hi @nick-cromwell 

Thanks for the bug report. This will be fixed soon.
This is a duplicate issue of https://github.com/nginxinc/kubernetes-ingress/issues/950.

Sorry for the inconvenience

Issue #950, opened 2020-05-15
[Helm] Trying to install a second release


katlimruiz:
**Describe the bug**
I have a cluster where I want to setup many different websites, each with its own nginx ingress.
I already have one that is running correctly. I used this command:
```
helm install myapp1 nginx-stable/nginx-ingress --namespace myapp1 --set controller.ingressClass=nginx-myapp1
```
which the ingressClass matches with an `ingress` resource. This runs fine.

So I want to add another website. I want to use this command:
```
helm install myapp2 nginx-stable/nginx-ingress --namespace myapp2 --set controller.ingressClass=nginx-myapp2
```
However, this gives me the following error:
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: namespace: , name: transportservers.k8s.nginx.org, existing_kind: apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition, new_kind: apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition
```
**Expected behavior**
The second installation should be inserted correctly. Since it is under a different namespace and a different release name.

**Your environment**
* Version of the Ingress Controller = NGINX Ingress Controller 1.7.0
* Version of Kubernetes = 1.15
* Kubernetes platform (e.g. Mini-kube or GCP) = Azure
* Using NGINX or NGINX Plus = NGINX

Dean-Coakley:
@katlimruiz Thanks for the bug report. We will fix it!

katlimruiz:
cool, I thought I was doing something wrong 👍 

when is the next release?

Dean-Coakley:
Next release is planned for June/July.

The fix should be in master shortly. Which leaves you with some options for work arounds in the meantime:
* Download the chart locally and do similar edits
* Move to using the `nginx-edge/nginx-ingress` chart. **Note that this is not recommended for production use.**
* Do `helm template [chart]` and edit the output to no longer have the CRDs.

Sorry for the hassle

Dean-Coakley:
Additionally, If you are not using our CRDs(if you are using ingress resources only) the following workaround should work:
* Simply set `controller.enableCustomResources` to `false`.

katlimruiz:
How could I know if I'm using the CRDS or not? Sorry I'm still new in kubernetes.

Dean-Coakley:
Ah, no problem.

[...]

PR #1009, opened 2020-06-23
Fix multiple KIC deploys via helm

Dean-Coakley:
### Proposed changes
Fixes: https://github.com/nginxinc/kubernetes-ingress/issues/967
Fixes: https://github.com/nginxinc/kubernetes-ingress/issues/950

Allows multiple installs of the Ingress Controller to be installed in the same cluster.

### Behaviour changes
* CustomResourceDefinitions are no longer managed by helm releases. Deleting a release will leave CRDs behind. - In line with helm chart best practices. - https://helm.sh/docs/chart_best_practices/custom_resource_definitions/
* CustomResourceDefinitions are installed unconditionally. Unless `--skip-crds` is provided to `helm install`.

#### Upgrade path with `controller.enableCustomResources=true`
* > _Given a deployed `1.7.x` IC_
* `helm upgrade my-release deployments/helm-chart`
  > **_Note: Existing CRDs and CRs will be deleted. You will also see listers temporarily fail until CRDs are recreated._**
* Recreate CRDs
  * `kubectl apply -f deployments/helm-chart/crds`
* Recreate CRs.

#### Upgrade path with `controller.enableCustomResources=false`
* > _Given a deployed `1.7.x` IC_
* `helm upgrade my-release deployments/helm-chart`

### Checklist

### DIFF
--- a/deployments/helm-chart/crds/transportserver.yaml
+++ b/deployments/helm-chart/crds/transportserver.yaml
@@ -1,10 +1,11 @@
-{{- if .Values.controller.enableCustomResources }}
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: transportservers.k8s.nginx.org
   labels:
-    {{- include "nginx-ingress.labels" . | nindent 4 }}
+    app.kubernetes.io/name: "nginx-ingress"
+  annotations:
+    "helm.sh/hook": crd-install
 spec:
   group: k8s.nginx.org
   versions:
@@ -76,4 +77,3 @@ spec:
                     type: integer
                   service:
                     type: string
-{{- end }}
--- a/deployments/helm-chart/README.md
+++ b/deployments/helm-chart/README.md
@@ -15,19 +15,21 @@ This chart deploys the NGINX Ingress controller in your Kubernetes cluster.
 
 ## Installing the Chart
 
-### Installing CRDs
+### Installing the CRDs
 
-The VirtualServer and VirtualServerRoute resources are new load balancing configuration, introduced in release 1.5 as an alternative to the Ingress resource. The resources enable use cases not supported with the Ingress resource, such as [traffic splitting](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples-of-custom-resources/traffic-splitting) and [advanced content-based routing](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples-of-custom-resources/advanced-routing).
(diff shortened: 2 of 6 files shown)


Rulox:
Looks good. I am a little bit worried by:
`Note: Existing CRDs and CRs will be deleted. You will also see listers temporarily fail until CRDs are recreated.`

I guess we'll have to document this very carefully on the new Release Notes.

deployments/helm-chart/crds/transportserver.yaml
```diff
@@ -1,10 +1,11 @@
-{{- if .Values.controller.enableCustomResources }}
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: transportservers.k8s.nginx.org
   labels:
-    {{- include "nginx-ingress.labels" . | nindent 4 }}
+    app.kubernetes.io/name: "nginx-ingress"
```
Rulox, line 6: I guess it doesn't make sense that this is somehow a parameter right?

Dean-Coakley: Well from my reading of: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/ I understand that templating in `/crds` is not permitted.
> "These CRDs are not templated, but will be installed by default when running a helm install for the chart. If the CRD already exists, it will be skipped with a warning."


Additionally, `nginx-ingress.labels` includes release specific labels. I don't believe these make much sense anymore. Helm does no life-cycle changes past initial install. 
Upgrades and deletion of CRDs must be done via `kubectl`.

deployments/helm-chart/crds/transportserver.yaml
```diff
@@ -1,10 +1,11 @@
...
+  annotations:
+    "helm.sh/hook": crd-install
```
pleshakov, line 8: are there any differences in how to install CRDs for Helm 2 vs Helm 3? Or it will work the same - just this annotation is needed for helm 2.

deployments/helm-chart/crds/transportserver.yaml
Rulox: Thanks! makes sense

deployments/helm-chart/crds/transportserver.yaml
Dean-Coakley: * Helm3 will install crds unless you set `helm install --skip-crds`.

* Helm2 will install crds unless you set `controller.enableCustomResources=false`

Aside from that the behaviour is the same. CRDs are installed before all other resources, are left behind after a release is removed for safety and must be upgraded via `kubectl`.

PR #1030, opened 2020-07-02
Fix port range

lucacome:
The lower end of the port range should be `1024`.

### DIFF
--- a/cmd/nginx-ingress/main_test.go
+++ b/cmd/nginx-ingress/main_test.go
@@ -7,7 +7,7 @@ import (
 )
 
 func TestValidatePort(t *testing.T) {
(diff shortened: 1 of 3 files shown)


cmd/nginx-ingress/main_test.go
```diff
@@ -7,15 +7,15 @@ import (
 )
 
 func TestValidatePort(t *testing.T) {
-	badPorts := []int{80, 443, 1, 1022, 65536}
```
LorcanMcVeigh, line 10: did this one really need to change?

lucacome: No, just wanted it to check the first invalid port

PR #1044, opened 2020-07-08
Add prometheus worker process metric

LorcanMcVeigh:
### Proposed changes
Add prometheus metric that gives the number of worker processes running  

### Checklist


internal/k8s/controller.go
```diff
@@ -837,6 +848,83 @@ func (lbc *LoadBalancerController) sync(task task) {
...
+	grepCmd.Stdin = reader
+
+	secReader, secWriter := io.Pipe()
+	grepCmd.Stdout = secWriter
+	awkCmd.Stdin = secReader
+
+	var buff bytes.Buffer
+	awkCmd.Stdout = &buff
+
+	lsCmd.Start()
```
Rulox, line 874: All these commands (Start, Wait, Close) can throw an error, not handling the errors is not a good idea. This means, this will end up in a huge method with a lot of `if err != nil`....

Apart form that, I find a little bit difficult to follow the fact that this is all in parts/lines. I am thinking of an approach where you can write all of this using Go code, for example, for the first part of getting PIDs, something like this:

```golang

        files, err := ioutil.ReadDir("/proc")
        if err != nil {
                fmt.Println(fmt.Sprintf("error %v", err))
        }

        for _, f := range files {
                u, _ := user.LookupId(fmt.Sprint(f.Sys().(*syscall.Stat_t).Uid))
                if u.Name == "nginx user" {
                        // store pids of the processes
                        fmt.Println(u.Name)
                }
        }
```

And for the second part, follow the same approach of trying to find the Go code that does what you want instead of having to call shell 

What do you think?

internal/nginx/manager.go
```diff
@@ -87,7 +86,7 @@ type LocalManager struct {
 	dhparamFilename              string
 	tlsPassthroughHostsFilename  string
 	verifyConfigGenerator        *verifyConfigGenerator
-	verifyClient                 *verifyClient
+	verifyClient                 *VerifyClient
```
Rulox, line 89: This won't be necessary if we use the Manager collector, another reason to use it.

internal/nginx/manager.go
```diff
@@ -289,6 +289,7 @@ func (lm *LocalManager) Reload() error {
 	}
 
 	lm.metricsCollector.IncNginxReloadCount()
+	lm.metricsCollector.UpdateWorkerProcessCount("new")
```
Rulox, line 292: Sorry, I told you it was `new` but it should be `curent` 🙏 
```suggestion
	lm.metricsCollector.UpdateWorkerProcessCount("current")
```

internal/nginx/manager.go
```diff
@@ -271,6 +270,7 @@ func (lm *LocalManager) Start(done chan error) {
 // Reload reloads NGINX.
 func (lm *LocalManager) Reload() error {
 	// write a new config version
+	lm.metricsCollector.UpdateWorkerProcessCount("old")
```
Rulox, line 273: What about storing the metric float in the collector struct
```
type LocalManagerMetricsCollector struct {
	// Metrics
	reloadsTotal     prometheus.Counter
	reloadsError     prometheus.Counter
	lastReloadStatus prometheus.Gauge
	lastReloadTime   prometheus.Gauge
	processTotal     *prometheus.GaugeVec
         workerProcessesValue float64
} 
```

This way, we don't need to call this twice, but we can always manually set the `old` at the beginning of the `UpdateWorkerProcessCount` method, then calculate the new one, and update the `workerProcessesValue` of the struct with the new value.

This way we avoid calling this twice. The reason is that I/O operations like opening files are not efficient, so if we can reduce them that'll be nice.

Let me know what you think

internal/nginx/manager.go
```diff
@@ -117,7 +116,7 @@ func NewLocalManager(confPath string, binaryFilename string, mc collectors.Manag
 		binaryFilename:              binaryFilename,
 		verifyConfigGenerator:       verifyConfigGenerator,
 		configVersion:               0,
-		verifyClient:                newVerifyClient(timeout),
+		verifyClient:                NewVerifyClient(timeout),
```
Rulox, line 119: I believe we don't need to be exportable anymore?

internal/nginx/verify.go
```diff
@@ -14,14 +14,14 @@ import (
 	"github.com/golang/glog"
 )
 
-// verifyClient is a client for verifying the config version.
+// VerifyClient is a client for verifying the config version.
```
Rulox, line 17:
```suggestion
// verifyClient is a client for verifying the config version.
```

internal/metrics/collectors/manager.go
```diff
@@ -21,10 +30,12 @@ type LocalManagerMetricsCollector struct {
 	reloadsError     prometheus.Counter
 	lastReloadStatus prometheus.Gauge
 	lastReloadTime   prometheus.Gauge
+	processTotal     *prometheus.GaugeVec
 }
 
 // NewLocalManagerMetricsCollector creates a new LocalManagerMetricsCollector
 func NewLocalManagerMetricsCollector(constLabels map[string]string) *LocalManagerMetricsCollector {
+	labelNames := []string{"generations"}
```
Rulox, line 38: Per definition of the task:
```suggestion
	labelNames := []string{"generation"}
```

internal/metrics/collectors/manager.go
```diff
@@ -88,12 +109,54 @@ func (nc *LocalManagerMetricsCollector) UpdateLastReloadTime(duration time.Durat
...
+		glog.Errorf("error %v", err)
+	}
+	processes := []string{}
+	for _, f := range procFolders {
+		u, err := user.LookupId(fmt.Sprint(f.Sys().(*syscall.Stat_t).Uid))
+		if err != nil {
+			glog.Errorf("Error %v", err)
+		}
+		if u.Name == "nginx user" {
+			// store pids of the processes
```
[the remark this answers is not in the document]
LorcanMcVeigh: the name of folders here are the PIDs, here they are the same thing but i forgot to remove this comment

internal/metrics/collectors/manager.go
```diff
@@ -88,12 +109,54 @@ func (nc *LocalManagerMetricsCollector) UpdateLastReloadTime(duration time.Durat
...
+	workerProcesses := 0
+	procFolders, err := ioutil.ReadDir("/proc")
+	if err != nil {
+		glog.Errorf("error %v", err)
+	}
+	processes := []string{}
+	for _, f := range procFolders {
+		u, err := user.LookupId(fmt.Sprint(f.Sys().(*syscall.Stat_t).Uid))
+		if err != nil {
+			glog.Errorf("Error %v", err)
```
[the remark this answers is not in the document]
Dean-Coakley: I believe both returns are required.

Ref:
https://golang.org/pkg/os/user/#LookupId
https://golang.org/src/os/user/lookup.go

Agree about encapsulation too, but I think it's an optional/preference thing.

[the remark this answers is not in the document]
Dean-Coakley: More meaningful error context should be added instead of `Error: %v` too.

internal/metrics/collectors/manager.go
```diff
@@ -88,12 +110,64 @@ func (nc *LocalManagerMetricsCollector) UpdateLastReloadTime(duration time.Durat
...
+		scanner := bufio.NewScanner(f)
+
+		for scanner.Scan() {
+			if strings.Contains(scanner.Text(), "PPid") {
+				words := strings.Split(scanner.Text(), "\t")
+				ppid := words[len(words)-1]
+
+				if ppid != "0" {
+					workerProcesses++
+				}
```
Dean-Coakley, line 157: This captures too many processes. This will capture any process created by the nginx user. This includes `ls` for `/proc` etc. Which leads to wrong results reported in the metric.

Instead you must filter by `ppid == masterProcessPid`

Unfortunately I can't advise on how to get `masterPid`.

internal/metrics/collectors/manager.go
```diff
@@ -58,6 +70,16 @@ func NewLocalManagerMetricsCollector(constLabels map[string]string) *LocalManage
 				ConstLabels: constLabels,
 			},
 		),
+
```
Dean-Coakley, line 73:
```suggestion
```

internal/metrics/collectors/manager.go
```diff
@@ -88,12 +110,64 @@ func (nc *LocalManagerMetricsCollector) UpdateLastReloadTime(duration time.Durat
...
+		statusFile := fmt.Sprintf("/proc/%v/status", file)
+		f, err := os.Open(statusFile)
+		if err != nil {
+			return 0, fmt.Errorf("Error in opening file %v", statusFile)
+		}
+
+		scanner := bufio.NewScanner(f)
+
+		for scanner.Scan() {
+			if strings.Contains(scanner.Text(), "PPid") {
```
Dean-Coakley, line 151: Nitpick: I think the follow would be clearer due to being more specific and I think it should work the same:
```suggestion
			if strings.HasPrefix(scanner.Text(), "PPid") {
```

internal/metrics/collectors/manager.go
Dean-Coakley: `masterProcessPid` should be able to be read from
`/var/lib/nginx/nginx.pid`

internal/metrics/collectors/manager.go
```diff
@@ -88,12 +110,64 @@ func (nc *LocalManagerMetricsCollector) UpdateLastReloadTime(duration time.Durat
...
+				words := strings.Split(scanner.Text(), "\t")
+				ppid := words[len(words)-1]
+
+				if ppid != "0" {
+					workerProcesses++
+				}
+			}
+		}
```
Dean-Coakley, line 159: Afaik you should handle scanner errors for non EOF errors:
```go
for scanner.Scan {
...
}
if err := scanner.Err(); err != nil {
	// return 0, fmt.Errorf...
}

File: internal/metrics/collectors/manager.go [hunk 14]
```diff
@@ -88,12 +110,64 @@ func (nc *LocalManagerMetricsCollector) UpdateLastReloadTime(duration time.Durat
...
+		if err != nil {
+			return 0, fmt.Errorf("Error could not find user for file %v : %v", f.Name(), err)
+		}
+		if u.Name == "nginx user" {
+			processes = append(processes, f.Name())
+		}
+	}
+	for _, file := range processes {
+		statusFile := fmt.Sprintf("/proc/%v/status", file)
+		f, err := os.Open(statusFile)
```
Dean-Coakley, line 143: Should this file should be closed at some point?
`f.Close()`

internal/metrics/collectors/manager.go
```diff
@@ -88,12 +109,76 @@ func (nc *LocalManagerMetricsCollector) UpdateLastReloadTime(duration time.Durat
...
+	}
+	defer masterPidFile.Close()
+
+	var masterPid string
+	masterPidScanner := bufio.NewScanner(masterPidFile)
+	for masterPidScanner.Scan() {
+		masterPid = masterPidScanner.Text()
+	}
+	if err := masterPidScanner.Err(); err != nil {
+		return 0, fmt.Errorf("could not read /var/lib/nginx/ngin.pid : %v", err)
```
Dean-Coakley, line 142:
```suggestion
		return 0, fmt.Errorf("could not read /var/lib/nginx/nginx.pid : %v", err)
```

(Or create `nginxPidPath`)

internal/metrics/collectors/manager.go
```diff
@@ -88,12 +109,76 @@ func (nc *LocalManagerMetricsCollector) UpdateLastReloadTime(duration time.Durat
...
+			defer f.Close()
+
+			scanner := bufio.NewScanner(f)
+
+			for scanner.Scan() {
+				if strings.HasPrefix(scanner.Text(), "PPid") {
+					ppidLineSplit := strings.Split(scanner.Text(), "\t")
+					ppid := ppidLineSplit[len(ppidLineSplit)-1]
+
+					if ppid != masterPid {
```
Dean-Coakley, line 165: In the case of a worker process: The `ppid` of the process should **match** the pid of the master process.
```suggestion
					if ppid == masterPid {
```

PR #1080, opened 2020-07-29
Implement ingress changes

LorcanMcVeigh:
### Proposed changes
Implement v1 k8s 1.18 ingress changes 
- update API in codebase 
- add exact path matching & document exact, prefix and implementationSpecific path matching 
- wildcard host domain names 

### Checklist

### DIFF
--- a/deployments/helm-chart/templates/rbac.yaml
+++ b/deployments/helm-chart/templates/rbac.yaml
@@ -71,7 +71,7 @@ rules:
   - watch
 {{- if .Values.controller.reportIngressStatus.enable }}
 - apiGroups:
-  - extensions
+  - networking.k8s.io
   resources:
   - ingresses/status
   verbs:
--- a/deployments/helm-chart/templates/rbac.yaml
+++ b/deployments/helm-chart/templates/rbac.yaml
@@ -62,7 +62,7 @@ rules:
   - patch
   - list
 - apiGroups:
-  - extensions
+  - networking.k8s.io
   resources:
   - ingresses
   verbs:
--- a/deployments/rbac/rbac.yaml
+++ b/deployments/rbac/rbac.yaml
@@ -4,7 +4,7 @@ metadata:
   name: nginx-ingress
 rules:
 - apiGroups:
-  - ""
+  - "networking.k8s.io"
   resources:
   - services
   - endpoints
@@ -46,7 +46,7 @@ rules:
   - patch
   - list
 - apiGroups:
-  - extensions
+  - "networking.k8s.io"
   resources:
   - ingresses
   verbs:
@@ -54,7 +54,7 @@ rules:
   - watch
   - get
 - apiGroups:
-  - "extensions"
+  - "networking.k8s.io"
   resources:
   - ingresses/status
   verbs:
--- a/deployments/rbac/rbac.yaml
+++ b/deployments/rbac/rbac.yaml
@@ -4,7 +4,7 @@ metadata:
   name: nginx-ingress
 rules:
 - apiGroups:
-  - "networking.k8s.io"
+  - ""
   resources:
   - services
   - endpoints
@@ -46,7 +46,7 @@ rules:
   - patch
   - list
 - apiGroups:
-  - "networking.k8s.io"
+  - networking.k8s.io
   resources:
   - ingresses
   verbs:
@@ -54,7 +54,7 @@ rules:
   - watch
   - get
 - apiGroups:
-  - "networking.k8s.io"
+  - networking.k8s.io
   resources:
   - ingresses/status
   verbs:
--- a/docs-web/configuration/ingress-resources/basic-configuration.md
+++ b/docs-web/configuration/ingress-resources/basic-configuration.md
@@ -2,7 +2,7 @@
 
 The example below shows a basic Ingress resource definition. It load balances requests for two services -- coffee and tea -- comprising a hypothetical *cafe* app hosted at `cafe.example.com`:
 ```yaml
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: cafe-ingress
@@ -15,11 +15,18 @@ spec:
   - host: cafe.example.com
     http:
       paths:
-      - path: /tea
+      - path: "/tea"
+        pathType: Prefix
+        backend:
+          serviceName: tea-svc
+          servicePort: 80
+      - path: "/tea/green"
+        pathType: Exact
         backend:
           serviceName: tea-svc
           servicePort: 80
       - path: /coffee
+        pathType: ImplementationSpecific # default
         backend:
           serviceName: coffee-svc
           servicePort: 80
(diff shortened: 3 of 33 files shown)


lucacome:
Should we change all the other examples in the docs that have `extensions/v1beta1`?

docs-web/configuration/ingress-resources/basic-configuration.md
```diff
@@ -30,10 +37,11 @@ Here is a breakdown of what this Ingress resource definition means:
...
 * In the `paths` field, we define two path‑based rules:
-  * The rule with the path `/tea` instructs NGINX to distribute the requests with the `/tea` URI among the pods of the *tea* service, which is deployed with the name `tea‑svc` in the cluster.
-  * The rule with the path `/coffee` instructs NGINX to distribute the requests with the `/coffee` URI among the pods of the *coffee* service, which is deployed with the name `coffee‑svc` in the cluster.
+  * The rule with the path `/tea` instructs NGINX to distribute the requests with the prefix matching `/tea` URI among the pods of the *tea* service, which is deployed with the name `tea‑svc` in the cluster.
+  * The rule with the path `/tea/green` instructs NGINX to distribute the requests with the exact matching `/tea/green` URI among the pods of the *tea* service, which is deployed with the name `tea-svc` in the cluster.
+  * The rule with the path `/coffee` instructs NGINX to distribute the requests with the implementation specific matching `/coffee` URI among the pods of the *coffee* service, which is deployed with the name `coffee‑svc` in the cluster. Implementation specific matching is the default prefix matching.
```
pleshakov, line 44: Because those are new features, but we still continue to support older Kubernetes versions, let's document them in a separate section. This way people will see that the feature is only available since k8s 1.18.

What do you think about the following section? We can put it after before the Restrictions section. I used your example, but removed unnecessary `""`.

## New Features Available in Kubernetes 1.18 and Above

Starting from Kubernetes 1.18, you can use the following new features:
* The host field supports wildcard domain names, such as `*.example.com`.
* The path supports different matching rules with the new field `PathType`, which takes the following values: `Prefix` for prefix-based matching, `Exact` for exact matching and `ImplementationSpecific`, which is the default type and is the same as  `Prefix`. For example:
  ```yaml
     - path: /tea
        pathType: Prefix
        backend:
          serviceName: tea-svc
          servicePort: 80
      - path: /tea/green
        pathType: Exact
        backend:
          serviceName: tea-svc
          servicePort: 80
     - path: /coffee
        pathType: ImplementationSpecific # default
        backend:
          serviceName: coffee-svc
          servicePort: 80
   ```

internal/configs/ingress.go
```diff
@@ -181,8 +181,14 @@ func generateNginxCfg(ingEx *IngressEx, pems map[string]string, apResources map[
 
 			ssl := sslServices[path.Backend.ServiceName] || staticParams.SpiffeCerts
 			proxySSLName := generateProxySSLName(path.Backend.ServiceName, ingEx.Ingress.Namespace)
+			var pathType extensions.PathType
```
pleshakov, line 184: since extensions/v1beta1 is depreciated, shall we start using `networking.k8s.io/v1beta1` (k8s.io/api/networking/v1beta1) in the code as well?

internal/configs/ingress.go
```diff
@@ -181,8 +181,14 @@ func generateNginxCfg(ingEx *IngressEx, pems map[string]string, apResources map[
...
+			if path.PathType == nil {
```
pleshakov, line 185: I suggest placing all the path related logic into a singe function ```func generateIngressPath(path, pathType) string```, which takes a path and pathType (could be nil) and generate the appropriate path we can use in a location.  The function can have a unit test, similar to `TestGeneratePath` in  internal/configs/virtualserver_test.go:4481 

You can use this function in `createLocation`.

It will be similar to `generatePath` in internal/configs/virtualserver.go:752

internal/configs/ingress.go
lucacome: I'm doing that part

internal/configs/ingress.go
LorcanMcVeigh: some of the logic will have to be before the `generateLocation` function is called because `Pathtype` is a pointer and passing that as a parameter ( if its nil ) will cause a `nil pointer dereference` error

deployments/rbac/rbac.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: nginx-ingress
 rules:
 - apiGroups:
-  - ""
+  - "networking.k8s.io"
```
pleshakov, line 7: this is a bug - the rule for services and endpoints is broken now

deployments/rbac/rbac.yaml
```diff
@@ -46,15 +46,15 @@ rules:
   - patch
   - list
 - apiGroups:
-  - extensions
+  - "networking.k8s.io"
```
pleshakov, line 49: no need to use `""` here and in the other changes

deployments/helm-chart/templates/rbac.yaml
```diff
@@ -62,7 +62,7 @@ rules:
   - patch
   - list
 - apiGroups:
-  - extensions
+  - networking.k8s.io
```
pleshakov, line 65: the changes to the rbac in Helm must be the same as the ones in deployments/rbac/rbac.yaml

docs-web/configuration/ingress-resources/basic-configuration.md
```diff
@@ -15,7 +15,7 @@ spec:
   - host: cafe.example.com
     http:
       paths:
-      - path: /tea
+      - path: "/tea"
```
pleshakov, line 18: no need for `""`

docs-web/configuration/ingress-resources/basic-configuration.md
```diff
@@ -32,14 +32,38 @@ Here is a breakdown of what this Ingress resource definition means:
...
 
 > For complete instructions on deploying the Ingress and Secret resources in the cluster, see the [complete-example](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/complete-example) in our GitHub repo.
 
 > To learn more about the Ingress resource, see the [Ingress resource documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/) in the Kubernetes docs.
 
+## New Features Available in Kubernetes 1.18 and Above
+
+Starting from Kubernetes 1.18, you can use the following new features:
+
+* The host field supports wildcard domain names, such as *.example.com.
```
pleshakov, line 47: could you possibly add the monospace styling from the original suggestion?
image

docs-web/configuration/ingress-resources/basic-configuration.md
```diff
@@ -32,14 +32,38 @@ Here is a breakdown of what this Ingress resource definition means:
...
+* The path supports different matching rules with the new field PathType, which takes the following values: Prefix for prefix-based matching, Exact for exact matching and ImplementationSpecific, which is the default type and is the same as Prefix. For example:
+```yaml
```
pleshakov, line 49: this snippet belongs to the second list element. It is needed to be indented properly. It should look like this:
image

docs-web/configuration/ingress-resources/basic-configuration.md
```diff
@@ -32,14 +32,38 @@ Here is a breakdown of what this Ingress resource definition means:
...
+   - path: /tea
```
pleshakov, line 50: please fix indentation in this excerpt here. For example, the second `- path` has more leading white space than the first and the third. In my original suggestion, it wasn't indented properly (I should've made a note for you, sry!). Please reindent.

As an ultimate test, double check that if somebody copy and paste that excerpt, they will not get an error! That would be annoying for our users.

In general, every excerpt we document should be validated.

internal/configs/ingress.go
pleshakov: @LorcanMcVeigh 
> some of the logic will have to be before the generateLocation function is called because Pathtype is a pointer and passing that as a parameter ( if its nil ) will cause a nil pointer dereference error

it will not cause a a nil pointer dereference error use `*networking.PathType` as a parameter and if you check for nil in `generateIngressPath`:
```
func generateIngressPath(path string, pathType *networking.PathType) string {
	if pathType == nil {
           . . .
        }
       . . .
```

Or perhaps I'm missing something? 🤔

internal/configs/ingress.go
```diff
@@ -250,9 +257,18 @@ func generateNginxCfg(ingEx *IngressEx, pems map[string]string, apResources map[
...
+	}
+
+	return path
+}
+
+func createLocation(path string, upstream version1.Upstream, cfg *ConfigParams, websocket bool, rewrite string, ssl bool, grpc bool, proxySSLName string, pathType *networking.PathType) version1.Location {
+	newPath := generateIngressPath(path, *pathType)
 	loc := version1.Location{
-		Path:                 path,
+		Path:                 newPath,
```
pleshakov, line 271: style: better to inline `newPath`:
```
Path: generateIngressPath(...)
```

internal/configs/ingress_test.go
```diff
@@ -126,6 +126,26 @@ func TestPathOrDefaultReturnActual(t *testing.T) {
 	}
 }
 
+func TestGenerateIngressPathExact(t *testing.T) {
```
pleshakov, line 129: PathTypePrefix is not tested. It is an omission.

Also, Since we'll have 3 different input (and the forth for `nil` check as per previous suggestion), better to use a table driven test.  For example, take a look at `TestGeneratePath` from internal/configs/virtualserver_test.go:4481.

internal/configs/ingress_test.go
```diff
@@ -126,6 +126,41 @@ func TestPathOrDefaultReturnActual(t *testing.T) {
...
+			path:     "/path/to/resource",
+			expected: "/path/to/resource",
+		},
+		{
+			pathType: v1beta1.PathTypeImplementationSpecific,
+			path:     "/path/to/resource",
+			expected: "/path/to/resource",
+		},
+		{
+			pathType: v1beta1.PathType(""),
```
pleshakov, line 151: is the case when pathType is an empty string possible? could you double check?

internal/configs/ingress_test.go
```diff
@@ -126,6 +126,41 @@ func TestPathOrDefaultReturnActual(t *testing.T) {
...
+			path:     "/path/to/resource",
+			expected: "/path/to/resource",
```
pleshakov, line 153: Could you add a case when `pathType` is `nil` ? this case is possible

internal/configs/ingress_test.go
LorcanMcVeigh: i dont think this is possible

PR #1099, opened 2020-08-17
Add reason label to total_reload metric 

LorcanMcVeigh:
### Proposed changes
Add a label `reason` to the prometheus metric `total_reloads` which specifies if the reload was caused by a endpoint update or another reason

### Checklist

### DIFF
--- a/internal/configs/configurator.go
+++ b/internal/configs/configurator.go
@@ -15,6 +15,7 @@ import (
 	"github.com/spiffe/go-spiffe/workload"
 
 	"github.com/nginxinc/kubernetes-ingress/internal/configs/version2"
+	"github.com/nginxinc/kubernetes-ingress/internal/metrics/collectors"
 	conf_v1alpha1 "github.com/nginxinc/kubernetes-ingress/pkg/apis/configuration/v1alpha1"
 
 	"github.com/golang/glog"
@@ -186,7 +187,7 @@ func (cnf *Configurator) AddOrUpdateIngress(ingEx *IngressEx) error {
 		return fmt.Errorf("Error adding or updating ingress %v/%v: %v", ingEx.Ingress.Namespace, ingEx.Ingress.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(true); err != nil {
+	if err := cnf.nginxManager.Reload(collectors.ReloadForOtherUpdate); err != nil {
 		return fmt.Errorf("Error reloading NGINX for %v/%v: %v", ingEx.Ingress.Namespace, ingEx.Ingress.Name, err)
 	}
 
@@ -222,7 +223,7 @@ func (cnf *Configurator) AddOrUpdateMergeableIngress(mergeableIngs *MergeableIng
 		return fmt.Errorf("Error when adding or updating ingress %v/%v: %v", mergeableIngs.Master.Ingress.Namespace, mergeableIngs.Master.Ingress.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(true); err != nil {
+	if err := cnf.nginxManager.Reload(collectors.ReloadForOtherUpdate); err != nil {
 		return fmt.Errorf("Error reloading NGINX for %v/%v: %v", mergeableIngs.Master.Ingress.Namespace, mergeableIngs.Master.Ingress.Name, err)
 	}
 
@@ -309,7 +310,7 @@ func (cnf *Configurator) AddOrUpdateVirtualServer(virtualServerEx *VirtualServer
 		return warnings, fmt.Errorf("Error adding or updating VirtualServer %v/%v: %v", virtualServerEx.VirtualServer.Namespace, virtualServerEx.VirtualServer.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(true); err != nil {
+	if err := cnf.nginxManager.Reload(collectors.ReloadForOtherUpdate); err != nil {
 		return warnings, fmt.Errorf("Error reloading NGINX for VirtualServer %v/%v: %v", virtualServerEx.VirtualServer.Namespace, virtualServerEx.VirtualServer.Name, err)
 	}
 
@@ -356,7 +357,7 @@ func (cnf *Configurator) AddOrUpdateVirtualServers(virtualServerExes []*VirtualS
 		allWarnings.Add(warnings)
 	}
 
-	if err := cnf.nginxManager.Reload(false); err != nil {
+	if err := cnf.nginxManager.Reload(collectors.ReloadForOtherUpdate); err != nil {
 		return allWarnings, fmt.Errorf("Error when reloading NGINX when updating Policy: %v", err)
 	}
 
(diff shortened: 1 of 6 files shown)


internal/configs/configurator.go
```diff
@@ -186,7 +186,7 @@ func (cnf *Configurator) AddOrUpdateIngress(ingEx *IngressEx) error {
 		return fmt.Errorf("Error adding or updating ingress %v/%v: %v", ingEx.Ingress.Namespace, ingEx.Ingress.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(); err != nil {
+	if err := cnf.nginxManager.Reload(true); err != nil {
```
pleshakov, line 189: should be false

internal/configs/configurator.go
```diff
@@ -222,7 +222,7 @@ func (cnf *Configurator) AddOrUpdateMergeableIngress(mergeableIngs *MergeableIng
 		return fmt.Errorf("Error when adding or updating ingress %v/%v: %v", mergeableIngs.Master.Ingress.Namespace, mergeableIngs.Master.Ingress.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(); err != nil {
+	if err := cnf.nginxManager.Reload(true); err != nil {
```
pleshakov, line 225: should be false

internal/configs/configurator.go
```diff
@@ -309,7 +309,7 @@ func (cnf *Configurator) AddOrUpdateVirtualServer(virtualServerEx *VirtualServer
 		return warnings, fmt.Errorf("Error adding or updating VirtualServer %v/%v: %v", virtualServerEx.VirtualServer.Namespace, virtualServerEx.VirtualServer.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(); err != nil {
+	if err := cnf.nginxManager.Reload(true); err != nil {
```
pleshakov, line 312: should be false

internal/configs/configurator.go
```diff
@@ -371,7 +371,7 @@ func (cnf *Configurator) AddOrUpdateTransportServer(transportServerEx *Transport
 		return fmt.Errorf("Error adding or updating TransportServer %v/%v: %v", transportServerEx.TransportServer.Namespace, transportServerEx.TransportServer.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(); err != nil {
+	if err := cnf.nginxManager.Reload(true); err != nil {
```
pleshakov, line 374: should be false

internal/configs/configurator.go
```diff
@@ -617,7 +617,7 @@ func (cnf *Configurator) DeleteIngress(key string) error {
 		cnf.deleteIngressMetricsLabels(key)
 	}
 
-	if err := cnf.nginxManager.Reload(); err != nil {
+	if err := cnf.nginxManager.Reload(true); err != nil {
```
pleshakov, line 620: should be false

internal/configs/configurator.go
```diff
@@ -634,7 +634,7 @@ func (cnf *Configurator) DeleteVirtualServer(key string) error {
 		cnf.deleteVirtualServerMetricsLabels(fmt.Sprintf(key))
 	}
 
-	if err := cnf.nginxManager.Reload(); err != nil {
+	if err := cnf.nginxManager.Reload(true); err != nil {
```
pleshakov, line 637: should be false

internal/configs/configurator.go
```diff
@@ -648,7 +648,7 @@ func (cnf *Configurator) DeleteTransportServer(key string) error {
 		return fmt.Errorf("Error when removing TransportServer %v: %v", key, err)
 	}
 
-	err = cnf.nginxManager.Reload()
+	err = cnf.nginxManager.Reload(true)
```
pleshakov, line 651: should be false

LorcanMcVeigh:
> @LorcanMcVeigh
> Please see my comments and suggestions
> 
> Additionally, after you start the IC, until it gets reloaded, you will not get metrics for one or both labels. For "other" label, as the first reload happens quickly after the start, it is hard to notice its absence. However, for the "endpoints" label, you can see its absence. For example, here the IC was reloaded but no endpoints update happened yet:
> 
> ```
> # HELP nginx_ingress_controller_nginx_reloads_total Number of successful NGINX reloads
> # TYPE nginx_ingress_controller_nginx_reloads_total counter
> nginx_ingress_controller_nginx_reloads_total{class="nginx",reason="other"} 2
> ```
> 
> Note that the metric nginx_ingress_controller_nginx_reloads_total{class="nginx",reason="endpoint"} is not present here at all.
> 
> The solution is to initialize those labels before the IC starts. Could you add that?

could you possibly tell me how you achieved this? I haven't been able to replicate this

internal/nginx/manager.go
```diff
@@ -268,7 +268,7 @@ func (lm *LocalManager) Start(done chan error) {
 }
 
 // Reload reloads NGINX.
-func (lm *LocalManager) Reload() error {
+func (lm *LocalManager) Reload(isEndPoint bool) error {
```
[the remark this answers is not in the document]
LorcanMcVeigh: could we not just define the constants in configurator because they wont be used outside of configurator and have no dependancies in manager.go?

pleshakov:
@LorcanMcVeigh 

> could you possibly tell me how you achieved this? I haven't been able to replicate this

yep. So just deploy an Ingress Controller. Without deploying any backends and Ingress/VS resources.

internal/nginx/manager.go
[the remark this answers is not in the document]
pleshakov: those constants are part of the API of the manager. so it make sense to define them there.

internal/configs/configurator.go
```diff
@@ -186,7 +187,7 @@ func (cnf *Configurator) AddOrUpdateIngress(ingEx *IngressEx) error {
 		return fmt.Errorf("Error adding or updating ingress %v/%v: %v", ingEx.Ingress.Namespace, ingEx.Ingress.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(); err != nil {
+	if err := cnf.nginxManager.Reload(collectors.ReloadForOtherUpdate); err != nil {
```
pleshakov, line 190: it is a bit strange that we call nginxManager from `nginx` package here and at the same time we use a constant from a different package. Could you move those  two constants to `nginx` package? (that was the original suggestion)

internal/metrics/collectors/manager.go
```diff
@@ -63,8 +71,14 @@ func NewLocalManagerMetricsCollector(constLabels map[string]string) *LocalManage
 }
 
 // IncNginxReloadCount increments the counter of successful NGINX reloads and sets the last reload status to true
-func (nc *LocalManagerMetricsCollector) IncNginxReloadCount() {
-	nc.reloadsTotal.Inc()
+func (nc *LocalManagerMetricsCollector) IncNginxReloadCount(isEndPointUpdate bool) {
+	if isEndPointUpdate {
+		nc.reloadsTotal.WithLabelValues("endpoints").Inc()
+		nc.reloadsTotal.WithLabelValues("other")
```
pleshakov, line 77: could you do these initilizations only once in `NewLocalManagerMetricsCollector` ?

```
nc.reloadsTotal.WithLabelValues("other")
nc.reloadsTotal.WithLabelValues("endpoints")
```

internal/nginx/manager.go
```diff
@@ -268,7 +268,7 @@ func (lm *LocalManager) Start(done chan error) {
 }
 
 // Reload reloads NGINX.
-func (lm *LocalManager) Reload() error {
+func (lm *LocalManager) Reload(isEndPointUpdate bool) error {
```
pleshakov, line 271: Could you use one away to name this parameter across all methods for consistency? Better to use plural as this is how the object is named in Kubernetes. Also, in the rest of the code, endpoint related ids don't use mixed caps (no `endPoint`).  `isEndpointsUpdate` would be a good name

Dean-Coakley (approved):

PR #1119, opened 2020-08-31
Gh actions edge

martialonline:
### Proposed changes
Add edge pipeline workflow, use latest Kubernetes (1.19.0), update Helm template RBAC endpoint to v1.

### DIFF
--- a/deployments/helm-chart/templates/rbac.yaml
+++ b/deployments/helm-chart/templates/rbac.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.rbac.create }}
 kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ include "nginx-ingress.name" . }}
   labels:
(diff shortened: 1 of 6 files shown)


tellet (approved):
Lgtm

deployments/helm-chart/templates/rbac.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.rbac.create }}
 kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
```
Dean-Coakley, line 3: Can you make the same change?:
https://github.com/nginxinc/kubernetes-ingress/blob/gh-actions-edge/deployments/rbac/rbac.yaml#L2

Dean-Coakley (approved):
LGTM. Pending apiVersion upgrade for rbac manifests.

deployments/helm-chart/templates/rbac.yaml
martialonline: Done, thanks!

PR #1133, opened 2020-09-09
Add support for IngressClass resources

### LINKED ISSUE #924: Support for k8s 1.18 Ingress Class?

lucacome:
Supporting `IngressClass` resource and `ingressClassName` for versions of k8s >= 1.18.

When using version of k8s >= 1.18 the `use-ingress-class-only` is now ignored and the Ingress Controller will only handle resources with the appropriate class set.

Fixes #924


LorcanMcVeigh (approved):

Dean-Coakley (commented):
@lucacome This looks great!

Just one thing to make you aware of, is that you can set a compatible kubernetes version in your Chart.yaml. https://helm.sh/docs/topics/charts/#the-chartyaml-file

Example: `kubeVersion: 1.15-1.19`.

This would reject the installation of a helm release at the helm client level. I think this could be a nice improvement to include as it shortens the feedback loop for a user. (They don't have to install incompatible resources into their cluster to see logs that their cluster is incompatible.)
```
$ helm install test deployments/helm-chart
Error: chart requires kubeVersion: 1.15-1.19 which is incompatible with Kubernetes v1.14.0

$ helm ls
$
```

Feel free to ignore, we can always refer back to: https://docs.nginx.com/nginx-ingress-controller/technical-specifications/#supported-kubernetes-versions and hope most users read that.

docs-web/installation/running-multiple-ingress-controllers.md
```diff
@@ -12,6 +12,7 @@ This document explains the following topics:
 The smooth coexistence of multiple Ingress Controllers in one cluster is provided by the Ingress class concept, which mandates the following:
 * Every Ingress Controller must only handle Ingress resources for its particular class. 
 * Ingress resources should be annotated with the `kubernetes.io/ingress.class` annotation set to the value, which corresponds to the class of the Ingress Controller the user wants to use. 
+* When using versions of Kubernetes >= 1.18, Ingress resources should have the `ingressClassName` field set to the value, which corresponds to the class of the Ingress Controller the user wants to use. 
 * VirtualServer and VirtualServerRoute resources should have the `ingressClassName` field set to the value, which corresponds to the class of the Ingress Controller the user wants to use.
 
 ### Configuring Ingress Class
```
pleshakov, line 18: You need to update the note in this section. Perhaps something like this? (please double check it, I might not have got it right with VirtualServers)

**Notes**: 
* For Kubernetes < 1.18,  if the class is not set in an Ingress configuration resource, the Ingress Controller will handle the resource. This is controlled via the `-use-ingress-class-only` argument. 
* For Kubernetes >= 1.18,  if the class is not set in an Ingress resource, Kubernetes will set it to the class of the default Ingress Controller. To make the Ingress Controller the default one, the `ingressclass.kubernetes.io/is-default-class` must be set on the IngressClass resource. See the Step 3 *Create an IngressClass resource* of the [Create Common Resources](https://deploy-preview-1133--nginx-kubernetes-ingress.netlify.app/nginx-ingress-controller/installation/installation-with-manifests/#create-common-resources) section.
* For VirtualServer and VirtualServerRoute resources, the Ingress Controller will always handle resources with an empty  class.

docs-web/installation/installation-with-helm.md
```diff
@@ -274,10 +274,21 @@ The following tables lists the configurable parameters of the NGINX Ingress cont
      - The number of replicas of the Ingress controller deployment.
      - 1
    * - ``controller.ingressClass``
-     - A class of the Ingress controller. The Ingress controller only processes Ingress resources that belong to its class - i.e. have the annotation ``"kubernetes.io/ingress.class"``  or the ``"ingressClassName"`` field in VirtualServer/VirtualServerRoute equal to the class. Additionally, the Ingress controller processes Ingress resources that do not have that annotation which can be disabled by setting the "-use-ingress-class-only" flag.
```
pleshakov, line 277: the preview for this table wasn't generated -- https://deploy-preview-1133--nginx-kubernetes-ingress.netlify.app/nginx-ingress-controller/installation/installation-with-helm/
I suspect some errors

deployments/helm-chart/Chart.yaml
```diff
@@ -2,6 +2,7 @@ name: nginx-ingress
 version: edge
 appVersion: edge
 apiVersion: v1
+kubeVersion: ">= 1.15.0"
```
pleshakov, line 5: In the main.go, we support 1.14. Here in helm, we don't support it. Was it intentional?  🤔

pleshakov (changes requested):
Hi @lucacome 

Please see my comments and suggestions. Most of them about docs. Looks good overall.

Additionally,

(1) We're changing behavior for VirtualServer/VirtualServerRoutes, right? Previously, if `"-use-ingress-class-only" was enabled, the IC would ignore VS/VSRs with empty class. Now it won't ignore them. If that's the case, could you add the change label to the PR? We will also need to document that change in the changelog.

(2) Could you include in the description that you're fixing this https://github.com/nginxinc/kubernetes-ingress/issues/924 ?

Dean-Coakley (approved):
LGTM pending https://github.com/nginxinc/kubernetes-ingress/pull/1133#discussion_r485818176

(Sorry if my example misled you @lucacome 😓 )

docs-web/installation/running-multiple-ingress-controllers.md
```diff
@@ -11,15 +11,18 @@ This document explains the following topics:
...
 
 The default Ingress class of NGINX Ingress Controller is `nginx`, which means that it only handles configuration resources with the `class` set to `nginx`. You can customize the class through the `-ingress-class` command-line argument.
 
-**Note**: By default, if the `class` is not set in a configuration resource, the Ingress Controller will handle the resource. This is controlled via the `-use-ingress-class-only` argument.
+**Notes**: 
+* For Kubernetes < 1.18, if the class is not set in an Ingress configuration resource, the Ingress Controller will handle the resource. This is controlled via the `-use-ingress-class-only` argument. 
+* For Kubernetes >= 1.18, if the class is not set in an Ingress resource, Kubernetes will set it to the class of the default Ingress Controller. To make the Ingress Controller the default one, the `ingressclass.kubernetes.io/is-default-class` must be set on the IngressClass resource. See Step 3 *Create an IngressClass resource* of the [Create Common Resources](https://deploy-preview-1133--nginx-kubernetes-ingress.netlify.app/nginx-ingress-controller/installation/installation-with-manifests/#create-common-resources) section.
```
pleshakov, line 24: same problem with an absolute link

pleshakov (approved):
@lucacome  looks good. Could you fix the mentioned link problems before merging?

PR #1147, opened 2020-09-16
add option to specify other log destinations in AP

rafwegv:
### Proposed changes
This change allows for configuring additional values as log destinations, when setting remote logging via apLogConf

### Checklist

### DIFF
--- a/internal/k8s/app_protect_resources.go
+++ b/internal/k8s/app_protect_resources.go
@@ -55,13 +55,21 @@ func ValidateAppProtectLogConf(logConf *unstructured.Unstructured) error {
 	return nil
 }
 
-var logDstEx = regexp.MustCompile(`syslog:server=((?:\d{1,3}\.){3}\d{1,3}|localhost):\d{1,5}`)
+var logDstEx = regexp.MustCompile(`(?:syslog:server=((?:\d{1,3}\.){3}\d{1,3}|localhost):\d{1,5})|stderr|(?:\/[\S]+)+`)
+var logDstFileEx = regexp.MustCompile(`(?:\/[\S]+)+`)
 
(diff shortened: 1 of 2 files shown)


internal/k8s/app_protect_resources.go
```diff
@@ -55,14 +55,22 @@ func ValidateAppProtectLogConf(logConf *unstructured.Unstructured) error {
 	return nil
 }
 
-var logDstEx = regexp.MustCompile(`syslog:server=((?:\d{1,3}\.){3}\d{1,3}|localhost):\d{1,5}`)
+var logDstEx = regexp.MustCompile(`(?:syslog:server=((?:\d{1,3}\.){3}\d{1,3}|localhost):\d{1,5})|stderr|(?:\/[\S]+)+`)
+var logDstFileEx = regexp.MustCompile(`(?:\/[\S]+)+`)
 
 // ValidateAppProtectLogDestinationAnnotation validates annotation for log destination configuration
 func ValidateAppProtectLogDestinationAnnotation(dstAntn string) error {
```
pleshakov, line 62: Could you possibly add a unit test for this function?  It be great to test for each of the possible types of valid values as well as invalid values (a case per each`return  fmt.Errorf`)

rafwegv: added tests, also removed one return fmt.Errorf since i found it is not reachable.

pleshakov (approved):
@rafwegv 

the test looks good. 

One comment though - for maintainability, we usually don't put the test input into global variables - we put them into the test function.  could you make `posDstAntns ` and  `negDstAntns ` local variables of `ValidateAppProtectLogDestinationAnnotation` ?

Dean-Coakley (approved):

PR #1148, opened 2020-09-16
Update docs and helm with latency metric

LorcanMcVeigh:
### Proposed changes
Update helm and documentation with the new latency metric. Fixed incorrect comments as well  

### Checklist

### DIFF
--- a/deployments/helm-chart/values.yaml
+++ b/deployments/helm-chart/values.yaml
@@ -280,6 +280,9 @@ controller:
     ## Set the port where the readiness endpoint is exposed.
     port: 8081
 
+  ## Enable collection of latency metrics for upstreams. Requires enable-prometheus-metrics
+  enableLatencyMetrics: true
+
 rbac:
   ## Configures RBAC.
   create: true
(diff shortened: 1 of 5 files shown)


deployments/helm-chart/values.yaml
```diff
@@ -280,6 +280,9 @@ controller:
     ## Set the port where the readiness endpoint is exposed.
     port: 8081
 
+  ## Enable collection of latency metrics for upstreams. Requires enable-prometheus-metrics
```
Dean-Coakley, line 283: In helm values file if you grep around for "Requires" you'll see that we reference fields by where they are in the yaml structure. 

This makes it easier to find other fields in the same file where they have differing names between the yaml and the cli arg, and lets people copy and paste directly to a `--set` command, for example: `helm install --set prometheus.create=true`.

```suggestion
  ## Enable collection of latency metrics for upstreams. Requires prometheus.create.
```

Dean-Coakley (approved):
💨

lucacome (approved):

PR #1158, opened 2020-09-24
Support variables in action proxy headers

pleshakov:
### Proposed changes
This PR adds support for variables in action proxy header. For example:
```yaml
proxy:
  upstream: coffee
  requestHeaders:
    pass: true
    set:
    - name: client-cert-id
      value: ${ssl_client_i_dn} # using variable
  responseHeaders:
    add:
    - name: upstream-addresses
      value: ${upstream_addr} # using variable
      always: true
```

Note: docs will be added in a separate PR


Dean-Coakley (approved):
🚀

LorcanMcVeigh (approved):

PR #1174, opened 2020-10-05
Document how policies are applied and validated

pleshakov:
### Proposed changes
Document how policies are applied and validated

### DIFF
--- a/docs-web/configuration/policy-resource.md
+++ b/docs-web/configuration/policy-resource.md
@@ -23,6 +23,8 @@ This document is the reference documentation for the Policy resource. An example
     - [IngressMTLS](#ingressmtls)
       - [IngressMTLS Merging Behavior](#ingressmtls-merging-behavior)
   - [Using Policy](#using-policy)
+    - [Applying Policies](#applying-policies)
+    - [Invalid Policies](#invalid-policies)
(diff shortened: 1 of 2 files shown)


docs-web/configuration/policy-resource.md
```diff
@@ -352,6 +354,86 @@ webapp-policy   27m
...
+          pass: coffee
+      ```
+
+      For VirtualServer, you can apply a policy:
+      - to all routes (spec policies)
+      - to a specific route (route policies)
+
+      Route policies of the *same type* override spec policies. In the example above, if the type of the policies `policy-1` and `policy-3` is `accessControl`, then for requests to `cafe.example.com/coffee`, NGINX will apply `policy-3`.
+
+      The overriding is enforced by NGINX: the spec policies are implemented in the `server` context of the config, the route policies --  the `location` context. As a result, the route policies of the same type win.
```
jputrino, line 395:
```suggestion
      The overriding is enforced by NGINX: the spec policies are implemented in the `server` context of the config, and the route policies are implemented in the `location` context. As a result, the route policies of the same type win.
```

jputrino (approved):

vepatel (approved):

PR #1176, opened 2020-10-06
Remove ValidateSecret and add tests for GetSecretKind

lucacome:
Removed `ValidateSecret` and add tests for `GetSecretKind`, `ValidateTLSSecret`, `ValidateIngressMTLSSecret` and `ValidateJWKSecret`

### DIFF
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -1609,7 +1609,7 @@ func (lbc *LoadBalancerController) handleRegularSecretDeletion(key string, ings 
 func (lbc *LoadBalancerController) handleSecretUpdate(secret *api_v1.Secret, ings []networking.Ingress, virtualServers []*conf_v1.VirtualServer) {
 	secretNsName := secret.Namespace + "/" + secret.Name
 
-	err := lbc.ValidateSecret(secret)
+	kind, err := GetSecretKind(secret)
 	if err != nil {
 		// Secret becomes Invalid
 		glog.Errorf("Couldn't validate secret %v: %v", secretNsName, err)
@@ -1625,9 +1625,6 @@ func (lbc *LoadBalancerController) handleSecretUpdate(secret *api_v1.Secret, ing
 	title := "Updated"
 	message := fmt.Sprintf("Configuration was updated due to updated secret %v", secretNsName)
 	state := conf_v1.StateValid
-
-	// we can safely ignore the error because the secret is valid in this function
-	kind, _ := GetSecretKind(secret)
 
 	if kind == JWK {
 		virtualServerExes := lbc.virtualServersToVirtualServerExes(virtualServers)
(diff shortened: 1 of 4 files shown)


LorcanMcVeigh (approved):
Looks Good, don't forget to `gofmt` 
I think it's ok to ignore the warnings in the files you didn't modify. we can `gofmt` them later

internal/k8s/controller.go
```diff
@@ -3302,21 +3302,25 @@ func (lbc *LoadBalancerController) isHealthCheckEnabled(ing *networking.Ingress)
...
-// For NGINX Plus, it also checks if the secret follows the JWK Secret format.
-func (lbc *LoadBalancerController) ValidateSecret(secret *api_v1.Secret) error {
-	err1 := ValidateTLSSecret(secret)
-	if !lbc.isNginxPlus {
-		return err1
-	}
-
-	err2 := ValidateJWKSecret(secret)
+// ValidateSecret validates that the secret follows the TLS, MTLS or JWK Secret format.
+func (lbc *LoadBalancerController) ValidateSecret(secret *api_v1.Secret) (err error) {
```
LorcanMcVeigh, line 3306: this should have unit tests (?)

LorcanMcVeigh (changes requested):
Sorry for the re-review, i just noticed this after I did my review

internal/k8s/handlers.go
```diff
@@ -154,6 +154,7 @@ func createSecretHandlers(lbc *LoadBalancerController) cache.ResourceEventHandle
 		AddFunc: func(obj interface{}) {
 			secret := obj.(*v1.Secret)
 			if err := lbc.ValidateSecret(secret); err != nil {
+				glog.V(3).Infof("Error validating secret %v: %v", secret.Name, err)
```
pleshakov, line 157: This error is not always an error (and in line 173) Consider other secrets in the cluster not used by the Ingress Controller.  For every secret like that, the error debug message will be printed. Will it be ok? For debug purposes, perhaps it makes sense to change the message, to something like `Ignoring secret of unknown kind`

I think the problem her is that we should have had a different function here in the first place. Something like `IsIngressControllerSecret` or `HasKnownSecretKind`, where it would return true/false if the Ingress Controller was interested in that secret. (But the name is not needed to be addressed in this PR though)

internal/k8s/handlers.go
```diff
@@ -174,16 +175,19 @@ func createSecretHandlers(lbc *LoadBalancerController) cache.ResourceEventHandle
...
 			glog.V(3).Infof("Removing Secret: %v", secret.Name)
 			lbc.AddSyncQueue(obj)
 		},
 		UpdateFunc: func(old, cur interface{}) {
+			curSecret := cur.(*v1.Secret)
 			errOld := lbc.ValidateSecret(old.(*v1.Secret))
-			errCur := lbc.ValidateSecret(cur.(*v1.Secret))
+			errCur := lbc.ValidateSecret(curSecret)
 			if errOld != nil && errCur != nil {
+				glog.V(3).Infof("Error validating secret %v: %v", curSecret.Name, errCur)
```
pleshakov, line 190: this `if` handles the case when the secret is not related to the Ingress Controller - it's not TLS, JWT or mTLS.  In that case both old and cur error are non nil. This doesn't mean it is an error -  it's valid case - the Ingress Controller needs to ignore those secrets. That's why I suggest changing the message to something like `Ignoring secret of unknown kind`. Note that I think it makes sense to print both errors here, since we have two non-nil errors.

internal/k8s/controller.go
pleshakov: unit test would help! 
I think the function could be simplified to
```
	_, err := GetSecretKind(secret)
	return err
```

internal/k8s/secret.go
```diff
@@ -23,11 +23,17 @@ const (
 
 // ValidateTLSSecret validates the secret. If it is valid, the function returns nil.
 func ValidateTLSSecret(secret *v1.Secret) error {
-	if _, exists := secret.Data[v1.TLSCertKey]; !exists {
+	_, certExists := secret.Data[v1.TLSCertKey]
+	_, keyExists := secret.Data[v1.TLSPrivateKeyKey]
+
+	if certExists && keyExists {
```
pleshakov, line 29: because of this check, I think line 40 (`return nil`) is never invoked

pleshakov (approved):
👍

internal/k8s/secret.go
lucacome: yeah but it's complaining if I don't add a return at the end 😅

LorcanMcVeigh (approved):

PR #1179, opened 2020-10-07
Fix TransportServers in debian AppProtect image

pleshakov:
### Proposed changes
Previously, the DockerfileWithAppProtectForPlus didn't create the folder /etc/nginx/stream-conf.d. As a result, the IC wasn't able to write generated config for TransportServers, because the directory didn't exist.

 The PR fixes the Dockerfile to create that folder.


lucacome (approved):
LGTM. Maybe a more descriptive commit message? It's more related to nginx and config than TransportServer itself imo

LorcanMcVeigh (approved):

PR #1180, opened 2020-10-08
Add support for EgressMTLS

lucacome:
Adding support for Egress MTLS. Docs will come in another PR.


pkg/apis/configuration/validation/policy.go
```diff
@@ -146,6 +151,24 @@ func validateIngressMTLS(ingressMTLS *v1alpha1.IngressMTLS, fieldPath *field.Pat
 
 }
 
+func validateEgressMTLS(egressMTLS *v1alpha1.EgressMTLS, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, validateSecretName(egressMTLS.TLSSecret, fieldPath.Child("tlsSecret"))...)
+
+	if egressMTLS.VerifyServer != nil && *egressMTLS.VerifyServer == true && egressMTLS.TrustedCertSecret == "" {
```
LorcanMcVeigh, line 159:
```suggestion
	if egressMTLS.VerifyServer != nil && *egressMTLS.VerifyServer && egressMTLS.TrustedCertSecret == "" {
```

LorcanMcVeigh (approved):
Looks Good, just spotted this because of the linter

internal/configs/virtualserver.go
```diff
@@ -586,6 +594,30 @@ func (vsc *virtualServerConfigurator) generatePolicies(owner runtime.Object, own
...
 				}
 
+			} else if pol.Spec.EgressMTLS != nil {
+				if context != routeContext {
+					vsc.addWarningf(owner, `EgressMTLS policy is not allowed in the %v context`, context)
+					policyError = true
+					break
+				}
+				if egressMTLS != nil {
+					vsc.addWarningf(owner, "Multiple egressMTLS policies are not allowed. EgressMTLS policy %q will be ignored", key)
```
pleshakov, line 604: They are not allowed in the same context. I suggest re-using the JWT message:
```suggestion
					vsc.addWarningf(owner, "Multiple egressMTLS policies in the same context is not valid. EgressMTLS policy %q will be ignored", key)
```

examples-of-custom-resources/egress-mtls/virtual-server.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  host: webapp.example.com
+  upstreams:
+  - name: secure-app
+    service: secure-app
+    port: 8443
+  routes:
+  - path: /
+    policies:
+        - name: egress-mtls-policy
```
pleshakov, line 14: Better to indent arrays consistently: 
```
    policies:
    - name ...
```

examples-of-custom-resources/egress-mtls/egress-mtls.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: egress-mtls-policy
+spec:
+  egressMTLS:
+    tlsSecret: egress-mtls-secret
+    trustedCertSecret: egress-trusted-ca-secret
+    verifyServer: on
+    verifyDepth: 2
+    serverName: on
+    sslName: "secure-app.example.com"
```
pleshakov, line 12: I don't think we need `"` for hosts
```suggestion
    sslName: secure-app.example.com
```

examples-of-custom-resources/egress-mtls/README.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+## Step 5 - Configure Load Balancing
+
+Create a VirtualServer resource for the web application:
+```
+$ kubectl apply -f virtual-server.yaml
+```
+
+Note that the VirtualServer references the policy `egress-mtls-policy` created in Step 3.
```
pleshakov, line 52:
```suggestion
Note that the VirtualServer references the policy `egress-mtls-policy` created in Step 4.
```

examples-of-custom-resources/egress-mtls/README.md
```diff
@@ -0,0 +1,60 @@
...
+1. Save the HTTP port of the Ingress Controller into a shell variable:
+    ```
+    $ IC_HTTP_PORT=<port number>
+    ```
+
+## Step 1 - Deploy a Secure Web Application
+
+Create the application deployment, service and secret:
+```
+$ kubectl apply -f secure-webapp.yaml
```
pleshakov, line 21: a bug
```suggestion
$ kubectl apply -f secure-app.yaml
```

examples-of-custom-resources/egress-mtls/README.md
```diff
@@ -0,0 +1,60 @@
...
+1. Save the public IP address of the Ingress Controller into a shell variable:
+    ```
+    $ IC_IP=XXX.YYY.ZZZ.III
+    ```
+1. Save the HTTP port of the Ingress Controller into a shell variable:
+    ```
+    $ IC_HTTP_PORT=<port number>
+    ```
+
+## Step 1 - Deploy a Secure Web Application
```
pleshakov, line 17: could we add an explanatory paragraph here. Something like.

The application requires clients to use TLS and present a client TLS certificate which it will verify.

examples-of-custom-resources/egress-mtls/README.md
```diff
@@ -0,0 +1,60 @@
+# Egress MTLS
+
+In this example, we deploy a web application, configure load balancing for it via a VirtualServer, and apply an Egress MTLS policy.
```
pleshakov, line 3:
```suggestion
In this example, we deploy a secure web application, configure load balancing for it via a VirtualServer, and apply an Egress MTLS policy.
```

examples-of-custom-resources/egress-mtls/README.md
```diff
@@ -0,0 +1,60 @@
...
+## Step 1 - Deploy a Secure Web Application
+
+Create the application deployment, service and secret:
+```
+$ kubectl apply -f secure-webapp.yaml
+```
+
+## Step 2 - Deploy the Egress MLTS Secret
+
+Create a secret with the name `egress-mtls-secret` that will be used for authentication to the Secure Web Application:
```
pleshakov, line 26: I think we can use the application here and going forward, similarly to the styling in the other examples:
```suggestion
Create a secret with the name `egress-mtls-secret` that will be used for authentication to application:
```

pleshakov (changes requested):
Hi @lucacome 

Please see my comments and suggestions.

internal/configs/virtualserver.go
```diff
@@ -351,6 +353,10 @@ func (vsc *virtualServerConfigurator) GenerateVirtualServerConfig(vsEx *VirtualS
 		} else {
 			upstreamName := virtualServerUpstreamNamer.GetNameForUpstreamFromAction(r.Action)
 			upstream := crUpstreams[upstreamName]
+			if routePoliciesCfg.EgressMTLS != nil {
+				upstream.TLS.Enable = true
```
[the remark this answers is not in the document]
lucacome: >I'm thinking about cases when a single Egress policy is applied at the spec level and then only it is in certain upstreams we need to enable it.

I'm not sure I understand the case here. If it's at the spec level isn't it applied to all the upstreams?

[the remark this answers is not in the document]
lucacome: I think it makes sense to say that the user needs to use 
```yaml
   tls:
     enable: true
```
at least for now

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -155,6 +156,19 @@ type IngressMTLS struct {
 	VerifyDepth      *int   `json:"verifyDepth"`
 }
 
+// EgressMTLS defines an Egress MTLS policy.
+type EgressMTLS struct {
+	TLSSecret         string `json:"tlsSecret"`
+	VerifyServer      *bool  `json:"verifyServer"`
```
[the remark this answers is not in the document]
lucacome: I was consistent in the struct with all `*bool`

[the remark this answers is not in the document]
pleshakov: I see. Well, the idea here is that if the default value of a field matches the default value of the type of the field, there is no need to use a pointer. We apply that for other types as well like string or int.

internal/k8s/controller_test.go
```diff
@@ -1435,3 +1486,189 @@ func TestGetIngressMTLSSecret(t *testing.T) {
...
+		lbc := LoadBalancerController{
+			secretLister: storeToSecretLister{
+				&cache.FakeCustomStore{
+					GetByKeyFunc: func(key string) (item interface{}, exists bool, err error) {
+						switch key {
+						case "default/valid-egress-mtls-secret":
+							return validSecret, true, nil
+						case "default/valid-egress-trusted-secret":
+							return validSecret2, true, nil
+						case "default/invalid-ingress-mtls-secret":
```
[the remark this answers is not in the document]
lucacome: Yeah should be egress and it's tested in the last case.

pleshakov (approved):
🚀

PR #1225, opened 2020-11-14
Introduce secret types

pleshakov:
### Proposed changes

This PR introduces secret types that the Ingress Controller can support:
- kubernetes.io/tls for TLS secrets. Such secrets must have the tls.crt and tls.key keys.
- nginx.org/ca - for CA secrets. Such secrets must have the ca.crt key with a CA.
- nginx.org/jwk - for JWK secrets. Such secrets must include the jwk key with a JWK.

The Ingress Controller will no longer support secrets without the type field set to the types above.


lucacome (approved):

mikestephen (approved):
Looking good!

PR #1229, opened 2020-11-17
Refactor generatePolicies func to be more easily extended for new policy types

mikestephen:
### Proposed changes
Extracted methods for each policy type which are called from the original function to iteratively build up the policiesConfig struct.

### Checklist


internal/configs/virtualserver.go
```diff
@@ -490,169 +546,200 @@ type policiesCfg struct {
...
+type validationResults struct {
+	isError  bool
+	warnings []string
+}
+
+func newValidationResults() *validationResults {
+	return &validationResults{}
+}
+
+func (r *validationResults) addWarning(msgFmt string, args ...interface{}) {
```
pleshakov, line 576: The names of the other similar functions that take a format string as a first argument end with "f". I suggest naming this function similarly:

```suggestion
func (r *validationResults) addWarningf(msgFmt string, args ...interface{}) {
```

internal/configs/virtualserver.go
```diff
@@ -490,169 +546,200 @@ type policiesCfg struct {
...
+			polNamespace = ownerDetails.ownerNamespace
+		}
+
+		key := fmt.Sprintf("%s/%s", polNamespace, p.Name)
 
+		if pol, exists := policies[key]; exists {
+			res := newValidationResults()
+			switch {
+			case pol.Spec.AccessControl != nil:
+				config.addAccessControlConfig(res, pol.Spec.AccessControl)
```
pleshakov, line 718: I wonder if we really need to pass `res` as a parameter. Could those add...Config functions return validationResults?

for example,
```go
func (p *policiesCfg) addAccessControlConfig(accessControl *conf_v1alpha1.AccessControl) validationResults
```
so that
```go

var res validationResults
. . .
res = config.addAccessControlConfig(pol.Spec.AccessControl)
```

imho it's easier to read when errors are returned this way.

pleshakov (commented):
Hi @mikestephen I have a few suggestions/comments. Otherwise, looks good!

internal/configs/virtualserver.go
mikestephen: Thanks for that!

internal/configs/virtualserver.go
mikestephen: Looks much better, I agree. I think I went for this approach because I mistakenly thought the results were being built up over multiple iterations of the for loop, but they aren't.

pleshakov (approved):

internal/configs/virtualserver.go
```diff
@@ -490,169 +546,212 @@ type policiesCfg struct {
...
+	res := newValidationResults()
+	rlZoneName := fmt.Sprintf("pol_rl_%v_%v_%v_%v", polNamespace, polName, vsNamespace, vsName)
+	p.LimitReqs = append(p.LimitReqs, generateLimitReq(rlZoneName, rateLimit))
+	p.LimitReqZones = append(p.LimitReqZones, generateLimitReqZone(rlZoneName, rateLimit))
+	if len(p.LimitReqs) == 1 {
+		p.LimitReqOptions = generateLimitReqOptions(rateLimit)
+	} else {
+		curOptions := generateLimitReqOptions(rateLimit)
+		if curOptions.DryRun != p.LimitReqOptions.DryRun {
+			res.addWarningf("RateLimit policy %q with limit request option dryRun=%v is overridden to dryRun=%v by the first policy reference in this context", polKey, curOptions.DryRun, p.LimitReqOptions.DryRun)
```
Dean-Coakley, line 600: Optional suggestion: 
* Split this line over multiple lines to reduce line length

```suggestion
			res.addWarningf(
				"RateLimit policy %q with limit request option dryRun=%v is overridden to dryRun=%v by the first policy reference in this context",
				polKey, curOptions.DryRun, p.LimitReqOptions.DryRun,
			)
```

internal/configs/virtualserver.go
```diff
@@ -490,169 +546,212 @@ type policiesCfg struct {
...
+	res := newValidationResults()
+	p.Allow = append(p.Allow, accessControl.Allow...)
+	p.Deny = append(p.Deny, accessControl.Deny...)
+	if len(p.Allow) > 0 && len(p.Deny) > 0 {
+		res.addWarningf("AccessControl policy (or policies) with deny rules is overridden by policy (or policies) with allow rules")
+	}
+	return res
+}
+
+func (p *policiesCfg) addRateLimitConfig(rateLimit *conf_v1alpha1.RateLimit, polKey string, polNamespace string, polName string, vsNamespace string, vsName string) *validationResults {
```
Dean-Coakley, line 590: Optional suggestion:
* Reduce line length

```suggestion
func (p *policiesCfg) addRateLimitConfig(rateLimit *conf_v1alpha1.RateLimit, polKey string, polNs string, polName string, vsNs string, vsName string) *validationResults {
```
**Note**: Probably makes sense to also do for `addJWTAuthConfig` if you do this change.

Or
```suggestion
func (p *policiesCfg) addRateLimitConfig(rateLimit *conf_v1alpha1.RateLimit, polKey string, polNamespace string,
	polName string, vsNamespace string, vsName string) *validationResults {
```

Dean-Coakley (approved):
🚀

internal/configs/virtualserver.go
mikestephen: Thanks, I really wasn't sure what the practice is in go to split long lines. 
I figured if `gofmt` didn't split it then I should leave it alone, but some googling found this tool: https://github.com/segmentio/golines
Any idea if the line-splitting recommended there is what's usually done in go code?

lucacome: There's not actually a standard and it's up to personal preference I guess. We could start using golines and agree on a line length. I'd vote for 120 😄

Dean-Coakley: Maybe: `golangci-lint run --enable lll`

`lll` will enable line length at 120 and can easily be added to our pipelines.
(https://github.com/nginxinc/kubernetes-ingress/blob/master/.golangci.yml)

Ref:
https://github.com/walle/lll

mikestephen: I used golines locally on the files in this PR to limit length to 120 chars.
I think it's a great suggestion to start linting for this in the pipeline, or incorporate golines into the pipeline somehow in a future PR.

PR #1231, opened 2020-11-17
Update IC log format

LorcanMcVeigh:
### Proposed changes
This PR updates the Ingress Controller log format for both Ingress and VirtualServer resources with k8s object details. 
- resource name 
- resource type 
- resource namespace
- service

### Checklist

### DIFF
--- a/deployments/helm-chart/values.yaml
+++ b/deployments/helm-chart/values.yaml
@@ -20,7 +20,7 @@ controller:
     enable: false
 
   ## Enables the Ingress controller pods to use the host's network namespace.
-  hostNetwork: false
+  hostNetwork: true
 
   ## Enables debugging for NGINX. Uses the nginx-debug binary. Requires error-log-level: debug in the ConfigMap via `controller.config.entries`.
   nginxDebug: false
@@ -168,7 +168,7 @@ controller:
 
   ## Add a location based on the value of health-status-uri to the default server. The location responds with the 200 status code for any request.
   ## Useful for external health-checking of the Ingress controller.
-  healthStatus: false
+  healthStatus: true
 
   ## Sets the URI of health status location in the default server. Requires contoller.healthStatus.
   healthStatusURI: "/nginx-health"
--- a/deployments/helm-chart/values.yaml
+++ b/deployments/helm-chart/values.yaml
@@ -20,7 +20,7 @@ controller:
     enable: false
 
   ## Enables the Ingress controller pods to use the host's network namespace.
-  hostNetwork: true
+  hostNetwork: false
 
   ## Enables debugging for NGINX. Uses the nginx-debug binary. Requires error-log-level: debug in the ConfigMap via `controller.config.entries`.
   nginxDebug: false
@@ -168,7 +168,7 @@ controller:
 
   ## Add a location based on the value of health-status-uri to the default server. The location responds with the 200 status code for any request.
   ## Useful for external health-checking of the Ingress controller.
-  healthStatus: true
+  healthStatus: false
 
   ## Sets the URI of health status location in the default server. Requires contoller.healthStatus.
   healthStatusURI: "/nginx-health"
--- a/docs-web/configuration/global-configuration/configmap-resource.md
+++ b/docs-web/configuration/global-configuration/configmap-resource.md
@@ -205,7 +205,7 @@ See the doc about [VirtualServer and VirtualServerRoute resources](/nginx-ingres
    * - ``log-format``
      - Sets the custom `log format <https://nginx.org/en/docs/http/ngx_http_log_module.html#log_format>`_ for HTTP and HTTPS traffic. For convenience, it is possible to define the log format across multiple lines (each line separated by ``\n``). In that case, the Ingress Controller will replace every ``\n`` character with a space character. All ``'`` characters must be escaped.
      - See the `template file <https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/version1/nginx.tmpl>`_ for the access log.
-     - 
+     - 127.0.0.1 - - [01/Jun/2020:18:54:31 +0000] "GET /coffee HTTP/1.1" 200 156 "-" "curl/7.54.0" "-" resource_namespace=default resource_name=cafe resource_type=ingress service_namespace=default service_name=coffee-svc pod_name=coffee-8c8ff9b4f-r99h9
    * - ``log-format-escaping``
      - Sets the characters escaping for the variables of the log format. Supported values: ``json`` (JSON escaping), ``default`` (the default escaping) ``none`` (disables escaping).
      - ``default``
--- a/docs-web/configuration/global-configuration/configmap-resource.md
+++ b/docs-web/configuration/global-configuration/configmap-resource.md
@@ -205,7 +205,22 @@ See the doc about [VirtualServer and VirtualServerRoute resources](/nginx-ingres
    * - ``log-format``
      - Sets the custom `log format <https://nginx.org/en/docs/http/ngx_http_log_module.html#log_format>`_ for HTTP and HTTPS traffic. For convenience, it is possible to define the log format across multiple lines (each line separated by ``\n``). In that case, the Ingress Controller will replace every ``\n`` character with a space character. All ``'`` characters must be escaped.
      - See the `template file <https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/version1/nginx.tmpl>`_ for the access log.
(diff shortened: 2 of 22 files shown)


internal/configs/version2/nginx.virtualserver.tmpl
```diff
@@ -44,6 +44,11 @@ server {
 
     server_name {{ $s.ServerName }};
```
pleshakov, line 45: same comments made for Plus template also apply to this template

pleshakov (changes requested):
Hi @LorcanMcVeigh 

Please see my feedback

I think it's better to rename the PR to something like "Support k8s objects variables in log format",  rather than "Update IC log format " to better reflect the PR contents.

deployments/helm-chart/values.yaml
```diff
@@ -20,7 +20,7 @@ controller:
     enable: false
 
   ## Enables the Ingress controller pods to use the host's network namespace.
-  hostNetwork: false
+  hostNetwork: true
```
mikestephen, line 23: Hi Lorcan, why does this and the health check below need to be true now?

LorcanMcVeigh: these were a mistake, I forgot to remove them

mikestephen: No problem :)

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -60,6 +61,10 @@ server {
 
 	status_zone {{$server.StatusZone}};
 
+	set $resource_type "ingress";
+	set $resource_name "{{$ingress.Name}}";
```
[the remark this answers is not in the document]
mikestephen: Looks like a good suggestion to me

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -39,12 +39,12 @@ http {
     {{if .LogFormat -}}
     log_format  main {{if .LogFormatEscaping}}escape={{ .LogFormatEscaping }} {{end}}
                      {{range $i, $value := .LogFormat -}}
-                     {{with $value}}'{{if $i}} {{end}}{{$value}}'
+                     {{with $value}}'{{if $i}} {{end}}{{$value}} resource_type="$resource_type" resource_name="$resource_name" resource_namespace="$resource_namespace" service="$service"'
```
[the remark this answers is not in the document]
mikestephen: So this change should be reverted, and it's up to the user to specify these placeholders in their log format in the ConfigMap if they desire?

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -39,12 +39,12 @@ http {
...
                      {{end}}{{end}};
     {{- else -}}
     log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                       '$status $body_bytes_sent "$http_referer" '
-                      '"$http_user_agent" "$http_x_forwarded_for"';
+                      '"$http_user_agent" "$http_x_forwarded_for" resource_type="$resource_type" resource_name="$resource_name" resource_namespace="$resource_namespace" service="$service"';
```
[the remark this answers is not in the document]
mikestephen: So this would become?:
```suggestion
                      '"$http_user_agent" "$http_x_forwarded_for" '
                      'resource_type="$resource_type" '
                      'resource_name="$resource_name" '
                      'resource_namespace="$resource_namespace" '
                      'service="$service"';
```

internal/configs/version1/nginx-plus.tmpl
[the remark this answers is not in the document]
LorcanMcVeigh: yes

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -136,15 +141,19 @@ server {
 
 	{{- range $location := $server.JWTRedirectLocations}}
 	location {{$location.Name}} {
+		set $service "nil";
```
mikestephen, line 143: Looks like in some places such as here, the default value is "nil", whereas in other places it's "". Is there a need to have "nil" instead of ""?

LorcanMcVeigh: you need a value in between the `""` otherwise nginx conf will generate that like as `set $service ;` and cause errors and not load/reload properly, `nil` was the first one that came to mind. But it turns out that its not needed there or in some other places anyway so I removed them

docs-web/configuration/global-configuration/configmap-resource.md
```diff
@@ -205,7 +205,22 @@ See the doc about [VirtualServer and VirtualServerRoute resources](/nginx-ingres
    * - ``log-format``
      - Sets the custom `log format <https://nginx.org/en/docs/http/ngx_http_log_module.html#log_format>`_ for HTTP and HTTPS traffic. For convenience, it is possible to define the log format across multiple lines (each line separated by ``\n``). In that case, the Ingress Controller will replace every ``\n`` character with a space character. All ``'`` characters must be escaped.
      - See the `template file <https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/version1/nginx.tmpl>`_ for the access log.
-     - 
+     - ```kind: ConfigMap
```
Dean-Coakley, line 208: This is invalid RST syntax and doesn't render correctly. Also, note that all other examples link to some example in https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/ which will give you access to markdown syntax instead.

I think it will be better to add an example there instead. - What do you think?

I assume that's what @pleshakov had in mind from https://github.com/nginxinc/kubernetes-ingress/pull/1231/files#r525617497

examples-of-custom-resources/basic-tcp-udp/global-configuration.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: k8s.nginx.org/v1alpha1
 kind: GlobalConfiguration 
 metadata:
   name: nginx-configuration
-  namespace: nginx-ingress
```
Dean-Coakley, line 5: From debugging? Revert?

examples/complete-example/cafe-ingress.yaml
```diff
@@ -2,8 +2,9 @@ apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: cafe-ingress
+  namespace: nginx-ingress
 spec:
-  # ingressClassName: nginx # use only with k8s version >= 1.18.0
+  ingressClassName: nginx # use only with k8s version >= 1.18.0
```
Dean-Coakley, line 7: From debugging? Revert?

docs-web/configuration/global-configuration/configmap-resource.md
```diff
@@ -205,7 +205,7 @@ See the doc about [VirtualServer and VirtualServerRoute resources](/nginx-ingres
    * - ``log-format``
      - Sets the custom `log format <https://nginx.org/en/docs/http/ngx_http_log_module.html#log_format>`_ for HTTP and HTTPS traffic. For convenience, it is possible to define the log format across multiple lines (each line separated by ``\n``). In that case, the Ingress Controller will replace every ``\n`` character with a space character. All ``'`` characters must be escaped.
      - See the `template file <https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/version1/nginx.tmpl>`_ for the access log.
-     - 
+     - `<https://github.com/nginxinc/kubernetes-ingress/tree/update-log-format/examples/custom-log-format>`
```
Dean-Coakley, line 208: This is invalid RST syntax. (See [deploy preview](https://deploy-preview-1231--nginx-kubernetes-ingress.netlify.app/nginx-ingress-controller/configuration/global-configuration/configmap-resource/#logging) )

I think it has to be something like:
```suggestion
     - `Custom Log Format <https://github.com/nginxinc/kubernetes-ingress/tree/update-log-format/examples/custom-log-format>`_.
```

examples/complete-example/cafe-ingress.yaml
Dean-Coakley: Namespace still needs revert here.

internal/configs/virtualserver.go
```diff
@@ -359,6 +359,7 @@ func (vsc *virtualServerConfigurator) GenerateVirtualServerConfig(
 	var vsrErrorPagesRouteIndex = make(map[string]int)
 	var vsrLocationSnippetsFromVs = make(map[string]string)
 	var vsrPoliciesFromVs = make(map[string][]conf_v1.PolicyReference)
+	IsVSR := false
```
Dean-Coakley, line 362: There doesn't seem to be a need to export `IsVSR` in this context?
```suggestion
	isVSR := false
```

Dean-Coakley (approved):

internal/configs/version1/nginx-plus.ingress.tmpl
mikestephen: Thanks for explaining @LorcanMcVeigh !

lucacome (approved):

PR #1232, opened 2020-11-17
Change imagePullSecrets of helm chart to type string

Dean-Coakley:
### Proposed changes
In order to be listed in the [Rancher App Catalog](https://github.com/rancher/helm3-charts) we need to be able to set `imagePullSecrets` so that deployments created by Rancher can pull from private docker registries.

However, Rancher only support the following datatypes: `string`, `multiline`, `boolean`, `int`, `enum`, `password`, `storageclass`, `hostname`, `pvc`, `secret`.

This does not allow to specify a string array as required. This PR changes to accept one string instead of an array. This doesn't affect any known use cases of this field.

### Checklist


pleshakov:
@Dean-Coakley 
will this be added to the release 1.9 branch also?
If that is the case, it's better to have both the old and new parameters, so that we don't introduce any breaking changes in 1.9.* releases.

Dean-Coakley:
@pleshakov Yeah that works for me. 

Then I assume we are in agreement that the field `controller.serviceAccount.imagePullSecrets` will be removed in 1.10 release.

pleshakov (approved):

lucacome (approved):

PR #1233, opened 2020-11-17
Annotate tcp metrics with k8s object labels

LorcanMcVeigh:
### Proposed changes
Add k8s object labels to tcp metrics. 

### Checklist


internal/k8s/controller.go
```diff
@@ -2324,11 +2325,21 @@ func (lbc *LoadBalancerController) createTransportServer(transportServer *conf_v
 
 		endps := getIPAddressesFromEndpoints(podEndps)
 		endpoints[endpointsKey] = endps
+
+		if lbc.isNginxPlus || lbc.isLatencyMetricsEnabled {
```
Dean-Coakley, line 2329: Should this ensure that prometheus is enabled too?

internal/configs/configurator.go
```diff
@@ -454,6 +468,86 @@ func (cnf *Configurator) AddOrUpdateVirtualServers(virtualServerExes []*VirtualS
...
+			newPeers[labelKey] = true
+			newPeersIPs = append(newPeersIPs, labelKey)
+		}
+	}
+
+	key := fmt.Sprintf("%v/%v", transportServerEx.TransportServer.Namespace, transportServerEx.TransportServer.Name)
+
+	removedPeers := findRemovedKeys(cnf.metricLabelsIndex.transportServerUpstreamPeers[key], newPeers)
+	cnf.metricLabelsIndex.transportServerUpstreamPeers[key] = newPeersIPs
+	cnf.latencyCollector.UpdateUpstreamServerPeerLabels(upstreamServerPeerLabels)
```
pleshakov, line 504: could you remove all the code related to the `latencyCollector`? That's because `latencyCollector` doesn't support TransportServers

internal/configs/configurator.go
```diff
@@ -454,6 +468,86 @@ func (cnf *Configurator) AddOrUpdateVirtualServers(virtualServerExes []*VirtualS
...
+	cnf.latencyCollector.DeleteUpstreamServerPeerLabels(removedPeers)
+
+	removedUpstreams := findRemovedKeys(cnf.metricLabelsIndex.transportServerUpstreams[key], newUpstreams)
+	cnf.latencyCollector.UpdateUpstreamServerLabels(labels)
+	cnf.metricLabelsIndex.transportServerUpstreams[key] = newUpstreamsNames
+
+	cnf.latencyCollector.DeleteUpstreamServerLabels(removedUpstreams)
+	cnf.latencyCollector.DeleteMetrics(removedPeers)
+
+	if cnf.isPlus {
```
pleshakov, line 514: cnf.isPlus will always be true because we only need to update labels for NGINX Plus. Could you remove this `if`?

internal/configs/configurator.go
```diff
@@ -454,6 +468,86 @@ func (cnf *Configurator) AddOrUpdateVirtualServers(virtualServerExes []*VirtualS
...
+		cnf.labelUpdater.DeleteStreamServerZoneLabels(removedZones)
+	}
+}
+
+func (cnf *Configurator) deleteTransportServerMetricsLabels(key string) {
+	cnf.latencyCollector.DeleteUpstreamServerLabels(cnf.metricLabelsIndex.transportServerUpstreams[key])
+	cnf.latencyCollector.DeleteUpstreamServerPeerLabels(cnf.metricLabelsIndex.transportServerUpstreamPeers[key])
+	cnf.latencyCollector.DeleteMetrics(cnf.metricLabelsIndex.transportServerUpstreamPeers[key])
+
+	if cnf.isPlus {
```
pleshakov, line 540: this will always be true, because we only need to update labels for NGINX Plus.  could you remove this `if`?

internal/configs/configurator.go
```diff
@@ -454,6 +468,86 @@ func (cnf *Configurator) AddOrUpdateVirtualServers(virtualServerExes []*VirtualS
...
+		newZones[transportServerEx.TransportServer.Name] = true
+		removedZones := findRemovedKeys(cnf.metricLabelsIndex.transportServerServerZones[key], newZones)
+		cnf.metricLabelsIndex.transportServerServerZones[key] = newZonesNames
+		cnf.labelUpdater.UpdateStreamServerZoneLabels(streamServerZoneLabels)
+		cnf.labelUpdater.DeleteStreamServerZoneLabels(removedZones)
+	}
+}
+
+func (cnf *Configurator) deleteTransportServerMetricsLabels(key string) {
+	cnf.latencyCollector.DeleteUpstreamServerLabels(cnf.metricLabelsIndex.transportServerUpstreams[key])
```
pleshakov, line 536: similarly to the previous comment, we don't need to update latency collector because it doesn't support TransportServers

internal/configs/configurator.go
```diff
@@ -454,6 +468,86 @@ func (cnf *Configurator) AddOrUpdateVirtualServers(virtualServerExes []*VirtualS
...
+
+		newZones[transportServerEx.TransportServer.Name] = true
+		removedZones := findRemovedKeys(cnf.metricLabelsIndex.transportServerServerZones[key], newZones)
+		cnf.metricLabelsIndex.transportServerServerZones[key] = newZonesNames
+		cnf.labelUpdater.UpdateStreamServerZoneLabels(streamServerZoneLabels)
+		cnf.labelUpdater.DeleteStreamServerZoneLabels(removedZones)
+	}
+}
+
+func (cnf *Configurator) deleteTransportServerMetricsLabels(key string) {
```
pleshakov, line 535: this function is not used. Could you start using it? :) (In `DeleteTransportServer` )

internal/configs/configurator.go
```diff
@@ -480,6 +574,10 @@ func (cnf *Configurator) addOrUpdateTransportServer(transportServerEx *Transport
 		return fmt.Errorf("Error generating TransportServer config %v: %v", name, err)
 	}
 
+	if (cnf.isPlus && cnf.isPrometheusEnabled) || cnf.isLatencyMetricsEnabled {
```
pleshakov, line 577: we don't need to check for ` cnf.isLatencyMetricsEnabled`, because the latency collector doesn't support TransportServer

internal/k8s/controller.go
pleshakov: @Dean-Coakley you're right. But there are other similar checks in `createIngressEx` and `createVirtualServerEx`. Better to change in those places as well (the change is small, ok to do it along the way)

go.sum
```diff
@@ -240,6 +240,7 @@ github.com/google/go-cmp v0.3.1 h1:Xye71clBPdm5HgqGwUkwhbynsUJZhDbS20FvLhQ2izg=
 github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
 github.com/google/go-cmp v0.4.0 h1:xsAVV57WRhGj6kEIi8ReJzQlHHqcBYCElAvkovg3B/4=
 github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/google/go-cmp v0.5.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
```
pleshakov, line 243: Could you run `go mod tidy` to clean up unused dependencies?

pleshakov (changes requested):
Hi @LorcanMcVeigh 

Please see my comments and suggestions

internal/configs/configurator.go
```diff
@@ -148,10 +154,18 @@ func createUpstreamServerLabels(svcName string, resourceType string, resourceNam
...
+	return []string{svcName, resourceType, resourceName, resourceNamespace}
+}
+
 func createServerZoneLabels(resourceType string, resourceName string, resourceNamespace string) []string {
 	return []string{resourceType, resourceName, resourceNamespace}
 }
 
+func createStreamServerZoneLabels(resourceType string, resourceName string, resourceNamespace string) []string {
+	return []string{resourceType, resourceName, resourceNamespace}
+}
```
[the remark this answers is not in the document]
LorcanMcVeigh: I agree these functions can be removed, all they do is return the parameters passed which is almost redundant because it can easily be done with one line of code and not calling a function e.g. ```labels = []string{label1, label2, label3......}```
I don't see how it would prevent bugs because there are no checks in these functions, they simply return the parameters ?

internal/k8s/controller.go
```diff
@@ -2324,11 +2325,21 @@ func (lbc *LoadBalancerController) createTransportServer(transportServer *conf_v
...
+			for _, endpoint := range podEndps {
+				podsByIP[endpoint.Address] = configs.PodInfo{
+					Name:         endpoint.PodName,
+					MeshPodOwner: endpoint.MeshPodOwner,
```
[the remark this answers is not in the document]
LorcanMcVeigh: so should I just change this to be `podsByIP[endpoint.Address] = endpoint.PodName` since we don't need the pod name from the `PodInfo` struct ?

LorcanMcVeigh:
force push was accidental, rebased against master

internal/configs/configurator.go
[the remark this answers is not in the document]
Dean-Coakley: These funcs serve the following purposes/bug preventions:
* Ensures the correct number of labelNames are set per nginx context. Prevents prometheus label cardinality error. (e.g: 3 label names, 4 label values)
* Ensures label names are always set in the correct order. Prevents label mismatch (e.g: resourceType=myResourceName)

From my perspective, I'm happy for this to be marked as resolved and proceed - we can take a look at a better solution at a later date. 

Possibly just removing these funcs and finding a way to add unit test coverage for future mistakes when extending labels.

pleshakov (changes requested):
@LorcanMcVeigh 
Please see my comments and suggestions. I also discovered a bug, which I hand't noticed during the previous review :(

internal/configs/configurator.go
```diff
@@ -454,6 +452,70 @@ func (cnf *Configurator) AddOrUpdateVirtualServers(virtualServerExes []*VirtualS
...
+		cnf.labelUpdater.DeleteStreamUpstreamServerLabels(removedUpstreams)
+
+		streamServerZoneLabels := make(map[string][]string)
+		newZones := make(map[string]bool)
+		newZonesNames := []string{transportServerEx.TransportServer.Spec.Host}
+
+		streamServerZoneLabels[transportServerEx.TransportServer.Spec.Host] = []string{
+			"transportserver", transportServerEx.TransportServer.Name, transportServerEx.TransportServer.Namespace}
+
+		newZones[transportServerEx.TransportServer.Spec.Host] = true
```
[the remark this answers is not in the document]
LorcanMcVeigh: Im confused @pleshakov , 
`transportServerEx.TransportServer.Spec.Host` will not be empty for non-TLSPassthrough transport servers? Is [this](https://github.com/nginxinc/kubernetes-ingress/blob/annotate-tcp/internal/configs/configurator_test.go#L1008) example TS wrong so? and I would've thought it would be the other way around. Using `transporterServerEx.TransportServer.Spec.Listener.Name` would be bad as it would be empty for non-TLSPassthrough TS's.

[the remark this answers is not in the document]
LorcanMcVeigh: Also for both [VS](https://github.com/nginxinc/kubernetes-ingress/blob/annotate-tcp/internal/configs/configurator.go#L358) and [Ingress](https://github.com/nginxinc/kubernetes-ingress/blob/annotate-tcp/internal/configs/configurator.go#L202) we use the host name as the map key, is there a reason to not follow that for TransportServers?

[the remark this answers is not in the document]
pleshakov: Please take a look at the examples:
- TLSPassthrogh transport server -- https://github.com/nginxinc/kubernetes-ingress/blob/master/examples-of-custom-resources/tls-passthrough/transport-server-passthrough.yaml
- TCP and UDP transport servers -- https://github.com/nginxinc/kubernetes-ingress/tree/master/examples-of-custom-resources/basic-tcp-udp

> Also for both VS and Ingress we use the host name as the map key, is there a reason to not follow that for TransportServers?

yep. there is no host for TCP/UDP load balancing

internal/configs/configurator.go
```diff
@@ -902,6 +904,10 @@ func (cnf *Configurator) DeleteVirtualServer(key string) error {
 
 // DeleteTransportServer deletes NGINX configuration for the TransportServer resource.
 func (cnf *Configurator) DeleteTransportServer(key string) error {
+	if cnf.isPlus && cnf.isPrometheusEnabled {
+		cnf.deleteTransportServerMetricsLabels(fmt.Sprintf(key))
```
pleshakov, line 908:
```suggestion
		cnf.deleteTransportServerMetricsLabels(key)
```

internal/configs/configurator_test.go
```diff
@@ -1126,4 +1128,82 @@ func TestUpdateTransportServerMetricsLabels(t *testing.T) {
...
+				ResourceNamespace: tsExTLS.TransportServer.Namespace,
+			},
+		},
+	}
+
+	streamUpstreamServerLabels = map[string][]string{
+		"upstream-3": {"service-3", "transportserver", "test-transportserver-tls", "default"},
+	}
+
+	streamUpstreamServerLabels = map[string][]string{
```
pleshakov, line 1172: streamUpstreamServerLabels is initialized twice with the same value

pleshakov (approved):
🚀

Dean-Coakley (approved):

PR #1235, opened 2020-11-17
Use buildkit secret flag for NGINX plus images

ciarams87:
The new --secret flag for docker build allows the user to pass secret
information to be used in the Dockerfile for building docker images
in a safe way that will not end up stored in the final image. This commit
introduces this change for transferring the cert and key for building
the NGINX Plus images.

### Checklist

- [] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,16 @@ ifneq ($(BUILD_IN_CONTAINER),1)
 	CGO_ENABLED=0 GO111MODULE=on GOFLAGS='$(GOFLAGS)' GOOS=linux go build -installsuffix cgo -ldflags "-w -X main.version=${VERSION} -X main.gitCommit=${GIT_COMMIT}" -o nginx-ingress github.com/nginxinc/kubernetes-ingress/cmd/nginx-ingress
 endif
 
-container: test verify-codegen verify-crds binary certificate-and-key
+prepare-license-secrets:
+ifneq (,$(findstring PlusForOpenShift,$(DOCKERFILE)))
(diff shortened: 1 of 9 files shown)


sundaram1995 (approved):

Makefile
```diff
@@ -55,7 +55,19 @@ ifneq ($(BUILD_IN_CONTAINER),1)
...
+DOCKER_BUILD_OPTIONS += --secret id=nginx-repo.crt,src=tempdir/nginx-repo.crt --secret id=nginx-repo.key,src=tempdir/nginx-repo.key
+endif
+
+install-plus-openshift: install-plus
+ifneq (,$(findstring PlusForOpenShift,$(DOCKERFILE)))
+	base64 rhel_license > tempdir/rhel_license
+DOCKER_BUILD_OPTIONS += --secret id=rhel_license,src=tempdir/rhel_license
+endif
+
+container: test verify-codegen verify-crds binary certificate-and-key install-plus-openshift
```
pleshakov, line 70: i would choose a slightly different names for the new targets. Something like `prepare-plus-license-secret` and `prepare-rhel-license-secret` or a combined `prepare-license-secrets`.  Imho `install` reminds of `make install`, which typically installs compiled software on the machine.

pleshakov (approved):
@ciarams87 Looks good!!

Makefile
ciarams87: Thanks, great point, naming of things is not one of my strengths :) I've updated it now!

PR #1238, opened 2020-11-18
Update controller-gen and CRDs

lucacome:
### Proposed changes
- Updated controller-gen to `0.4.1`
- Removed verify-crds hack and used github workflow to detect changes


.github/workflows/edge.yml
```diff
@@ -55,6 +55,9 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: '${{ env.GOLANG_VERSION }}'
+      - name: Check if CRDs changed
+        run: |
+          make update-crds && git diff --name-only --exit-code deployments/common/crds*
```
Dean-Coakley, line 60: `update-crds` also updates helm-chart/crds. Is it safe to assume that if manifest crds are up to date, helm-chart ones are too?

Dean-Coakley: I guess this matches the behaviour of `verify-crds` anyway. We made that assumption, even before `update-crds` updated helm crds.

Dean-Coakley (approved):

mikestephen (approved):

PR #1239, opened 2020-11-18
Add imagePullSecretName support to helm chart

Dean-Coakley:
### Proposed changes
Continuation of https://github.com/nginxinc/kubernetes-ingress/pull/1232 for stable release.

* Add imagePullSecretName so we can support Rancher App Catalog integration.
* `imagePullSecrets` remains for backwards compatibility, which will be removed in a later release.

### Checklist


Dean-Coakley:
Some outstanding questions:
* I don't see any no hard requirement for this to be included in 1.9.1 for the Rancher integration. @pleshakov Do you have any preference for this to be included or excluded in that release.
* Do we want to mark `imagePullSecrets` as deprecated? Or it will just be in 1.10 release notes?
  * If we do, should something like the following be added to all doc locations and values?: `"DEPRECATED: To be removed in future release. Please migrate to "imagePullSecretName"`

pleshakov:
@Dean-Coakley 
> I don't see any no hard requirement for this to be included in 1.9.1 for the Rancher integration. @pleshakov Do you have any preference for this to be included or excluded in that release?

Are we launching Rancher integration any time soon? If we don't include this now, does it mean we will need to create another minor release like 1.9.2 for the rancher integration?

> Do we want to mark imagePullSecrets as deprecated? Or it will just be in 1.10 release notes?

Yeah, I think it makes sense to add 👍

Dean-Coakley:
@pleshakov While not strictly required as the changes will be prepared on a branch on https://github.com/nginxinc/helm3-charts it makes sense to have all PRs there map closely to KIC release tag. 

So it would be best to include this in 1.9.1 for consistency.

pleshakov (approved):

deployments/helm-chart/templates/controller-serviceaccount.yaml
```diff
@@ -10,4 +10,8 @@ metadata:
 imagePullSecrets:
 {{ toYaml .Values.controller.serviceAccount.imagePullSecrets }}
 {{- end }}
+{{- if .Values.controller.serviceAccount.imagePullSecretName }}
```
mikestephen, line 13: If someone were to specify both `controller.serviceAccount.imagePullSecrets` and `controller.serviceAccount.imagePullSecrets` you'd end up with two `imagePullSecrets` blocks in the file. Should the behaviour be one takes precedence over the other? Maybe `controller.serviceAccount.imagePullSecretName` could have precedence since `controller.serviceAccount.imagePullSecrets` is being deprecated?

Dean-Coakley: Yeah it's good to protect against this.

mikestephen (approved):
LGTM!

PR #1240, opened 2020-11-18
Validate TLS and CA secrets

pleshakov:
### Proposed changes

Previously, the contents of the TLS and CA secrets wasn't validated. As a result, NGINX could be configured to use an invalid TLS cert and/or key. In that case, NGINX would fail to reload. More over, NGINX would keep failing to reload until that secret was fixed/removed or a resource referencing that secret was removed.

This PR brings validation of the contents of secrets:
* cert and key of TLS secret
* cert of CA secret

If a secret is invalid, NGINX will not be configured to use it.

### DIFF
--- a/internal/k8s/controller_test.go
+++ b/internal/k8s/controller_test.go
@@ -1364,7 +1364,7 @@ func TestGetIngressMTLSSecret(t *testing.T) {
 			Namespace: "default",
 		},
 		Type: SecretTypeCA,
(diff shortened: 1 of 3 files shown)


lucacome (approved):

mikestephen (approved):
Looks great to me!

Dean-Coakley (approved):

PR #1255, opened 2020-11-30
Add validation to ingress annotations and config map keys

mikestephen:
Validation of some annotations was not checking for valid times, sizes etc.
resulting in the user being able to specify settings which would cause Nginx
to fail to apply the changes.

### Proposed changes

### Checklist

### DIFF
--- a/pkg/apis/configuration/validation/common.go
+++ b/pkg/apis/configuration/validation/common.go
@@ -2,6 +2,7 @@ package validation
 
 import (
 	"fmt"
+	"github.com/nginxinc/kubernetes-ingress/internal/configs"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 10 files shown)


pkg/apis/configuration/validation/common.go
```diff
@@ -2,6 +2,7 @@ package validation
 
 import (
 	"fmt"
+	"github.com/nginxinc/kubernetes-ingress/internal/configs"
 	"regexp"
 	"strings"
```
lucacome, line 7:
```suggestion
import (
	"fmt"
	"regexp"
	"strings"

	"github.com/nginxinc/kubernetes-ingress/internal/configs"
	
```

lucacome (approved):
LGTM! Just one suggestion about imports so the linter is happy 😄

mikestephen:
Closing, as there is a better way to add validation which provides better user feedback in k8s status messages.

PR #1256, opened 2020-11-30
Add LocalSecretStore

pleshakov:
### Proposed changes

This PR introduces LocalSecretStore for storing secrets (TLS, CA, JWK).
LocalSecretStore validates secrets and updates the secret contents on the file system when a secret is updated.
The Ingress Controller code was simplified as it is no longer necessary to update the secret contents on the file system on every regeneration of Ingress (or VS) config.

### DIFF
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -154,6 +154,7 @@ type LoadBalancerController struct {
 	isNginxReady                  bool
 	isLatencyMetricsEnabled       bool
 	configuration                 *Configuration
+	secretStore                   SecretStore
 }
 
 var keyFunc = cache.DeletionHandlingMetaNamespaceKeyFunc
@@ -288,6 +289,8 @@ func NewLoadBalancerController(input NewLoadBalancerControllerInput) *LoadBalanc
 	}
 
 	lbc.configuration = NewConfiguration(lbc.HasCorrectIngressClass, input.IsNginxPlus, input.VirtualServerValidator)
+
+	lbc.secretStore = NewLocalSecretStore(lbc.configurator)
 
 	lbc.updateIngressMetrics()
 	return lbc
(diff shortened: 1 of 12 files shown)


mikestephen (approved):
Looks good to me Michael!

internal/k8s/controller.go
```diff
@@ -1794,37 +1728,38 @@ func (lbc *LoadBalancerController) createIngressEx(ing *networking.Ingress, vali
...
 			if err != nil {
-				glog.Warningf("Error retrieving secret %v for Ingress %v: %v", secretName, ing.Name, err)
-				secret = nil
-			} else {
-				err = ValidateJWKSecret(secret)
-				if err != nil {
-					glog.Warningf("Error validating secret %v for Ingress %v: %v", secretName, ing.Name, err)
-					secret = nil
-				}
+				glog.Warningf("Error trying to get the secret %v for Ingress %v: %v", secretName, ing.Name, err)
```
Dean-Coakley, line 1756: Should we add some context about what namespace to at least one of the logs about the resources?
```suggestion
				glog.Warningf("Error trying to get the secret %v for Ingress %v/%v: %v", secretKey, ing.Namespace, ing.Name, err)
```

internal/k8s/controller.go
```diff
@@ -2227,58 +2164,84 @@ func (lbc *LoadBalancerController) getPolicies(policies []conf_v1.PolicyReferenc
 	return result, errors
 }
 
-func (lbc *LoadBalancerController) addJWTSecrets(policies []*conf_v1alpha1.Policy, jwtKeys map[string]*api_v1.Secret) error {
+func (lbc *LoadBalancerController) addJWTSecrets(secrets map[string]*configs.SecretReference, policies []*conf_v1alpha1.Policy) error {
```
Dean-Coakley, line 2167: Should we call this var `secretRefs` so the reader can assume it's type to be `map[string]*configs.SecretReference` instead of `map[string]*api_v1.Secret` without reading the func signature? 

```suggestion
func (lbc *LoadBalancerController) addJWTSecrets(secretRefs map[string]*configs.SecretReference, policies []*conf_v1alpha1.Policy) error {
```

Dean-Coakley (approved):
LGTM.

Some optional suggestions/questions:

tests/suite/test_jwt_policies_vsr.py
```diff
@@ -352,7 +352,7 @@ def test_jwt_policy_delete_secret(
         assert f"Request ID:" in resp1.text
         assert crd_info["status"]["state"] == "Warning"
         assert (
-            f'"{v_s_route_setup.route_m.namespace}/{secret}" which does not exist'
+            "references an invalid Secret: secret doesn't exist or of an unsupported type"
```
vepatel, line 355: No more secret name in warning message?

pleshakov: good point. I can improve those messages in the next PR that implements remaining warnings

internal/k8s/controller.go
pleshakov: renamed

PR #1265, opened 2020-12-04
Support warnings for secrets

pleshakov:
### Proposed changes
- Support warnings for TLS and JWK secrets in Ing
- Support warnings for TLS secrets in VS

### DIFF
--- a/internal/configs/configurator.go
+++ b/internal/configs/configurator.go
@@ -230,19 +230,20 @@ func (cnf *Configurator) deleteIngressMetricsLabels(key string) {
 }
 
 // AddOrUpdateIngress adds or updates NGINX configuration for the Ingress resource.
-func (cnf *Configurator) AddOrUpdateIngress(ingEx *IngressEx) error {
-	if err := cnf.addOrUpdateIngress(ingEx); err != nil {
-		return fmt.Errorf("Error adding or updating ingress %v/%v: %v", ingEx.Ingress.Namespace, ingEx.Ingress.Name, err)
+func (cnf *Configurator) AddOrUpdateIngress(ingEx *IngressEx) (Warnings, error) {
+	warnings, err := cnf.addOrUpdateIngress(ingEx)
+	if err != nil {
+		return warnings, fmt.Errorf("Error adding or updating ingress %v/%v: %v", ingEx.Ingress.Namespace, ingEx.Ingress.Name, err)
 	}
 
 	if err := cnf.nginxManager.Reload(nginx.ReloadForOtherUpdate); err != nil {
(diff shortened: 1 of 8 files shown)


mikestephen (approved):
Looks good to me!

internal/configs/configurator.go
```diff
@@ -1188,80 +1201,92 @@ func generateApResourceFileContent(apResource *unstructured.Unstructured) []byte
...
 		if err != nil {
-			return fmt.Errorf("Error adding or updating mergeableIngress %v/%v: %v", m.Master.Ingress.Namespace, m.Master.Ingress.Name, err)
+			return allWarnings, fmt.Errorf("Error adding or updating mergeableIngress %v/%v: %v", m.Master.Ingress.Namespace, m.Master.Ingress.Name, err)
 		}
+		allWarnings.Add(warnings)
 	}
 
 	if err := cnf.nginxManager.Reload(nginx.ReloadForOtherUpdate); err != nil {
-		return fmt.Errorf("Error when reloading NGINX when updating %v: %v", resource.GetKind(), err)
+		return allWarnings, fmt.Errorf("Error when reloading NGINX when updating %v: %v", resource.GetKind(), err)
```
Dean-Coakley, line 1224: Would it be good to include resNs/resName here?
```suggestion
		return allWarnings, fmt.Errorf("Error when reloading NGINX when updating %v %v/%v: %v", resource.GetKind(), resource.GetNamespace(), resource.GetName(), err)
```

internal/configs/ingress.go
```diff
@@ -246,16 +252,19 @@ func generateNginxCfg(ingEx *IngressEx, apResources map[string]string, isMinion
 			Annotations: ingEx.Ingress.Annotations,
 		},
 		SpiffeClientCerts: staticParams.NginxServiceMesh && !cfgParams.SpiffeServerCerts,
-	}
+	}, allWarnings
 }
 
-func generateJWTConfig(secretRefs map[string]*secrets.SecretReference, cfgParams *ConfigParams, redirectLocationName string) (*version1.JWTAuth, *version1.JWTRedirectLocation) {
+func generateJWTConfig(owener runtime.Object, secretRefs map[string]*secrets.SecretReference, cfgParams *ConfigParams,
```
Dean-Coakley, line 258:
```suggestion
func generateJWTConfig(owner runtime.Object, secretRefs map[string]*secrets.SecretReference, cfgParams *ConfigParams,
```

Dean-Coakley (approved):

internal/configs/ingress.go
pleshakov: good catch

internal/configs/configurator.go
pleshakov: same enhancement can be applied to the other app protect related methods. better to do it in a separate PR

PR #1266, opened 2020-12-04
Add workqueue prometheus metrics

Dean-Coakley:
### Proposed changes
Add the following prometheus metrics:
* `nginx_ingress_controller_workqueue_depth`
*  `nginx_ingress_controller_workqueue_queue_duration_seconds`
* `nginx_ingress_controller_workqueue_work_duration_seconds`

### Checklist

### DIFF
--- a/docs-web/logging-and-monitoring/prometheus.md
+++ b/docs-web/logging-and-monitoring/prometheus.md
@@ -37,12 +37,11 @@ The Ingress Controller exports the following metrics:
   * `controller_ingress_resources_total`. Number of handled Ingress resources. This metric includes the label type, that groups the Ingress resources by their type (regular, [minion or master](/nginx-ingress-controller/configuration/ingress-resources/cross-namespace-configuration)). **Note**: The metric doesn't count minions without a master.
   * `controller_virtualserver_resources_total`. Number of handled VirtualServer resources.
   * `controller_virtualserverroute_resources_total`. Number of handled VirtualServerRoute resources. **Note**: The metric counts only VirtualServerRoutes that have a reference from a VirtualServer.
-* Kubernetes Cluster metrics
-  * `controller_workqueue_depth` Current depth of workqueue.
-  * `controller_workqueue_queue_duration_second`. How long in seconds an item stays in workqueue before being requested.
-  * `controller_workqueue_work_duration_seconds`. How long in seconds processing an item from workqueue takes.
+  * Workqueue metrics. **Note**: the workqueue is a queue used by the Ingress Controller to process changes to the relevant resources in the cluster like Ingress resources. The Ingress Controller uses only one queue. The metrics for that queue will have the label `name="taskQueue"`
+    * `workqueue_depth`. Current depth of the workqueue.
+    * `workqueue_queue_duration_second`. How long in seconds an item stays in the workqueue before being requested.
+    * `workqueue_work_duration_seconds`. How long in seconds processing an item from the workqueue takes.
 
-
-**Note**: all metrics have the namespace nginx_ingress. For example, nginx_ingress_controller_nginx_reloads_total.
+**Note**: all metrics have the namespace `nginx_ingress`. For example, `nginx_ingress_controller_nginx_reloads_total`.
 
 **Note**: all metrics include the label `class`, which is set to the class of the Ingress Controller. The class is configured via the `-ingress-class` command-line argument.
--- a/docs-web/logging-and-monitoring/prometheus.md
+++ b/docs-web/logging-and-monitoring/prometheus.md
@@ -27,7 +27,7 @@ The Ingress Controller exports the following metrics:
   * Exported by NGINX/NGINX Plus. Refer to the [NGINX Prometheus Exporter developer docs](https://github.com/nginxinc/nginx-prometheus-exporter#exported-metrics) to find more information about the exported metrics.
   * There is a Grafana dashboard for NGINX Plus metrics located in the root repo folder.
   * Calculated by the Ingress Controller:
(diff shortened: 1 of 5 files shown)


internal/k8s/task_queue.go
```diff
@@ -103,30 +102,19 @@ func (tq *taskQueue) Shutdown() {
 // kind represents the kind of the Kubernetes resources of a task
 type kind int
 
+// resources
 const (
-	// ingress resource
```
mikestephen, line 107: Nice job, I don't think these comments communicated anything the constant name didn't.

internal/metrics/collectors/workqueue.go
```diff
@@ -0,0 +1,117 @@
...
+				ConstLabels: constLabels,
+			},
+			[]string{"name"},
+		),
+		latency: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Namespace:   metricsNamespace,
+				Subsystem:   workqueueSubsystem,
+				Name:        "queue_duration_seconds",
+				Help:        "How long in seconds an item stays in workqueue before being requested",
```
mikestephen, line 35: Should this be something like:
`How long in seconds an item stays in workqueue before being processed`

internal/metrics/collectors/workqueue.go
```diff
@@ -0,0 +1,117 @@
...
+				Help:        "Current depth of workqueue",
+				ConstLabels: constLabels,
+			},
+			[]string{"name"},
+		),
+		latency: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Namespace:   metricsNamespace,
+				Subsystem:   workqueueSubsystem,
+				Name:        "queue_duration_seconds",
```
mikestephen, line 34: I wonder should this metric and the next one be recorded in milliseconds instead of seconds? I'm just thinking processing might often take less than a second.

mikestephen (approved):
I'm not sure yet how the queue depth will work, because I'm worried that the metrics will be scraped say every second, and the depth metric will fluctuate a lot to reflect the queue depth at any moment in time. I have a feeling we might need a high-water mark or something here, so the metric more reflects the largest depth recorded in the last 10 seconds or something like that.
That looks like it'll be in a future PR?, so approving this one.

pleshakov (commented):
Hi @Dean-Coakley 

Looks good! Mostly docs comments/suggestions

@mikestephen 
> I'm not sure yet how the queue depth will work, because I'm worried that the metrics will be scraped say every second, and the depth metric will fluctuate a lot to reflect the queue depth at any moment in time. I have a feeling we might need a high-water mark or something here, so the metric more reflects the largest depth recorded in the last 10 seconds or something like that.
That looks like it'll be in a future PR?, so approving this one.


it is common that scape intervals are much more than 1 second, like 15 seconds (for example, https://www.robustperception.io/keep-it-simple-scrape_interval-id )

I think if there is a problem with processing items, the other two metrics will also show that - `work_duration_seconds` and `queue_duration_seconds` will increase.  Perhaps `depth` metric is more useful for real-time troubleshooting? Ex.  an IC pod for some reason doesn't process changes in the cluster -> admins can take a look at the metric via `/metrics` on the IC pod and they will see that the queue keeps growing and growing`

At the same time, perhaps the [counter](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/metrics/prometheus/workqueue/metrics.go#L47) for all the added elements can help? this way it will be possible to calculate the processing rate in Prometheus based on that metric  and see if there was an influx of changes at some interval.

Also, right now we add to the queue changes to all endpoints in the cluster (which change frequently), even the ones the IC not interested in. The IC  processes them very quickly and quickly ignores. So perhaps in the future we can also filter them out, so that those changes don't skew our queue metrics.

internal/metrics/collectors/processes.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-// NginxProcessesMetricsCollector implements NginxPorcessesCollector interface and prometheus.Collector interface
+// NginxProcessesMetricsCollector implements NginxProcessesCollector interface and prometheus.Collector interface
```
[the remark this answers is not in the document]
Dean-Coakley: Removed this reference.

docs-web/logging-and-monitoring/prometheus.md
```diff
@@ -27,17 +27,22 @@ The Ingress Controller exports the following metrics:
...
+  * `ingress_resources_total`. Number of handled Ingress resources. This metric includes the label type, that groups the Ingress resources by their type (regular, [minion or master](/nginx-ingress-controller/configuration/ingress-resources/cross-namespace-configuration)). **Note**: The metric doesn't count minions without a master.
+  * `virtualserver_resources_total`. Number of handled VirtualServer resources.
+  * `virtualserverroute_resources_total`. Number of handled VirtualServerRoute resources. **Note**: The metric counts only VirtualServerRoutes that have a reference from a VirtualServer.
+* Kubernetes Cluster metrics
+  * `workqueue_depth` Current depth of workqueue.
+  * `workqueue_queue_duration_second`. How long in seconds an item stays in workqueue before being requested.
+  * `workqueue_work_duration_seconds`. How long in seconds processing an item from workqueue takes.
+
+
+**Note**: all metrics have the namespace `nginx_ingress_controller`. For example, `nginx_ingress_controller_nginx_reloads_total`.
```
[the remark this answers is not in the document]
Dean-Coakley: Right, I saw https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/metrics/collectors/collectors.go#L3 which seemed to applied to all metrics but I missed that nginx/nginxplus metrics are exported in the same place.

pleshakov (approved):

Dean-Coakley:
@mikestephen Are you still happy for this to be merged?

https://github.com/nginxinc/kubernetes-ingress/pull/1266#discussion_r537382313 seems resolved to me.

And sounds like you're happy to take a look at high-water mark in a later PR if determined necessary?

internal/metrics/collectors/workqueue.go
mikestephen: Yep, that makes sense, thanks.

PR #1271, opened 2020-12-07
K8s log variables for splits & matches

LorcanMcVeigh:
### Proposed changes
This PR adds the k8s object variables to the log format for splits and matches.

### Checklist

### DIFF
--- a/examples/custom-log-format/README.md
+++ b/examples/custom-log-format/README.md
@@ -6,14 +6,16 @@ This example lets you set the log-format for NGINX using the configmap reosurce
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: oss-nginx-ingress
+  name: nginx-config
+  namespace: nginx-ingress
 data:
-  log-format:  $remote_addr - $remote_user [$time_local] \"$request\" $status $body_bytes_sent \"$http_referer\"  \"$http_user_agent\" \"$http_x_forwarded_for\" $resource_name $resource_type $resource_namespace $service;
+  log-format:  '$remote_addr - $remote_user [$time_local] "$request" $status $body_bytes_sent "$http_referer"  "$http_user_agent" "$http_x_forwarded_for" "$resource_name" "$resource_type" "$resource_namespace" "$service";'
 ```
 
-In addition to the built-in NGINX variables, you can also use the variables that the Ingress Controller configures:
-
-- $resource_type - The type of k8s resource. 
-- $resource_name - The name of the k8s resource
+In addition to the [built-in NGINX variables](https://nginx.org/en/docs/varindex.html), you can also use the variables that the Ingress Controller configures:
+- $resource_type - The type of kubernetes resource that handled the client request.
+- $resource_name - The name of the resource
 - $resource_namespace - The namespace the resource exists in.
-- $service - The service that exposes the resource.
+- $service - The name of the server the client request was sent to.
+
+**note** These variables are only available for Ingress, VirtualServer and VirtualServerRoute resources.
--- a/examples/custom-log-format/README.md
+++ b/examples/custom-log-format/README.md
@@ -1,6 +1,6 @@
 # Custom NGINX log format
 
-This example lets you set the log-format for NGINX using the configmap reosurce 
+This example lets you set the log-format for NGINX using the configmap resource
 
 ```yaml 
 kind: ConfigMap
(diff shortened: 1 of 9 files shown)


mikestephen (approved):

mikestephen (approved):

examples/custom-log-format/README.md
```diff
@@ -6,12 +6,9 @@ This example lets you set the log-format for NGINX using the configmap reosurce
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: nginx-config
```
pleshakov, line 9: the name we use in the default  `manifests` is `nginx-config` (see https://github.com/nginxinc/kubernetes-ingress/blob/master/deployments/common/nginx-config.yaml#L4 ). could we use the same name for consistency? I also suggest adding the `nginx-ingress` namespace. Also, please see a similar example with the configmap -- https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/custom-templates

examples/custom-log-format/README.md
```diff
@@ -6,12 +6,9 @@ This example lets you set the log-format for NGINX using the configmap reosurce
...
+  name: oss-nginx-ingress
 data:
-  log-format: |
-      compression '$remote_addr - $remote_user [$time_local] '
-                       '"$request" $status $bytes_sent '
-                       '"$http_referer" "$http_user_agent" "$gzip_ratio"'
+  log-format:  $remote_addr - $remote_user [$time_local] \"$request\" $status $body_bytes_sent \"$http_referer\"  \"$http_user_agent\" \"$http_x_forwarded_for\" $resource_name $resource_type $resource_namespace $service;
```
pleshakov, line 11: since  `$resource_name $resource_type $resource_namespace $service` variables can be empty, better to wrap them in `"` like `"$resource_name"`. This way when logs are parsed (by some tool), it will be possible to understand which variable are empty - in that case in the logs empty variables will become empty strings enclosed in `""`

also, you can wrap the whole format in `'` so that you don't need to escape every `"`. For example:
`log-format: '$remote_addr ... "$service"'`

examples/custom-log-format/README.md
```diff
@@ -6,12 +6,9 @@ This example lets you set the log-format for NGINX using the configmap reosurce
 kind: ConfigMap
```
pleshakov, line 6: few more notes about the example:

(1)the description has a typo in `reosurce`

(2) a few suggestions  and a correction about the variable descriptions:

In addition to the [built-in NGINX variables](https://nginx.org/en/docs/varindex.html), you can also use the variables that the Ingress Controller configures:

* `$resource_type` - The type of the Kubernetes resource, which handled the client request. Could be `ingress`, `virtualserver` or `virtualserverroute`.
* `$resource_name` - The name of the resource.
* `$resource_namespace` - The namespace the resource exists in.
* `$service` - The name of the service the client request was sent to.

internal/configs/version2/http.go
```diff
@@ -71,6 +71,7 @@ type Server struct {
 	EgressMTLS                *EgressMTLS
 	PoliciesErrorReturn       *Return
 	Namespace                 string
+	Name                      string
```
pleshakov, line 74: Would it make sense to rename these variables to `VSName` and `VSNamespace`? That will be consistent with `VSR*` variables. Also, the `Server.Name` field is a bit confusing because we have the `ServerName` variable, so `Server.VSName` will clear that confusion.

pleshakov (changes requested):
Hi @LorcanMcVeigh 

Please see my comments. Mostly about the example

examples/custom-log-format/README.md
```diff
@@ -6,14 +6,16 @@ This example lets you set the log-format for NGINX using the configmap reosurce
...
-In addition to the built-in NGINX variables, you can also use the variables that the Ingress Controller configures:
-
-- $resource_type - The type of k8s resource. 
-- $resource_name - The name of the k8s resource
+In addition to the [built-in NGINX variables](https://nginx.org/en/docs/varindex.html), you can also use the variables that the Ingress Controller configures:
+- $resource_type - The type of kubernetes resource that handled the client request.
+- $resource_name - The name of the resource
 - $resource_namespace - The namespace the resource exists in.
-- $service - The service that exposes the resource.
+- $service - The name of the server the client request was sent to.
```
pleshakov, line 19: service, not a server
```suggestion
- $service - The name of the service the client request was sent to.
```

examples/custom-log-format/README.md
```diff
@@ -6,14 +6,16 @@ This example lets you set the log-format for NGINX using the configmap reosurce
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: oss-nginx-ingress
+  name: nginx-config
+  namespace: nginx-ingress
 data:
-  log-format:  $remote_addr - $remote_user [$time_local] \"$request\" $status $body_bytes_sent \"$http_referer\"  \"$http_user_agent\" \"$http_x_forwarded_for\" $resource_name $resource_type $resource_namespace $service;
+  log-format:  '$remote_addr - $remote_user [$time_local] "$request" $status $body_bytes_sent "$http_referer"  "$http_user_agent" "$http_x_forwarded_for" "$resource_name" "$resource_type" "$resource_namespace" "$service";'
 ```
```
pleshakov, line 14: the typo in `reosurce` in the example description wasn't fixed (

examples/custom-log-format/README.md
```diff
@@ -6,14 +6,16 @@ This example lets you set the log-format for NGINX using the configmap reosurce
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: oss-nginx-ingress
+  name: nginx-config
+  namespace: nginx-ingress
 data:
```
pleshakov, line 11: `;` at the end of the log format is new. But it is not needed. Could you remove it?

pleshakov (approved):
Hi @LorcanMcVeigh 

Changes look good. There are still a few small problems. But I don't think another review is needed. Could you fix them before merging?

PR #1281, opened 2020-12-14
support for App Protect UDS

rafwegv:
### Proposed changes
Introduce support for App Protect User Defined Signatures. Add new CRD and the required code to handle them.

### Checklist


internal/k8s/app_protect_configuration.go
```diff
@@ -0,0 +1,459 @@
...
+					if err != nil {
+						errMsg := fmt.Sprintf("Error Parsing time from maxRevisionDatetime in %s: %v", policyObj.GetName(), err)
+						return &AppProtectPolicyEx{Obj: policyObj, IsValid: false, ErrorMsg: invalidTimestampErrorMsg}, fmt.Errorf(errMsg)
+					}
+					timeReq.MaxRevTime = &maxRevTime
+				}
+				sigReqs = append(sigReqs, SignatureReq{Tag: reqTag.(string), RevTimes: &timeReq})
+			}
+		}
+	}
```
Dean-Coakley, line 170: Could something like the following be done to reduce nesting?
```suggestion
	if !found {
		return nil, nil
	}
	for _, req := range list {
		requirement := req.(map[string]interface{})
		if reqTag, ok := requirement["tag"]; ok {
			timeReq := RevTimes{}
			if minRev, ok := requirement["minRevisionDatetime"]; ok {
				minRevTime, err := time.Parse(timeLayout, minRev.(string))
				if err != nil {
					errMsg := fmt.Sprintf("Error Parsing time from minRevisionDatetime in %s: %v", policyObj.GetName(), err)
					return &AppProtectPolicyEx{Obj: policyObj, IsValid: false, ErrorMsg: invalidTimestampErrorMsg}, fmt.Errorf(errMsg)
				}
				timeReq.MinRevTime = &minRevTime
			}
			if maxRev, ok := requirement["maxRevisionDatetime"]; ok {
				maxRevTime, err := time.Parse(timeLayout, maxRev.(string))
				if err != nil {
					errMsg := fmt.Sprintf("Error Parsing time from maxRevisionDatetime in %s: %v", policyObj.GetName(), err)
					return &AppProtectPolicyEx{Obj: policyObj, IsValid: false, ErrorMsg: invalidTimestampErrorMsg}, fmt.Errorf(errMsg)
				}
				timeReq.MaxRevTime = &maxRevTime
			}
			sigReqs = append(sigReqs, SignatureReq{Tag: reqTag.(string), RevTimes: &timeReq})
		}
	}
```

internal/k8s/app_protect_configuration.go
```diff
@@ -0,0 +1,459 @@
...
+				}
+				if maxRev, ok := requirement["maxRevisionDatetime"]; ok {
+					maxRevTime, err := time.Parse(timeLayout, maxRev.(string))
+					if err != nil {
+						errMsg := fmt.Sprintf("Error Parsing time from maxRevisionDatetime in %s: %v", policyObj.GetName(), err)
+						return &AppProtectPolicyEx{Obj: policyObj, IsValid: false, ErrorMsg: invalidTimestampErrorMsg}, fmt.Errorf(errMsg)
+					}
+					timeReq.MaxRevTime = &maxRevTime
+				}
+				sigReqs = append(sigReqs, SignatureReq{Tag: reqTag.(string), RevTimes: &timeReq})
```
Dean-Coakley, line 167: Alternatively, this could be extracted out to a function to reduce nesting.

Dean-Coakley (commented):
Looks pretty good to me, left some comments about spelling errors and some suggestions for code cleanliness.

internal/k8s/app_protect_configuration.go
Dean-Coakley: Actually, `!found` should return the `AppProtectPolicyEx` just without sigReqs, right?

build/appprotect/DockerfileWithAppProtectForPlus
```diff
@@ -92,7 +94,7 @@ RUN  mkdir -p /var/lib/nginx \
 			    /var/log/nginx/ \
 	&& apt-get remove --purge -y libcap2-bin \
 	&& rm /etc/nginx/conf.d/*
-
+
```
pleshakov, line 97: whitespace alert :)

internal/k8s/app_protect_configuration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+	}
+
+	return &t
+}
+
+func sliceCmpFunc(x, y *unstructured.Unstructured) bool {
+	return x.GetUID() > y.GetUID()
+}
+
+var unstructuredSliceCmpOpts = cmpopts.SortSlices(sliceCmpFunc)
```
pleshakov, line 25: Could you possibly clarify why this is needed? I tried removing it and the tests still passed.

rafwegv: Sure
It passed for me as well, so i pushed and got the errors. when rerunning localy and disabling cache i saw one in 10 tries fail, due to the slices being in incorect order.
This additional option sorts the *unstructured.Unstructured slices in udschanges to prevent this. Reran ~20 times after this change and had 100% sucess

internal/k8s/app_protect_configuration.go
rafwegv: yep

build/appprotect/DockerfileWithAppProtectForPlus
rafwegv: all the RUN statements seem to be seperated by a whitespace from the others

internal/k8s/app_protect_configuration_test.go
pleshakov: oh, got it. thanks for clarification 👍 
that's probably because that list is constructed here https://github.com/nginxinc/kubernetes-ingress/blob/24ed921e3f1e136d407f0afcfa824506b7d03f56/internal/k8s/app_protect_configuration.go#L431 from a map. However, when iterating a map, the keys are not guaranteed to be returned in the same order. 

In a similar type, we sort the keys when iterating over the map -- https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/k8s/configuration.go#L549

updating the test to use a custom comparator looks good to me, as long as the different doesn't cause any issues when IC is running

build/appprotect/DockerfileWithAppProtectForPlus
Dean-Coakley: @rafwegv Right, but I think @pleshakov is saying it can just be an empty line, so the indentation can be removed just leaving an empty line remaining.

rafwegv: ok i see it now.

internal/k8s/app_protect_configuration_test.go
rafwegv: right, the order maters only in tests so i thought i fix it there.

pleshakov (approved):
🚀

Dean-Coakley (approved):

PR #1290, opened 2020-12-17
Graduate policy resource and accessControl policy to GA

Dean-Coakley:
### Proposed changes
* Update policy resource apiVersion `v1alpha1` -> `v1`
* Update accessControl apiVersion `v1alpha1` -> `v1`
* Update documentation to reflect policy and accessControl graduation to GA.
* Fix integration tests.

### Checklist


pkg/apis/configuration/v1/types.go
```diff
@@ -313,3 +314,43 @@ type VirtualServerRouteStatus struct {
...
+
+	Spec PolicySpec `json:"spec"`
+}
+
+// PolicySpec is the spec of the Policy resource.
+// The spec includes multiple fields, where each field represents a different policy.
+// Only one policy (field) is allowed.
+type PolicySpec struct {
+	AccessControl *AccessControl        `json:"accessControl"`
+	RateLimit     *v1alpha1.RateLimit   `json:"rateLimit"`
```
pleshakov, line 336: shall we move those types from `v1alpha1` to this package as well?  since they belong to this type and can't exist without it. we can add a comment that those fields are an alpha feature

pleshakov (approved):

pkg/apis/configuration/v1/types.go
Dean-Coakley: Done in https://github.com/nginxinc/kubernetes-ingress/pull/1290/commits/0f347250cb8719e7de409e5f510c4d1470e386c2

Dean-Coakley:
@pleshakov This is ready for review

docs-web/installation/installation-with-helm.md
```diff
@@ -238,6 +238,9 @@ The following tables lists the configurable parameters of the NGINX Ingress cont
    * - ``controller.enableCustomResources``
      - Enable the custom resources.
      - true
+   * - ``controller.enablePreviewPolicies``
```
pleshakov, line 241: the regular helm  readme needs to be updated similarly

docs-web/configuration/policy-resource.md
```diff
@@ -140,6 +138,8 @@ policies:
 
 ### RateLimit
 
+> **Feature Status**: Rate-Limiting is available as a preview feature: it is suitable for experimenting and testing; however, it must be used with caution in production environments. Additionally, while the feature is in preview status, we might introduce some backward-incompatible changes to the resource specification in the next releases.
```
pleshakov, line 141: Better to include the information about how to enable the preview policies (here and for the other preview policies)

**Feature Status**: Rate-Limiting is available as a preview feature: it is suitable for experimenting and testing; however, it must be used with caution in production environments. Additionally, while the feature is in preview status, we might introduce some backward-incompatible changes to the resource specification in the next releases. The feature is disabled by default. To enable it, set the [`enable-preview-policies`]() command-line argument of the Ingress Controller.

pleshakov (approved):

pkg/apis/configuration/validation/policy.go
```diff
@@ -7,18 +7,18 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/nginxinc/kubernetes-ingress/pkg/apis/configuration/v1alpha1"
+	v1 "github.com/nginxinc/kubernetes-ingress/pkg/apis/configuration/v1"
```
mikestephen, line 10: Is `v1` at the start of this line unnecessary, since the package is named `v1` anyway?

mikestephen (approved):
Looks really good Dean!

pkg/apis/configuration/validation/policy.go
Dean-Coakley: This comes directly from running `goimports`

If I remove it and run `goimports pkg/apis/configuration/validation/policy.go -w` again it will re-add it.

My guess is that `goimports` does this with small pkg names as it is more likely that you'll come across import collisions. And this would make it more clear and convenient to change an import alias as you see fit?

Issue #1295, opened 2020-12-20
Custom 404 if subdomain does not match


030:
**Is your feature request related to a problem? Please describe.**

There is a wildcard DNS entry, i.e. boo.some-domain and foo.some-domain point to the same IP address. The host in the ingress routes the traffic to the correct service. When the subdomain is neither boo, nor foo, a default 404 nginx page is returned.

**Describe the solution you'd like**

An annotation that can be added to the ingress-controller and that overrides the custom 404 page by a html page that one could create themselves.

**Describe alternatives you've considered**

#752, but in this case, there are two subdomains instead of two URIs, e.g. `/tea`. `/coffee`

**Additional context**

030:
Thank you for merging the PR. Could you indicate whether it will be part of the 1.10 release?

030:
Related PR: https://github.com/nginxinc/kubernetes-ingress/pull/1297

pleshakov:
Hi @030 

Unfortunately, we can't include it in the 1.10 (which is released this week), because of our code freeze.

however, it will be included in 1.11 release, targeted for late March

030:
@pleshakov Would it not be possible to release it earlier? We have a production that is running on the edge version right now. I was assuming that it would be merged into the 1.10.0.

pleshakov:
@030 
I can suggest a temporal workaround with a custom template. Please see https://github.com/nginxinc/kubernetes-ingress/tree/v1.10.0/examples/custom-templates

The idea is to put the whole template in `main-template` confgimap key. The `return 404` of the default server can be changed according to your requirements.

Note that for minor releases (1.10.x) custom template doesn't change, as we don't introduce any new features. So it will be safe for upgrades.

ogarrett:
closing - implemented in PR: #1297

PR #1297, opened 2021-01-04
[#1295] Add feature redirect if host does not exist

030:
This allows to redirect to another URL instead of returning a 404

### Proposed changes

### Checklist

### DIFF
--- a/docs-web/configuration/global-configuration/configmap-resource.md
+++ b/docs-web/configuration/global-configuration/configmap-resource.md
@@ -114,6 +114,10 @@ See the doc about [VirtualServer and VirtualServerRoute resources](/nginx-ingres
      - Enables or disables the `real_ip_recursive <https://nginx.org/en/docs/http/ngx_http_realip_module.html#real_ip_recursive>`_ directive.
      - ``False``
      - 
+   * - ``default-server-return``
+     - Configures the `return <https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#return>`_  directive in the default server, which handles a client request if none of the hosts of Ingress or VirtualServer resources match. The default value configures NGINX to return a 404 error page. You can configure a fixed response or a redirect. For example,  ``default-server-return: 302 https://nginx.org`` will redirect a client to https://nginx.org.
+     - ``404``
+     -
    * - ``server-tokens``
      - Enables or disables the `server_tokens <https://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens>`_ directive. Additionally, with the NGINX Plus, you can specify a custom string value, including the empty string value, which disables the emission of the “Server” field.
      - ``True``
--- a/docs-web/configuration/global-configuration/configmap-resource.md
+++ b/docs-web/configuration/global-configuration/configmap-resource.md
@@ -115,7 +115,7 @@ See the doc about [VirtualServer and VirtualServerRoute resources](/nginx-ingres
      - ``False``
      - 
    * - ``default-server-return``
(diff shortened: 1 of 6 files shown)


pleshakov:
Hi @030 

Thanks for the PR! We'll look into it and get back to you shortly.

pleshakov:
Hi @030 

Most of the global configuration that affects the generated NGINX config we implement through the ConfigMap. This allows changing that configuration without re-creating the IC pods and also minimizes the number of cli args (and as a result helm chart parameters).

If this feature was implemented through the ConfigMap, I think the following ConfigMap key would be enough:

```yaml
default-server-redirect-url: https://nginx.org
```

At the same time, another way to solve this problem could be through a more general mechanism -- server snippets. We support snippets for Ingress resources via annotations and the ConfigMap. However, we don't support them for the default server.

For example, here we configure a redirect:
```yaml
default-server-server-snippets: location / { return 302 https://nginx.org; }
```

More use cases could be supported. Here we tell NGINX to close the connection:
```yaml
default-server-server-snippets: location / { return 444; }
```

Here we tell NGINX to return a static response: 
```yaml
default-server-server-snippets: location / { return 200 "welcome from the Ingress Controller!\n"; }
```

The default value would be  `location / { return 404; }`, for backward compatibilty.

I wonder what you think about the both ConfigMap approaches?

030:
@pleshakov Thank you for the feedback. Changes have been applied and the redirect works in conjunction with a configMap as well.

pleshakov (commented):
Hi @030 

Thanks for addressing the feedback.

Because in the implementation `default-server-redirect-url` configures the `return` directive, could you rename the key to `default-server-return` ? Note that the supported use cases of the return directive are broader than redirecting. That's why `default-server-return` suites better.  Could you also make the corresponding changes to the identifiers in the code?

In my original suggestion, I assumed that `default-server-redirect-url` would only configures the redirect. For example, `return 302 {{.DefaultServerRedirectURL}};`.  However, current the implementation with `return {{.DefaultServerRedirectURL};` makes sense and covers more use cases. So I think it is better to keep it, although with the mentioned renaming.

There are also a few additional comments below.

030:
@pleshakov Variable has been renamed and the documentation has been moved to the general customization paragraph

docs-web/configuration/global-configuration/configmap-resource.md
```diff
@@ -114,6 +114,10 @@ See the doc about [VirtualServer and VirtualServerRoute resources](/nginx-ingres
      - Enables or disables the `real_ip_recursive <https://nginx.org/en/docs/http/ngx_http_realip_module.html#real_ip_recursive>`_ directive.
      - ``False``
      - 
+   * - ``default-server-return``
+     - Configures the `return <https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#return>`_  directive in the default server, which handles a client request if none of the hosts of Ingress or VirtualServer resources match. The default value configures NGINX to return a 404 error page. You can configure a fixed response or a redirect. For example,  ``default-server-return: 302 https://nginx.org`` will redirect a client to https://nginx.org.
```
jputrino, line 118:
```suggestion
     - Configures the `return <https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#return>`_  directive in the default server, which handles a client request if none of the hosts of Ingress or VirtualServer resources match. The default value configures NGINX to return a 404 error page. You can configure a fixed response or a redirect. For example, ``default-server-return: 302 https://nginx.org`` will redirect a client to https://nginx.org.
```

pleshakov (approved):
Hi @030 

This looks good to me 👍  (with this comment applied https://github.com/nginxinc/kubernetes-ingress/pull/1297/files#r560548874 )

If you could also update the commit message, as it still refers to the redirect feature. Also, for consistency, could you remove the `[#1295]` from the beginning of a commit message? we typically don't put the issue number in the title -- https://github.com/nginxinc/kubernetes-ingress/commits/master

I will ask for one more approval from our project maintainers.

Dean-Coakley (approved):
Thanks very much for your continued efforts on this @030 !

LGTM.

This looks ready to be merged once the feedback from @jputrino is resolved.

PR #1302, opened 2021-01-11
Fix path in code-gen script

lucacome:
Fix output path


pleshakov (approved):

Dean-Coakley (approved):
🚀 

Thanks @lucacome !

PR #1304, opened 2021-01-14
Add OIDC policy

lucacome:
Add OIDC policy

TODO:
- docs


internal/configs/oidc/oidc.conf
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    location @oidc_error {
+        # This location is called when oidcAuth() or oidcCodeExchange() returns an error
+        status_zone "OIDC error";
+        default_type text/plain;
+        return 500 $internal_error_message;
+    }
+
+    # location /api/ {
```
mikestephen, line 89: Should this commented section be removed?

mikestephen (approved):
Looks good to me Luca!

internal/configs/oidc/oidc.conf
Dean-Coakley: I think this was intentionally left so it is the only modification made to https://github.com/nginxinc/nginx-openid-connect/blob/master/openid_connect.server_conf . 

This way, if we want to apply updates to a new version of the file from that repo: the fact that `location /api/` should be removed should be clear from the diff.

internal/configs/virtualserver.go
```diff
@@ -597,6 +611,7 @@ type policiesCfg struct {
 	JWTAuth         *version2.JWTAuth
 	IngressMTLS     *version2.IngressMTLS
 	EgressMTLS      *version2.EgressMTLS
+	OIDC            *bool
```
LorcanMcVeigh, line 614: does this need to be a pointer? ( honestly idk )

LorcanMcVeigh (approved):

internal/configs/configurator.go
```diff
@@ -1428,6 +1439,8 @@ func (cnf *Configurator) AddOrUpdateSecret(secret *api_v1.Secret) string {
 		return cnf.addOrUpdateCASecret(secret)
 	case secrets.SecretTypeJWK:
 		return cnf.addOrUpdateJWKSecret(secret)
+	case secrets.SecretTypeOIDC:
+		return cnf.addOrUpdateOIDCSecret(secret)
```
Dean-Coakley, line 1443: I think it's misleading to have a func called `addOrUpdateOIDCSecret` which doesn't add or update a secret.

Can we just do `return ""`/ `return "" # OIDC ClientSecret is not required on the filesystem, it is written directly to the config file.` here? And remove `addOrUpdateOIDCSecret`.

Dean-Coakley (approved):

internal/nginx/manager.go
```diff
@@ -28,6 +28,9 @@ const TLSSecretFileMode = 0600
 // JWKSecretFileMode defines the default filemode for files with JWK Secrets.
 const JWKSecretFileMode = 0644
 
+// OIDCSecretFileMode defines the default filemode for files with OIDC Secrets.
+const OIDCSecretFileMode = 0644
```
Dean-Coakley, line 32: After changes in https://github.com/nginxinc/kubernetes-ingress/pull/1304/commits/a2c724f7b3e934a5be97aca6450f7fa2d5f412e7 this can be removed

```suggestion
```

Dean-Coakley: (Amazed that `make lint` didn't find this. We should probably revisit our lint config soon.)

internal/configs/oidc/oidc.conf
mikestephen: Ah makes sense, thanks.

internal/configs/configurator.go
mikestephen: Yeah I was on the fence about that too, I'll change it.

internal/configs/oidc/oidc.conf
lucacome: yep, that was the idea 🙂

pleshakov (commented):
Hi @lucacome 
Please see my feedback. 
I will provide feedback about the example shortly

pleshakov:
@lucacome 
didn't catch it right away. after  running the example, I noticed the `findPoliciesForSecret` in internal/k8s/controller.go:2483 doesn't support oidc secret. This means updating the secret doesn't update it in the config

internal/k8s/controller.go
```diff
@@ -2492,6 +2492,8 @@ func findPoliciesForSecret(policies []*conf_v1.Policy, secretNamespace string, s
 			res = append(res, pol)
 		} else if pol.Spec.EgressMTLS != nil && pol.Spec.EgressMTLS.TrustedCertSecret == secretName && pol.Namespace == secretNamespace {
 			res = append(res, pol)
+		} else if pol.Spec.OIDC != nil && pol.Spec.OIDC.ClientSecret == secretName && pol.Namespace == secretNamespace {
```
pleshakov, line 2495: would it make sense to add a test case to  TestFindPoliciesForSecret ?

internal/configs/virtualserver_test.go
```diff
@@ -2974,6 +3140,14 @@ func TestGeneratePoliciesFails(t *testing.T) {
 				test.msg,
 			)
 		}
+		if vsc.oidcPolCfg.oidc != nil {
```
pleshakov, line 3143: what if `vsc.oidcPolCfg.oidc` returns nil as a bug ?

examples-of-custom-resources/oidc/oidc.yaml
```diff
@@ -6,7 +6,7 @@ spec:
   oidc:
     clientID: nginx-plus
     clientSecret: oidc-secret
-    authEndpoint: http://idp.example.com:8080/auth/realms/master/protocol/openid-connect/auth
+    authEndpoint: http://keycloak.example.com/auth/realms/master/protocol/openid-connect/auth
```
pleshakov, line 9:
```suggestion
    authEndpoint: https://keycloak.example.com/auth/realms/master/protocol/openid-connect/auth
```

pleshakov (approved):

PR #1315, opened 2021-01-20
Add User ID to webapp response

lucacome:
Using the `username` header to show that the user is authenticated.


pleshakov (approved):
@lucacome 

looks good!

Would it make sense to update the doc here https://github.com/nginxinc/kubernetes-ingress/blob/master/docs-web/configuration/policy-resource.md#oidc ? Something like below:

```yaml
spec:
  oidc:
    clientID: nginx-plus
    clientSecret: oidc-secret
    authEndpoint: https://idp.example.com/openid-connect/auth
    tokenEndpoint: https://idp.example.com/openid-connect/token
    jwksURI: https://idp.example.com/openid-connect/certs
```` 

NGINX Plus will pass the ID of an authenticated user to the backend in the HTTP header `username`.

Dean-Coakley (approved):

PR #1317, opened 2021-01-21
Add status field to policy resource 

LorcanMcVeigh:
### Proposed changes
This PR adds the status field, with the sub-fields 'state', 'reason' and 'message', to the policy resource.
It also prints out the state of the policy when `kubectl get policy <policy-name>` is ran. 

### Checklist


docs-web/configuration/global-configuration/reporting-resources-status.md
```diff
@@ -110,4 +110,44 @@ The Ingress controller must be configured to report a VirtualServer or VirtualSe
...
+
+```eval_rst
+.. list-table::
+   :header-rows: 1
+
+   * - Field
+     - Description
+     - Type
+   * - ``State``
+     - Current state of the resource. Can be ``Valid``, ``Warning`` an ``Invalid``. For more information, refer to the ``message`` field.
```
mikestephen, line 145: Should this read:
Can be ``Valid``, ``Warning`` or ``Invalid``
?

internal/k8s/controller.go
```diff
@@ -2378,13 +2396,25 @@ func (lbc *LoadBalancerController) getPolicies(policies []conf_v1.PolicyReferenc
 		}
 
 		policy := policyObj.(*conf_v1.Policy)
-
+		msg := ""
```
mikestephen, line 2399: Is it more conventional to use `var msg string` here instead when initialising a variable to its zero-value?
I'm fairly new to golang so I could be wrong.

internal/k8s/status.go
```diff
@@ -558,3 +559,66 @@ func (su *statusUpdater) generateExternalEndpointsFromStatus(status []api_v1.Loa
 
 	return externalEndpoints
 }
+
+func hasPolicyStatusChanged(pol *v1.Policy, state string, reason string, message string) bool {
+	if pol.Status.State != state {
```
mikestephen, line 564: I think this function could be?:
```
func hasPolicyStatusChanged(pol *v1.Policy, state string, reason string, message string) bool {
        return pol.Status.State != state || pol.Status.Reason != reason || pol.Status.Message != message
}
```

pkg/client/clientset/versioned/typed/configuration/v1/fake/fake_policy.go
```diff
@@ -86,6 +86,18 @@ func (c *FakePolicies) Update(ctx context.Context, policy *configurationv1.Polic
 	return obj.(*configurationv1.Policy), err
 }
 
+// UpdateStatus was generated because the type contains a Status member.
+// Add a +genclient:noStatus comment above the type to avoid generating UpdateStatus().
+func (c *FakePolicies) UpdateStatus(ctx context.Context, policy *configurationv1.Policy, opts v1.UpdateOptions) (*configurationv1.Policy, error) {
```
mikestephen, line 91: Since this is generated code, it seems odd that it contains a reference to `FakePolicies`. Isn't this a test struct?

mikestephen (approved):
Just left a couple of comments, but looking great Lorcan.

pkg/client/clientset/versioned/typed/configuration/v1/fake/fake_policy.go
LorcanMcVeigh: yeah but this code is, as you said, generated by kubernetes `https://github.com/nginxinc/kubernetes-ingress/tree/master/vendor/k8s.io/code-generator` so i don't think we can control what it generates.

docs-web/configuration/global-configuration/reporting-resources-status.md
LorcanMcVeigh: I actually think this was a typo as well, I dont think a policy can have a `Warning` state. so it would be `Can be ''Valid'' or ''Invalid''. `

pleshakov (commented):
Hi @LorcanMcVeigh 

We need to consider a few more cases related to leader election:

(1) only report the status from the leader. this is needed so that all Ingress Controller pods don't try to report statuses, but only one, to avoid duplicated work/errors related to collisions of updates.

For example, for VS/VSR, before reporting a status, the following check is performed:
```
if lbc.reportVsVsrStatusEnabled() {
```
we need to do a similar check for policies.
we can use that function. Better to rename it so it doesn't apply only for VS/VSR.  For example, `reportCustomResourceStatusEnabled()`

(2) when an Ingress Controller pod becomes the leader, it needs to report statuses of the resources it handles. Otherwise, the resources that were processed during the period when there was no elected leader (can happen during upgrades or when the leader pod fails) will not have the status reported.

We already do that for VirtualServer/VirtualServerRoute and Ingresses. See https://github.com/nginxinc/kubernetes-ingress/blob/2cdc714918e9153364319d91c4edb38a22eefefe/internal/k8s/leader.go#L51 For Policies, we simply need to look go through every policy in the lister and report the status based on the validation.

internal/k8s/controller.go
```diff
@@ -2344,7 +2389,8 @@ func (lbc *LoadBalancerController) getAllPolicies() []*conf_v1.Policy {
 
 		err := validation.ValidatePolicy(pol, lbc.isNginxPlus, lbc.enablePreviewPolicies)
 		if err != nil {
-			glog.V(3).Infof("Skipping invalid Policy %s/%s: %v", pol.Namespace, pol.Name, err)
```
pleshakov, line 2347: I don't think we no longer need to have a `msg` variable here

internal/k8s/controller.go
```diff
@@ -1861,6 +1878,34 @@ func (lbc *LoadBalancerController) updateVirtualServerRoutesStatusFromEvents() e
 	return nil
 }
 
+func (lbc *LoadBalancerController) updatePolicyStatus() error {
```
pleshakov, line 1881: for consistency with similar `updateVirtualServersStatusFromEvents`
```suggestion
func (lbc *LoadBalancerController) updatePoliciesStatus() error {
```

internal/k8s/controller.go
```diff
@@ -1861,6 +1878,34 @@ func (lbc *LoadBalancerController) updateVirtualServerRoutesStatusFromEvents() e
...
+		pol := obj.(*conf_v1.Policy)
+
+		err := validation.ValidatePolicy(pol, lbc.isNginxPlus, lbc.enablePreviewPolicies)
+		if err != nil {
+			msg := fmt.Sprintf("Policy %v/%v is invalid and was rejected: %v", pol.Namespace, pol.Name, err)
+			err = lbc.statusUpdater.UpdatePolicyStatus(pol, conf_v1.StateInvalid, "Rejected", msg)
+			if err != nil {
+				allErrs = append(allErrs, err)
+			}
+		}
```
pleshakov, line 1893: there is a bug here - the code for reporting "AddedOrUpdated" will always be executed, even if the the error is not nil.

Shall it be like below?

```
if err != nil {
  // policy is rejected
} else {
 // policy is added or updated
}
```

Dean-Coakley (approved):

pleshakov (approved):

PR #1367, opened 2021-02-09
Fixes writing of time values to templates

mikestephen:
### Proposed changes
Time values as described here https://nginx.org/en/docs/syntax.html are valid for time-based nginx directives, including values such as `1m 30s` (note the space).

If a directive such as `proxy_connect_timeout 1m 30s` is written to the nginx conf file it will be invalid, instead it must be quoted, like `proxy_connect_timeout "1m 30s"`

This fix adds this quoting of time values.

### Checklist


pleshakov:
Hi @mikestephen 

 a few more directives are validated using that ParseTime function. Do we need to "fix" them as well?
 
 - TransportServer timeout
 - TransportServer nextUpstreamTimeout
 - VirtualServer Upstream Queue
 - VirtualServer Upstream.SessionCookie expires 
 - VirtualServer Upstream HealthCheck  interval and jitter
 - VirtualServer Upstream next-upstream-timeout

mikestephen:
Closing this, because a simpler fix is to disallow whitespace in time expressions in KIC. This will remove the need for quoting time expressions in the nginx config.

PR #1382, opened 2021-02-16
Add passive health checks to transportServer

LorcanMcVeigh:
### Proposed changes
Add new parameters to upstream server to allow for passive health checks 

### Checklist


internal/configs/transportserver.go
```diff
@@ -103,7 +103,10 @@ func generateStreamUpstreams(transportServerEx *TransportServerEx, upstreamNamer
 		endpointsKey := GenerateEndpointsKey(transportServerEx.TransportServer.Namespace, u.Service, nil, uint16(u.Port))
 		endpoints := transportServerEx.Endpoints[endpointsKey]
 
-		ups := generateStreamUpstream(name, endpoints, isPlus)
+		maxFails := generateIntFromPointer(u.MaxFails, 1)
```
pleshakov, line 106: should we move the generation of maxFails and failTimeout into `generateStreamUpstream`, which can take `upstream` as a parameter?  this way we don't need to pass every upstream related parameter to `generateStreamUpstream`, we can pass the whole upstream

LorcanMcVeigh: yeah I think that would make sense

Dean-Coakley (approved):

pleshakov (approved):

PR #1384, opened 2021-02-17
Add health checks to transport server

LorcanMcVeigh:
### Proposed changes
This PR adds active health checks support to Transport Servers for NGINX Plus. It adds a health check field to yaml and fields for better configuring the health checks. 

### Checklist


docs-web/configuration/transportserver-resource.md
```diff
@@ -181,6 +181,64 @@ port: 8443
...
+   * - ``enable``
+     - Enables a health check for an upstream server. The default is ``false``.
+     - ``boolean``
+     - No
+   * - ``interval``
+     - The interval between two consecutive health checks. The default is ``5s``.
+     - ``string``
+     - No
+   * - ``timeout``
+     - This overrides the timeout set by `proxy_timeout <http://nginx.org/en/docs/stream/ngx_stream_proxy_module.html#proxy_timeout>`_ which is set in [SessionParameters](#sessionparameters) for health checks. The default value is ``5s``.
```
pleshakov, line 221: the link is not generated properly:
image

docs-web/configuration/transportserver-resource.md
```diff
@@ -181,6 +181,64 @@ port: 8443
      - Yes
 ```
 
+
+### Upstream.Healthcheck
```
pleshakov, line 185: could you also add the field `healthCheck` to the table for https://deploy-preview-1384--nginx-kubernetes-ingress.netlify.app/nginx-ingress-controller/configuration/transportserver-resource/#upstream which links to this section ?

pkg/apis/configuration/validation/transportserver_test.go
```diff
@@ -387,6 +387,79 @@ func TestValidateListenerProtocol(t *testing.T) {
...
+			healthCheck: &v1alpha1.HealthCheck{
+				Enabled:  true,
+				Timeout:  "-30s",
+				Jitter:   "5s",
+				Port:     88,
+				Interval: "10",
+				Passes:   3,
+				Fails:    4,
+			},
+			msg: "invalid timeout",
```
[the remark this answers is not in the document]
LorcanMcVeigh: I just thought that the other fields using the same validation methods as those 2, either validating a string or validating an int that it would be redundant to test for all the fields, no?

internal/configs/version2/nginx-plus.transportserver.tmpl
```diff
@@ -30,10 +30,10 @@ server {
 
     proxy_pass {{ $s.ProxyPass }};
 
-    {{ if $s.HealthCheck }}
-    health_check interval={{ $s.HCInterval }} port={{ $s.HCPort }} 
-        passes={{ $s.HCPasses }} jitter={{ $s.HCJitter }} fails={{ $s.HCFails }} {{ if $s.UDP }}udp{{ end }};
-    health_check_timeout {{ $s.HCTimeout }};
+    {{ if $s.HealthCheck.Enabled }}
```
pleshakov, line 33: the Enabled field is unnecessary. We can just have  this condition, which will be true if HealthCheck is not nil. 

```
{{ if $s.HealthCheck }}
   # hc stuff 
{{ end }}
```

internal/configs/version2/stream.go
```diff
@@ -35,13 +35,17 @@ type StreamServer struct {
...
-	HCInterval               string
-	HCPort                   int
-	HCPasses                 int
-	HCJitter                 string
-	HCFails                  int
-	HCTimeout                string
+	HealthCheck              *StreamHealthCheck
+}
+
+type StreamHealthCheck struct {
```
pleshakov, line 41: // StreamHealthCheck defines a health check for a StreamUpstream in a StreamServer.

internal/configs/transportserver_test.go
```diff
@@ -366,3 +365,89 @@ func TestGenerateUnixSocket(t *testing.T) {
...
+				Passes:   1,
+				Fails:    1,
+			},
+		},
+	}
+
+	for _, test := range tests {
+		result := generateTransportServerHealthCheck(test.healthCheck)
+		if !reflect.DeepEqual(result, test.expected) {
+			fmt.Printf("Input : %+v\n", test.healthCheck)
```
pleshakov, line 448: for comparing large structs, better to use this approach, as it prints a nice diff, so it's easy to spot the mismatch:
```
		if diff := cmp.Diff(test.expected, result); diff != "" {
			t.Errorf("generatePolicies() '%v' mismatch (-want +got):\n%s", test.msg, diff)
		}
```

internal/configs/transportserver_test.go
```diff
@@ -366,3 +365,89 @@ func TestGenerateUnixSocket(t *testing.T) {
...
+			fmt.Printf("Result : %+v\n", result)
```
pleshakov, line 449: those seem to be debug statements.  
cmp.Diff will print the diff nicely, no need for any additional debug prints

internal/configs/transportserver_test.go
```diff
@@ -366,3 +365,89 @@ func TestGenerateUnixSocket(t *testing.T) {
 		t.Errorf("generateUnixSocket() returned %q but expected %q", result, expected)
 	}
 }
+
+func TestGenerateTransportServerHealthChecks(t *testing.T) {
+	tests := []struct {
+		healthCheck *conf_v1alpha1.HealthCheck
+		expected    *version2.StreamHealthCheck
+	}{
```
pleshakov, line 373: I suggest adding the msg field to describe each test case, as it is not immediately clear what test cases we test here

pleshakov (commented):
Hi @LorcanMcVeigh 

Thanks for addressing feedback. I got a few more comments/suggestion related to the new code.

pkg/apis/configuration/validation/transportserver_test.go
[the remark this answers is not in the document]
pleshakov: in this case, we're testing `validateTSUpstreamHealthChecks`,  not `validateTime` and `validatePositiveIntOrZero`. Essentially, our cases here  test that validateTSUpstreamHealthChecks propagates errors from `validateTime` and `validatePositiveIntOrZero`, not that `validateTime` and `validatePositiveIntOrZero` work correctly - that is tested in the tests for those functions.

pkg/apis/configuration/validation/transportserver_test.go
```diff
@@ -387,6 +387,127 @@ func TestValidateListenerProtocol(t *testing.T) {
...
+				Passes:   3,
+				Fails:    4,
+			},
+			msg: "valid Health check",
+		},
+	}
+	for _, test := range tests {
+		allErrs := validateTSUpstreamHealthChecks(test.healthCheck, field.NewPath("healthCheck"))
+		if len(allErrs) > 0 {
+			t.Errorf("validateUpstreamHealthChecks() returned errors %v  for valid input for the case of %s", allErrs, test.msg)
```
Dean-Coakley, line 419:
```suggestion
			t.Errorf("validateTSUpstreamHealthChecks() returned errors %v for valid input for the case of %s", allErrs, test.msg)
```

pkg/apis/configuration/validation/transportserver_test.go
```diff
@@ -387,6 +387,127 @@ func TestValidateListenerProtocol(t *testing.T) {
...
+				Fails:    4,
+			},
+			msg: "invalid jitter value",
+		},
+	}
+
+	for _, test := range tests {
+		allErrs := validateTSUpstreamHealthChecks(test.healthCheck, field.NewPath("healthCheck"))
+		if len(allErrs) == 0 {
+			t.Errorf("validateUpstreamHealthChecks() returned no error for invalid input %v", test.msg)
```
Dean-Coakley, line 506:
```suggestion
			t.Errorf("validateTSUpstreamHealthChecks() returned no error for invalid input %v", test.msg)
```

Dean-Coakley (approved):

internal/configs/transportserver.go
```diff
@@ -112,7 +116,55 @@ func generateStreamUpstreams(transportServerEx *TransportServerEx, upstreamNamer
...
+func generateTransportServerHealthCheck(healthCheck *conf_v1alpha1.HealthCheck) *version2.StreamHealthCheck {
+	if healthCheck == nil || !healthCheck.Enabled {
+		return nil
+	}
+
+	hc := generateTransportServerHealthCheckWithDefaults()
+
+	hc.Enabled = healthCheck.Enabled
+
+	if healthCheck.Interval != "" {
```
pleshakov, line 131: sorry, didn't catch it during the previous round.

for time fields - interval, jitter and timeout --  we need to use `generateTime`, as that function also handles transformation cases like this  "10m 1s" -> "10m1s" , which prevents NGINX from failing to reload

internal/configs/transportserver_test.go
```diff
@@ -345,6 +350,61 @@ func TestGenerateUnixSocket(t *testing.T) {
...
+				Fails:    5,
+			},
+			msg: "valid health checks",
+		},
+	}
+
+	for _, test := range tests {
+		result := generateTransportServerHealthCheck(test.healthCheck)
+		if diff := cmp.Diff(result, test.expected); diff != "" {
+			t.Errorf("generateTransportServerHealthCheck() returned %v but expected %v", result, test.expected)
```
pleshakov, line 403: could you print the diff here, so that it is clear which fields are not equal?

An example from a similar test:
```
		if diff := cmp.Diff(test.expected, result); diff != "" {
			t.Errorf("generatePolicies() '%v' mismatch (-want +got):\n%s", test.msg, diff)
		}
```

docs-web/configuration/transportserver-resource.md
pleshakov: could you update the upstream fields table with the health check field? 

this is what we have for the VS:
image

internal/configs/transportserver.go
```diff
@@ -93,15 +94,18 @@ func generateUnixSocket(transportServerEx *TransportServerEx) string {
 	return ""
 }
 
-func generateStreamUpstreams(transportServerEx *TransportServerEx, upstreamNamer *upstreamNamer, isPlus bool) []version2.StreamUpstream {
+func generateStreamUpstreamsAndHealthCheck(transportServerEx *TransportServerEx, upstreamNamer *upstreamNamer, isPlus bool) ([]version2.StreamUpstream, *version2.StreamHealthCheck) {
 	var upstreams []version2.StreamUpstream
+	var hc *version2.StreamHealthCheck
 
 	for _, u := range transportServerEx.TransportServer.Spec.Upstreams {
```
soneillf5, line 101: Can we remove the health check generation from this method?

At the moment, it appears to create a healthcheck for each upstream, but there is only one pointer. I think the value is overridden so it only returns the healthcheck for the last upstream, is this correct?

How about a new function, which takes the upstreams (transportServerEx.TransportServer.Spec.Upstreams ?) and returns the health checks for each one?

internal/configs/transportserver_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
```
soneillf5, line 7: I mentioned previously about the pointer override, assuming this is not intended behaviour, can we add a test to confirm that two upstreams generates two health checks ? I think there's multiple health checks in the tests below, but some of them are invalid ?

internal/configs/transportserver.go
LorcanMcVeigh: > At the moment, it appears to create a healthcheck for each upstream, but there is only one pointer. I think the value is overridden so it only returns the healthcheck for the last upstream, is this correct?
 
Yeah that's right, at the moment we don't support multiple health checks, so I'll refactor so that the only health check that's generated is the one that referenced the right upstream

internal/configs/transportserver.go
```diff
@@ -116,43 +115,38 @@ func generateStreamUpstreamsAndHealthCheck(transportServerEx *TransportServerEx,
...
+			}
+
+			if u.HealthCheck.Port > 0 {
+				hc.Port = u.HealthCheck.Port
+			}
+		}
 	}
-
 	return hc
+
```
pleshakov, line 149: this whitespace seems necessary

internal/configs/transportserver_test.go
```diff
@@ -351,56 +351,110 @@ func TestGenerateUnixSocket(t *testing.T) {
...
 			},
-			msg: "valid health checks",
+			msg: "valid 2 health checks",
 		},
 	}
 
 	for _, test := range tests {
-		result := generateTransportServerHealthCheck(test.healthCheck)
+		result := generateTransportServerHealthCheck(upstreamName, test.upstreams)
 		if diff := cmp.Diff(result, test.expected); diff != "" {
```
pleshakov, line 456: otherwise "-" will mean "+" and vice versa 
```suggestion
		if diff := cmp.Diff(test.expected, result); diff != "" {
```

pleshakov (approved):

soneillf5 (approved):
Looks great!

Dean-Coakley (approved):

PR #1407, opened 2021-02-23
Try switching AP smoke tests to mac runner on nightly

ciarams87:
### Proposed changes
The AP smoke tests have errors frequently, we think because tests are timing out due to the infrastructure not being powerful enough (the default linux and windows runners are based on on Standard_DS2_v2 virtual machines in Microsoft Azure). This commit switches the runner for these jobs to macOS, as it has a more powerful spec (see https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources for more info)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -254,35 +254,35 @@ jobs:
             marker: ''
             type: plus
             k8s: 1.16.15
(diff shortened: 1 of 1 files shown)


ciarams87:
Happening on more images than just plus - closing the pr for now

PR #1409, opened 2021-02-24
Fix virtual server health check to generate time correctly

soneillf5:
### Proposed changes

### Checklist

### DIFF
--- a/internal/configs/virtualserver_test.go
+++ b/internal/configs/virtualserver_test.go
@@ -5838,6 +5838,33 @@ func TestGenerateHealthCheck(t *testing.T) {
 			expected:     nil,
 			msg:          "HealthCheck not enabled",
 		},
+		{
+			upstream: conf_v1.Upstream{
+				HealthCheck: &conf_v1.HealthCheck{
+					Enable:         true,
+					Interval:       "1m 5s",
+					Jitter:         "2m 3s",
+					ConnectTimeout: "1m 10s",
+					SendTimeout:    "1m 20s",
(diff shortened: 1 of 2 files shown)


internal/configs/virtualserver_test.go
```diff
@@ -5854,6 +5854,78 @@ func TestGenerateHealthCheck(t *testing.T) {
...
 
+func TestGenerateHealthCheckTimeParameter(t *testing.T) {
+	upstreamName := "test-upstream"
+	tests := []struct {
+		upstream     conf_v1.Upstream
+		upstreamName string
+		expected     *version2.HealthCheck
+		msg          string
+	}{
+		{
```
pleshakov, line 5865: would it make sense to move this case to `TestGenerateHealthCheck`? otherwise, I don't think we need a loop since we have only one test case here. At the same  time, I think we can add an additional test case -- since we also call generateTime in `newHealthCheckWithDefaults`, do we need a test case that exercises that ?

I wonder if we really need to set all the fields in the input? Perhaps only time-related and Enable would be enough for this case?



For the name of the test,  perhaps "HealthCheck with time parameters with spaces" could better reflect the case?

soneillf5: * I've moved the unit test to be part of the existing collection of health check tests.
* Now it just tests the time values with the rest as defaults.
* I've renamed the test to suit.

pleshakov (approved):

ciarams87 (approved):
Looks good 👍🏼

vepatel (approved):

PR #1411, opened 2021-02-24
Add App Protect unary gRPC support 

rafwegv:
### Proposed changes
This bumps the App Protect version to the last release and also introduces support for unary grpc in the Policy crd.

### Checklist

### DIFF
--- a/deployments/helm-chart/crds/appprotect.f5.com_appolicies.yaml
+++ b/deployments/helm-chart/crds/appprotect.f5.com_appolicies.yaml
@@ -15,1484 +15,1642 @@ spec:
   preserveUnknownFields: false
   scope: Namespaced
   versions:
-    - name: v1beta1
-      schema:
-        openAPIV3Schema:
-          description: APPolicyConfig is the Schema for the APPolicyconfigs API
-          properties:
-            apiVersion:
-              description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
-              type: string
-            kind:
-              description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
-              type: string
-            metadata:
-              type: object
-            spec:
-              description: APPolicySpec defines the desired state of APPolicy
-              properties:
-                modifications:
-                  items:
-                    properties:
-                      action:
-                        type: string
-                      description:
-                        type: string
-                      entity:
-                        properties:
-                          name:
-                            type: string
-                        type: object
-                      entityChanges:
-                        properties:
-                          type:
-                            type: string
-                        type: object
-                    type: object
-                    x-kubernetes-preserve-unknown-fields: true
-                  type: array
-                modificationsReference:
+  - name: v1beta1
+    schema:
+      openAPIV3Schema:
+        description: APPolicyConfig is the Schema for the APPolicyconfigs API
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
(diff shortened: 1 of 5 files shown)


Dean-Coakley (approved):

Dean-Coakley:
@rafwegv We now have some automation for changelog generation: https://github.com/nginxinc/kubernetes-ingress/actions/workflows/release-drafter.yml

The resulting changelog draft should be visible to you: https://github.com/nginxinc/kubernetes-ingress/releases

I think it would be good if you could rename the PR to note that this change relates to AppProtect.

deployments/common/crds-v1beta1/appprotect.f5.com_appolicies.yaml
```diff
@@ -1058,6 +1199,8 @@ spec:
                 urls:
                   items:
                     properties:
+                      $action:
```
pleshakov, line 1202: just to double check, is `$` in front of the action is correct?

deployments/common/crds-v1beta1/appprotect.f5.com_appolicies.yaml
```diff
@@ -480,6 +495,68 @@ spec:
...
+                        type: string
+                      hasIdlFiles:
+                        type: boolean
+                      idlFiles:
+                        items:
+                          properties:
+                            idlFile:
+                              properties:
+                                $ref:
+                                  pattern: ^http
```
pleshakov, line 523: what will happen if the file can't be downloaded from the URL? For example, the server rejects the request or times out. What is the timeout value? 

can a failure that prevent NGINX Plus from reloading?

what happens if the download is slow? let's say it takes 60s. does it mean NGINX Plus will not be able to reload until the file is downloaded?

If this can cause too many problems, perhaps we should not allow external references

deployments/common/crds-v1beta1/appprotect.f5.com_appolicies.yaml
rafwegv: yes, it means that this field should be deleted from the base policy rather than added

deployments/common/crds-v1beta1/appprotect.f5.com_appolicies.yaml
rafwegv: Valid points. Let's discuss offline.

deployments/helm-chart/crds/appprotect.f5.com_appolicies.yaml
```diff
@@ -965,6 +1100,12 @@ spec:
                 signature-requirements:
                   items:
                     properties:
+                      maxRevisionDatetime:
```
pleshakov, line 1103: does this mean we're restoring this https://github.com/nginxinc/kubernetes-ingress/commit/119617464cd9c8826980e746b0e99e48beefe533 ?

pleshakov (commented):
Hi @rafwegv 

Please see additional feedback:

(1) Do we need to update the CRD for `APLogConf`?  I see at least one change - the format can support `grpc`.
 https://docs.nginx.com/nginx-app-protect/configuration/#factory-configuration-files

(2) Do we need to use different default files for logging and/or policies for gRPC?
(based on https://docs.nginx.com/nginx-app-protect/configuration/#logging  )
If that is the case,  I think we need to update the template https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/version1/nginx-plus.ingress.tmpl#L67-L71

rafwegv:
re: 

> Hi @rafwegv
> 
> Please see additional feedback:
> 
> (1) Do we need to update the CRD for `APLogConf`? I see at least one change - the format can support `grpc`.
> https://docs.nginx.com/nginx-app-protect/configuration/#factory-configuration-files
> 
> (2) Do we need to use different default files for logging and/or policies for gRPC?
> (based on https://docs.nginx.com/nginx-app-protect/configuration/#logging )
> If that is the case, I think we need to update the template https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/version1/nginx-plus.ingress.tmpl#L67-L71

Hi @pleshakov 
(1) yes, will add grpc to the format enum, thanks for noticing.
(2) I think theese should remain as examples, and if Customers would like to create configs of of them they can.

deployments/helm-chart/crds/appprotect.f5.com_appolicies.yaml
rafwegv: not yet, this must have gotten reenabled in  merge somwhere. will disable.

rafwegv:
Hi @pleshakov,
I removed the revision date times as well as links from grpc resources. Discussed with internally and decided to drop them. Cons in this use case outweigh the pros.

pleshakov (commented):
Hi @rafwegv 

Could you possibly rebase against the master? Note that we finally removed v1beta1 CRDS.

I also noticed a few seemingly unrelated to gRPC fields that were added to the CRDs. Like `signatureOverrides`. Just to double check, are we adding any other features as part of this PR?

rafwegv:
@pleshakov 
Sometimes fields will be updated elswhere in the schema for various reasons. Theese in particular are kind of meta fields that are not part of a feature per-se. Alternatively we could have seperate PRs for such bug fixes in the future.

pleshakov (approved):

PR #1413, opened 2021-02-26
Add snippet feature to TransportServer

LorcanMcVeigh:
### Proposed changes
This PR adds Snippet feature to the TransportServer resource. 

### Checklist

### DIFF
--- a/docs-web/configuration/transportserver-resource.md
+++ b/docs-web/configuration/transportserver-resource.md
@@ -19,6 +19,7 @@ This document is the reference documentation for the TransportServer resource. T
     - [SessionParameters](#sessionparameters)
     - [Action](#action)
   - [Using TransportServer](#using-transportserver)
+    - [Usings Snippets](#using-snippets)
     - [Validation](#validation)
       - [Structural Validation](#structural-validation)
       - [Comprehensive Validation](#comprehensive-validation)
@@ -369,6 +370,41 @@ secure-app   46sm
 ```
 
 In the kubectl get and similar commands, you can also use the short name `ts` instead of `transportserver`.
+
+### Using Snippets
+
+Snippets allow you to insert raw NGINX config into different contexts of NGINX configuration. In the example below, we use snippets to configure [access control](http://nginx.org/en/docs/stream/ngx_stream_access_module.html) in a TransportServer:
+
+```yaml
+apiVersion: k8s.nginx.org/v1alpha1
+kind: TransportServer
+metadata:
+  name: cafe
+spec:
+  host: cafe.example.com
+  serverSnippets: |
+    deny  192.168.1.1;
+    allow 192.168.1.0/24;
+  upstreams:
+  - name: tea
+    service: tea-svc
+    port: 80
+```
+
+Snippets are intended to be used by advanced NGINX users who need more control over the generated NGINX configuration.
+
+However, because of the disadvantages described below, snippets are disabled by default. To use snippets, set the [`enable-snippets`](/nginx-ingress-controller/configuration/global-configuration/command-line-arguments#cmdoption-enable-snippets) command-line argument.
+
+Disadvantages of using snippets:
+* *Complexity*. To use snippets, you will need to:
+  * Understand NGINX configuration primitives and implement a correct NGINX configuration.
+  * Understand how the IC generates NGINX configuration so that a snippet doesn't interfere with the other features in the configuration.
+* *Decreased robustness*. An incorrect snippet makes the NGINX config invalid which will lead to a failed reload. This will prevent any new configuration updates, including updates for the other TransportServer resource until the snippet is fixed.
+* *Security implications*. Snippets give access to NGINX configuration primitives and those primitives are not validated by the Ingress Controller.
+
+
+> Note that during a period when the NGINX config includes an invalid snippet, NGINX will continue to operate with the latest valid configuration.
+
 
 ### Validation
 
--- a/internal/configs/configurator.go
+++ b/internal/configs/configurator.go
@@ -559,7 +559,7 @@ func (cnf *Configurator) addOrUpdateTransportServer(transportServerEx *Transport
 
 	tsCfg := generateTransportServerConfig(transportServerEx, transportServerEx.ListenerPort, cnf.isPlus)
 
-	content, err := cnf.templateExecutorV2.ExecuteTransportServerTemplate(&tsCfg)
+	content, err := cnf.templateExecutorV2.ExecuteTransportServerTemplate(tsCfg)
 	if err != nil {
 		return fmt.Errorf("Error generating TransportServer config %v: %v", name, err)
 	}
@@ -1073,6 +1073,7 @@ func (cnf *Configurator) UpdateConfig(cfgParams *ConfigParams, ingExes []*Ingres
 	return allWarnings, nil
 }
 
+// UpdateTransportServers updates TransportServers.
 func (cnf *Configurator) UpdateTransportServers(updatedTSExes []*TransportServerEx, deletedKeys []string) error {
 	for _, tsEx := range updatedTSExes {
 		err := cnf.addOrUpdateTransportServer(tsEx)
--- a/internal/configs/transportserver.go
+++ b/internal/configs/transportserver.go
@@ -30,7 +30,7 @@ func (tsEx *TransportServerEx) String() string {
 }
 
 // generateTransportServerConfig generates a full configuration for a TransportServer.
-func generateTransportServerConfig(transportServerEx *TransportServerEx, listenerPort int, isPlus bool) version2.TransportServerConfig {
+func generateTransportServerConfig(transportServerEx *TransportServerEx, listenerPort int, isPlus bool) *version2.TransportServerConfig {
 	upstreamNamer := newUpstreamNamerForTransportServer(transportServerEx.TransportServer)
 
 	upstreams := generateStreamUpstreams(transportServerEx, upstreamNamer, isPlus)
@@ -59,14 +59,14 @@ func generateTransportServerConfig(transportServerEx *TransportServerEx, listene
 		proxyTimeout = transportServerEx.TransportServer.Spec.SessionParameters.Timeout
 	}
 
-	statusZone := ""
+	serverSnippets := generateSnippets(true, transportServerEx.TransportServer.Spec.ServerSnippets, []string{})
+
+	statusZone := transportServerEx.TransportServer.Spec.Listener.Name
 	if transportServerEx.TransportServer.Spec.Listener.Name == conf_v1alpha1.TLSPassthroughListenerName {
 		statusZone = transportServerEx.TransportServer.Spec.Host
-	} else {
-		statusZone = transportServerEx.TransportServer.Spec.Listener.Name
 	}
 
-	return version2.TransportServerConfig{
+	tsConfig := &version2.TransportServerConfig{
 		Server: version2.StreamServer{
 			TLSPassthrough:           transportServerEx.TransportServer.Spec.Listener.Name == conf_v1alpha1.TLSPassthroughListenerName,
 			UnixSocket:               generateUnixSocket(transportServerEx),
@@ -84,9 +84,12 @@ func generateTransportServerConfig(transportServerEx *TransportServerEx, listene
 			ProxyNextUpstreamTimeout: generateTimeWithDefault(nextUpstreamTimeout, "0s"),
 			ProxyNextUpstreamTries:   nextUpstreamTries,
 			HealthCheck:              healthCheck,
+			Snippets:                 serverSnippets,
 		},
 		Upstreams: upstreams,
 	}
+
+	return tsConfig
 }
 
 func generateUnixSocket(transportServerEx *TransportServerEx) string {
--- a/internal/configs/transportserver_test.go
+++ b/internal/configs/transportserver_test.go
@@ -1,7 +1,6 @@
 package configs
 
 import (
-	"reflect"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -62,6 +61,82 @@ func TestTransportServerExString(t *testing.T) {
 	}
 }
 
+func TestGenerateTransportServerConfigForTCPSnippets(t *testing.T) {
+	transportServerEx := TransportServerEx{
+		TransportServer: &conf_v1alpha1.TransportServer{
+			ObjectMeta: meta_v1.ObjectMeta{
+				Name:      "tcp-server",
+				Namespace: "default",
+			},
+			Spec: conf_v1alpha1.TransportServerSpec{
+				Listener: conf_v1alpha1.TransportServerListener{
+					Name:     "tcp-listener",
+					Protocol: "TCP",
+				},
+				Upstreams: []conf_v1alpha1.Upstream{
+					{
+						Name:    "tcp-app",
+						Service: "tcp-app-svc",
+						Port:    5001,
+					},
+				},
+				Action: &conf_v1alpha1.Action{
+					Pass: "tcp-app",
+				},
+				ServerSnippets: "deny  192.168.1.1;\nallow 192.168.1.0/24;",
+			},
+		},
+		Endpoints: map[string][]string{
+			"default/tcp-app-svc:5001": {
+				"10.0.0.20:5001",
+			},
+		},
+	}
+
+	listenerPort := 2020
+
+	expected := &version2.TransportServerConfig{
+		Upstreams: []version2.StreamUpstream{
+			{
+				Name: "ts_default_tcp-server_tcp-app",
+				Servers: []version2.StreamUpstreamServer{
+					{
+						Address:     "10.0.0.20:5001",
+						MaxFails:    1,
+						FailTimeout: "10s",
+					},
+				},
+				UpstreamLabels: version2.UpstreamLabels{
+					ResourceName:      "tcp-server",
+					ResourceType:      "transportserver",
+					ResourceNamespace: "default",
+					Service:           "tcp-app-svc",
+				},
+			},
+		},
+		Server: version2.StreamServer{
+			Port:                     listenerPort,
+			UDP:                      false,
+			StatusZone:               "tcp-listener",
+			ProxyPass:                "ts_default_tcp-server_tcp-app",
+			Name:                     "tcp-server",
+			Namespace:                "default",
+			ProxyConnectTimeout:      "60s",
+			ProxyNextUpstream:        false,
+			ProxyNextUpstreamTries:   0,
+			ProxyNextUpstreamTimeout: "0s",
+			ProxyTimeout:             "10m",
+			HealthCheck:              nil,
+			Snippets:                 []string{"deny  192.168.1.1;", "allow 192.168.1.0/24;"},
+		},
+	}
+
+	result := generateTransportServerConfig(&transportServerEx, listenerPort, true)
+	if diff := cmp.Diff(expected, result); diff != "" {
+		t.Errorf("generateTransportServerConfig() mismatch (-want +got):\n%s", diff)
+	}
+}
+
 func TestGenerateTransportServerConfigForTCP(t *testing.T) {
 	transportServerEx := TransportServerEx{
 		TransportServer: &conf_v1alpha1.TransportServer{
@@ -104,7 +179,7 @@ func TestGenerateTransportServerConfigForTCP(t *testing.T) {
 
 	listenerPort := 2020
 
-	expected := version2.TransportServerConfig{
+	expected := &version2.TransportServerConfig{
 		Upstreams: []version2.StreamUpstream{
 			{
 				Name: "ts_default_tcp-server_tcp-app",
@@ -136,14 +211,15 @@ func TestGenerateTransportServerConfigForTCP(t *testing.T) {
 			ProxyNextUpstreamTimeout: "0s",
 			ProxyTimeout:             "50s",
 			HealthCheck:              nil,
+			Snippets:                 []string{},
 		},
 	}
 
-	isPlus := true
-	result := generateTransportServerConfig(&transportServerEx, listenerPort, isPlus)
-	if !reflect.DeepEqual(result, expected) {
-		t.Errorf("generateTransportServerConfig() returned \n%+v but expected \n%+v", result, expected)
+	result := generateTransportServerConfig(&transportServerEx, listenerPort, true)
+	if diff := cmp.Diff(expected, result); diff != "" {
+		t.Errorf("generateTransportServerConfig() mismatch (-want +got):\n%s", diff)
 	}
+
 }
 
 func TestGenerateTransportServerConfigForTLSPasstrhough(t *testing.T) {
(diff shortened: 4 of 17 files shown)


internal/configs/transportserver_test.go
```diff
@@ -62,6 +62,164 @@ func TestTransportServerExString(t *testing.T) {
...
+			HealthCheck:              nil,
+			Snippets:                 []string{"deny  192.168.1.1;", "allow 192.168.1.0/24;"},
+		},
+	}
+	tsc := &transportServerConfigurator{
+		isPlus:         true,
+		enableSnippets: true,
+	}
+	result := generateTransportServerConfig(&transportServerEx, listenerPort, tsc)
+	if !reflect.DeepEqual(result, expected) {
```
Dean-Coakley, line 139: Consider using https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff here and in `TestGenerateTransportServerConfigForTCPSnippetsNotEnabled`.

Otherwise fail output can be hard to spot the error, example:
```
--- FAIL: TestGenerateTransportServerConfigForTCPSnippets (0.00s)
    /go/src/github.com/nginxinc/kubernetes-ingress/internal/configs/transportserver_test.go:139: generateTransportServerConfig() returned 
        {Server:{TLSPassthrough:false UnixSocket: Port:2020 UDP:false StatusZone:tcp-listener ProxyRequests:<nil> ProxyResponses:<nil> ProxyPass:ts_default_tcp-server_tcp-app Name:tcp-server Namespace:default ProxyTimeout:10m ProxyConnectTimeout:60s ProxyNextUpstream:false ProxyNextUpstreamTimeout:0s ProxyNextUpstreamTries:0 HealthCheck:<nil> Snippets:[deny  192.168.1.1; allow 192.168.1.0/24;]} Upstreams:[{Name:ts_default_tcp-server_tcp-app Servers:[{Address:10.0.0.20:5001 MaxFails:1 FailTimeout:10s}] UpstreamLabels:{Service:tcp-app-svc ResourceType:transportserver ResourceName:tcp-server ResourceNamespace:default}}]} but expected 
        {Server:{TLSPassthrough:false UnixSocket: Port:2020 UDP:false StatusZone:tcp-listener ProxyRequests:<nil> ProxyResponses:<nil> ProxyPass:ts_default_tcp-server_tcp-app Name:tcp-server Namespace:default ProxyTimeout:10m ProxyConnectTimeout:60s ProxyNextUpstream:false ProxyNextUpstreamTimeout: ProxyNextUpstreamTries:0 HealthCheck:<nil> Snippets:[deny  192.168.1.1; allow 192.168.1.0/24;]} Upstreams:[{Name:ts_default_tcp-server_tcp-app Servers:[{Address:10.0.0.20:5001 MaxFails:1 FailTimeout:10s}] UpstreamLabels:{Service:tcp-app-svc ResourceType:transportserver ResourceName:tcp-server ResourceNamespace:default}}]}
```

Dean-Coakley (approved):

docs-web/configuration/transportserver-resource.md
```diff
@@ -366,6 +366,41 @@ secure-app   46sm
 
 In the kubectl get and similar commands, you can also use the short name `ts` instead of `transportserver`.
 
+### Using Snippets
```
pleshakov, line 369: could you update the TOC with this section?

docs-web/configuration/transportserver-resource.md
```diff
@@ -366,6 +366,41 @@ secure-app   46sm
...
+
+Snippets allow you to insert raw NGINX config into different contexts of NGINX configuration. In the example below, we use snippets to configure several NGINX features in a TransportServer:
```
pleshakov, line 371:
```suggestion
Snippets allow you to insert raw NGINX config into different contexts of NGINX configuration. In the example below, we use snippets to configure [access control](http://nginx.org/en/docs/stream/ngx_stream_access_module.html) in a TransportServer:
```

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -62,6 +62,7 @@ type TransportServer struct {
 // TransportServerSpec is the spec of the TransportServer resource.
 type TransportServerSpec struct {
 	Listener           TransportServerListener `json:"listener"`
+	Snippets           string                  `json:"snippets"`
```
pleshakov, line 65: Could we name it `serverSnippets`?  (both go field and YAML field) because such snippets apply to the `server` context. This would be consistent with the snippets for HTTP resources - http, server and location snippets.

internal/configs/transportserver.go
```diff
@@ -16,6 +17,25 @@ type TransportServerEx struct {
 	PodsByIP        map[string]string
 }
 
+// TransportServerConfigurator generates a transportServer configuration
```
pleshakov, line 20:
```suggestion
// transportServerConfigurator generates a transportServer configuration
```

internal/configs/transportserver.go
```diff
@@ -29,10 +49,10 @@ func (tsEx *TransportServerEx) String() string {
 }
 
 // generateTransportServerConfig generates a full configuration for a TransportServer.
-func generateTransportServerConfig(transportServerEx *TransportServerEx, listenerPort int, isPlus bool) version2.TransportServerConfig {
+func generateTransportServerConfig(transportServerEx *TransportServerEx, listenerPort int, tsc *transportServerConfigurator) version2.TransportServerConfig {
```
pleshakov, line 52: this really looks a method of `transportServerConfigurator`. Could we convert the function to a method rather than passing `*transportServerConfigurator` as an argument?

internal/configs/transportserver_test.go
```diff
@@ -62,6 +62,164 @@ func TestTransportServerExString(t *testing.T) {
...
+			Snippets:                 []string{},
+		},
+	}
+	tsc := &transportServerConfigurator{
+		isPlus:         true,
+		enableSnippets: false,
+		warnings:       newWarnings(),
+	}
+	result := generateTransportServerConfig(&transportServerEx, listenerPort, tsc)
+	if !reflect.DeepEqual(result, expected) {
```
pleshakov, line 218: we need to check that warnings are generated in this case. for example, like we do here https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/virtualserver_test.go#L3341

internal/configs/configurator.go
```diff
@@ -551,7 +551,9 @@ func (cnf *Configurator) addOrUpdateTransportServer(transportServerEx *Transport
 	name := getFileNameForTransportServer(transportServerEx.TransportServer)
 
 	listener := cnf.globalCfgParams.Listeners[transportServerEx.TransportServer.Spec.Listener.Name]
-	tsCfg := generateTransportServerConfig(transportServerEx, listener.Port, cnf.isPlus)
+	tsc := newTransportServerConfigurator(cnf.isPlus, cnf.staticCfgParams)
+
+	tsCfg := generateTransportServerConfig(transportServerEx, listener.Port, tsc)
```
pleshakov, line 556: warnings need to be propagated back to Controller and reported there

internal/configs/configurator.go
```diff
@@ -1085,23 +1087,21 @@ func (cnf *Configurator) UpdateConfig(cfgParams *ConfigParams, ingExes []*Ingres
...
 	cnf.globalCfgParams = ParseGlobalConfiguration(globalConfiguration, cnf.staticCfgParams.TLSPassthrough)
 
+	updatedTransportServerExes := []*TransportServerEx{}
+	deletedTransportServerExes := []*TransportServerEx{}
+
 	for _, tsEx := range transportServerExes {
 		if cnf.CheckIfListenerExists(&tsEx.TransportServer.Spec.Listener) {
 			updatedTransportServerExes = append(updatedTransportServerExes, tsEx)
-
 			err := cnf.addOrUpdateTransportServer(tsEx)
```
pleshakov, line 1099: warnings from config generation need to be propagated back to the Controller

please note that this function is removed in https://github.com/nginxinc/kubernetes-ingress/pull/1415
a new one is created https://github.com/nginxinc/kubernetes-ingress/blob/8db9f1029fa4ab2ddd9c66ee8985f6c574616626/internal/configs/configurator.go#L1076

this function will also need to propagate warnings -- https://github.com/nginxinc/kubernetes-ingress/blob/8db9f1029fa4ab2ddd9c66ee8985f6c574616626/internal/configs/configurator.go#L634

internal/configs/transportserver_test.go
```diff
@@ -62,6 +62,164 @@ func TestTransportServerExString(t *testing.T) {
 	}
 }
 
+func TestGenerateTransportServerConfigForTCPSnippets(t *testing.T) {
```
pleshakov, line 65: input and output for TestGenerateTransportServerConfigForTCPSnippets 
 and TestGenerateTransportServerConfigForTCPSnippetsNotEnabled are almost the same. 
 
 is there a way to reduce duplication ?

internal/configs/transportserver.go
pleshakov, line 52: generateTransportServerConfig needs to report warnings, similarly to https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/configs/virtualserver.go#L264

internal/configs/transportserver.go
```diff
@@ -16,6 +17,25 @@ type TransportServerEx struct {
...
+type transportServerConfigurator struct {
+	isPlus         bool
+	enableSnippets bool
+	warnings       Warnings
+}
+
+// newTransportServerConfigurator creates a new TransportServerConfigurator
+func newTransportServerConfigurator(
+	isPlus bool,
+	staticParams *StaticConfigParams,
```
pleshakov, line 30: because we use only one parameter of `*StaticConfigParams`, perhaps it doesn't make sense to pass `*StaticConfigParams`? but instead, just a boolean flag?

internal/configs/transportserver.go
```diff
@@ -16,6 +17,25 @@ type TransportServerEx struct {
...
+) *transportServerConfigurator {
+	return &transportServerConfigurator{
+		isPlus:         isPlus,
+		enableSnippets: staticParams.EnableSnippets,
+		warnings:       make(map[runtime.Object][]string),
```
pleshakov, line 35: why not `newWarnings()` like in `transportserver_test.go` ?

pleshakov (commented):
Hi @LorcanMcVeigh 

Shall we also add `streamSnippets` field that configures snippets in the `stream` context ? similar to `http-snippets` https://deploy-preview-1413--nginx-kubernetes-ingress.netlify.app/nginx-ingress-controller/configuration/virtualserver-and-virtualserverroute-resources/#using-snippets

ogarrett:
> Hi @LorcanMcVeigh
> 
> Shall we also add `streamSnippets` field that configures snippets in the `stream` context ? similar to `http-snippets` https://deploy-preview-1413--nginx-kubernetes-ingress.netlify.app/nginx-ingress-controller/configuration/virtualserver-and-virtualserverroute-resources/#using-snippets

I agree - there are some directives that can only be specified at the `stream` context (including geo, keyval and map) and there may be cases where it's convenient to apply a directive there so that all stream servers inherit it

internal/configs/transportserver_test.go
LorcanMcVeigh: sure this makes sense

internal/configs/transportserver.go
soneillf5: I think it's best to keep these types of functions as is. 

A method operates on a struct, which can hold state. If the function is stateless, then we can more easily reason about and test the function. When we provide specific inputs, we always get the same outputs. If it converted to a method, then the outputs also depend on the state of the struct, which can change and is shared.

internal/configs/transportserver_test.go
soneillf5: We confined Warnings to `generateTransportServerConfig()` as the supporting infrastructure isn't available. Warnings are handled at a higher level and the TransportServer status is updated. Waiting on https://github.com/nginxinc/kubernetes-ingress/pull/1415 for this.

internal/configs/configurator.go
soneillf5: We confined Warnings to `generateTransportServerConfig()` as the supporting infrastructure isn't available. Warnings are handled at a higher level and the TransportServer status is updated. Waiting on https://github.com/nginxinc/kubernetes-ingress/pull/1415 for this.

internal/configs/configurator.go
soneillf5: We confined Warnings to `generateTransportServerConfig()` as the supporting infrastructure isn't available. Warnings are handled at a higher level and the TransportServer status is updated. Waiting on https://github.com/nginxinc/kubernetes-ingress/pull/1415 for this.

internal/configs/transportserver.go
soneillf5: We confined Warnings to `generateTransportServerConfig()` as the supporting infrastructure isn't available. Warnings are handled at a higher level and the TransportServer status is updated. Waiting on https://github.com/nginxinc/kubernetes-ingress/pull/1415 for this.

internal/configs/transportserver_test.go
soneillf5: I think in the case of two tests, the cost of the increased complexity doesn't help. If there were several tests, it would be a great idea.

internal/configs/transportserver.go
LorcanMcVeigh: Done

internal/configs/transportserver.go
LorcanMcVeigh: Done

internal/configs/transportserver.go
pleshakov: in that case, I think it's better to call it something like `transportServerConfigurationOptions`, not `transportServerConfigurator`. 

Also, warnings in that case -- better to make `generateTransportServerConfig` return them - rather than store them in `transportServerConfigurator`. Otherwise, we'll have a type that acts both as input and output, which also makes it  harder to reason about it.

internal/configs/transportserver_test.go
pleshakov: they are confined, but not tested - generateTransportServerConfig doesn't check that they were generated? Could you add a test?

internal/configs/transportserver.go
LorcanMcVeigh: done

pleshakov:
Hi @LorcanMcVeigh 

the changes look good to me. 

However, because of the future streamSnippets, we need to rename `snippets` to `serverSnippets, which would be similar to http and server snippets for VS :

```yaml
apiVersion: k8s.nginx.org/v1alpha1
kind: TransportServer
metadata:
  name: dns-tcp
spec:
  streamSnippets: |
    limit_conn_zone $binary_remote_addr zone=addr:10m;
  listener:
    name: dns-tcp 
    protocol: TCP
  serverSnippets: |
    limit_conn addr 1;
    limit_conn_log_level error;
  upstreams:
  - name: dns-app
    service: coredns
    port: 5353
  action:
    pass: dns-app
```

docs-web/configuration/transportserver-resource.md
```diff
@@ -366,6 +367,41 @@ secure-app   46sm
...
+Snippets allow you to insert raw NGINX config into different contexts of NGINX configuration. In the example below, we use snippets to configure [access control](http://nginx.org/en/docs/stream/ngx_stream_access_module.html) in a TransportServer:
+
+```yaml
+apiVersion: k8s.nginx.org/v1alpha1
+kind: TransportServer
+metadata:
+  name: cafe
+spec:
+  host: cafe.example.com
+  snippets: |
```
pleshakov, line 381: the field needs to be updated

internal/configs/transportserver.go
```diff
@@ -59,14 +59,17 @@ func generateTransportServerConfig(transportServerEx *TransportServerEx, listene
 		proxyTimeout = transportServerEx.TransportServer.Spec.SessionParameters.Timeout
 	}
 
-	statusZone := ""
+	serverSnippets := generateSnippets(snippetsEnabled, transportServerEx.TransportServer.Spec.ServerSnippets, []string{})
+	if !snippetsEnabled && (transportServerEx.TransportServer.Spec.ServerSnippets != "") {
+		return nil, fmt.Errorf("snippet specified but snippets feature is not enabled")
```
pleshakov, line 64: when we have a warning, we don't stop processing a resource - we record a warning and continue. Could you not return an error here, as it is not an error? This would be similar to how we handle warnings for VirtualServer 

I suggest removing this check from this PR completely and address it (with unit tests) in the next PR when warnings will be propagated fully to the Controller 

sorry, I just noticed this change :(

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -62,6 +62,7 @@ type TransportServer struct {
 // TransportServerSpec is the spec of the TransportServer resource.
 type TransportServerSpec struct {
 	Listener           TransportServerListener `json:"listener"`
+	ServerSnippets     string                  `json:"server-snippets"`
```
pleshakov, line 65: with new resources (post-VirtualServer,  like TransortServer and Policies), we follow th camel-case convention for field names. Could we have `serverSnippets` here?

internal/configs/transportserver.go
soneillf5: I think this is an error. If you specify a snippets in a resource, then you expect them to be enabled and working, otherwise why would the user add them?

LorcanMcVeigh: Yes i agree with Sean. If the user tries to use a snippet and the feature isn't enabled then that means the resource doesn't work and is invalid. We don't want to deploy half a resource or a resource with half the functionality that the user defined in the yaml configuration.

pleshakov: it makes sense: better to report an error and reject a TS in this case. This would be different from VS, but it's better user experience.

however, such validation happens here https://github.com/nginxinc/kubernetes-ingress/blob/master/pkg/apis/configuration/validation/transportserver.go#L25 not in `generateTransportServerConfig`.

docs-web/configuration/transportserver-resource.md
pleshakov: @LorcanMcVeigh there is another omission in the docs - we need to have snippets added to the table in this section https://docs.nginx.com/nginx-ingress-controller/configuration/transportserver-resource/#transportserver-specification Could you add that?

internal/configs/transportserver.go
LorcanMcVeigh: yes but snippets don't get validated and we don't provide such data, like if a command line arg is set, to that validation process.

pleshakov: yeah, but if you want users to get an error when they use snippets when snippets are disabled, validation package is the right place - this is where we do validation. that will involve propagating enable snippets status to transport server validator

docs-web/configuration/transportserver-resource.md
LorcanMcVeigh: I've added the section `using-snippets` that describes them and how they're used. It's unnecessary to add it twice. Is there something missing or is there a benefit to adding it in twice?

internal/configs/transportserver.go
LorcanMcVeigh: Users do get an error when they use snippets when snippets are disabled. The `generateTransportServerConfig` validation error is propagated to `updateTransportServerEvents` function in `controller.go` which will throw an event that will notify the user. 

This validation of `snippets` requires the k8s resource and the context of the IC. This package handles the validation of the k8s resource fields and it is not like the resource is not formed correctly or has incorrect fields. It's a valid string and thats all the context this package should have. The CLI arg is an implementation detail and is independent of the resource so this validation package shouldn't know about it.

docs-web/configuration/transportserver-resource.md
pleshakov: the tables describe all the supported fields of the resource. so this is for completeness and consistency

internal/configs/transportserver.go
pleshakov: The added approach with handling errors is inconsistent with the approach we use for config generation: we always rely on the latest version of the resource, not the latest good version.

When a resource is rejected with an error, but its config is not removed or regenerated, this means an IC replica keeps using the latest good version.  However, other replicas might have a different good version stored. More over, new replicas (which could be created in case of replicas failures, node draining, scaling, upgrades) will only see the latest version in the cluster. As a result, the configuration will skew among replicas.

> This package handles the validation of the k8s resource fields and it is not like the resource is not formed correctly or has incorrect fields.

the validation depends on the enabled features. For example, validation for Policies and VirtualServers rejects resources that use Plus features, validation for TransportServers rejects TLS Passthrough resources if TLS Passthrough is not enabled.

Note that Plus or Passthrough are configured via cli args - so they are constants, because changing CLI arguments requires an upgrade of a deployment - it is not something that could be configured via ConfigMap or GlobalConfiguration. That's why it is possible to make validation decisions based on those conditions.

if you don't think a configured snippet is an error, then it should be a warning and handled like a warning - in that case the config is still generated, but the warning is reported to the users as an event and in the status of the resource.

soneillf5: Think I have it:
```
Status:
  Message:  TransportServer default/dns-tcp was rejected with error: spec.serverSnippets: Forbidden: snippet specified but snippets feature is not enabled
  Reason:   Rejected
  State:    Invalid
Events:
  Type     Reason    Age                From                      Message
  ----     ------    ----               ----                      -------
  Warning  Rejected  20s (x2 over 21s)  nginx-ingress-controller  TransportServer default/dns-tcp was rejected with error: spec.serverSnippets: Forbidden: snippet specified but snippets feature is not enabled
```

internal/configs/transportserver.go
```diff
@@ -59,14 +59,14 @@ func generateTransportServerConfig(transportServerEx *TransportServerEx, listene
 		proxyTimeout = transportServerEx.TransportServer.Spec.SessionParameters.Timeout
 	}
 
-	statusZone := ""
+	serverSnippets := generateSnippets(snippetsEnabled, transportServerEx.TransportServer.Spec.ServerSnippets, []string{})
```
pleshakov, line 62: because of the recent change, we no longer need to pass  `snippetsEnabled` to `generateTransportServerConfig`, because after validation `transportServerEx.TransportServer.Spec.ServerSnippets` is (1) either empty when snippets are disabled or (2) it is either empty or set to something when snippets are enabled. This means there is no way snippetsEnabled is false and transportServerEx.TransportServer.Spec.ServerSnippets is set to something.

Perhaps we can have a special function for TransportServer, like
```
serverSnippets := generateSnippetsForTransportServer(transportServerEx.TransportServer.Spec.ServerSnippets, []string{})
```

or simply:
```
	var serverSnippets []string
	if transportServerEx.TransportServer.Spec.ServerSnippets != "" {
		serverSnippets = []string{strings.Split(transportServerEx.TransportServer.Spec.ServerSnippets, "\n")}
	}
```

that will also remove the need for the unit test `TestGenerateTransportServerConfigForTCPSnippetsNotEnabled`

pleshakov (commented):
@soneillf5 
the last change looks good!  however, because of the change, there is some remaining code and logic, which is not really used - that's why a few comments below

internal/configs/transportserver.go
soneillf5: I'm leaving the function as is and passing in false, removing the enable flag from the ts config.

PR #1419, opened 2021-03-02
Add golangci-lint as a dependency

lucacome:
In the continued effort to make it easier to manage dependencies, adding golang-ci to `go.mod` since now `golangci-lint-action` supports reading the version from it. This way local and CI are aligned (`make lint` now uses `go run`).


pleshakov:
not fully sure about vendoring a big project like golang ci.

would it lead to increased build time? especially, if we consider instructions here https://docs.nginx.com/nginx-ingress-controller/installation/building-ingress-controller-image/

perhaps another way of doing it is for us to write a GH action that would update goland ci version and things like go version, which dependabot doesn't do ?  :)  we can also have a check in the Makefile that can ensure the version, for local golang ci runs

lucacome:
it's part of `tools.go` that has a build constraint, it doesn't get pulled in at build time

lucacome:
@ciarams87 @soneillf5 any opinions about this?

soneillf5 (approved):

PR #1420, opened 2021-03-02
Support IngressClassName in TransportServer

pleshakov:
### Proposed changes
The TransportServer resource now includes the `spec.ingressClassName`, which users can set to choose which NGINX Ingress Controller they want to handle their TransportServer resources.  This is useful when multiple NGINX Ingress Controllers are running in the cluster (for example, one for prod and one for dev).


cmd/nginx-ingress/main.go
```diff
@@ -76,12 +76,13 @@ var (
...
-	For Kubernetes < 1.18, the Ingress Controller only processes resources that belong to its class
-	- i.e have the annotation "kubernetes.io/ingress.class" equal to the class.
+	For Kubernetes < 1.18, the Ingress Controller only processes resources that belong to its class - 
+	i.e have the annotation "kubernetes.io/ingress.class" (for Ingress resources)
+	or field "ingressClassName" (for VirtualServer/VirtualServerRoute/TransportServer resources) equal to the class.
 	Additionally, the Ingress Controller processes resources that do not have the class set,
 	which can be disabled by setting the "-use-ingress-class-only" flag
 
-	The Ingress Controller processes all the VirtualServer/VirtualServerRoute resources that do not have the "ingressClassName" field for all versions of kubernetes.`)
+	The Ingress Controller processes all the VirtualServer/VirtualServerRoute/TransportServer resources that do not have the "ingressClassName" field for all versions of kubernetes.`)
```
soneillf5, line 85: Can you clarify this sorry, it processes the resources that *do not* have ingressClassName ?

internal/k8s/configuration_test.go
```diff
@@ -2171,6 +2171,70 @@ func TestAddInvalidTransportServer(t *testing.T) {
 	}
 }
 
+func TestAddTransportServerWithIncorrectClass(t *testing.T) {
```
soneillf5, line 2174: Great test!

internal/k8s/controller_test.go
```diff
@@ -229,108 +230,94 @@ func TestHasCorrectIngressClass(t *testing.T) {
 	}
 }
 
-func TestHasCorrectIngressClassVS(t *testing.T) {
-	ingressClass := "ing-ctrl"
-	lbcIngOnlyTrue := &LoadBalancerController{
-		ingressClass:        ingressClass,
+func deepCopyWithIngressClass(obj interface{}, class string) interface{} {
```
soneillf5, line 233: Not sure about this function, all the paths looks the same.
Can it be replaced on line 315 by 
```
                objCopy := obj.DeepCopy()
 		objCopy.Spec.IngressClass = class
```

soneillf5 (changes requested):
All good except for the deepCopy() in the test.

cmd/nginx-ingress/main.go
pleshakov: @soneillf5 
 "The Ingress Controller processes all the VirtualServer/VirtualServerRoute/TransportServer resources with an empty "ingressClassName" field for all versions of kubernetes."  ?

internal/k8s/controller_test.go
pleshakov: @soneillf5 
we need to reference a concrete type to apply `objCopy.Spec.IngressClass = class`  , that's why we have a a type switch with 3 cases

pleshakov: if deepCopy is confusing, I can create a concrete type instead, but still do that three times:
```
	case *conf_v1.VirtualServer:
		vs := conf_v1.VirtualServer{}
		vs.Spec.IngressClass = class
		return &vs
```

soneillf5: No that's fine as it is, thanks for the explanation.

soneillf5 (approved):

docs-web/installation/running-multiple-ingress-controllers.md
```diff
@@ -22,7 +22,7 @@ The default Ingress class of NGINX Ingress Controller is `nginx`, which means th
 **Notes**: 
 * For Kubernetes < 1.18, if the class is not set in an Ingress configuration resource, the Ingress Controller will handle the resource. This is controlled via the `-use-ingress-class-only` argument. 
 * For Kubernetes >= 1.18, if the class is not set in an Ingress resource, Kubernetes will set it to the class of the default Ingress Controller. To make the Ingress Controller the default one, the `ingressclass.kubernetes.io/is-default-class` must be set on the IngressClass resource. See Step 3 *Create an IngressClass resource* of the [Create Common Resources](/nginx-ingress-controller/installation/installation-with-manifests/#create-common-resources) section.
-* For VirtualServer and VirtualServerRoute resources, the Ingress Controller will always handle resources with an empty class.
+* For VirtualServer, VirtualServerRoute, and TransportServer resources the Ingress Controller will always handle resources with an empty class.
```
lucacome, line 25:
```suggestion
* For VirtualServer, VirtualServerRoute and TransportServer resources the Ingress Controller will always handle resources with an empty class.
```

lucacome (approved):

pstackle:
@pleshakov Is there an estimate of when this change will be included in a release?

pleshakov:
Hi @pstackle 

it will go into 1.11.0, scheduled for the end of march

PR #1426, opened 2021-03-04
Fix python tests

pleshakov:
### Proposed changes
One way of fixing the failing python tests, which was due to  https://github.com/nginxinc/kubernetes-ingress/pull/1371


soneillf5 (commented):
Does this change fix the tests? If so, why change the behaviour of generateTime instead of creating a new function? Does that break different tests ?

vepatel (approved):

pleshakov:
Hi @soneillf5 

> Does this change fix the tests? 

yes. the pipeline is green

>  If so, why change the behaviour of generateTime instead of creating a new function? Does that break different tests ?

The change broke one test for generateTransportServer configuration, that was fixed with using a different default 0 -> 0s -- https://github.com/nginxinc/kubernetes-ingress/blob/3af06bf3fd9e6a0a4f882fab676357390a679863/internal/configs/transportserver.go#L84

I don't see a point of keeping the old function. `generateTime` is a new (this release cycle) function, so we have only new code that rely one it

soneillf5 (approved):

Dean-Coakley (approved):

PR #1444, opened 2021-03-11
Fix setting host header in action proxy

pleshakov:
### Proposed changes
Previously, a configuration like below will make NGINX send two host headers to the backend - the default "$host" and the configured "bar.example.com":
```yaml
apiVersion: k8s.nginx.org/v1
kind: VirtualServer
metadata:
  name: foo 
spec:
  host: foo.example.com
  upstreams:
  - name: foo
    port: 8080
    service: backend-svc
  routes:
  - path: "/"
    action:
      proxy:
        upstream: foo
        requestHeaders:
          set:
          - name: Host
            value: bar.example.com
```
Now NGINX will only send one host header: by default - "$host", or the one configured in `requestHeaders`

I also updated some of the tests to print diffs. Otherwise, it was hard to see the mismatch in results.

### DIFF
--- a/internal/configs/virtualserver.go
+++ b/internal/configs/virtualserver.go
@@ -1476,23 +1476,24 @@ func generateLocation(path string, upstreamName string, upstream conf_v1.Upstrea
 func generateProxySetHeaders(proxy *conf_v1.ActionProxy) []version2.Header {
 	var headers []version2.Header
 
-	hostHeaderValue := "$host"
+	hasHostHeader := false
 
 	if proxy != nil && proxy.RequestHeaders != nil {
 		for _, h := range proxy.RequestHeaders.Set {
-			if strings.ToLower(h.Name) == "host" {
-				hostHeaderValue = h.Value
-				continue
-			}
-
 			headers = append(headers, version2.Header{
 				Name:  h.Name,
 				Value: h.Value,
 			})
+
+			if strings.ToLower(h.Name) == "host" {
+				hasHostHeader = true
+			}
 		}
 	}
 
-	headers = append(headers, version2.Header{Name: "Host", Value: hostHeaderValue})
+	if !hasHostHeader {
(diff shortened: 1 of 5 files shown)


internal/configs/virtualserver.go
```diff
@@ -1474,18 +1474,26 @@ func generateLocation(path string, upstreamName string, upstream conf_v1.Upstrea
...
-	}
-
 	var headers []version2.Header
-	for _, h := range proxy.RequestHeaders.Set {
-		headers = append(headers, version2.Header{
-			Name:  h.Name,
-			Value: h.Value,
-		})
+
+	hostHeaderValue := "$host"
```
soneillf5, line 1479: This code is unclear. I think you're trying to add a default host header with a value of '$host', if no such header exists. 

```
hasHostHeader := false

if proxy != nil && proxy.RequestHeaders != nil {
     for _, h := range proxy.RequestHeaders.Set {
 	    headers = append(headers, version2.Header{
 		    Name:  h.Name,
 		    Value: h.Value,
 	    })
 	    if strings.ToLower(h.Name) == "host" {
 		    hasHostHeader = true
 	    }
     }
}


if !hasHostHeader {
     headers = append(headers, version2.Header{Name: "Host", Value: "$host"})
}
 	
```

soneillf5 (changes requested):

internal/configs/virtualserver.go
pleshakov: @soneillf5 
the intention was also to prevent setting duplicate host headers, if a user configures that. In that case, the last one in the list would win.

however, since we didn't report any warning/error in that case and ignored the duplicate(s) silently, I think it's better to allow it(them) and make it a user error. Such configuration is valid from NGINX perspective. 

I used your suggestion.

soneillf5: Brilliant, thanks.

soneillf5 (approved):

lucacome (approved):

PR #1449, opened 2021-03-11
Add support for ClusterIP in upstreams

lucacome:
Add support for ClusterIP in upstreams

TODO:
- [x] docs

### DIFF
--- a/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
+++ b/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
@@ -428,7 +428,7 @@ tls:
      - ``map[string]string``
      - No
    * - ``use-cluster-ip``
-     - Enables using Cluster IP and port instead of the default behavior of using Pod IP and port. When this field is enabled, the fields that configure NGINX behavior related to multiple upstream servers (like ``lb-method`` and ``next-upstream``) will have no effect, as the Ingress Controller will configure NGINX with only one upstream server that will match the service Cluster IP.
+     - Enables using the Cluster IP and port of the service instead of the default behavior of using the IP and port of the pods. When this field is enabled, the fields that configure NGINX behavior related to multiple upstream servers (like ``lb-method` and `next-upstream``) will have no effect, as the Ingress Controller will configure NGINX with only one upstream server that will match the service Cluster IP.
      - ``boolean``
      - No
    * - ``port``
--- a/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
+++ b/docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
@@ -189,9 +189,9 @@ name: access-control
    * - ``name``
      - The name of a policy. If the policy doesn't exist or invalid, NGINX will respond with an error response with the `500` status code.
      - ``string``
-     - Yes 
+     - Yes
    * - ``namespace``
-     - The namespace of a policy. If not specified, the namespace of the VirtualServer resource is used. 
+     - The namespace of a policy. If not specified, the namespace of the VirtualServer resource is used.
      - ``string``
      - No
 ```
@@ -218,7 +218,7 @@ The route defines rules for matching client requests to actions like passing a r
      - ``string``
      - Yes
    * - ``policies``
-     - A list of policies. The policies override the policies of the same type defined in the ``spec`` of the VirtualServer. See `Applying Policies </nginx-ingress-controller/configuration/policy-resource/#applying-policies>`_ for more details. 
+     - A list of policies. The policies override the policies of the same type defined in the ``spec`` of the VirtualServer. See `Applying Policies </nginx-ingress-controller/configuration/policy-resource/#applying-policies>`_ for more details.
      - `[]policy <#virtualserver-policy>`_
      - No
    * - ``action``
@@ -351,7 +351,7 @@ action:
      - ``string``
      - Yes
    * - ``policies``
-     - A list of policies. The policies override *all* policies defined in the route of the VirtualServer that references this resource. The policies also override the policies of the same type defined in the ``spec`` of the VirtualServer. See `Applying Policies </nginx-ingress-controller/configuration/policy-resource/#applying-policies>`_ for more details. 
+     - A list of policies. The policies override *all* policies defined in the route of the VirtualServer that references this resource. The policies also override the policies of the same type defined in the ``spec`` of the VirtualServer. See `Applying Policies </nginx-ingress-controller/configuration/policy-resource/#applying-policies>`_ for more details.
      - `[]policy <#virtualserver-policy>`_
      - No
    * - ``action``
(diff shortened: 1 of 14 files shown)


pkg/apis/configuration/validation/virtualserver.go
```diff
@@ -409,9 +410,13 @@ func (vsv *VirtualServerValidator) validateUpstreams(upstreams []v1.Upstream, fi
 		} else {
 			upstreamNames.Insert(u.Name)
 		}
+		if u.UseClusterIP && u.Subselector != nil {
```
pleshakov, line 413: would it make sense to extend unit tests to cover this new condition ?

internal/k8s/controller.go
```diff
@@ -2277,35 +2278,48 @@ func (lbc *LoadBalancerController) createVirtualServerEx(virtualServer *conf_v1.
...
-		var podEndps []podEndpoint
-		var err error
+		var endps []string
+		if u.UseClusterIP {
+			s, err := lbc.getServiceForUpstream(virtualServer.Namespace, u.Service, u.Port)
+			if err != nil {
+				glog.Warningf("Error getting Service for Upstream %v: %v", u.Service, err)
+			}
+			port := strconv.Itoa(int(u.Port))
+			endps = append(endps, fmt.Sprintf("%s:%s", s.Spec.ClusterIP, port))
```
pleshakov, line 2288: if err != nil (to reproduce,  we can reference a non-existing svc in the VS/VSR),  s will be `nil` , so it will be a panic  
 
(same in the VSR part)

pleshakov, line 2288: so that there is no need in `strconv.Itoa`
(same in the VSR part)
```suggestion
			endps = append(endps, fmt.Sprintf("%s:%d", s.Spec.ClusterIP, u.Port))
```

internal/k8s/controller.go
```diff
@@ -2366,33 +2380,44 @@ func (lbc *LoadBalancerController) createVirtualServerEx(virtualServer *conf_v1.
 		for _, u := range vsr.Spec.Upstreams {
 			endpointsKey := configs.GenerateEndpointsKey(vsr.Namespace, u.Service, u.Subselector, u.Port)
 
-			var podEndps []podEndpoint
-			var err error
-			if len(u.Subselector) > 0 {
-				podEndps, err = lbc.getEndpointsForSubselector(vsr.Namespace, u)
+			var endps []string
+			if u.UseClusterIP {
+				s, err := lbc.getServiceForUpstream(virtualServer.Namespace, u.Service, u.Port)
```
pleshakov, line 2385: we need to use the VSR namespace here

internal/k8s/controller.go
```diff
@@ -2277,35 +2278,48 @@ func (lbc *LoadBalancerController) createVirtualServerEx(virtualServer *conf_v1.
 	for _, u := range virtualServer.Spec.Upstreams {
 		endpointsKey := configs.GenerateEndpointsKey(virtualServer.Namespace, u.Service, u.Subselector, u.Port)
 
-		var podEndps []podEndpoint
-		var err error
+		var endps []string
+		if u.UseClusterIP {
```
pleshakov, line 2282: I think it makes sense to refactor the getting endpoints logic into a function so that we can use it both for VS and VSR (prob after this PR) - the code looks identical

docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -507,6 +507,10 @@ tls:
      - Sets the size of the buffer used for reading the first part of a response received from the upstream server. See the `proxy_buffer_size <https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffer_size>`_ directive. The default is set in the ``proxy-buffer-size`` ConfigMap key.
      - ``string``
      - No
+   * - ``use-cluster-ip``
+     - Enables using Cluster IP and port instead of the default behavior of using Pod IP and port. When this field is enabled, the fields that configure NGINX behavior related to multiple upstream servers (like `lb-method` and `next-upstream`) will have no effect, as the Ingress Controller will configure NGINX with only one upstream server that will match the service Cluster IP.
```
pleshakov, line 511:
```suggestion
     - Enables using the Cluster IP and port of the service instead of the default behavior of using the IP and port of the pods. When this field is enabled, the fields that configure NGINX behavior related to multiple upstream servers (like `lb-method` and `next-upstream`) will have no effect, as the Ingress Controller will configure NGINX with only one upstream server that will match the service Cluster IP.
```

docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -507,6 +507,10 @@ tls:
      - Sets the size of the buffer used for reading the first part of a response received from the upstream server. See the `proxy_buffer_size <https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffer_size>`_ directive. The default is set in the ``proxy-buffer-size`` ConfigMap key.
      - ``string``
      - No
+   * - ``use-cluster-ip``
```
pleshakov, line 510: would it make sense to place it under service or subselector? Cause those options are related. this way for a user it would be easier to locate the `use-cluster-ip` field.

docs-web/configuration/virtualserver-and-virtualserverroute-resources.md
pleshakov, line 511: Note that "``" are rendered as italic, not monospace in this context. in case your intention was to use monospace
image
https://deploy-preview-1449--nginx-kubernetes-ingress.netlify.app/nginx-ingress-controller/configuration/virtualserver-and-virtualserverroute-resources/#upstream

pleshakov (commented):
Hi @lucacome 

Please see my feedback

There is one more issue - when the new field is enabled, since the IC no longer uses endpoints, but rather uses  the service ClusterIP,  there is no need to update and reload NGINX config (or update API for Plus) for the affected VirtualServers/VirtualServerRoutes, when the endpoints related to the service are updated.
I think we can easily prevent it if we update `FindResourcesForEndpoints` in `Configuration` to use a different referenceChecker - the one that could take into consideration the upstream `use-cluster-ip` field. What do you think?

internal/k8s/controller.go
lucacome: yeah I thought about that but decided to leave it for later 👍

pkg/apis/configuration/validation/virtualserver.go
lucacome: I had a feeling I was missing something! 😅

internal/k8s/reference_checkers.go
```diff
@@ -121,6 +123,9 @@ func (rc *serviceReferenceChecker) IsReferencedByVirtualServer(svcNamespace stri
 	}
 
 	for _, u := range vs.Spec.Upstreams {
+		if rc.hasClusterIP && u.UseClusterIP {
```
pleshakov, line 126: there is an edge case:
```yaml
  upstreams:
  - name: tea-1
    service: tea-svc
    has-cluster-ip: true
    port: 80
  - name: tea-2
    service: tea-svc
    port: 80
```    

in this case, IsReferencedByVirtualServer and IsReferencedByVirtualServerRoute need to return true.

So we need `continue` rather then `return false` in case of `if rc.hasClusterIP && u.UseClusterIP`

pleshakov (approved):

internal/k8s/reference_checkers_test.go
```diff
@@ -1044,3 +1044,129 @@ func TestIsPolicyIsReferenced(t *testing.T) {
 		}
 	}
 }
+
+func TestEndpointIsReferencedByVirtualServerAndVirtualServerRoutes(t *testing.T) {
```
soneillf5, line 1048: great tests!

soneillf5 (approved):

PR #1457, opened 2021-03-13
Wait for caches to sync when IC starts

pleshakov:
### Proposed changes

When the Ingress Controller is ready, its readiness probe succeeds, and it starts accepting client traffic.

The IC becomes ready when it drains the work queue at the start - this means the IC generated the configuration for all Ingress and other relevant resources in the cluster.

However, because the IC didn't wait for the caches to sync before starting processing the work queue, the work queue could contain only a fraction of the resources in the cluster. As a result, the IC would drain the queue quickly and mistakenly become ready. This is a bug, as not every relevant resource was processed.

Now, before starting the work queue, the IC waits for the caches to sync. This fixes the problem

Changes:
- Use shared informer factories for most of the controllers (informers), instead of creating them manually. 
- For controllers that need special parameters, continue creating them manually
- Wait for the caches of all controllers to sync before starting processing the work queue

### DIFF
--- a/tests/suite/fixtures.py
+++ b/tests/suite/fixtures.py
@@ -537,9 +537,11 @@ def crd_ingress_controller_with_ap(
         print("------------------------- Register AP CRD -----------------------------------")
         ap_pol_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/appprotect.f5.com_appolicies.yaml")
         ap_log_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/appprotect.f5.com_aplogconfs.yaml")
+        ap_uds_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/appprotect.f5.com_apusersigs.yaml")
         vs_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/k8s.nginx.org_virtualservers.yaml")
         vsr_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/k8s.nginx.org_virtualserverroutes.yaml")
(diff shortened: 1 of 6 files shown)


soneillf5 (approved):
Approved subject to tests passing.

tests/suite/fixtures.py
```diff
@@ -537,9 +537,11 @@ def crd_ingress_controller_with_ap(
         print("------------------------- Register AP CRD -----------------------------------")
         ap_pol_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/appprotect.f5.com_appolicies.yaml")
         ap_log_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/appprotect.f5.com_aplogconfs.yaml")
+        ap_uds_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/appprotect.f5.com_apusersigs.yaml")
         vs_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/k8s.nginx.org_virtualservers.yaml")
         vsr_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/k8s.nginx.org_virtualserverroutes.yaml")
         pol_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/k8s.nginx.org_policies.yaml")
+        ts_crd_name = get_name_from_yaml(f"{DEPLOYMENTS}/common/crds-v1beta1/k8s.nginx.org_transportservers.yaml")
```
vepatel, line 544: do we need ts crd in AP?

pleshakov: whenever `-enable-custom-resources` is set to `true`, it is necessary to register all custom resources CRDs. otherwise, the informers will not sync, and the IC will not become ready

lucacome (approved):

vepatel (approved):

PR #1466, opened 2021-03-19
Feature transportserver stream snippets

soneillf5:
### Proposed changes
Add stream level snippets to transport servers.

### Checklist

### DIFF
--- a/docs-web/configuration/transportserver-resource.md
+++ b/docs-web/configuration/transportserver-resource.md
@@ -123,6 +123,10 @@ The TransportServer resource defines load balancing configuration for TCP, UDP, 
      - Specifies which Ingress Controller must handle the TransportServer resource.
      - ``string``
      - No
+   * - ``streamSnippets``
+     - Sets a custom snippet in the ``stream`` context.
+     - ``string``
+     - No
    * - ``serverSnippets``
      - Sets a custom snippet in the ``server`` context.
      - ``string``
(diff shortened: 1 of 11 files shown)


docs-web/configuration/transportserver-resource.md
```diff
@@ -391,6 +391,27 @@ spec:
     port: 80
 ```
 
+Snippets can also be specified for a stream. In the example below, we use snippets to configure [access control](http://nginx.org/en/docs/stream/ngx_stream_access_module.html) and we set the [logging level](http://nginx.org/en/docs/http/ngx_http_limit_conn_module.html#limit_conn_zone) for cases when the server limits the number of connections in a TransportServer:
```
pleshakov, line 394: better to configure a complete feature, rather than just a parameter. otherwise, we configure a parameter for a feature that is not enabled and configured.

(I also fixed the link - it was for http module, not for stream)

```suggestion
Snippets can also be specified for a stream. In the example below, we use snippets to configure [access control](http://nginx.org/en/docs/stream/ngx_stream_access_module.html) and [limit the number of connections](https://nginx.org/en/docs/stream/ngx_stream_limit_conn_module.html):
```

docs-web/configuration/transportserver-resource.md
```diff
@@ -391,6 +391,27 @@ spec:
...
+
+
+```yaml
+apiVersion: k8s.nginx.org/v1alpha1
+kind: TransportServer
+metadata:
+  name: cafe
+spec:
+  host: cafe.example.com
+  streamSnippets: |
```
pleshakov, line 404: could you also add an entry for `streamSnippets` to the table for the spec here https://docs.nginx.com/nginx-ingress-controller/configuration/transportserver-resource/#transportserver-specification ? `Sets a custom snippet in the stream context.` would do the job

pleshakov (commented):
@soneillf5 
code looks good! 
some comments on the docs

lucacome (approved):

docs-web/configuration/transportserver-resource.md
soneillf5: done

pleshakov (approved):

ciarams87 (approved):
👏🏼

PR #1479, opened 2021-03-23
Add load balancing method to transport server config

soneillf5:
### Proposed changes
This change allows a user to specify the load balancing method used by the TransportServer resource.

It changes the defaults to match the Nginx defaults of round robin. 

It has different validation from the http load balancing method.

### Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/master/CONTRIBUTING.md) doc
- [x ] I have added tests that prove my fix is effective or that my feature works
- [ x] I have updated necessary documentation
- [x ] I have rebased my branch onto master
- [x ] I will ensure my PR is targeting the master branch and pulling from my branch from my own fork

### DIFF
--- a/docs-web/configuration/transportserver-resource.md
+++ b/docs-web/configuration/transportserver-resource.md
@@ -174,6 +174,7 @@ port: 8443
 maxFails: 3
 maxConns: 100
 failTimeout: 30s
+loadBalancingMethod: least_conn
 ```
 
 ```eval_rst
@@ -211,6 +212,10 @@ failTimeout: 30s
    * - ``healthCheck``
      - The health check configuration for the Upstream. See the `health_check <https://nginx.org/en/docs/stream/ngx_stream_upstream_hc_module.html#health_check>`_ directive. Note: this feature is supported only in NGINX Plus.
      - `healthcheck <#upstream-healthcheck>`_
+     - No
+   * - ``loadBalancingMethod``
+     - The method used to load balance the upstream servers. By default, connections are distributed between the servers using a weighted round-robin balancing method. See the `upstream <http://nginx.org/en/docs/stream/ngx_stream_upstream_module.html#upstream>`_ section for available methods and their details.
+     - ``string``
      - No
 
 ```
--- a/internal/configs/version2/nginx-plus.transportserver.tmpl
+++ b/internal/configs/version2/nginx-plus.transportserver.tmpl
@@ -2,7 +2,9 @@
 upstream {{ $u.Name }} {
     zone {{ $u.Name }} 256k;
 
-    random two least_conn;
+    {{ if $u.LoadBalancingMethod }}
+    {{ $u.LoadBalancingMethod }};
+    {{ end }}
 
     {{ range $s := $u.Servers }}
     server {{ $s.Address }} max_fails={{ $s.MaxFails }} fail_timeout={{ $s.FailTimeout }} max_conns={{ $s.MaxConnections }};
--- a/pkg/apis/configuration/v1alpha1/types.go
+++ b/pkg/apis/configuration/v1alpha1/types.go
@@ -85,13 +85,14 @@ type TransportServerListener struct {
 
 // Upstream defines an upstream.
 type Upstream struct {
-	Name        string       `json:"name"`
-	Service     string       `json:"service"`
-	Port        int          `json:"port"`
-	FailTimeout string       `json:"failTimeout"`
-	MaxFails    *int         `json:"maxFails"`
-	MaxConns    *int         `json:"maxConns"`
-	HealthCheck *HealthCheck `json:"healthCheck"`
+	Name                string       `json:"name"`
+	Service             string       `json:"service"`
+	Port                int          `json:"port"`
+	FailTimeout         string       `json:"failTimeout"`
+	MaxFails            *int         `json:"maxFails"`
+	MaxConns            *int         `json:"maxConns"`
+	HealthCheck         *HealthCheck `json:"healthCheck"`
+	LoadBalancingMethod string       `json:"loadBalancingMethod"`
 }
 
 // HealthCheck defines the parameters for active Upstream HealthChecks.
--- a/pkg/apis/configuration/validation/transportserver.go
+++ b/pkg/apis/configuration/validation/transportserver.go
@@ -2,8 +2,10 @@ package validation
 
 import (
 	"fmt"
+	"regexp"
+	"strings"
 
-	v1alpha1 "github.com/nginxinc/kubernetes-ingress/pkg/apis/configuration/v1alpha1"
+	"github.com/nginxinc/kubernetes-ingress/pkg/apis/configuration/v1alpha1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -13,13 +15,15 @@ import (
 type TransportServerValidator struct {
 	tlsPassthrough  bool
 	snippetsEnabled bool
+	isPlus          bool
 }
 
 // NewTransportServerValidator creates a new TransportServerValidator.
-func NewTransportServerValidator(tlsPassthrough bool, snippetsEnabled bool) *TransportServerValidator {
+func NewTransportServerValidator(tlsPassthrough bool, snippetsEnabled bool, isPlus bool) *TransportServerValidator {
 	return &TransportServerValidator{
 		tlsPassthrough:  tlsPassthrough,
 		snippetsEnabled: snippetsEnabled,
+		isPlus:          isPlus,
 	}
 }
 
@@ -37,7 +41,7 @@ func (tsv *TransportServerValidator) validateTransportServerSpec(spec *v1alpha1.
 	isTLSPassthroughListener := isPotentialTLSPassthroughListener(&spec.Listener)
 	allErrs = append(allErrs, validateTransportServerHost(spec.Host, fieldPath.Child("host"), isTLSPassthroughListener)...)
 
-	upstreamErrs, upstreamNames := validateTransportServerUpstreams(spec.Upstreams, fieldPath.Child("upstreams"))
+	upstreamErrs, upstreamNames := validateTransportServerUpstreams(spec.Upstreams, fieldPath.Child("upstreams"), tsv.isPlus)
 	allErrs = append(allErrs, upstreamErrs...)
 
 	allErrs = append(allErrs, validateTransportServerUpstreamParameters(spec.UpstreamParameters, fieldPath.Child("upstreamParameters"), spec.Listener.Protocol)...)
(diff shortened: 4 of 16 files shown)


ciarams87 (approved):
🔥

pkg/apis/configuration/validation/transportserver.go
```diff
@@ -169,11 +172,79 @@ func validateTransportServerUpstreams(upstreams []v1alpha1.Upstream, fieldPath *
...
+	"random least_time=last_byte":      true,
+	"least_time connect":               true,
+	"least_time first_byte":            true,
+	"least_time last_byte":             true,
+	"least_time last_byte inflight":    true,
+}
+
+func validateHashLoadBalancingMethod(method string) error {
+	keyWords := strings.Split(method, " ")
+	if len(keyWords) > 0 && keyWords[0] == "hash" {
```
pleshakov, line 239: this code is similar to the validation for http lb method, which has a bug. 
A value like this `hash $bla` will make NGINX fail to reload, because the variable `bla` doesn't exist in the Config - it's neither built-in nor user defined.

Another example of invalid value - `hash abc;breakingout;`, where the `;` after `abc` 

There is a solution for that - have a list of allowed variables and validate the passed value to make sure only the allowed variables are used. This will be similar to the validation done here https://github.com/nginxinc/kubernetes-ingress/blob/b47bc251a0dae8e8bd8c4f9adab9be0eeb33da10/pkg/apis/configuration/validation/virtualserver.go#L1000 . We just need to choose variables. Also, we can prevent escaping by placing the value in `""`.

Would it make sense to do that in a different PR and address this issue for both http and stream upstreams?

pleshakov (changes requested):
Hi @soneillf5 
Please see my feedback

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -84,12 +84,13 @@ type TransportServerListener struct {
...
-	FailTimeout string       `json:"failTimeout"`
-	MaxFails    *int         `json:"maxFails"`
-	HealthCheck *HealthCheck `json:"healthCheck"`
+	Name                string       `json:"name"`
+	Service             string       `json:"service"`
+	Port                int          `json:"port"`
+	FailTimeout         string       `json:"failTimeout"`
+	MaxFails            *int         `json:"maxFails"`
+	HealthCheck         *HealthCheck `json:"healthCheck"`
+	LoadBalancingMethod string       `json:"loadBalancingMethod"`
```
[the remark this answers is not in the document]
soneillf5: The name is suitable as is.

docs-web/configuration/transportserver-resource.md
```diff
@@ -203,6 +204,9 @@ failTimeout: 30s
      - The health check configuration for the Upstream. See the `health_check <https://nginx.org/en/docs/stream/ngx_stream_upstream_hc_module.html#health_check>`_ directive. Note: this feature is supported only in NGINX Plus.
      - `healthcheck <#upstream-healthcheck>`_
      - No
+   * - ``loadBalancingMethod``
```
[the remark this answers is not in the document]
soneillf5: Yes, good catch

internal/configs/version2/nginx-plus.transportserver.tmpl
```diff
@@ -2,7 +2,9 @@
 upstream {{ $u.Name }} {
     zone {{ $u.Name }} 256k;
 
-    random two least_conn;
+    {{ if $u.LoadBalancingMethod }}
+    $u.LoadBalancingMethod;
```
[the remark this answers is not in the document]
soneillf5: I can make the default `random two least_conn;` but it will effect things.

pkg/apis/configuration/v1alpha1/types.go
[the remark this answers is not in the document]
soneillf5: I've updated it to allow round robin. It's a shame though, as it's inconsistent with Nginx.

internal/configs/version2/nginx-plus.transportserver.tmpl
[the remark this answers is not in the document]
soneillf5: great catch, I keep messing that up.

pkg/apis/configuration/validation/transportserver.go
```diff
@@ -169,11 +172,79 @@ func validateTransportServerUpstreams(upstreams []v1alpha1.Upstream, fieldPath *
...
+	"random two":            true,
+	"random two least_conn": true,
+}
+
+var nginxPlusStreamLoadBalanceValidInput = map[string]bool{
+	"least_conn":                       true,
+	"random":                           true,
+	"random two":                       true,
+	"random two least_conn":            true,
+	"random two least_time":            true,
```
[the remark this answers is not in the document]
soneillf5: removed

pkg/apis/configuration/validation/transportserver.go
soneillf5: I'll add the validation for escapes, it's important to catch that.

soneillf5: The 'hash' method is non structured. I.e the load balancing method's type is a string. If we want to modify the value of a string, in this case splitting the hash method into parts and wrapping the value in quotes, it would be better to specify the hash load balancing method as a richer type. We can then safely parse and modify it's parameters.

soneillf5: @pleshakov I've updated the validation logic, it will match either `hash $var` or `hash $var consistent`. The documentation is unclear https://nginx.org/en/docs/stream/ngx_stream_upstream_module.html?&_ga=2.207794642.1395714857.1620389921-1881412668.1619606515#hash as it mentions the hash key can be a text, number,variable or combination of them. I've simplified the validation to say if the key contains a `$` then it must be a valid var. 

I'm also not sure which vars are valid, I've put one in to show the logic. Do you know how we can get a list of valid vars for the hash method ? http://nginx.org/en/docs/varindex.html

pleshakov: > I'm also not sure which vars are valid, I've put one in to show the logic. Do you know how we can get a list of valid vars for the hash method ? http://nginx.org/en/docs/varindex.html

From NGINX perspective, all variables available in the stream module are ok to use. However, the most common use case is to use `$remote_addr`. For HTTP load balancing, $uri, headers, args, etc. could be used. However, since it is TCP load balancing, we don't have them here. So we can just start `$remote_addr` and extend that list later if any new use cases are discovered.

One use case when you may want to add some text is if you want change the distribution of the IP addresses for the existing configuration. For example, you can change `hash ${remote_addr}AAA`  to `"hash ${remote_addr}AAB"`  and that will cause re-distribution. 

If we want to support a combination of variables and text, then we should use the existing `validateStringWithVariables` function, which supports validating a string containing variables and text.

soneillf5: @pleshakov I've updated the code to use `validateStringWithVariables` 
Also updated unit tests and confirmed it's working as intended

docs-web/configuration/transportserver-resource.md
```diff
@@ -207,6 +208,10 @@ failTimeout: 30s
...
+     - The method used to load balance the upstream servers. By default, connections are distributed between the servers using a weighted round-robin balancing method. See the `upstream <http://nginx.org/en/docs/stream/ngx_stream_upstream_module.html#upstream>`_ section for available methods and their details.
```
pleshakov, line 212:
```suggestion
     - The method used to load balance the upstream servers. See the `upstream <https://nginx.org/en/docs/stream/ngx_stream_upstream_module.html#upstream>`_ section for available methods and their details. To use the round-robin method, specify ``round_robin``. The ``hash`` method only supports ``$remote_addr`` variable with optional text and the consistent parameter. For example, ``hash ${remote_addr}``,  ``hash ${remote_addr}AAA``, ``hash $remote_addr consistent``. The default is ``random two least_conn``. 
```

pkg/apis/configuration/validation/transportserver.go
```diff
@@ -171,11 +175,87 @@ func validateTransportServerUpstreams(upstreams []v1alpha1.Upstream, fieldPath *
...
+	allErrs := field.ErrorList{}
+	if method == "" {
+		return allErrs
+	}
+
+	method = strings.TrimSpace(method)
+
+	if strings.HasPrefix(method, "hash") {
+		errs := validateHashLoadBalancingMethod(method, fieldPath, isPlus)
+		return append(allErrs, errs...)
```
pleshakov, line 195: return validateHashLoadBalancingMethod(method, fieldPath, isPlus) ?

pkg/apis/configuration/validation/transportserver_test.go
```diff
@@ -206,6 +206,122 @@ func TestValidateTransportServerHost(t *testing.T) {
...
+			isPlus:   false,
+			hasError: true,
+		},
+		{
+			method:   "hash ${remote_addr}",
+			isPlus:   false,
+			hasError: false,
+		},
+		{
+			method:   "hash ${remote_addr}AAA",
```
pleshakov, line 236: Could you add one more case
```
		},
		{
			method:   `hash ${remote_addr}"`,
			isPlus:   false,
			hasError: true,
		},
```
This will cover `if !escapedStringsFmtRegexp.MatchString(method) {`

pkg/apis/configuration/validation/transportserver_test.go
```diff
@@ -206,6 +206,122 @@ func TestValidateTransportServerHost(t *testing.T) {
...
+			hasError: true,
+		}, {
+			method:   "random two least_time=connect",
+			isPlus:   true,
+			hasError: true,
+		},
+	}
+
+	for _, test := range tests {
+		allErrs := getErrors(validateLoadBalancingMethod(test.method, field.NewPath("method"), test.isPlus))
```
pleshakov, line 307: what's the motivation for using `getErrors`?  

with
```
=== RUN   TestValidateTransportServerLoadBalancingMethod
    transportserver_test.go:315: validateLoadBalancingMethod("random two least_conn /", false) returned errors [method: Invalid value: "random two least_conn /": load balancing method is not valid: random two least_conn /] for valid input
--- FAIL: TestValidateTransportServerLoadBalancingMethod (0.00s)
```

without
```
=== RUN   TestValidateTransportServerLoadBalancingMethod
    transportserver_test.go:315: validateLoadBalancingMethod("random two least_conn /", false) returned errors [method: Invalid value: "random two least_conn /": load balancing method is not valid: random two least_conn /] for valid input
--- FAIL: TestValidateTransportServerLoadBalancingMethod (0.00s)

```

deployments/helm-chart/crds/appprotect.f5.com_aplogconfs.yaml
```diff
@@ -15,55 +15,51 @@ spec:
   preserveUnknownFields: false
   scope: Namespaced
   versions:
```
pleshakov, line 17: could you exclude the app protect crds changes?

I added a PR for that https://github.com/nginxinc/kubernetes-ingress/pull/1587

pleshakov (commented):
Hi @soneillf5  Please see a few final comments. Otherwise, looks good

pkg/apis/configuration/validation/transportserver_test.go
soneillf5: done

pleshakov (approved):
🚀

PR #1480, opened 2021-03-23
Add max connections config to TransportServer

soneillf5:
### Proposed changes
This change allows users to specify the max number of connections to the proxied server. https://nginx.org/en/docs/stream/ngx_stream_upstream_module.html#max_conns

### Checklist

### DIFF
--- a/internal/configs/transportserver.go
+++ b/internal/configs/transportserver.go
@@ -169,6 +169,7 @@ func generateStreamUpstream(upstream *conf_v1alpha1.Upstream, upstreamNamer *ups
 
 	name := upstreamNamer.GetNameForUpstream(upstream.Name)
 	maxFails := generateIntFromPointer(upstream.MaxFails, 1)
+	maxConnections := generateIntFromPointer(upstream.MaxConnections, 0)
 	failTimeout := generateTimeWithDefault(upstream.FailTimeout, "10s")
 
 	for _, e := range endpoints {
@@ -183,9 +184,10 @@ func generateStreamUpstream(upstream *conf_v1alpha1.Upstream, upstreamNamer *ups
 
 	if !isPlus && len(endpoints) == 0 {
 		upsServers = append(upsServers, version2.StreamUpstreamServer{
-			Address:     nginxNonExistingUnixSocket,
-			MaxFails:    maxFails,
-			FailTimeout: failTimeout,
+			Address:        nginxNonExistingUnixSocket,
+			MaxFails:       maxFails,
+			FailTimeout:    failTimeout,
+			MaxConnections: maxConnections,
 		})
 	}
 
--- a/internal/configs/version2/nginx.transportserver.tmpl
+++ b/internal/configs/version2/nginx.transportserver.tmpl
@@ -5,7 +5,7 @@ upstream {{ $u.Name }} {
     random two least_conn;
 
     {{ range $s := $u.Servers }}
-    server {{ $s.Address }} max_fails={{ $s.MaxFails }} fail_timeout={{ $s.FailTimeout }};
+    server {{ $s.Address }} max_fails={{ $s.MaxFails }} fail_timeout={{ $s.FailTimeout }} max_conns=$s.MaxConnections;
     {{ end }}
 }
 {{ end }}
--- a/internal/configs/version2/stream.go
+++ b/internal/configs/version2/stream.go
@@ -15,9 +15,10 @@ type StreamUpstream struct {
 
 // StreamUpstreamServer defines a stream upstream server.
 type StreamUpstreamServer struct {
-	Address     string
-	MaxFails    int
-	FailTimeout string
+	Address        string
+	MaxFails       int
+	FailTimeout    string
+	MaxConnections int
 }
 
 // StreamServer defines a server in the stream module.
--- a/pkg/apis/configuration/v1alpha1/types.go
+++ b/pkg/apis/configuration/v1alpha1/types.go
@@ -84,12 +84,13 @@ type TransportServerListener struct {
 
 // Upstream defines an upstream.
 type Upstream struct {
-	Name        string       `json:"name"`
-	Service     string       `json:"service"`
-	Port        int          `json:"port"`
-	FailTimeout string       `json:"failTimeout"`
-	MaxFails    *int         `json:"maxFails"`
-	HealthCheck *HealthCheck `json:"healthCheck"`
+	Name           string       `json:"name"`
+	Service        string       `json:"service"`
+	Port           int          `json:"port"`
+	FailTimeout    string       `json:"failTimeout"`
+	MaxFails       *int         `json:"maxFails"`
+	MaxConnections *int         `json:"maxConnections"`
+	HealthCheck    *HealthCheck `json:"healthCheck"`
 }
 
 // HealthCheck defines the parameters for active Upstream HealthChecks.
--- a/deployments/common/crds-v1beta1/k8s.nginx.org_transportservers.yaml
+++ b/deployments/common/crds-v1beta1/k8s.nginx.org_transportservers.yaml
@@ -114,6 +114,8 @@ spec:
                         type: integer
                       timeout:
                         type: string
+                  maxConnections:
+                    type: integer
                   maxFails:
                     type: integer
                   name:
--- a/deployments/common/crds/k8s.nginx.org_transportservers.yaml
+++ b/deployments/common/crds/k8s.nginx.org_transportservers.yaml
@@ -113,6 +113,8 @@ spec:
                             type: integer
                           timeout:
                             type: string
+                      maxConnections:
+                        type: integer
                       maxFails:
                         type: integer
                       name:
--- a/deployments/helm-chart/crds/k8s.nginx.org_transportservers.yaml
+++ b/deployments/helm-chart/crds/k8s.nginx.org_transportservers.yaml
@@ -113,6 +113,8 @@ spec:
                             type: integer
                           timeout:
                             type: string
+                      maxConnections:
+                        type: integer
                       maxFails:
                         type: integer
                       name:
--- a/docs-web/configuration/transportserver-resource.md
+++ b/docs-web/configuration/transportserver-resource.md
@@ -168,6 +168,7 @@ name: secure-app
 service: secure-app
 port: 8443
 maxFails: 3
+maxConnections: 100
 failTimeout: 30s
 ```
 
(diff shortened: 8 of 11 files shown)


pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -84,12 +84,13 @@ type TransportServerListener struct {
...
-	Port        int          `json:"port"`
-	FailTimeout string       `json:"failTimeout"`
-	MaxFails    *int         `json:"maxFails"`
-	HealthCheck *HealthCheck `json:"healthCheck"`
+	Name           string       `json:"name"`
+	Service        string       `json:"service"`
+	Port           int          `json:"port"`
+	FailTimeout    string       `json:"failTimeout"`
+	MaxFails       *int         `json:"maxFails"`
+	MaxConnections *int         `json:"maxConnections"`
```
pleshakov, line 92: could we use `maxConns` as the name of the field.
With new resources - Policies and TransportServer - we started using camelCase for field names. However, we still want to be consistent with the VS/VSR resources. For max conns, VS/VSR upstreams have `max-conns` field. So could we have a similar name here?

internal/configs/version2/nginx.transportserver.tmpl
```diff
@@ -5,7 +5,7 @@ upstream {{ $u.Name }} {
     random two least_conn;
 
     {{ range $s := $u.Servers }}
-    server {{ $s.Address }} max_fails={{ $s.MaxFails }} fail_timeout={{ $s.FailTimeout }};
+    server {{ $s.Address }} max_fails={{ $s.MaxFails }} fail_timeout={{ $s.FailTimeout }} max_conns=$s.MaxConnections;
```
pleshakov, line 8: there is a a bug here -- this will generate `max_conns=$s.MaxConnections;`

additionally, in this the Plus template, `if` is used, but here it doesn't use. Could we have a consistent approach?

`max_conns={{ $s.MaxConnections }};` should work without if, since the default is `0`.

internal/configs/version2/stream.go
```diff
@@ -15,9 +15,10 @@ type StreamUpstream struct {
 
 // StreamUpstreamServer defines a stream upstream server.
 type StreamUpstreamServer struct {
-	Address     string
-	MaxFails    int
-	FailTimeout string
+	Address        string
+	MaxFails       int
+	FailTimeout    string
+	MaxConnections int
```
pleshakov, line 21: In general, we try to follow the names of NGINX directives/parameter in the config structs, because they represent NGINX configuration at NGINX level - so better if we have 1 to 1 match where possible for simplicity. Could you rename to `MaxConns` ?

pleshakov (changes requested):
Hi @soneillf5 
Please see my feedback

pkg/apis/configuration/v1alpha1/types.go
soneillf5: I think the name is fine. I'd prefer if it was consistent within the resource.

internal/configs/version2/nginx.transportserver.tmpl
soneillf5: ah yes, will update.

internal/configs/version2/stream.go
soneillf5: I think the name is fine as it is.

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -249,3 +251,69 @@ def test_tcp_request_load_balanced_missing_service(
...
+            transport_server_setup.namespace,
+        )
+        wait_before_test()
+
+        # step 2 - make the number of allowed connections
+        port = transport_server_setup.public_endpoint.tcp_server_port
+        host = transport_server_setup.public_endpoint.public_ip
+
+        clients = []
+        for i in range(6):
```
vepatel, line 287: Is there a reason the max possible concurrent connections is 6 here but yaml says `maxConnections: 2`

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -249,3 +251,69 @@ def test_tcp_request_load_balanced_missing_service(
...
+            c = self.make_holding_connection(host, port)
+            clients.append(c)
+
+        # step 3 - assert the next connection fails
+        try:
+            c = self.make_holding_connection(host, port)
+            # making a connection should fail and throw an exception
+            assert c is None
```
vepatel, line 295: will there be an exception on exceeding the amount if connection?

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -249,3 +251,69 @@ def test_tcp_request_load_balanced_missing_service(
...
+        print(f"sending tcp requests to: {host}:{port}")
+        client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        client.connect((host, port))
+        client.sendall(b'hold')
+        response = client.recv(4096)
+        endpoint = response.decode()
+        print(f'response: {endpoint}')
+        return client
+
+    def test_tcp_request_max_connections(
```
vepatel, line 265: Although adding a test which test the functionality is great but I think the one which checks the key, value in configurtion i.e. stream-conf should take precedence. What do you thing @soneillf5 @pleshakov ?

tests/tcp-server/README.md
```diff
@@ -21,8 +21,8 @@ If you make changes to the server:
   local version ```-> imagePullPolicy: Never```
    * Test the changes
  * Include the change as part of the commit that requires the tcp-server change
-   * Build the docker image with an increased version number ```docker build -t tcp-server:v2```
-   * Push the docker image to the public repo
+   * Build the docker image with an increased version number ```docker build -t tcp-server:2.1```
+   * Push the docker image to the public repo ```docker push seanoneillf5/tcp-server:2.1```
```
vepatel, line 25: should we not specify the exact repo name and replace it with `<public_repo_name>` as user will be creating their own?

internal/configs/transportserver.go
```diff
@@ -172,13 +172,15 @@ func generateStreamUpstream(upstream *conf_v1alpha1.Upstream, upstreamNamer *ups
...
 
 	for _, e := range endpoints {
 		s := version2.StreamUpstreamServer{
-			Address:     e,
-			MaxFails:    maxFails,
-			FailTimeout: failTimeout,
+			Address:        e,
+			MaxFails:       maxFails,
+			FailTimeout:    failTimeout,
+			MaxConnections: maxConnections,
```
vepatel, line 183: @soneillf5 respective unit tests in https://github.com/nginxinc/kubernetes-ingress/blob/feature-max-connections-transportserver-upstream-server/internal/configs/transportserver_test.go mainly around value validation and errors on wrong config.

internal/configs/transportserver.go
```diff
@@ -172,13 +172,15 @@ func generateStreamUpstream(upstream *conf_v1alpha1.Upstream, upstreamNamer *ups
 
 	name := upstreamNamer.GetNameForUpstream(upstream.Name)
 	maxFails := generateIntFromPointer(upstream.MaxFails, 1)
+	maxConnections := generateIntFromPointer(upstream.MaxConnections, 0)
```
vepatel, line 175: @pleshakov Is there a default for max_conn in http://nginx.org/en/docs/stream/ngx_stream_upstream_module.html similar to `fail_timeout` ?

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: there's 6 here as there's 3 replicas of 2 max connections.

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: We catch the specific exception thrown for exceeding the max number of connections `except ConnectionResetError as E:`

If there's a different type of exception or no exception, the test will fail.

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: I think we can add a test for that. If the max connections functionality starts to fail, it will help diagnose the problem.

tests/tcp-server/README.md
soneillf5: https://hub.docker.com/repository/docker/seanoneillf5/tcp-server

I think seanoneillf5 is the repo name?

tests/suite/test_transport_server_tcp_load_balance.py
vepatel: okay, in that case is there a possibility to assert something in error text or `E.library` or `E.reason`?

internal/configs/transportserver.go
soneillf5: The default is 0 according to http://nginx.org/en/docs/stream/ngx_stream_upstream_module.html
Which means there's no limit

vepatel: perfect, my bad for not reading the text :grimacing:

internal/configs/transportserver.go
soneillf5: I've added a little unit test. I'm not sure what else to check as it's a very simple piece of config.

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: I've added a test for the config itself.

tests/settings.py
```diff
@@ -8,7 +8,7 @@
 TEST_DATA = f"{PROJECT_ROOT}/data"
 DEFAULT_IMAGE = "nginx/nginx-ingress:edge"
 DEFAULT_PULL_POLICY = "IfNotPresent"
-DEFAULT_IC_TYPE = "nginx-ingress"
+DEFAULT_IC_TYPE = "nginx-plus-ingress"
```
vepatel, line 11: @soneillf5 can we revert this back?

tests/tcp-server/README.md
pleshakov: Perhaps we can create a team/company repo?
cc @ciarams87

pleshakov (commented):
Hi @soneillf5 

The PR looks good to me, expect for the name of the field.

Could we use `maxConns` as the name of the field? There are two reasons:

(1) With new resources - Policies and TransportServer - we started using camelCase for field names. However, we still want to be consistent with the VS/VSR resources. For max conns, VS/VSR upstreams have `max-conns` field.  This is similar to how NGINX uses the same directive and parameters in HTTP and TCP configuration.
(2) For fields related to parameters and tunables (for example, load balancing parameters, rate-limiting parameters, etc.), we reuse NGINX nomenclature. For example, the existing `maxFails`, `failTimeout` in TransportServer.

tests/settings.py
soneillf5: done

soneillf5:
@pleshakov I've updated it to 'maxConns'. I think it's better not to, but since you've blocked the PR, I've made the change.

vepatel (approved):

pleshakov (approved):

PR #1492, opened 2021-03-31
Update the IC so GlobalConfig is not a dependency

soneillf5:
### Proposed changes
The globalConfig parameter for the IC was a dependency, it was parsed and validated on starting the IC.
This meant it was requirement for the helm chart and the operator. This change removes the parsing
but keeps the namespace/name definition so watchers are setup correctly, enabling GlobalConfig
resources to be added/updated/removed while the IC is running.

### Checklist


soneillf5:
This change means we don't need a GlobalConfig resource to start the IC.

We do need to specify the namespace/name though. What do the reviewers think about changing the default from empty string to a valid namespace/name ?

internal/configs/configurator.go
```diff
@@ -96,7 +96,6 @@ type Configurator struct {
 	nginxManager            nginx.Manager
 	staticCfgParams         *StaticConfigParams
 	cfgParams               *ConfigParams
-	globalCfgParams         *GlobalConfigParams
```
pleshakov, line 99: good catch!

deployments/helm-chart/values.yaml
```diff
@@ -154,19 +154,8 @@ controller:
   ## Enable TLS Passthrough on port 443. Requires controller.enableCustomResources.
   enableTLSPassthrough: false
 
-  globalConfiguration:
```
pleshakov, line 157: This is a convenient feature - helm allows admins to manage every aspect of KIC configuration via helm parameter, including GlobalConfiguration. Could you revert it back?

If you want to allow admins to reference the GlobalConfiguration, rather than creating it, you can add `name` parameter. So something like `globalConfiguration.name`. In this case, if `create` is `false`, Helm will configure KIC to use that Global configuration . The default name should be autogenerated as before - this is to allow installing multiple KICs in the same namespace, so that they don't have any collisions among their resources.

pleshakov (changes requested):
Hi @soneillf5 

> We do need to specify the namespace/name though. What do the reviewers think about changing the default from empty string to a valid namespace/name ?

Better to not have a default. The GlobalConfiguration is only required for TCP/UDP load balancing. Most of the people use only HTTP load balancing in the Ingress Controller, so there is no need to monitor for GlobalConfiguration resource.
Also, because multiple KICs could be deployed in the same namespace, having the same default will make those KICs depend on the same configuration, which isn't the usual case - because those are different KICs, their ConfigMaps and GlobalConfiguration will probably be different

deployments/helm-chart/values.yaml
soneillf5: Done

ciarams87 (approved):
LGTM

pleshakov:
Hi @soneillf5 

A few docs issues:

could you restore the docs for helm (both in deployments/helm-chart/README.md and https://github.com/nginxinc/kubernetes-ingress/pull/1492/files ) ?

image

Also, in the installation with manifests, we have 
image

However, the file `deployments/common/global-configuration.yaml` was removed.  It's better to leave it  - it could be used as the base to add necessary listeners.

soneillf5:
@pleshakov I've removed the reference to creating the GlobalConfig from `installation-with-manifests.md` to be consistent with the other customer resources.

pleshakov (approved):

PR #1500, opened 2021-04-02
Support ssl_reject_handshake in Ingress and VS

pleshakov:
### Proposed changes

To handle missing or invalid TLS Secrets in Ingress and VirtualServer
resources, previously the Ingress Controller would generate the
following configuration:
```
ssl_certificate /etc/nginx/secrets/default;
ssl_certificate_key /etc/nginx/secrets/default;
ssl_ciphers NULL;
```

The configuration will break any attempts for clients to establish
TLS connections for the affected server in NGINX.

The configuration has the following limitations:
- It requires a TLS cert and key (we used the default server Secret as
it was always present on the file system)
- It doesn't work if clients and NGINX use TLS v1.3: NGINX will terminate
TLS connection.

This commit introduces the new ssl_reject_handshake directive, which
configures NGINX to break any attempt to establish a TLS connection:
```
ssl_reject_handshake on;
```

The directive addresses the mentioned limitations: it doesn't require
a TLS cert and key and works with TLS v1.3.

## Changes for Changelog

* Previously, to handle missing or invalid TLS Secrets in Ingress and VirtualServer resources, the Ingress Controller would configure NGINX to break any attempts for clients to establish TLS connections to the affected hosts using `ssl_ciphers NULL;` in the NGINX configuration. The method didn't work for TLS v1.3. Now the Ingress Controller uses `ssl_reject_handshake on;`, which works for TLS v1.3.

### DIFF
--- a/internal/configs/ingress.go
+++ b/internal/configs/ingress.go
@@ -313,6 +313,7 @@ func addSSLConfig(server *version1.Server, owner runtime.Object, host string, in
 	}
 
 	var pemFile string
+	var rejectHandshake bool
 
 	if tlsSecret != "" {
 		secretRef := secretRefs[tlsSecret]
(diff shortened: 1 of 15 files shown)


internal/configs/ingress.go
```diff
@@ -313,6 +313,7 @@ func addSSLConfig(server *version1.Server, owner runtime.Object, host string, in
 	}
 
 	var pemFile string
+	var rejectHandshake bool
```
soneillf5, line 316: Consider assigning `rejectHandshake` as 'true' and handle the single case it's false? If I'm reading it right.

soneillf5 (approved):

lucacome (approved):

internal/configs/ingress.go
pleshakov: it's 2 cases for false vs 3 cases for true

PR #1503, opened 2021-04-03
Remove NGINX_VERSION from base images

lucacome:
- Let dependabot find new updates for base images.
- `NGINX_VERSION` is set in the base images, so it can be used where needed

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,9 +1,8 @@
 # syntax=docker/dockerfile:1.0-experimental
-ARG NGINX_VERSION=1.19.9
 ARG BUILD_OS=debian
 
 ############################################# Base image for Debian #############################################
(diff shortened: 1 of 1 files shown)


build/Dockerfile
```diff
@@ -138,9 +137,10 @@ COPY --chown=nginx:0 LICENSE /licenses
 
 ############################################# Base image for Openshift OSS #############################################
 FROM openshift-base AS openshift
-ARG NGINX_VERSION
 ARG IC_VERSION
 
+ENV NGINX_VERSION 1.19.9
```
lucacome, line 142: Unfortunately, the base image for openshift is not specific for NGINX and doesn't have the variable so we need to set it manually for now.

soneillf5 (approved):

ciarams87 (approved):
👍🏼

PR #1512, opened 2021-04-05
Don't require default server TLS secret

pleshakov:
### Proposed changes

If the default server TLS secret is not configured via -default-server-tls-secret cli arg or it is not present on the filesystem at /etc/nginx/secrets/default, the Ingress Controller will configure NGINX to reject TLS connections to the default server.

Note: The default server is used in NGINX configuration to handle HTTP and HTTPS request to hosts that are not configured by any Ingress, VirtualServer or TransportServer resources. The default server simply returns 404 responses.

Also note: because currently handling missing/invalid TLS secret relies on the default server TLS secret being present on the file system, this PR https://github.com/nginxinc/kubernetes-ingress/pull/1500 (which changes the handling) needs to be merged first.

### DIFF
--- a/cmd/nginx-ingress/main.go
+++ b/cmd/nginx-ingress/main.go
@@ -91,9 +91,10 @@ var (
 
 	defaultServerSecret = flag.String("default-server-tls-secret", "",
 		`A Secret with a TLS certificate and key for TLS termination of the default server. Format: <namespace>/<name>.
-	If not set, certificate and key in the file "/etc/nginx/secrets/default" are used. If a secret is set,
-	but the Ingress controller is not able to fetch it from Kubernetes API or a secret is not set and
-	the file "/etc/nginx/secrets/default" does not exist, the Ingress controller will fail to start`)
+	If not set, than the certificate and key in the file "/etc/nginx/secrets/default" are used. 
+	If "/etc/nginx/secrets/default" doesn't exist, the Ingress Controller will configure NGINX to reject TLS connections to the default server.
+	If a secret is set, but the Ingress controller is not able to fetch it from Kubernetes API or it is not set and the Ingress Controller 
+	fails to read the file "/etc/nginx/secrets/default", the Ingress controller will fail to start.`)
 
 	versionFlag = flag.Bool("version", false, "Print the version and git-commit hash and exit")
 
@@ -429,6 +430,8 @@ func main() {
 		nginxManager.AppProtectPluginStart(aPPluginDone)
 	}
 
+	var sslRejectHandshake bool
+
 	if *defaultServerSecret != "" {
 		secret, err := getAndValidateSecret(kubeClient, *defaultServerSecret)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


soneillf5 (approved):

cmd/nginx-ingress/main.go
```diff
@@ -440,9 +443,11 @@ func main() {
 		bytes := configs.GenerateCertAndKeyFileContent(secret)
 		nginxManager.CreateSecret(configs.DefaultServerSecretName, bytes, nginx.TLSSecretFileMode)
 	} else {
-		_, err = os.Stat("/etc/nginx/secrets/default")
+		_, err = os.Stat(configs.DefaultServerSecretPath)
 		if os.IsNotExist(err) {
-			glog.Fatalf("A TLS cert and key for the default server is not found")
+			sslRejectHandshake = true
+		} else {
+			glog.Fatalf("Error reading the default server TLS cert and key from %s: %v", configs.DefaultServerSecretPath, err)
```
soneillf5, line 450: is this correct?

cmd/nginx-ingress/main.go
```diff
@@ -440,9 +443,11 @@ func main() {
 		bytes := configs.GenerateCertAndKeyFileContent(secret)
 		nginxManager.CreateSecret(configs.DefaultServerSecretName, bytes, nginx.TLSSecretFileMode)
 	} else {
-		_, err = os.Stat("/etc/nginx/secrets/default")
+		_, err = os.Stat(configs.DefaultServerSecretPath)
```
soneillf5, line 446: Can we write a unit test for this logic?

soneillf5 (changes requested):

cmd/nginx-ingress/main.go
pleshakov: good idea 👍  there was a bug

cmd/nginx-ingress/main.go
```diff
@@ -886,3 +888,11 @@ func minVersion(min string) (v *util_version.Version) {
 
 	return minVer
 }
+
+func checkDefaultServerSecretStatErr(statErr error) (exists bool, err error) {
+	if os.IsNotExist(statErr) {
+		return false, nil
+	}
+
+	return statErr == nil, statErr
```
soneillf5, line 897: Sorry for the additional comments, I don't want to be too harsh, but this doesn't look correct yet. 

if os.IsNotExist() -> is true then the file doesn't exist, however you're returning (statErr == nil), shouldn't that just be ` return true, statErr` ? In which case the code becomes:
```
if os.IsNotExist(statErr) {
    return false, nil
}
return true, statErr
```
which simplifies to:
```
return os.IsNotExist(statErr), statErr
```

I don't really understand this function, it appears to be just wrapping os.IsNotExist. 

In terms of idiomatic error handling in Go, I think the usual practice is something like:
```
   if os.IsTimeout(err) {
      ...
   }
   if os.IsNotExist(err) {
       ...
   }
   ...
   // handle valid case
   return value
```

soneillf5 (changes requested):

cmd/nginx-ingress/main.go
pleshakov: resolved offline

lucacome (approved):

soneillf5 (approved):

PR #1513, opened 2021-04-06
AP: add support for multiple log configs

rafwegv:
### Proposed changes
Will be introduced in NAP 3.2
Support multiple log destinations and formats. Since we split log config into two annotations, allowing admins to setup multiple log destination is implemented as a coma separated value to the existing log conf annotations, where a config corresponds to the same index of the destination annotation.
for example:
`appprotect.f5.com/app-protect-security-log: "default/logconf,default/logconf_splunk" ` 
`appprotect.f5.com/app-protect-security-log-destination: "syslog:server=127.0.0.1:514,syslog:server=192.168.1.1:514"`

Both lists need to be the same length, if one of the referenced configs is ivalid/nonexistent no logconfs are aplied.

### Checklist


internal/k8s/controller.go
```diff
@@ -2050,7 +2050,7 @@ func (lbc *LoadBalancerController) createIngressEx(ing *networking.Ingress, vali
 			if apLogConfAntn, exists := ingEx.Ingress.Annotations[configs.AppProtectLogConfAnnotation]; exists {
 				logConf, logDst, err := lbc.getAppProtectLogConfAndDst(ing)
 				if err != nil {
-					glog.Warningf("Error Getting App Protect policy %v for Ingress %v/%v: %v", apLogConfAntn, ing.Namespace, ing.Name, err)
+					glog.Warningf("Error Getting App Protect Log Comnfig %v for Ingress %v/%v: %v", apLogConfAntn, ing.Namespace, ing.Name, err)
```
Dean-Coakley, line 2053:
```suggestion
					glog.Warningf("Error Getting App Protect Log Config %v for Ingress %v/%v: %v", apLogConfAntn, ing.Namespace, ing.Name, err)
```

internal/k8s/appprotect/app_protect_resources.go
```diff
@@ -131,6 +131,15 @@ func ParseResourceReferenceAnnotation(ns, antn string) string {
 	return antn
 }
 
+// ParseResourceReferenceAnnotationList returns a slice of ns/names strings
+func ParseResourceReferenceAnnotationList(ns, antns string) []string {
```
Dean-Coakley, line 135: Optional: Maybe this should be just called `annotations`? For a moment I understand `antns` to mean `annotationNamespace`
```suggestion
func ParseResourceReferenceAnnotationList(ns, annotations string) []string {
```

```suggestion
func ParseResourceReferenceAnnotationList(ns, antnRefs string) []string {
```

Dean-Coakley (approved):

internal/k8s/controller.go
```diff
@@ -2160,27 +2160,35 @@ func (lbc *LoadBalancerController) createIngressEx(ing *networking.Ingress, vali
...
 	if _, exists := ing.Annotations[configs.AppProtectLogConfDstAnnotation]; !exists {
-		return nil, "", fmt.Errorf("Error: %v requires %v in %v", configs.AppProtectLogConfAnnotation, configs.AppProtectLogConfDstAnnotation, ing.Name)
+		return nil, []string{""}, fmt.Errorf("Error: %v requires %v in %v", configs.AppProtectLogConfAnnotation, configs.AppProtectLogConfDstAnnotation, ing.Name)
 	}
 
-	logDst = ing.Annotations[configs.AppProtectLogConfDstAnnotation]
+	logDsts = strings.Split(ing.Annotations[configs.AppProtectLogConfDstAnnotation], ",")
+	logConfNsNs := appprotect.ParseResourceReferenceAnnotationList(ing.Namespace, ing.Annotations[configs.AppProtectLogConfAnnotation])
+	if len(logDsts) != len(logConfNsNs) {
+		return nil, []string{""}, fmt.Errorf("Error Validating App Protect Destination and config for Ingress %v: LogConf and LogDestination must be equal length", ing.Name)
```
soneillf5, line 2172: why do we have an empty string in this return value ?

internal/configs/configurator.go
```diff
@@ -1221,20 +1221,21 @@ func createSpiffeCert(certs []*x509.Certificate) []byte {
...
 
 	if ingEx.AppProtectLogConf != nil {
-		logConfFileName := appProtectLogConfFileNameFromUnstruct(ingEx.AppProtectLogConf)
-		logConfContent := generateApResourceFileContent(ingEx.AppProtectLogConf)
-		cnf.nginxManager.CreateAppProtectResourceFile(logConfFileName, logConfContent)
-		apRes[appProtectLogConfKey] = logConfFileName + " " + ingEx.AppProtectLogDst
+		for i, logConf := range ingEx.AppProtectLogConf {
+			logConfFileName := appProtectLogConfFileNameFromUnstruct(logConf)
+			logConfContent := generateApResourceFileContent(logConf)
+			cnf.nginxManager.CreateAppProtectResourceFile(logConfFileName, logConfContent)
```
soneillf5, line 1236: I just looked into this function, I'm surprised it doesn't return an error. @pleshakov is there any reason the manager doesn't return errors from most of it's functions?

internal/k8s/reference_checkers.go
```diff
@@ -217,10 +219,14 @@ func newAppProtectResourceReferenceChecker(annotation string) *appProtectResourc
 	return &appProtectResourceReferenceChecker{annotation}
 }
 
+// In App Protect logConfs can be a coma separated list.
 func (rc *appProtectResourceReferenceChecker) IsReferencedByIngress(namespace string, name string, ing *networking.Ingress) bool {
-	if pol, exists := ing.Annotations[rc.annotation]; exists {
-		if pol == namespace+"/"+name || (namespace == ing.Namespace && pol == name) {
-			return true
+	if resName, exists := ing.Annotations[rc.annotation]; exists {
+		resNames := strings.Split(resName, ",")
```
soneillf5, line 225: do we need to split and iterate? Would a `string.Contains(resName, namespace+"/"+name)` work?

soneillf5 (approved):

internal/configs/configurator.go
pleshakov: @soneillf5 those errors are file IO related and treated as unrecoverable to prevent NGINX from diverging from the desired configuration.

tests/suite/test_app_protect_integration.py
```diff
@@ -410,3 +406,78 @@ def test_ap_enable_true_policy_correct_uds(
...
+            doc["metadata"]["annotations"]["appprotect.f5.com/app-protect-policy"] = ap_policy
+            doc["metadata"]["annotations"]["appprotect.f5.com/app-protect-enable"] = "True"
+            doc["metadata"]["annotations"][
+                "appprotect.f5.com/app-protect-security-log-enable"
+            ] = "True"
+
+            # both lists need to be the same length, if one of the referenced configs is invalid/non-existent then no logconfs are applied.
+            doc["metadata"]["annotations"][
+                "appprotect.f5.com/app-protect-security-log"
+            ] = f"{test_namespace}/logconf,{test_namespace}/logconf"
```
pleshakov, line 442: @vepatel Perhaps we should also test that implicit namespace works?
```suggestion
            ] = f"{test_namespace}/logconf,logconf"
```

pleshakov (commented):
@rafwegv please see my comments and suggestions.

Additionally, could we update the docs about changed annotations https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-annotations/#app-protect  ?

Also, do we need to update WAF policy as well https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#waf   ? Perhaps in a different PR? I suggest sharing the changes to the fields before starting implementing so that we all agree on the new configuration.

tests/suite/test_app_protect_integration.py
vepatel: but I've deployed test resources in a separate namespace so need to reference by the same, nothing related to test goes to default.

internal/configs/ingress.go
```diff
@@ -30,8 +36,8 @@ type IngressEx struct {
 	ValidHosts        map[string]bool
 	ValidMinionPaths  map[string]bool
 	AppProtectPolicy  *unstructured.Unstructured
-	AppProtectLogConf *unstructured.Unstructured
-	AppProtectLogDst  string
+	AppProtectLogConf []*unstructured.Unstructured
```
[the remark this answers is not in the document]
rafwegv: yep that makes sense. imlemented.

internal/k8s/appprotect/app_protect_resources.go
```diff
@@ -118,6 +118,15 @@ func ParseResourceReferenceAnnotation(ns, antn string) string {
 	return antn
 }
 
+// ParseResourceReferenceAnnotationList returns a slice of ns/names strings
+func ParseResourceReferenceAnnotationList(ns, annotations string) []string {
```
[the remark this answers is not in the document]
rafwegv: yep, done

internal/configs/ingress.go
```diff
@@ -20,19 +20,29 @@ const emptyHost = ""
 const appProtectPolicyKey = "policy"
 const appProtectLogConfKey = "logconf"
 
+// AppProtectResources holds namespace names of App Protect resources relavant to an Ingress
+type AppProtectResources struct {
+	AppProtectPolicy   string
+	AppProtectLogconfs []string
+}
+
+type AppProtectLog struct {
```
pleshakov, line 29: neat: could you add a doc string for consistency? since the type is exported

internal/k8s/controller.go
```diff
@@ -2160,27 +2159,38 @@ func (lbc *LoadBalancerController) createIngressEx(ing *networking.Ingress, vali
 	return ingEx
 }
 
-func (lbc *LoadBalancerController) getAppProtectLogConfAndDst(ing *networking.Ingress) (logConf *unstructured.Unstructured, logDst string, err error) {
-	logConfNsN := appprotect.ParseResourceReferenceAnnotation(ing.Namespace, ing.Annotations[configs.AppProtectLogConfAnnotation])
+func (lbc *LoadBalancerController) getAppProtectLogConfAndDst(ing *networking.Ingress) (apLog []configs.AppProtectLog, err error) {
```
pleshakov, line 2162: because we use `err` variable throughout the code in the method body and without any intention to return its value, I suggest dropping the names in `(apLog []configs.AppProtectLog, err error)` -> `([]configs.AppProtectLog, error)` and inside the body, simply declare `var apLog []configs.AppProtectLog` . And a small suggestion, because we have a slice, better to use plural `apLogs`.

pleshakov (commented):
Hi @rafwegv 

the changes looks good!  just a few small suggestions.

> Additionally, could we update the docs about changed annotations https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-annotations/#app-protect ?

are you planning to update docs in a separate PR?

pleshakov (approved):
🚀

PR #1516, opened 2021-04-06
Feature transport server status tests

soneillf5:
### Proposed changes
Add tests for TransportServer status functionality.

### Checklist

### DIFF
--- a/tests/data/transport-server-status/rejected-warning.yaml
+++ b/tests/data/transport-server-status/rejected-warning.yaml
@@ -1,0 +1,15 @@
+apiVersion: k8s.nginx.org/v1alpha1
+kind: TransportServer
+metadata:
+  name: transport-server
+spec:
+  listener:
+    # we specify a missing listener to generate a 'Warning' state
+    name: invalid-listener
+    protocol: TCP
+    upstreams:
+      - name: dns-app
+        service: coredns
+        port: 5353
+    action:
+      pass: dns-app
--- a/tests/suite/custom_resources_utils.py
+++ b/tests/suite/custom_resources_utils.py
@@ -88,6 +88,36 @@ def read_custom_resource(custom_objects: CustomObjectsApi, namespace, plural, na
         raise
 
 
+def read_custom_resource_v1alpha1(custom_objects: CustomObjectsApi, namespace, plural, name) -> object:
+    """
+    Get CRD information (kubectl describe output)
+
+    :param custom_objects: CustomObjectsApi
+    :param namespace: The custom resource's namespace
+    :param plural: the custom resource's plural name
+    :param name: the custom object's name
+    :return: object
+    """
+    print(f"Getting info for v1alpha1 crd {name} in namespace {namespace}")
+    try:
+        response = custom_objects.get_namespaced_custom_object(
+            "k8s.nginx.org", "v1alpha1", namespace, plural, name
+        )
+        pprint(response)
+        return response
+
+    except ApiException:
+        logging.exception(f"Exception occurred while reading CRD")
+        raise
+
+
+def read_ts(custom_objects: CustomObjectsApi, namespace, name) -> object:
+    """
+    Read TransportService resource.
+    """
+    return read_custom_resource_v1alpha1(custom_objects, namespace, "transportservers", name)
+
+
 def read_ap_crd(custom_objects: CustomObjectsApi, namespace, plural, name) -> object:
     """
     Get AppProtect CRD information (kubectl describe output)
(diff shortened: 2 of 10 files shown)


tests/suite/test_transport_server_status.py
```diff
@@ -0,0 +1,114 @@
...
+        wait_before_test()
+        response = read_v1alpha1_crd(
+            kube_apis.custom_objects,
+            transport_server_setup.namespace,
+            "transportservers",
+            transport_server_setup.name,
+        )
+        self.restore_transport_server(kube_apis, transport_server_setup)
+        assert (
+            response["status"]
```
pleshakov, line 84: shall we check the `Message` field as well? (here and in the other similar checks)

tests/suite/test_transport_server_status.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        patch_src = f"{TEST_DATA}/transport-server-status/rejected-warning.yaml"
+        patch_transport_server_from_yaml(
+            kube_apis.custom_objects,
+            transport_server_setup.name,
+            patch_src,
+            transport_server_setup.namespace,
+        )
+        wait_before_test()
+        response = read_v1alpha1_crd(
```
pleshakov, line 76: we have a number functions specific for TransportServers -- `patch_transport_server_from_yaml`, `create_ts`, `delete_ts`. However, `read_v1alpha1_crd` is not TransportServer specific. 

Would it make sense to also have `read_ts`?

tests/suite/fixtures.py
```diff
@@ -727,6 +733,64 @@ def fin():
...
+
+@pytest.fixture(scope="class")
+def transport_server_setup(
+        request, kube_apis, test_namespace
+) -> TransportServerSetup:
+    """
+    Prepare Transport Server Example.
+
+    :param request: internal pytest fixture to parametrize this method:
+        {example: virtual-server|virtual-server-tls|..., app_type: simple|split|...}
```
pleshakov, line 758: do the examples in the docs need to match TransportServer specific things?

tests/suite/fixtures.py
```diff
@@ -727,6 +733,64 @@ def fin():
...
+    print(
+        "------------------------- Deploy Transport Server Example -----------------------------------"
+    )
+
+    # deploy global config
+    global_config_file = f"{TEST_DATA}/{request.param['example']}/standard/global-configuration.yaml"
+    gc_resource = create_gc(kube_apis.custom_objects, global_config_file, "nginx-ingress")
+
+    # deploy dns
+    dns_file = f"{TEST_DATA}/{request.param['example']}/standard/dns.yaml"
```
pleshakov, line 774: do we always need to deploy a DNS app? 
for status check, we don't need really to check the traffic flow `client->NGINX->backend`. so I wonder whether we should deploy the example app conditionally

tests/suite/fixtures.py
```diff
@@ -727,6 +733,64 @@ def fin():
...
+    :param test_namespace:
+    :return: TransportServerSetup
+    """
+    print(
+        "------------------------- Deploy Transport Server Example -----------------------------------"
+    )
+
+    # deploy global config
+    global_config_file = f"{TEST_DATA}/{request.param['example']}/standard/global-configuration.yaml"
+    gc_resource = create_gc(kube_apis.custom_objects, global_config_file, "nginx-ingress")
```
pleshakov, line 771: We also have TransportServer for TLS Passthrough, which doesn't require a GlobalConfiguration, but requires the `-enable-tls-passthrough` cli arg.  would it be wise to assume we will have a different fixture for that?

tests/data/transport-server-status/rejected-warning.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: k8s.nginx.org/v1alpha1
+kind: TransportServer
+metadata:
+  name: transport-server
+spec:
+  listener:
+    # we specify an invalid listener to generate a 'Warning' state
+    name: invalid-listener
```
pleshakov, line 8: technically it is a missing or non-existing listener, because it doesn't exist. this is in contrast with `protocol: invalid-protocol`, where the protocol is indeed invalid.
you use "Test TransportServer status with a missing listener." in tests/suite/test_transport_server_status.py

tests/suite/test_transport_server_status.py
```diff
@@ -0,0 +1,114 @@
...
+@pytest.mark.ts
+@pytest.mark.parametrize(
+    "crd_ingress_controller, transport_server_setup",
+    [
+        (
+            {
+                "type": "complete",
+                "extra_args":
+                    [
+                        "-enable-custom-resources",
```
pleshakov, line 20: because status reporting depends on the replica being the leader, and leader election can take time, I wonder if you had any flakes or problems when running the new tests? Note that in test_virtual_server_status.py,  leader election is disabled `-enable-leader-election=false`

tests/suite/test_transport_server_status.py
soneillf5: I'm not sure. If we change the message, do we want this test to break? @vepatel ?

tests/suite/test_transport_server_status.py
soneillf5: I've updated the method names, wrapping a read_custom_resource_v1alpha.

tests/suite/fixtures.py
soneillf5: I don't think it adds value, so I've removed it entirely.

tests/suite/fixtures.py
soneillf5: Good point. We should update the fixture/tests when we implement those tests. https://nginxsoftware.atlassian.net/browse/IND-23927

tests/suite/fixtures.py
soneillf5: I think we do. I tested it locally, removing the dns server but the tests don't pass as the pods never become ready. I think this is because nginx requires a valid listener ? So without the dns service, it doesn't start.

tests/data/transport-server-status/rejected-warning.yaml
soneillf5: fixed

tests/suite/test_transport_server_status.py
soneillf5: I've had no issues locally but that doesn't mean it isn't a problem. Do you suggest disabling leader election?

vepatel: @soneillf5 I'd suggest disabling for now, I've had issues in pipeline when enabled but not locally.

tests/suite/test_transport_server_status.py
vepatel: @soneillf5 I'd say `reason` and `state` is good enough.

tests/suite/custom_resources_utils.py
```diff
@@ -249,6 +279,117 @@ def create_virtual_server_from_yaml(
         raise
 
 
+def create_ts(custom_objects: CustomObjectsApi, yaml_manifest, namespace) -> dict:
```
vepatel, line 282: Minor: can we change it to `create_ts_from_yaml` just to keep it inline with other lib functions.

tests/suite/custom_resources_utils.py
```diff
@@ -249,6 +279,117 @@ def create_virtual_server_from_yaml(
...
+
+    :param custom_objects: CustomObjectsApi
+    :param yaml_manifest: an absolute path to file
+    :param namespace:
+    :return: a dictionary representing the resource
+    """
+    return create_resource_from_yaml(custom_objects, yaml_manifest, namespace, "transportservers")
+
+
+def create_gc(custom_objects: CustomObjectsApi, yaml_manifest, namespace) -> dict:
```
vepatel, line 294: same here `create_gc_from_yaml`

tests/suite/custom_resources_utils.py
```diff
@@ -423,6 +564,32 @@ def delete_virtual_server(custom_objects: CustomObjectsApi, name, namespace) ->
     print(f"VirtualServer was removed with name '{name}'")
 
 
+def delete_transport_server(custom_objects: CustomObjectsApi, name, namespace) -> None:
```
vepatel, line 567: do we need both this one and `delete_ts`?

vepatel (approved):
@soneillf5  lgtm, just a few suggestions related to definition names in comments if you can take a look.

tests/suite/test_transport_server_status.py
soneillf5: updated to add `-enable-leader-election=false`

tests/suite/custom_resources_utils.py
soneillf5: done

tests/suite/custom_resources_utils.py
soneillf5: done

tests/suite/custom_resources_utils.py
soneillf5: removed `delete_transport_server`

tests/suite/fixtures.py
pleshakov: we need the app related listeners in GlobalConfiguration, but not the app itslef (its deployment and service)
In that case, the IC will generate the NGINX config with zero upstream servers, which is legit. For example, for TCP TS:
```nginx
upstream ts_default_dns-tcp_dns-app {
    zone ts_default_dns-tcp_dns-app 256k;
    random two least_conn;
}
server {
    listen 5353;

    status_zone dns-tcp;
    proxy_pass ts_default_dns-tcp_dns-app;
    proxy_timeout 10m;
    proxy_connect_timeout 60s;
}
```
the status will be:
Status:
  Message:  Configuration for default/dns-tcp was added or updated
  Reason:   AddedOrUpdated
  State:    Valid
```

pleshakov (approved):

PR #1537, opened 2021-04-15
Add tests for default server

pleshakov:
### Proposed changes
Adds tests for the default server for two cases:
- when the default TLS secret is configured
- when the default TLS secret is not configured


tests/suite/test_default_server.py
```diff
@@ -0,0 +1,88 @@
...
+
+from suite.resources_utils import create_secret_from_yaml, is_secret_present, delete_secret, wait_before_test, \
+    ensure_connection, replace_secret
+from suite.ssl_utils import get_server_certificate_subject
+from settings import TEST_DATA, DEPLOYMENTS
+
+
+def assert_cn(endpoint, cn):
+    host = "random" # any host would work
+    subject_dict = get_server_certificate_subject(endpoint.public_ip, host, endpoint.port_ssl)
```
soneillf5, line 13: Can we pass null or empty string or update the function to take the parameter optionally ?

soneillf5 (approved):
LGTM

vepatel (changes requested):
@pleshakov can we mark them as `@pytest.mark.ingresses` as current one will only run them in nightly!

tests/suite/test_default_server.py
pleshakov: while to test the default server we can use any host value, I think it's better to control that host value here  in `assert_cn`, rather than coming up with the default value in `get_server_certificate_subject`. the other places we use that function (test_tls.py, test_virtual_server_tls.py and test_wildcard_tls_secret.py) all need to set a specific host header  and can't rely on any default value.

vepatel (approved):

PR #1543, opened 2021-04-19
Add test for TransportServer load balancing config

soneillf5:
### Proposed changes
Adds a basic check for the config change relating to the number of replicas updating the stream/tcp load balancing config on the nginx instance.

### Checklist

### DIFF
--- a/tests/suite/fixtures.py
+++ b/tests/suite/fixtures.py
@@ -45,6 +45,7 @@ from suite.resources_utils import (
     delete_service,
     replace_configmap_from_yaml,
     delete_testing_namespaces,
+    get_first_pod_name,
 )
 from suite.resources_utils import (
     create_ingress_controller,
@@ -120,12 +121,13 @@ class PublicEndpoint:
         port_ssl (int):
     """
 
-    def __init__(self, public_ip, port=80, port_ssl=443, api_port=8080, metrics_port=9113):
+    def __init__(self, public_ip, port=80, port_ssl=443, api_port=8080, metrics_port=9113, tcp_server_port=3333):
         self.public_ip = public_ip
         self.port = port
         self.port_ssl = port_ssl
         self.api_port = api_port
         self.metrics_port = metrics_port
+        self.tcp_server_port = tcp_server_port
 
 
 class IngressControllerPrerequisites:
@@ -252,10 +254,10 @@ def ingress_controller_endpoint(
             namespace,
             f"{TEST_DATA}/common/service/nodeport-with-additional-ports.yaml",
         )
-        port, port_ssl, api_port, metrics_port = get_service_node_ports(
+        port, port_ssl, api_port, metrics_port, tcp_server_port = get_service_node_ports(
             kube_apis.v1, service_name, namespace
         )
-        return PublicEndpoint(public_ip, port, port_ssl, api_port, metrics_port)
+        return PublicEndpoint(public_ip, port, port_ssl, api_port, metrics_port, tcp_server_port)
     else:
         create_service_from_yaml(
             kube_apis.v1,
@@ -754,16 +756,24 @@ class TransportServerSetup:
         namespace (str):
     """
 
-    def __init__(self, name, namespace):
+    def __init__(self, name, namespace, ingress_pod_name, ic_namespace, public_endpoint: PublicEndpoint, resource):
         self.name = name
         self.namespace = namespace
+        self.ingress_pod_name = ingress_pod_name
+        self.ic_namespace = ic_namespace
+        self.public_endpoint = public_endpoint
+        self.resource = resource
 
 
 @pytest.fixture(scope="class")
-def transport_server_setup(request, kube_apis, test_namespace) -> TransportServerSetup:
+def transport_server_setup(
+        request, kube_apis, ingress_controller_prerequisites, test_namespace, ingress_controller_endpoint
+) -> TransportServerSetup:
     """
     Prepare Transport Server Example.
 
+    :param ingress_controller_endpoint:
+    :param ingress_controller_prerequisites:
     :param request: internal pytest fixture to parametrize this method
     :param kube_apis: client apis
     :param test_namespace:
@@ -779,9 +789,9 @@ def transport_server_setup(request, kube_apis, test_namespace) -> TransportServe
     )
     gc_resource = create_gc_from_yaml(kube_apis.custom_objects, global_config_file, "nginx-ingress")
 
-    # deploy dns
-    dns_file = f"{TEST_DATA}/{request.param['example']}/standard/dns.yaml"
-    create_items_from_yaml(kube_apis, dns_file, test_namespace)
+    # deploy service_file
+    service_file = f"{TEST_DATA}/{request.param['example']}/standard/service_deployment.yaml"
+    create_items_from_yaml(kube_apis, service_file, test_namespace)
 
     # deploy transport server
     transport_server_file = f"{TEST_DATA}/{request.param['example']}/standard/transport-server.yaml"
@@ -794,12 +804,22 @@ def transport_server_setup(request, kube_apis, test_namespace) -> TransportServe
     def fin():
         print("Clean up TransportServer Example:")
         delete_ts(kube_apis.custom_objects, ts_resource, test_namespace)
-        delete_items_from_yaml(kube_apis, dns_file, test_namespace)
+        delete_items_from_yaml(kube_apis, service_file, test_namespace)
         delete_gc(kube_apis.custom_objects, gc_resource, "nginx-ingress")
 
     request.addfinalizer(fin)
 
-    return TransportServerSetup(ts_resource["metadata"]["name"], test_namespace)
+    ic_pod_name = get_first_pod_name(kube_apis.v1, ingress_controller_prerequisites.namespace)
+    ic_namespace = ingress_controller_prerequisites.namespace
+
+    return TransportServerSetup(
+        ts_resource['metadata']['name'],
+        test_namespace,
+        ic_pod_name,
+        ic_namespace,
+        ingress_controller_endpoint,
+        ts_resource,
+    )
 
 
 @pytest.fixture(scope="class")
--- a/tests/suite/resources_utils.py
+++ b/tests/suite/resources_utils.py
@@ -131,6 +131,36 @@ def create_deployment_from_yaml(apps_v1_api: AppsV1Api, namespace, yaml_manifest
     return create_deployment(apps_v1_api, namespace, dep)
 
 
+def patch_deployment_from_yaml(apps_v1_api: AppsV1Api, namespace, yaml_manifest) -> str:
+    """
+    Create a deployment based on yaml file.
+
+    :param apps_v1_api: AppsV1Api
+    :param namespace: namespace name
+    :param yaml_manifest: absolute path to file
+    :return: str
+    """
+    print(f"Load {yaml_manifest}")
+    with open(yaml_manifest) as f:
+        dep = yaml.safe_load(f)
+    return patch_deployment(apps_v1_api, namespace, dep)
+
+
+def patch_deployment(apps_v1_api: AppsV1Api, namespace, body) -> str:
+    """
+    Create a deployment based on a dict.
+
+    :param apps_v1_api: AppsV1Api
+    :param namespace: namespace name
+    :param body: dict
+    :return: str
+    """
+    print("Patch a deployment:")
+    apps_v1_api.patch_namespaced_deployment(body['metadata']['name'], namespace, body)
+    print(f"Deployment patched with name '{body['metadata']['name']}'")
+    return body['metadata']['name']
+
+
 def create_deployment(apps_v1_api: AppsV1Api, namespace, body) -> str:
     """
     Create a deployment based on a dict.
@@ -165,7 +195,7 @@ def create_deployment_with_name(apps_v1_api: AppsV1Api, namespace, name) -> str:
         return create_deployment(apps_v1_api, namespace, dep)
 
 
-def scale_deployment(apps_v1_api: AppsV1Api, name, namespace, value) -> None:
+def scale_deployment(apps_v1_api: AppsV1Api, name, namespace, value) -> int:
     """
     Scale a deployment.
 
@@ -173,13 +203,15 @@ def scale_deployment(apps_v1_api: AppsV1Api, name, namespace, value) -> None:
     :param namespace: namespace name
     :param name: deployment name
     :param value: int
-    :return:
+    :return: original: int the original amount of replicas
     """
     print(f"Scale a deployment '{name}'")
     body = apps_v1_api.read_namespaced_deployment_scale(name, namespace)
+    original = body.spec.replicas
     body.spec.replicas = value
     apps_v1_api.patch_namespaced_deployment_scale(name, namespace, body)
     print(f"Scale a deployment '{name}': complete")
+    return original
 
 
 def create_daemon_set(apps_v1_api: AppsV1Api, namespace, body) -> str:
@@ -310,7 +342,7 @@ def create_service_with_name(v1: CoreV1Api, namespace, name) -> str:
         return create_service(v1, namespace, dep)
 
 
-def get_service_node_ports(v1: CoreV1Api, name, namespace) -> (int, int, int, int):
+def get_service_node_ports(v1: CoreV1Api, name, namespace) -> (int, int, int, int, int):
     """
     Get service allocated node_ports.
 
@@ -320,11 +352,11 @@ def get_service_node_ports(v1: CoreV1Api, name, namespace) -> (int, int, int, in
     :return: (plain_port, ssl_port, api_port, exporter_port)
     """
     resp = v1.read_namespaced_service(name, namespace)
-    assert len(resp.spec.ports) == 4, "An unexpected amount of ports in a service. Check the configuration"
+    assert len(resp.spec.ports) == 5, "An unexpected amount of ports in a service. Check the configuration"
     print(f"Service with an API port: {resp.spec.ports[2].node_port}")
     print(f"Service with an Exporter port: {resp.spec.ports[3].node_port}")
     return resp.spec.ports[0].node_port, resp.spec.ports[1].node_port,\
-        resp.spec.ports[2].node_port, resp.spec.ports[3].node_port
+        resp.spec.ports[2].node_port, resp.spec.ports[3].node_port, resp.spec.ports[4].node_port
 
 
 def wait_for_public_ip(v1: CoreV1Api, namespace: str) -> str:
@@ -747,6 +779,21 @@ def get_ingress_nginx_template_conf(v1: CoreV1Api, ingress_namespace, ingress_na
     :return: str
     """
     file_path = f"/etc/nginx/conf.d/{ingress_namespace}-{ingress_name}.conf"
+    return get_file_contents(v1, file_path, pod_name, pod_namespace)
+
+
+def get_ts_nginx_template_conf(v1: CoreV1Api, resource_namespace, resource_name, pod_name, pod_namespace) -> str:
+    """
+    Get contents of /etc/nginx/stream-conf.d/ts_{namespace}-{resource_name}.conf in the pod.
+
+    :param v1: CoreV1Api
+    :param resource_namespace:
+    :param resource_name:
+    :param pod_name:
+    :param pod_namespace:
+    :return: str
+    """
+    file_path = f"/etc/nginx/stream-conf.d/ts_{resource_namespace}_{resource_name}.conf"
     return get_file_contents(v1, file_path, pod_name, pod_namespace)
 
 
--- a/tests/suite/test_transport_server_tcp_load_balance.py
+++ b/tests/suite/test_transport_server_tcp_load_balance.py
@@ -1,0 +1,251 @@
+import pytest
+import re
+import socket
+
+from suite.resources_utils import (
+    wait_before_test,
+    get_ts_nginx_template_conf,
+    scale_deployment
+)
+from suite.custom_resources_utils import (
+    patch_ts,
+    read_ts,
+    delete_ts,
+    create_ts_from_yaml,
+)
+from settings import TEST_DATA
+
+@pytest.mark.ts
+@pytest.mark.parametrize(
+    "crd_ingress_controller, transport_server_setup",
+    [
+        (
+            {
+                "type": "complete",
+                "extra_args":
+                    [
+                        "-global-configuration=nginx-ingress/nginx-configuration",
+                        "-enable-leader-election=false"
+                    ]
+            },
+            {"example": "transport-server-tcp-load-balance"},
+        )
+    ],
+    indirect=True,
+)
+class TestTransportServerTcpLoadBalance:
+
+    def restore_ts(self, kube_apis, transport_server_setup) -> None:
+        """
+        Function to revert a TransportServer resource to a valid state.
+        """
+        patch_src = f"{TEST_DATA}/transport-server-status/standard/transport-server.yaml"
+        patch_ts(
+            kube_apis.custom_objects,
+            transport_server_setup.name,
+            patch_src,
+            transport_server_setup.namespace,
+        )
+
+    def test_number_of_replicas(
+        self, kube_apis, crd_ingress_controller, transport_server_setup, ingress_controller_prerequisites
+    ):
+        """
+        The load balancing of TCP should result in 4 servers to match the 4 replicas of a service.
+        """
+        original = scale_deployment(kube_apis.apps_v1_api, "tcp-service", transport_server_setup.namespace, 4)
+        wait_before_test()
+
+        result_conf = get_ts_nginx_template_conf(
+            kube_apis.v1,
+            transport_server_setup.namespace,
+            transport_server_setup.name,
+            transport_server_setup.ingress_pod_name,
+            ingress_controller_prerequisites.namespace
+        )
+
+        pattern = 'server .*;'
+        num_servers = len(re.findall(pattern, result_conf))
+
+        assert num_servers is 4
+
+        scale_deployment(kube_apis.apps_v1_api, "tcp-service", transport_server_setup.namespace, original)
+        wait_before_test()
+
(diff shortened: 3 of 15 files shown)


tests/suite/fixtures.py
```diff
@@ -785,7 +790,11 @@ def fin():
 
     request.addfinalizer(fin)
 
-    return TransportServerSetup(ts_resource['metadata']['name'], test_namespace)
+    ingress_name = "nginx-ingress"
```
vepatel, line 793: ingress_name? or ic_name?

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,64 @@
...
+class TestTransportServerTcpLoadBalance:
+
+    def restore_service(self, kube_apis, transport_server_setup) -> None:
+        """
+        Function to revert a TransportServer resource to a valid state.
+        """
+        dns_file = f"{TEST_DATA}/transport-server-tcp-load-balance/standard/dns.yaml"
+        patch_deployment_from_yaml(kube_apis.apps_v1_api, transport_server_setup.namespace, dns_file)
+
+    @pytest.mark.sean
```
vepatel, line 39: `@pytest.mark.smoke`

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,64 @@
...
+    def test_number_of_replicas(
+        self, kube_apis, crd_ingress_controller, transport_server_setup, ingress_controller_prerequisites
+    ):
+        """
+        The load balancing of TCP should result in 3 servers to match the 3 replicas of a service.
+        """
+        dns_file = f"{TEST_DATA}/transport-server-tcp-load-balance/more-replicas.yaml"
+        patch_deployment_from_yaml(kube_apis.apps_v1_api, transport_server_setup.namespace, dns_file)
```
pleshakov, line 47: To control the number of replicas, another approach is to only to update the scale subresource using  `apps_v1_api.patch_namespaced_deployment_scale()`  ( which corresponds to  Patch Scale in https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#deployment-v1-apps )
This way, there will be no need to have multiple YAMLS for different number of replicas - it will be possible just to patch the number of replicas
This would be similar to the following kubectl command:
```
kubectl scale deploy coffee --replicas=1 -v=8
. . .
I0419 16:01:17.808008   45385 request.go:968] Request Body: {"spec":{"replicas":1}}
. . .

File: tests/suite/resources_utils.py [hunk 4]
```diff
@@ -750,6 +780,21 @@ def get_ingress_nginx_template_conf(v1: CoreV1Api, ingress_namespace, ingress_na
     return get_file_contents(v1, file_path, pod_name, pod_namespace)
 
 
+def get_ingress_nginx_template_ts_conf(v1: CoreV1Api, resource_namespace, resource_name, ingress_name, ingress_namespace) -> str:
```
pleshakov, line 783: for consistency with `get_ingress_nginx_template_conf`
```suggestion
def get_ts_nginx_template_conf(v1: CoreV1Api, resource_namespace, resource_name, ingress_name, ingress_namespace) -> str:
```

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,64 @@
...
+            transport_server_setup.ingress_pod_name,
+            ingress_controller_prerequisites.namespace
+        )
+
+        pattern = 'server.*max_fails=1 fail_timeout=10s;'
+        num_servers = len(re.findall(pattern, result_conf))
+
+        assert num_servers is 3
+
+        self.restore_service(kube_apis, transport_server_setup)
```
pleshakov, line 64: because restore is done after the rest of the code, this code will not be invoked if something goes wrong before it (like a failing assert). Could that cause any problems?

tests/suite/resources_utils.py
```diff
@@ -750,6 +780,21 @@ def get_ingress_nginx_template_conf(v1: CoreV1Api, ingress_namespace, ingress_na
...
+    """
+    Get contents of /etc/nginx/conf.d/{namespace}-{ingress_name}.conf in the pod.
```
pleshakov, line 785: the doc doesn't match the names used in the function

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,64 @@
...
+
+        result_conf = get_ingress_nginx_template_ts_conf(
+            kube_apis.v1,
+            transport_server_setup.namespace,
+            transport_server_setup.name,
+            transport_server_setup.ingress_pod_name,
+            ingress_controller_prerequisites.namespace
+        )
+
+        pattern = 'server.*max_fails=1 fail_timeout=10s;'
```
pleshakov, line 59: I wonder if it makes sense to match only the server and IP:port, and not match other parameters - for those we can have a separate test?
 
Also note that we have `max_conns` parameter in our backlog. So when we implement it, we might need to update this test if it matches everything until `;`.

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: done

tests/tcp-server/main.go
```diff
@@ -0,0 +1,48 @@
+package main
```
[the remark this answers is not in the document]
soneillf5: I've removed the python script, and since discovered `nc 192.168.64.2 3333`

tests/suite/resources_utils.py
soneillf5: fixed

tests/suite/fixtures.py
soneillf5: I'm not sure of the difference, luckily it's not needed anymore so I removed it!

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: Very good point. something equivalent to go's `defer` would be nice here.
I've looked into it briefly, https://docs.pytest.org/en/stable/fixture.html#yield-fixtures-recommended
but it would be a bigger change than this PR should make. Definitely a target for a refactor.

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: done

soneillf5: I should write a git hook to prevent my name getting committed

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: yup, it's nice to test for a specific thing that we care about. If `max_conns` or something else change, it shouldn't fail. I've updated the match for this.

tests/tcp-server/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"net"
+)
+
+func main() {
```
soneillf5, line 10: @vepatel @ciarams87 @lucacome @pleshakov There's a question for the team here. I've added a little tcp server to respond with it's IP. I'll be re-using it for the other TransportServer tests, for checking the lb method and such. 

What's the best way to engineer this?
Do we do a once off push of the docker image when we make a change to this server?
Do we add a pipeline/workflow?
Do we put it in a different repo?

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,114 @@
...
+                        "-global-configuration=nginx-ingress/nginx-configuration",
+                        "-enable-leader-election=false"
+                    ]
+            },
+            {"example": "transport-server-tcp-load-balance", "app_type": "simple"},
+        )
+    ],
+    indirect=True,
+)
+class TestTransportServerTcpLoadBalance:
```
soneillf5, line 40: Are there any other tests people would like to see?

tests/tcp-server/main.go
vepatel: +1 for once-off push in a public repo and then pull it as usual through manifest

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,114 @@
...
+        port = node_port_service.spec.ports[0].node_port
+        host = transport_server_setup.public_endpoint.public_ip
+
+        print(f"sending tcp requests to: {host}:{port}")
+
+        endpoints = {}
+        for i in range(20):
+            client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            client.connect((host, port))
+            response = client.recv(4096)
```
vepatel, line 100: is there something that can be asserted in response?

soneillf5: We could read the endpoints from the config as a set and make sure the two sets match?

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,114 @@
...
+    @pytest.mark.sean
+    def test_number_of_replicas(
+        self, kube_apis, crd_ingress_controller, transport_server_setup, ingress_controller_prerequisites
+    ):
+        """
+        The load balancing of TCP should result in 4 servers to match the 4 replicas of a service.
+        """
+        original = scale_deployment(kube_apis.apps_v1_api, "tcp-service", transport_server_setup.namespace, 4)
+
+        wait_before_test()
```
pleshakov, line 58: because we asked k8s to bring up new pods, better to wait until the new pods are running, rather than simply sleeping. I worry that 3 seconds  (the default in wait_before_test) will not be enough all the time for k8s to bring up the new pods.  note that we have `wait_until_all_pods_are_ready` in `resource_utils` to check if the pods are running

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,114 @@
...
+            if endpoint not in endpoints:
+                endpoints[endpoint] = 1
+            else:
+                endpoints[endpoint] = endpoints[endpoint] + 1
+
+        assert len(endpoints) is 3
+
+        delete_service(kube_apis.v1, node_port_name, transport_server_setup.namespace)
+
+        self.restore_transport_server(kube_apis, transport_server_setup)
```
pleshakov, line 112: the TS is not changed in the tests. Do we really need to call `restore_transport_server`? Do we really need that function at all?

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,114 @@
...
+        wait_before_test(3)
+
+        node_port_service = kube_apis.v1.read_namespaced_service(node_port_name, transport_server_setup.namespace)
+        port = node_port_service.spec.ports[0].node_port
+        host = transport_server_setup.public_endpoint.public_ip
+
+        print(f"sending tcp requests to: {host}:{port}")
+
+        endpoints = {}
+        for i in range(20):
```
pleshakov, line 97: in this test, we're not really making connections through NGINX in this test 🤔  perhaps I'm missing something?

the NodePort port needs to be configured in the Ingress Controller service. After configuring it, we can send a connection to the `public_endpoint.public_ip` and the nodeport. This way the connection will go through NGINX.

Currently, the connection goes like this:
```
client -> node -> kube-proxy -> backend
```
meaning we're not testing NGINX

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,114 @@
...
+            client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            client.connect((host, port))
+            response = client.recv(4096)
+            endpoint = response.decode()
```
pleshakov, line 101: do we need to close the connection after recv?

tests/suite/fixtures.py
```diff
@@ -767,9 +773,9 @@ def transport_server_setup(
     global_config_file = f"{TEST_DATA}/{request.param['example']}/standard/global-configuration.yaml"
     gc_resource = create_gc_from_yaml(kube_apis.custom_objects, global_config_file, "nginx-ingress")
 
-    # deploy dns
-    dns_file = f"{TEST_DATA}/{request.param['example']}/standard/dns.yaml"
-    create_items_from_yaml(kube_apis, dns_file, test_namespace)
+    # deploy service_file
+    service_file = f"{TEST_DATA}/{request.param['example']}/standard/service.yaml"
```
pleshakov, line 777: `service.yaml` includes a deployment and a service, not just a service. 
because service in Kubernetes has a meaning - it is one of the k8s resources - perhaps we should use a different name here? For example, tcp-app.yaml ? and app_file for the variable.   just to avoid any potential confusion.

tests/tcp-server/main.go
```diff
@@ -0,0 +1,56 @@
...
+	log.Printf("write data to connection: %v\n", response)
+	_, err := conn.Write([]byte(response))
+	if err != nil {
+		log.Printf("error writing to connection: %v", err)
+		return
+	}
+}
+
+// GetOutboundIP https://stackoverflow.com/questions/23558425/how-do-i-get-the-local-ip-address-in-go
+func GetOutboundIP() net.IP {
```
pleshakov, line 48: we can use the conn.LocalAddr instead of GetOutboundIP :
```
	_, err := conn.Write([]byte(conn.LocalAddr().String()))
```

The local address will include the address of which was used to accept the connection. Which is what we need - if we connect to a pod to "1.1.1.1:3333", the LocaLAddr will be `1.1.1.1:3333`

For example:
```
$ go run main.go
```
```
$ nc 127.0.0.1 3333
127.0.0.1:3333%
$ nc 10.0.0.24 3333
10.0.0.24:3333%
```

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: Great suggestion, thanks!

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: That's frustrating as this was the intent of the test!

I haven't been able to get it to work. Do you know how it should be setup?

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: removed, well spotted

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: yes! good spot, I've updated it with a close

tests/suite/fixtures.py
soneillf5: I was hoping to pick a name that's common across status and load balancing test, so I've updated to `service_deployment.yaml` if that's okay

tests/tcp-server/main.go
soneillf5: updated

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: I've tried this but it doesn't work I'm afraid.

tests/tcp-server/main.go
soneillf5: I've added a small readme. I'm not sure where to push the image so it can be used
tests/tcp-server/README.md

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: It's not clear what you mean when you say: `the NodePort port needs to be configured in the Ingress Controller service.`
Do you mean the Nginx-ingress deployment `deployments/deployment/nginx-ingress.yaml` ?
I think that's used by the tests `tests.suite.resources_utils.create_ingress_controller`

Or do you mean the `kind: service` with the name `nginx-ingress` here: `deployments/service/nodeport.yaml` ?

soneillf5: Okay I think I have it:
```
get services -A
NAMESPACE                      NAME            TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)                                                                   AGE
nginx-ingress                  nginx-ingress   NodePort    10.106.78.48    <none>        80:31426/TCP,443:31001/TCP,8080:32493/TCP,9113:31972/TCP,3333:32249/TCP   12s
test-namespace-1619176893926   tcp-service     ClusterIP   10.107.12.201   <none>        3333/TCP
```

```
k get endpoints -A
NAMESPACE                      NAME                       ENDPOINTS                                                     AGE
nginx-ingress                  nginx-ingress              172.17.0.3:9113,172.17.0.3:3333,172.17.0.3:8080 + 2 more...   73s
test-namespace-1619176893926   tcp-service                172.17.0.4:3333,172.17.0.5:3333,172.17.0.6:3333               68s
```

```
k describe service nginx-ingress -n nginx-ingress
Name:                     nginx-ingress
Namespace:                nginx-ingress
...
Port:                     tcp-server  3333/TCP
TargetPort:               3333/TCP
NodePort:                 tcp-server  32249/TCP
Endpoints:                172.17.0.3:3333
```
I think this means endpoint is pointing to the ingress-controller endpoint `172.17.0.3`.
If we contact the public endpoint, we get the tcp-service IPs
```
nc 192.168.64.2 32249
172.17.0.4:3333
nc 192.168.64.2 32249
172.17.0.5:3333%
```

So I think it's going:
```
public endpoint (192.168.64.2 32249) -> ingress controller (172.17.0.3:3333) -> tcp-server (172.17.0.4:3333)
```

tests/tcp-server/main.go
vepatel: docker-hub public repo?

tests/tcp-server/README.md
```diff
@@ -0,0 +1,29 @@
+##TCP Server
```
pleshakov, line 1: not rendered correctly
```suggestion
## TCP Server
```

tests/suite/resources_utils.py
```diff
@@ -750,6 +782,21 @@ def get_ingress_nginx_template_conf(v1: CoreV1Api, ingress_namespace, ingress_na
...
+
+    :param v1: CoreV1Api
+    :param resource_namespace:
+    :param resource_name:
+    :param ingress_name:
+    :param ingress_namespace:
+    :return: str
+    """
+    file_path = f"/etc/nginx/stream-conf.d/ts_{resource_namespace}_{resource_name}.conf"
+    return get_file_contents(v1, file_path, ingress_name, ingress_namespace)
```
pleshakov, line 797: just noticed, ingress_name and ingress_namespace are actually IC pod_name and pod_namespace. Could you rename them?
```
def get_file_contents(v1: CoreV1Api, file_path, pod_name, pod_namespace) -> str:
```

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,88 @@
+import pytest
```
pleshakov, line 1: there is a number of unused imports 

@lucacome @soneillf5 @vepatel @ciarams87 
perhaps it make sense to enable linting for python code in the pipeline?

pleshakov (approved):
worked well in kind!

```
suite/test_transport_server_tcp_load_balance.py::TestTransportServerTcpLoadBalance::test_number_of_replicas[crd_ingress_controller0-transport_server_setup0] PASSED [ 50%]
suite/test_transport_server_tcp_load_balance.py::TestTransportServerTcpLoadBalance::test_tcp_request_load_balanced[crd_ingress_controller0-transport_server_setup0] PASSED [100%]
```

IC logs:
```
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.003 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.004 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.044 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.022 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.020 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.001 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.022 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.032 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.013 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.004 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.007 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.002 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.002 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.001 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.004 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.008 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.002 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.004 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.003 "-"
10.244.0.1 [23/Apr/2021:16:26:14 +0000] TCP 200 16 0 0.011 "-"
```

tests/tcp-server/README.md
soneillf5: fixed.

tests/suite/resources_utils.py
soneillf5: fixed

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: removed unused imports

tests/tcp-server/main.go
soneillf5: done
`seanoneillf5/tcp-server:1.0`

tests/suite/test_transport_server_tcp_load_balance.py
vepatel: @pleshakov @lucacome @soneillf5 recomment black and flake8 for python linting:

`~/nginx/kubernetes-ingress/tests ingress-mtls-tests*
tests ❯ flake8 suite/test_ingress_mtls.py
suite/test_ingress_mtls.py:39: [F541] f-string is missing placeholders
suite/test_ingress_mtls.py:40: [F541] f-string is missing placeholders
suite/test_ingress_mtls.py:53: [F541] f-string is missing placeholders
suite/test_ingress_mtls.py:56: [F541] f-string is missing placeholders
suite/test_ingress_mtls.py:59: [F541] f-string is missing placeholders
suite/test_ingress_mtls.py:146: [F541] f-string is missing placeholders
suite/test_ingress_mtls.py:211: [F541] f-string is missing placeholders

~/nginx/kubernetes-ingress/tests ingress-mtls-tests*
tests ❯ black suite/test_ingress_mtls.py -l 100      
All done! ✨ 🍰 ✨
1 file left unchanged.
`

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: I've added endpoint checks.

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,175 @@
...
+        (
+            {
+                "type": "complete",
+                "extra_args":
+                    [
+                        "-global-configuration=nginx-ingress/nginx-configuration",
+                        "-enable-leader-election=false"
+                    ]
+            },
+            {"example": "transport-server-tcp-load-balance", "app_type": "simple"},
```
vepatel, line 28: is` "app_type": "simple"` needed here?

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+
+        pattern = 'server .*;'
+        servers = re.findall(pattern, result_conf)
+        for key in endpoints.keys():
+            found = False
+            for server in servers:
+                if key in server:
+                    found = True
+            assert found
```
vepatel, line 113: no way to assert a string?

vepatel: can we add a test where two TS have same listener port and the one deployed later causes port collision?

tests/tcp-server/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.16-alpine as builder
```
vepatel, line 1: can we remove docker and go files now that image is available in docker? or move them inside data/? @lucacome @pleshakov what would be a better approach?

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: removed

tests/tcp-server/Dockerfile
soneillf5: I'd prefer to leave it. It doesn't interact with the rest of the code or require maintenance, but it will really help the next person that tries to change the tcp-server

pleshakov: agree with Sean, better to leave it. If we have too many apps like that, we can put them in one folder like `test-apps` or something like that

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: I've added an additional test to cover port collision and resolution.

vepatel (approved):

PR #1544, opened 2021-04-19
Merge release branch to master

lucacome:

Issue #1310, opened 2021-01-17
Publish container images for ARM


c-x-berger:
**Is your feature request related to a problem? Please describe.**
My cluster is ARM-based (read: made of Raspberry Pis), and the [Docker image](https://hub.docker.com/r/nginx/nginx-ingress) is apparently only published for amd64.

As a result, I can't really use this controller. (Yes, I could build the images myself, but I personally don't really have the means for that.)

**Describe the solution you'd like**
ARM images (at the very least, 64-bit ARMv8) published under the various `nginx/nginx-ingress` tags

**Describe alternatives you've considered**
... I could use a different controller, I guess?

groundhog2k:
@c-x-berger : I made the same suggestion with #974  :-(
At the moment the best solution is to use https://github.com/kubernetes/ingress-nginx. It provides all platforms with a multi-arch. docker image and the community is very active to keep the controller and the helm chart (https://github.com/kubernetes/ingress-nginx/tree/master/charts/ingress-nginx) up-to-date. For me this is working great!

lucacome:
Hi @c-x-berger 

I just wanted to let you know that we now build debian and alpine images for arm and arm64.

Issue #974, opened 2020-05-30
Support for other platforms - ARM64v8


groundhog2k:
Please support other platforms like arm64v8 (maybe old arm32v.. too). Nginx web server is supported and there is an image arm64v8/nginx, but not for the nginx ingress-controller.

lucacome:
Hi @groundhog2k  

I just wanted to let you know that we now build debian and alpine images for arm and arm64.

Issue #944, opened 2020-05-11
[feature request] Support ppc64le


runlevel5:
I am using Linux (PowerPC64 LE) bare metal to host kubernetes cluster. The `docker.io/nginx/nginx-ingress` image does come in ppc64le so I am unable to use the ingress controller.

**Describe the solution you'd like**
1st-class support for ppc64le

**Describe alternatives you've considered**
Custom build the docker image locally as workaround for the time-being. My images are uploaded to docker hub if anyone is interested https://hub.docker.com/repository/docker/powerpc64el/nginx-ingress

**Additional context**
- None

runlevel5:
It is not stale

ogarrett:
NGINX (the team) don't actively develop on ppc64le or ship new packages for that platform.

We don't plan to ship official docker images for that paltform

lucacome:
Hi @runlevel5 

just wanted to let you know that we now build debian and apline images for ppc64le.

runlevel5:
@lucacome :+1: Great news. Thanks much for letting me know

PR #1549, opened 2021-04-21
Add GoReleaser

lucacome:
~This will get the wrong `tag` until #1544 is merged.~

Removed `edge` from the version name in the Makefile, it will now be something like `v1.11.1-SNAPSHOT-00f0053` for local builds (same that GoReleaser uses for `edge`), so it will remove the need to update the version in the Makefile.
(When we make a stable release, GoReleaser will use `v1.11.1`)

GoReleaser builds binaries for:
- amd64
- arm
- arm64
- ppc64le
- s390x

and we push Docker images for the same architectures.

Addresses https://github.com/nginxinc/kubernetes-ingress/issues/944, https://github.com/nginxinc/kubernetes-ingress/issues/974 and https://github.com/nginxinc/kubernetes-ingress/issues/1310

### DIFF
--- a/.github/workflows/edge.yml
+++ b/.github/workflows/edge.yml
@@ -36,7 +36,6 @@ env:
   HELM_CHART_VERSION: 0.0.0-edge
   GIT_NAME: NGINX Kubernetes Team
   GIT_MAIL: kubernetes@nginx.com
-  VERSION: edge
 
 jobs:
 
@@ -46,6 +45,8 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Output Variables
         id: commit
         run: |
@@ -56,24 +57,26 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: ${{ env.GO_VERSION }}
+      - name: Determine GOPATH
+        run: echo "GOPATH=$(go env GOPATH)" >> $GITHUB_ENV
       - name: Check if CRDs changed
         run: |
           make update-crds && git diff --name-only --exit-code deployments/common/crds*
       - name: Check if Codegen changed
         run: |
           make update-codegen && git diff --name-only --exit-code pkg/**/zz_generated.deepcopy.go
-      - name: Build Binary
-        run: >
-          make build
-          VERSION=${{ env.VERSION }}
-          IC_VERSION=${{ env.VERSION }}-${{ steps.commit.outputs.sha }}
-          GIT_COMMIT=${{ steps.commit.outputs.sha }}
+      - name: Build binaries
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          version: latest
+          args: --rm-dist --debug --skip-publish --snapshot
         env:
-          GOFLAGS: '-gcflags=-trimpath=${{ github.workspace }} -asmflags=-trimpath=${{ github.workspace }}'
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GOPATH: ${{ env.GOPATH }}
       - name: Store Artifacts in Cache
         uses: actions/cache@v2.1.5
         with:
-          path: ${{ github.workspace }}/nginx-ingress
+          path: ${{ github.workspace }}/dist
           key: nginx-ingress-${{ github.run_id }}-${{ github.run_number }}
 
   unit-tests:
@@ -171,7 +174,7 @@ jobs:
       - name: Fetch Cached Artifacts
         uses: actions/cache@v2.1.5
         with:
-          path: ${{ github.workspace }}/nginx-ingress
+          path: ${{ github.workspace }}/dist
           key: nginx-ingress-${{ github.run_id }}-${{ github.run_number }}
       - name: Docker Buildx
         uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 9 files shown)


.github/workflows/edge.yml
```diff
@@ -46,6 +45,8 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
soneillf5, line 49: why do we do this? does it break otherwise ?

.gitignore
```diff
@@ -53,3 +53,5 @@ hack/controller-gen-*
 docs-web/.netlify/state.json
 site/
 venv/
+
+dist/
```
soneillf5, line 57: cool!

soneillf5 (approved):
Really cool! It's so different from the stuff I learnt when starting go about targeting all the different architectures.

.github/workflows/edge.yml
lucacome: yeah, the docs for GoReleaser say:

> Note the fetch-depth: 0 option on the Checkout workflow step. It is required for GoReleaser to work properly. Without that, GoReleaser might fail or behave incorrectly.

this is the description of the parameter:

>Number of commits to fetch. 0 indicates all history for all branches and tags.

.goreleaser.yml
```diff
@@ -0,0 +1,31 @@
+env:
+  - GO111MODULE=on
+before:
+  hooks:
+    - go mod tidy
```
pleshakov, line 5: are `tidy` and `verify` commands really necessary?
if they are expected to change the repo contents, would it be better if our pipeline had a check in a separate stage - to check that the repo files do not change after tidy and verify?

build/Dockerfile
```diff
@@ -272,7 +272,7 @@ RUN mkdir -p /var/lib/nginx /etc/nginx/secrets /etc/nginx/stream-conf.d \
 	&& chown -R nginx:0 /etc/nginx /var/cache/nginx /var/lib/nginx \
 	&& rm -f /etc/nginx/conf.d/* /etc/apt/apt.conf.d/90nginx /etc/apt/sources.list.d/nginx-plus.list
 
-COPY internal/configs/version1/nginx$PLUS.ingress.tmpl \
+COPY --chown=nginx:0 internal/configs/version1/nginx$PLUS.ingress.tmpl \
```
pleshakov, line 275: why is this change necessary? 
for the  template files, it is ok if root owns it, because the IC only need to read it and only to read it on the start.

build/Dockerfile
```diff
@@ -284,28 +284,36 @@ COPY internal/configs/version1/nginx$PLUS.ingress.tmpl \
...
 COPY --chown=nginx:0 nginx-ingress /
 
-USER nginx
+
+############################################# Create image with nginx-ingress built by GoReleaser #############################################
+FROM common AS goreleaser
+ARG TARGETARCH
+ARG TARGETVARIANT
+
+COPY --chown=nginx:0 dist/kubernetes-ingress_linux_$TARGETARCH${TARGETVARIANT:+_7}/nginx-ingress /
```
pleshakov, line 319: similar question, why does it need to be owned by nginx user?

cmd/nginx-ingress/main.go
```diff
@@ -195,7 +196,7 @@ func main() {
 	}
 
 	if *versionFlag {
-		fmt.Printf("Version=%v GitCommit=%v\n", version, gitCommit)
+		fmt.Printf("Version=%v GitCommit=%v Date=%v\n", version, commit, date)
```
pleshakov, line 199: do we need to update the doc for the version flag?

pleshakov, line 199: is it clear from the context that it is a build date? Perhaps `BuildDate` ?

.goreleaser.yml
lucacome: they're not expected to change the repository's content, it's just a check to make sure everything is fine.

build/Dockerfile
lucacome: We had a problem not long ago with permissions and having nginx be the owner of the files will fix that. Since we set nginx as the user and run the binary as nginx user, it makes sense that nginx owns the files.

build/Dockerfile
lucacome: it's the same for the binaries in the other layers.

cmd/nginx-ingress/main.go
lucacome: I've updated the flag description, is that it?

lucacome: I think it's clear from the context, like for the git commit, but I'm fine either way.

.goreleaser.yml
pleshakov: `go mod tidy` exits with 0 status code even if it updates the go.mod. so it will silently fix the files.

build/Dockerfile
pleshakov: for the files that the IC or NGINX workers can change, it makes sense. But for the files that IC or NGINX only need to read, I think it makes sense to be owned by root and allow read/execute on those, to follow the principle of least privilege. For example, as NGINX binary:
```
ls -la /usr/sbin/nginx
-rwxr-xr-x 1 root root 1698544 Nov 26 13:00 /usr/sbin/nginx
```

cmd/nginx-ingress/main.go
pleshakov: looks good. thanks!
we have the flag docs here as well though https://docs.nginx.com/nginx-ingress-controller/configuration/global-configuration/command-line-arguments/

pleshakov (approved):

build/Dockerfile
lucacome: isn't it least privilege if the user is nginx instead of root?

PR #1553, opened 2021-04-26
Feature transportserver udp tests

soneillf5:
### Proposed changes
This change adds tests for TransportServer UDP load balancing functionality.
It uses a test server to respond to UDP requests, allowing us to tests the number of replicas and endpoints that the TransportServer works with.

### Checklist


tests/data/transport-server-udp-load-balance/missing-service-transport-server.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: transport-server
+spec:
+  listener:
+    name: udp-server
+    protocol: UDP
+  upstreams:
+    - name: udp-app
+      service: missing-service
+      port: 3333
```
vepatel, line 12: should this be 3334?

tests/udp-server/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.16-alpine as builder
```
vepatel, line 1: @soneillf5 minor one but  can we one readme and dockerfile for both tcp and udp server and ask the user to pass port as var in docker-compose?

tests/suite/test_transport_server_udp_load_balance.py
```diff
@@ -0,0 +1,186 @@
...
+            client.sendto("ping".encode('utf-8'), (host, port))
+            try:
+                client.recvfrom(4096)
+                # it should timeout
+                assert False
+            except socket.timeout:
+                print("successfully timed out")
+            client.close()
+
+        self.restore_ts(kube_apis, transport_server_setup)
```
vepatel, line 186: we should add a test for listener port collision similar to tcp one.

tests/suite/resources_utils.py
```diff
@@ -352,11 +352,12 @@ def get_service_node_ports(v1: CoreV1Api, name, namespace) -> (int, int, int, in
     :return: (plain_port, ssl_port, api_port, exporter_port)
     """
     resp = v1.read_namespaced_service(name, namespace)
-    assert len(resp.spec.ports) == 5, "An unexpected amount of ports in a service. Check the configuration"
+    assert len(resp.spec.ports) == 6, "An unexpected amount of ports in a service. Check the configuration"
```
vepatel, line 355: although it was a pre-existing one, but can we change assert to if-else as assert can cause an exit if false

tests/udp-server/main.go
```diff
@@ -0,0 +1,50 @@
...
+		n, err = listener.WriteTo([]byte(address), addr)
+		if err != nil {
+			log.Panicln(err)
+		}
+		fmt.Printf("packet-written: bytes=%d to=%s\n", n, addr.String())
+	}
+}
+
+func GetOutboundIP() net.IP {
+	conn, err := net.Dial("udp", "8.8.8.8:80")
```
pleshakov, line 43: I wonder if this can be source of any failures, considering also that it is called on every request handling.

If it can be, there is another way to get pod IP information - propagate it using a cli argument or env variable. https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables  and https://github.com/nginxinc/kubernetes-ingress/blob/master/deployments/deployment/nginx-ingress.yaml#L57

tests/data/transport-server-udp-load-balance/missing-service-transport-server.yaml
soneillf5: fixed

tests/suite/resources_utils.py
soneillf5: fixed

tests/suite/test_transport_server_udp_load_balance.py
soneillf5: fixed

tests/udp-server/Dockerfile
soneillf5: fixed

tests/suite/test_transport_server_udp_load_balance.py
```diff
@@ -0,0 +1,186 @@
...
+                client.recvfrom(4096)
+                # it should timeout, it's an error if it doesn't.
+                assert False
+            except socket.timeout:
+                print("successfully timed out")
+            client.close()
+
+        self.restore_ts(kube_apis, transport_server_setup)
+
+    def test_udp_request_load_balanced_missing_service(
```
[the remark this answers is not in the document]
soneillf5: I tried it out, it's much nicer. Great feedback thanks!

tests/udp-server/main.go
pleshakov: > I've reduced it to one call at the start of the service, to reduce the likelihood it fails.
I don't think the pod fields as values solution will work. We need the IP address which is not defined in the spec/metadata as far as I know.


I suggest trying that busybox example from https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables - you would see that the env variable `MY_POD_IP` will have the pod IP. That env variable can be used in application arg like  `... -ip=$(MY_POD_IP)`

> I'd rather try this for now and if we get failures in the pipeline, then we can invest more time ?

this makes sense

pleshakov (approved):

vepatel (approved):

tests/udp-server/main.go
soneillf5: I dug a little deeper, turns out the pod ip is available. I've updated the PR with POD_IP as an env var.

pleshakov (approved):

PR #1556, opened 2021-04-26
Updated regex for create tag release workflow

soneillf5:
### Proposed changes
Minor regex tweak

### Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,9 +10,8 @@ on:
       - 'examples/**'
       - 'examples-of-custom-resources/**'
       - '**.md'
-  create:
     tags:
-      - 'v[0-9].[0-9].[0-9]+*'
+      - 'v[0-9]+.[0-9]+.[0-9]+.*'
 
 defaults:
   run:


vepatel (approved):

.github/workflows/release.yml
```diff
@@ -12,7 +12,7 @@ on:
       - '**.md'
   create:
     tags:
-      - 'v[0-9].[0-9].[0-9]+*'
+      - 'v[0-9].[0-9].[0-9]+.*'
```
ciarams87, line 15: This isn't correct either - this translates to e.g. 'v1.1.1.*'.
We just need to remove the 'create' and just include tags in with the 'push' block (see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-branches-and-tags) 
```
on:
  push:
    branches:
      - release-1.11
    paths-ignore:
      - 'docs/**'
      - 'docs-web/**'
      - 'examples/**'
      - 'examples-of-custom-resources/**'
      - '**.md'
    tags:
      - 'v[0-9].[0-9].[0-9]+*'

ciarams87: Actually, this won't work now for our next release - if I'm regexing correctly, each `[0-9]` will match precisely one digit so it should probably be 'v[0-9].[0-9][0-9].[0-9]+*' or something, but I'll verify that.

ciarams87 (approved):

PR #1574, opened 2021-05-05
Add Docker image for Alpine with NGINX Plus

lucacome:

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -13,8 +13,21 @@ RUN apt-get update \
 # docker.io/library/nginx is a temporary workaround for Dependabot to see this as different from the one used in Debian
 FROM docker.io/library/nginx:1.19.10-alpine AS alpine
 
-RUN apk add --no-cache libcap \
-	&& rm -rf /var/cache/apk/*
+RUN apk add --no-cache libcap
+
+
+############################################# Base image for Alpine with NGINX Plus #############################################
+FROM alpine:3.13 as alpine-plus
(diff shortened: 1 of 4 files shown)


lucacome:
@pleshakov added the docs

build/Dockerfile
```diff
@@ -12,8 +12,21 @@ RUN apt-get update \
...
 FROM nginx:1.19.9-alpine AS alpine
 
-RUN apk add --no-cache libcap \
-	&& rm -rf /var/cache/apk/*
+RUN apk add --no-cache libcap
+
+
+############################################# Base image for Alpine with NGINX Plus #############################################
+FROM alpine:3.13 as alpine-plus
+ENV NGINX_PLUS_VERSION 24-r1
```
pleshakov, line 20: currently, the other Plus images all have R23. Could we have one version of N+ across all images?

lucacome: 3.13 is only supported in R24.

pleshakov: would it make sense to upgrade other images to R24 first then?

lucacome: Created #1604 to update to R24

pleshakov: @lucacome will we able to not use a specific patch version of N+ for Apline similarly to Debian in #1604?

lucacome: yep I'll update this PR once the other one is done

pleshakov (approved):

PR #1577, opened 2021-05-06
Add automatic response for issues

lucacome:
- Add automatic response pointing to the docs
- Additional text is added if there's an annotation from community IC in the text

### DIFF
--- a/.github/workflows/issues.yaml
+++ b/.github/workflows/issues.yaml
@@ -1,0 +1,26 @@
+on:
+  issues:
+    types: [opened]
+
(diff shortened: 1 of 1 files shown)


ciarams87 (approved):

.github/workflows/issues.yaml
```diff
@@ -0,0 +1,26 @@
+on:
+  issues:
+    types: [opened]
+
+jobs:
```
soneillf5, line 5: Awesome addition!

PR #1580, opened 2021-05-07
Add container scanning for OSS images

ciarams87:
### Proposed changes
Add trivy container scanning for OSS images. Scanning of the Plus and App Protect images will be done separately in GitLab.

### Checklist

### DIFF
--- a/.github/workflows/edge.yml
+++ b/.github/workflows/edge.yml
@@ -92,6 +92,64 @@ jobs:
       - name: Run Tests
         run: go test ./...
 
+  container-scan:
+    name: Container Scanning
+    runs-on: ubuntu-20.04
+    needs: binary
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Fetch Cached Artifacts
+        uses: actions/cache@v2.1.5
+        with:
+          path: ${{ github.workspace }}/dist
+          key: nginx-ingress-${{ github.run_id }}-${{ github.run_number }}
+      - name: Docker Buildx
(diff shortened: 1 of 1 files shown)


.github/workflows/edge.yml
```diff
@@ -92,6 +92,64 @@ jobs:
...
+          build-args: |
+            BUILD_OS=${{ matrix.image }}
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: '${{ matrix.image }}:${{ github.sha }}'
+          format: 'template'
+          template: '@/contrib/sarif.tpl'
+          output: 'trivy-results-${{ matrix.image }}.sarif'
+          severity: 'CRITICAL,HIGH'
```
lucacome, line 137: should we also include `LOW` and `MEDIUM`? I think if we enable `ignore-unfixed: true` it makes sense to have them all.

lucacome, line 137: we should add `ignore-unfixed: true` since if they're not fixed upstream there's not much we can do.

lucacome (changes requested):
Do we need to build the images again? Can't it just be a step in one of the other builds?

ciarams87:
> Do we need to build the images again? Can't it just be a step in one of the other builds?

It could just be a step in another build, but having it as a separate stage runs these jobs concurrently with the other helm and smoke tests, so it doesn't take any longer, and I think it's easier to follow this way.

pleshakov (approved):

PR #1601, opened 2021-05-17
Add gitlab pipeline and remove plus building from github

ciarams87:
### Proposed changes
Remove plus building from GitHub; remove some of the CI logic from the workflows to the test folder; add a gitlab-ci.yml file for the Gitlab integration,

### Checklist

### DIFF
--- a/.github/workflows/edge.yml
+++ b/.github/workflows/edge.yml
@@ -98,6 +98,9 @@ jobs:
     name: Container Scanning
     runs-on: ubuntu-20.04
     needs: binary
+    strategy:
+      matrix:
+        image: [debian, alpine, opentracing, ubi]
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
@@ -153,92 +156,24 @@ jobs:
           name: 'trivy-results-${{ matrix.image }}.sarif'
           path: 'trivy-results-${{ matrix.image }}.sarif'
         if: always()
-    strategy:
-      matrix:
-        image: [debian, alpine, opentracing, ubi]
 
   smoke-tests:
     name: Smoke Tests
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-20.04
     needs: [binary, unit-tests]
-    if:
-      github.event.pull_request.head.repo.full_name == 'nginxinc/kubernetes-ingress' ||
-      github.event_name == 'push'
     strategy:
       matrix:
         include:
-          - os: ubuntu-20.04
-            image: debian
-            tag: ${{ github.sha }}
+          - image: debian
             marker: 'ingresses'
-            type: oss
-            ic-type: nginx-ingress
-          - os: ubuntu-20.04
-            image: alpine
-            tag: ${{ github.sha }}
+          - image: alpine
             marker: 'vsr'
-            type: oss
-            ic-type: nginx-ingress
-          - os: ubuntu-20.04
-            image: opentracing
-            tag: ${{ github.sha }}
+          - image: opentracing
             marker: 'vs'
-            type: oss
-            ic-type: nginx-ingress
-          - os: ubuntu-20.04
-            image: ubi
-            tag: ${{ github.sha }}
+          - image: ubi
             marker: 'ts'
-            type: oss
-            ic-type: nginx-ingress
-          - os: ubuntu-20.04
-            image: debian
-            tag: ${{ github.sha }}
+          - image: debian
             marker: 'policies'
-            type: oss
-            ic-type: nginx-ingress
-          - os: ubuntu-20.04
-            image: ubi-plus
-            tag: ${{ github.sha }}
-            marker: 'ingresses'
-            type: plus
-            ic-type: nginx-plus-ingress
-          - os: ubuntu-20.04
-            image: debian-plus
-            tag: ${{ github.sha }}
-            marker: 'vsr'
-            type: plus
-            ic-type: nginx-plus-ingress
-          - os: ubuntu-20.04
-            image: debian-plus
-            tag: ${{ github.sha }}
-            marker: 'vs'
-            type: plus
-            ic-type: nginx-plus-ingress
-          - os: ubuntu-20.04
-            image: opentracing-plus
-            tag: ${{ github.sha }}
-            marker: 'ts'
-            type: plus
-            ic-type: nginx-plus-ingress
-          - os: ubuntu-20.04
-            image: debian-plus
-            tag: ${{ github.sha }}
-            marker: 'policies'
-            type: plus
-            ic-type: nginx-plus-ingress
-          - os: ubuntu-20.04
-            image: debian-plus
-            tag: ${{ github.sha }}
-            marker: 'ts'
-            type: plus
-            ic-type: nginx-plus-ingress
-          - os: ubuntu-20.04
-            image: debian-plus-nap
-            tag: ${{ github.sha }}-ap
-            marker: 'appprotect'
-            type: plus-ap
-            ic-type: nginx-plus-ingress
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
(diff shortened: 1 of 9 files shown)


.github/workflows/edge.yml
```diff
@@ -153,92 +284,25 @@ jobs:
...
-      matrix:
-        image: [debian, alpine, opentracing, ubi]
 
   smoke-tests:
     name: Smoke Tests
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-20.04
     needs: [binary, unit-tests]
-    if:
-      github.event.pull_request.head.repo.full_name == 'nginxinc/kubernetes-ingress' ||
```
lucacome, line 165: Wasn't this here to prevent builds from forks? We don't need it anymore?

lucacome (commented):
I'm not sure about having edge and nightly in the same file, I think it makes it harder to read. I don't see less duplicated code as you mentioned? It's only the `Build binary` step?

tests/docker/Dockerfile
```diff
@@ -1,9 +1,14 @@
 FROM python:3.9-slim
 
+ARG GCLOUD_VERSION=338.0.0
+
 RUN apt-get update && apt-get install -y curl \
 	&& curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl \
 	&& chmod +x ./kubectl \
-	&& mv ./kubectl /usr/local/bin
+	&& mv ./kubectl /usr/local/bin \
```
lucacome, line 8: Maybe we could have a different target for `google-cloud-sdk`? It's not needed when running tests locally or with kind in the pipeline, right?

lucacome:
I've noticed that the link to the pipeline doesn't bring you directly to the run if gitlab needs to reauthenticate, but works fine after that. We should report a bug 😄

.github/workflows/edge.yml
ciarams87: I left if for here to skip nightly, but I can change it to just pull requests and push events

ciarams87:
> I'm not sure about having edge and nightly in the same file, I think it makes it harder to read. I don't see less duplicated code as you mentioned? It's only the `Build binary` step?

@lucacome The build binary, helm tests, and notification are deduped. Also, I've noticed a lot of overhead as it is when we change something in the build process, we have too many places to change (i.e. for goreleaser, we needed to update edge, nightly, and release). Github doesn't allow for shared code in a 3rd place very easily (or at all for what I want it) so I figure at least sharing those stages is helpful. WDYT? I'll be doing another pass at refactoring the workflows after this also. (Side not, I did some analysis, and rebuilding the images is actually better than I had thought performance wise - all the stages using a shared cache so concurrent building of the same images in different stages actually speeds it up.

tests/docker/Dockerfile
ciarams87: I'll create a separate docker file for gitlab as we will probably further divergent dependencies (e.g. awscli etc)

.github/workflows/edge.yml
```diff
@@ -153,92 +203,25 @@ jobs:
...
-            image: ubi-plus
-            tag: ${{ github.sha }}
+          - image: debian
             marker: 'ingresses'
-            type: plus
-            ic-type: nginx-plus-ingress
-          - os: ubuntu-20.04
-            image: debian-plus
-            tag: ${{ github.sha }}
+          - image: alpine
```
vepatel, line 217: Should we run all tests on `debian` only? because of `runs-on: ubuntu-20.04`?

.github/workflows/release.yml
```diff
@@ -166,215 +131,110 @@ jobs:
...
-            marker: '-m appprotect'
-            type: plus-ap
-            k8s: 1.16.15
-          - os: ubuntu-20.04
-            image: nginx-plus-ingress
-            tag: ${{ needs.binary.outputs.version }}-ap
-            marker: '-m appprotect'
-            type: plus-ap
-            k8s: 1.15.12
+        k8s: [1.19.1, 1.18.8, 1.17.11, 1.16.15, 1.15.12]
```
vepatel, line 206: Can we remove 1.15 and include 1.20 similar to nightly https://github.com/nginxinc/kubernetes-ingress/pull/1601/files#diff-0d5658b415099a82c11c03a06ca4ec765b4003a1f4b2f3f1943980a882cf8aa6L278

.github/workflows/edge.yml
```diff
@@ -361,14 +301,15 @@ jobs:
           name: test-results-${{ steps.k8s.outputs.cluster }}
           path: ${{ github.workspace }}/tests/tests-${{ steps.k8s.outputs.cluster }}.html
         if: always()
-
-  build:
-    name: Build Docker Images
+  
+  nightly-smoke-tests:
```
vepatel, line 305: Is nightly and edge are both in single yaml now?

.github/workflows/edge.yml
```diff
@@ -23,6 +23,8 @@ on:
       - opened
       - reopened
       - synchronize
+  schedule:
+    - cron:  '0 4 * * *'
```
vepatel, line 27: shouldn't we run edge only on repo actions as usual

tests/docker/gitlab.Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM python:3.9-slim
```
vepatel, line 1: do we need to include gitlab specific files in github as sync would be one way (from gh to gitlab) assuming code will still be committed in gh but a hook on merge will sync it to gitlab triggering gitlab pipeline.

tests/ci-files/ci-kind-config.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Cluster
```
vepatel, line 1: is this going outside of pipleines?

tests/ci-files/helm-http-test.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
vepatel, line 1: same as above

.github/workflows/edge.yml
ciarams87: `runs-on: ubuntu-20.04` is the just the runner OS - we always used ubuntu-20.04 runner for all of our jobs. The image is which IC image version we are using in the KinD cluster.

.github/workflows/release.yml
ciarams87: Oh sure, no problem. I missed that, thanks 👍🏼

.github/workflows/edge.yml
ciarams87: Yes, but that is proving controversial so it might not stay that way (see Luca's comment above) 😄  I wanted to promote code re-use, and I dislike the overhead of having too many different workflow files that do the same thing - when we want to update the build process somehow, we have to do duplicate the effort and it increases the possibility of errors IMO. Which stages are run depend on the github event i.e. schedule/ push/ pull_request. I don't mind if I'm outvoted though if it seems harder to follow 😃

.github/workflows/edge.yml
ciarams87: This is because of putting nightly and edge in same file. The same pipeline steps will run in edge and nightly as they do now though, the only difference is how they are declared.

tests/ci-files/ci-kind-config.yaml
ciarams87: It's so we are using the same configuration in github and gitlab

tests/ci-files/helm-http-test.sh
ciarams87: Same answer as above 👍🏼

.github/workflows/edge.yml
lucacome: I've stumbled across [this](https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action) the other day. Maybe we can just create a simple action for the build part so we can re-use it. I haven't looked too much into it.

lucacome: Maybe we should just remove the plus builds for now (so they stop failing 😄) and merge this?

tests/docker/gitlab.Dockerfile
ciarams87: Discussed offline, synopsis is that the GitHub repo is still the Source Of Truth, I’m using the gitlab “includes” trick so that we don’t have to keep the gitlab pipeline code here.

.github/workflows/edge.yml
ciarams87: I returned edge and nightly to their separate ways 😄

.github/workflows/edge.yml
```diff
@@ -362,70 +254,15 @@ jobs:
           path: ${{ github.workspace }}/tests/tests-${{ steps.k8s.outputs.cluster }}.html
         if: always()
```
soneillf5, line 256: So much removed! Brilliant!

soneillf5 (approved):

vepatel (approved):

PR #1603, opened 2021-05-18
AppProtect gRPC automation tests

vepatel:
### Proposed changes
- Add automation tests for AppProtect gRPC support

### Checklist


tests/suite/test_app_protect_grpc.py
```diff
@@ -0,0 +1,168 @@
...
+    @pytest.mark.parametrize("backend_setup", [{"policy": "grpc-block-sayhello"}], indirect=True)
+    def test_responses_grpc_block(
+        self, kube_apis, crd_ingress_controller_with_ap, backend_setup, test_namespace
+    ):
+        """
+        Test grpc-block-hello AppProtect policy: Blocks /sayhello gRPC method only
+        Client sends request to /sayhello
+        """
+        syslog_pod = kube_apis.v1.list_namespaced_pod(test_namespace).items[-1].metadata.name
+        block_response = subprocess.run(["binaries/./grpc_client", "-address", f"{backend_setup.ingress_host}:{backend_setup.ssl_port}"], capture_output=True)
```
pleshakov, line 133: would it make sense to use https://github.com/grpc/grpc/blob/v1.37.1/examples/python/helloworld/greeter_client.py ? (providing we can make it work). this way there will be no need to add any binaries to the repo.

tests/suite/test_app_protect_grpc.py
```diff
@@ -0,0 +1,168 @@
...
+@pytest.mark.skip_for_nginx_oss
+@pytest.mark.appprotect
+@pytest.mark.smoke
+@pytest.mark.parametrize(
+    "crd_ingress_controller_with_ap",
+    [{"extra_args": [f"-enable-custom-resources", f"-enable-app-protect"]}],
+    indirect=["crd_ingress_controller_with_ap"],
+)
+class TestAppProtect:
+    @pytest.mark.parametrize("backend_setup", [{"policy": "grpc-block-sayhello"}], indirect=True)
```
pleshakov, line 124: I wonder if the approach of instantiating the `backend_setup` two times is optimal. Since we have multiple 40s sleeps there, I think calling `backend_setup` once instead of twice will save a reasonable amount of time. 

I wonder if it makes sense just to have a singe test that (1) tests the block sayhello policy (2) updates the policy to block saygoodbye (3) tests the block saygoodbye policy. Would it save time? Would the saving worth it?

tests/data/common/app/grpc/app.yaml
```diff
@@ -0,0 +1,32 @@
...
+    matchLabels:
+      app: greeter
+  template:
+    metadata:
+      labels:
+        app: greeter
+    spec:
+      containers:
+      - name: greeter
+        image: nginxkic/test-grpc-server:0.1
```
pleshakov, line 17: should we publish the sources similarly to how we published the sources for the TCP and UDP servers? so we don't loose it

tests/suite/test_app_protect_grpc.py
```diff
@@ -0,0 +1,168 @@
...
+            .addresses[0]
+            .ip
+        )
+    
+    print("------------------------- Deploy ingress -----------------------------")
+    ingress_host = {}
+    src_ing_yaml = f"{TEST_DATA}/appprotect/grpc/ingress.yaml"
+    create_ingress_with_ap_annotations(kube_apis, src_ing_yaml, test_namespace, policy, "True", "True", f"{syslog_ep}:514")
+    ingress_host = get_first_ingress_host_from_yaml(src_ing_yaml)
+    wait_before_test(40)
```
pleshakov, line 95: we should be able to rely on an event here - once the IC reloads N+ successfully, it will emit an event for that Ingress resource. This will allow us avoid waiting for 40 seconds. However, since this `wait` approach is used in the existing test code, perhaps it makes sense to address it in a separate PR.

tests/suite/test_app_protect_grpc.py
```diff
@@ -0,0 +1,168 @@
...
+    log_name = create_ap_logconf_from_yaml(kube_apis.custom_objects, src_log_yaml, test_namespace)
+
+    print(f"------------------------- Deploy appolicy: {policy} ---------------------------")
+    src_pol_yaml = f"{TEST_DATA}/appprotect/grpc/{policy}.yaml"
+    pol_name = create_ap_policy_from_yaml(kube_apis.custom_objects, src_pol_yaml, test_namespace)
+
+    print("------------------------- Deploy Syslog -----------------------------")
+    src_syslog_yaml = f"{TEST_DATA}/appprotect/syslog.yaml"
+    create_items_from_yaml(kube_apis, src_syslog_yaml, test_namespace)
+    wait_before_test(40)
```
pleshakov, line 82: not sure if this wait is necessary here. is it for syslog pods? if that is the case, can we wait until the pod is ready?

tests/suite/test_app_protect_grpc.py
```diff
@@ -0,0 +1,168 @@
...
+    delete_common_app,
+    replace_configmap_from_yaml,
+    create_ingress_with_ap_annotations,
+    wait_before_test,
+    get_file_contents,
+)
+from suite.yaml_utils import get_first_ingress_host_from_yaml
+
+
+ap_policies_under_test = ["grpc-block-saygoodbye", "grpc-bloc-sayhello"]
```
pleshakov, line 27: this variable isn't used. is it intentional?

pleshakov (commented):
Hi @vepatel 

please see my questions and comments

tests/data/common/app/grpc/app.yaml
vepatel: I'm not too sure about publishing sources for this one as the code isn't written by our team members unlike TCP/UDP server one which @soneillf5 wrote.

tests/suite/test_app_protect_grpc.py
vepatel: The approach you suggested is the correct one and there's already a ticket in the backlog to fix this across the suite.

tests/suite/test_app_protect_grpc.py
vepatel: this is redundant and will be removed, I added it in the beginning to parameterize the single test based on different policies but the code was turning out to be more convoluted and reduction in line of code and 40sec savings didn't seemed worth it.

tests/suite/test_app_protect_grpc.py
vepatel: @pleshakov I tried this approach of having one test before but imo It'd be better to have both of them independent of each other since we'd still need to wait after applying policy and ingress. There will be however only one wait needed for syslog pod but having separate syslog deployments makes the log more readable as its just a massive string not a structured log format like json. Also see: https://github.com/nginxinc/kubernetes-ingress/pull/1603#discussion_r635115886

tests/suite/test_app_protect_grpc.py
vepatel: great catch, I'll add waiting for pod! However there could be an issue at times when the pod is ready but endpoint hasn't been assigned which is why the wait is for.

tests/data/common/app/grpc/app.yaml
pleshakov: in the case of the backend, it's just a Dockerfile, right? we don't really modify the backend code?

tests/suite/test_app_protect_grpc.py
pleshakov: I see. maybe we can wait for that condition (we have 1 endpoint), instead of for the pod being ready?

tests/data/common/app/grpc/app.yaml
soneillf5: where does the test-grpc-server come from then ?

tests/suite/test_app_protect_grpc.py
```diff
@@ -0,0 +1,171 @@
...
+    delete_items_from_yaml,
+    delete_common_app,
+    replace_configmap_from_yaml,
+    create_ingress_with_ap_annotations,
+    wait_before_test,
+    get_file_contents,
+)
+from suite.yaml_utils import get_first_ingress_host_from_yaml
+
+my_hosts = Hosts()
```
soneillf5, line 25: Are these global? Generaly that's not a great idea, i.e.  `my_hosts` can't be declared in the same way in another file. I don't know Python though, so I could be wrong.

tests/suite/test_app_protect_grpc.py
soneillf5: I'm assuming the 40s that is catching us out is app protect ?

tests/suite/test_app_protect_grpc.py
vepatel: I'm checking https://pypi.org/project/grpc-requests/ which is based on the same, will update if it works!

soneillf5 (approved):

PR #1604, opened 2021-05-18
Update NGINX Plus to R24

lucacome:
- Updated to R24
- Simplified updating to a new version

For Debian there's a virtual package named with the R version, for example `nginx-plus-r24`, form there we can figure out the latest version of the actual NGINX package for that version and the modules.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,5 +1,6 @@
 # syntax=docker/dockerfile:1.2
 ARG BUILD_OS=debian
+ARG NGINX_PLUS_VERSION=r24
 
 ############################################# Base image for Debian #############################################
 FROM nginx:1.21.0 AS debian
@@ -33,8 +34,7 @@ COPY --chown=nginx:0 internal/configs/oidc/* /etc/nginx/oidc/
 ############################################# Base image for Debian with NGINX Plus #############################################
 FROM debian:buster-slim AS debian-plus
 ARG IC_VERSION
-ENV NGINX_PLUS_VERSION 23-1~buster
-ENV NGINX_NJS_VERSION 23+0.5.0-1~buster
+ARG NGINX_PLUS_VERSION
 
 RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode=0644 \
 	--mount=type=secret,id=nginx-repo.key,dst=/etc/ssl/nginx/nginx-repo.key,mode=0644 \
@@ -46,9 +46,10 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
 	&& gpg --no-default-keyring --keyring nginx_keyring.gpg --export > /etc/apt/trusted.gpg.d/nginx_signing.gpg \
 	&& wget -P /etc/apt/apt.conf.d https://cs.nginx.com/static/files/90pkgs-nginx \
 	&& echo "Acquire::https::pkgs.nginx.com::User-Agent  \"k8s-ic-$IC_VERSION-apt\";" >> /etc/apt/apt.conf.d/90pkgs-nginx \
-	&& printf "deb https://pkgs.nginx.com/plus/debian buster nginx-plus\n" > /etc/apt/sources.list.d/nginx-plus.list \
-	&& apt-get update && apt-get install --no-install-recommends --no-install-suggests -y \
-	nginx-plus=${NGINX_PLUS_VERSION} nginx-plus-module-njs=${NGINX_NJS_VERSION} \
+	&& printf "%s\n" "deb https://pkgs.nginx.com/plus/debian buster nginx-plus" > /etc/apt/sources.list.d/nginx-plus.list \
+	&& apt-get update \
+	&& apt-get install --no-install-recommends --no-install-suggests -y \
+	nginx-plus-${NGINX_PLUS_VERSION} nginx-plus-module-njs-${NGINX_PLUS_VERSION} \
 	&& apt-get purge --auto-remove -y apt-transport-https gnupg wget \
 	&& rm -rf /var/lib/apt/lists/*
 
@@ -58,13 +59,7 @@ COPY --chown=nginx:0 internal/configs/oidc/* /etc/nginx/oidc/
 ############################################# Base image for Debian with NGINX Plus and App Protect #############################################
 FROM debian-plus as debian-plus-nap
 ARG IC_VERSION
-
-ENV APPPROTECT_MODULE_VERSION 23+3.462.0-1~buster
-ENV APPPROTECT_PLUGIN_VERSION 3.462.0-1~buster
-ENV APPPROTECT_ENGINE_VERSION 6.53.1-1~buster
-ENV APPPROTECT_COMPILER_VERSION 6.53.1-1~buster
-ENV APPPROTECT_SIG_VERSION 2021.02.26-1~buster
-ENV APPPROTECT_THREAT_CAMPAIGNS_VERSION 2021.03.02-1~buster
+ARG NGINX_PLUS_VERSION
 
 RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode=0644 \
 	--mount=type=secret,id=nginx-repo.key,dst=/etc/ssl/nginx/nginx-repo.key,mode=0644 \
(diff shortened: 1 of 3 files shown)


build/Dockerfile
```diff
@@ -65,14 +61,17 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
...
+	&& apt-get update \
+	&& module_version=$(apt-cache showpkg nginx-plus-module-appprotect | awk -v ver="nginx-plus-$NGINX_PLUS_VERSION" '{ if ($6 == ver) {print $1; exit}}') \
+	engine_version=$(apt-cache showpkg app-protect | awk -v ver="$module_version" '{ if ($1 == ver && $3 == "nginx-plus-module-appprotect") {print substr($NF, 1, length($NF)-1); exit}}') \
+	&& apt-get install --no-install-recommends --no-install-suggests -y \
+	nginx-plus-module-appprotect=${module_version} \
+	app-protect-plugin=$(apt-cache showpkg nginx-plus-module-appprotect | awk -v ver="nginx-plus-$NGINX_PLUS_VERSION" '{ if ($6 == ver) {print substr($NF, 1, length($NF)-1); exit}}') \
+	app-protect-engine=${engine_version} \
+	app-protect-compiler=${engine_version} \
+	app-protect=${module_version} \
+	&& apt-get install -y app-protect-attack-signatures \
```
lucacome, line 73: we will always install the latest version of signatures and threats, I don't think it makes sense pin it to a specific version

build/Dockerfile
```diff
@@ -34,8 +34,10 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
 	&& wget -P /etc/apt/apt.conf.d https://cs.nginx.com/static/files/90pkgs-nginx \
 	&& echo "Acquire::https::pkgs.nginx.com::User-Agent  \"k8s-ic-$IC_VERSION-apt\";" >> /etc/apt/apt.conf.d/90pkgs-nginx \
 	&& printf "deb https://pkgs.nginx.com/plus/debian buster nginx-plus\n" > /etc/apt/sources.list.d/nginx-plus.list \
-	&& apt-get update && apt-get install --no-install-recommends --no-install-suggests -y \
-	nginx-plus=${NGINX_PLUS_VERSION} nginx-plus-module-njs=${NGINX_NJS_VERSION} \
+	&& apt-get update \
+	&& apt-get install --no-install-recommends --no-install-suggests -y \
+	nginx-plus=$(apt-cache showpkg nginx-plus-$NGINX_PLUS_VERSION | awk '{ if ($1 == "nginx-plus") {print $2; exit}}') \
```
lucacome, line 39: Form R24 the virtual package started working so this could be simplified to `nginx-plus-r24`, but doesn't work with `R23`.

This doesn't apply to NAP since they don't have R virtual packages.

build/Dockerfile
```diff
@@ -1,5 +1,6 @@
 # syntax=docker/dockerfile:1.2
 ARG BUILD_OS=debian
+ARG NGINX_PLUS_VERSION=r24
```
pleshakov, line 3: we should document the behavior here https://docs.nginx.com/nginx-ingress-controller/installation/building-ingress-controller-image/  : when  building an NGINX Plus image, the Dockerfile will install the latest patch version of the major NGINX Plus version supported in by the Ingress Controller release.

build/Dockerfile
pleshakov, line 39: would it make sense to add comments about why those bash scripts are necessary? For example, "figure out the latest patch version of the supported NGINX Plus version". I found those commands a bit confusing, but that might be just me though.

build/Dockerfile
```diff
@@ -65,14 +61,17 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
...
-	&& apt-get update && apt-get install --no-install-recommends --no-install-suggests -y \
-	nginx-plus-module-appprotect=$APPPROTECT_MODULE_VERSION \
-	app-protect-plugin=$APPPROTECT_PLUGIN_VERSION \
-	app-protect-engine=$APPPROTECT_ENGINE_VERSION \
-	app-protect-compiler=$APPPROTECT_COMPILER_VERSION \
-	app-protect=$APPPROTECT_MODULE_VERSION \
-	&& apt-get install -y app-protect-attack-signatures${APPPROTECT_SIG_VERSION:+=$APPPROTECT_SIG_VERSION} \
-	&& apt-get install -y app-protect-threat-campaigns${APPPROTECT_THREAT_CAMPAIGNS_VERSION:+=$APPPROTECT_THREAT_CAMPAIGNS_VERSION} \
+	&& apt-get update \
+	&& module_version=$(apt-cache showpkg nginx-plus-module-appprotect | awk -v ver="nginx-plus-$NGINX_PLUS_VERSION" '{ if ($6 == ver) {print $1; exit}}') \
```
pleshakov, line 65: what is version of app protect that gets used?

does it mean a new app protect release (say 3.3) will be automatically supported? if that is the case, is it reasonable? do we need to ensure that our pipelines pass tests against that new version?

build/Dockerfile
pleshakov: pinging @rafwegv to double check if there are any potential issues

build/Dockerfile
```diff
@@ -20,8 +21,7 @@ RUN apk add --no-cache libcap \
 ############################################# Base image for Debian with NGINX Plus #############################################
 FROM debian:buster-slim AS debian-plus
 ARG IC_VERSION
-ENV NGINX_PLUS_VERSION 23-1~buster
-ENV NGINX_NJS_VERSION 23+0.5.0-1~buster
+ARG NGINX_PLUS_VERSION
```
pleshakov, line 24: I wonder if should provide a fallback option - allow customers to use specific version. 
For example, imagine a problem is discovered in a new version of NGINX Plus patch version. Or in a module. Would it be possible for a customer to fall back to the previous versions?

build/Dockerfile
lucacome: yes, whatever is the latest compatible version with that `R` version

build/Dockerfile
lucacome: if we don't want to support building for older versions this can be simplified to just use r24 like for UBI, but I can add a comment otherwise (or in NAP)

build/Dockerfile
lucacome: hmm has this ever happened? I think if there was a known problem with a specific version of NGINX Plus, the team would release a new version right away?

I can't think of a way to allow for a specific version with this method, you can only change R version I think, like use `r23`. In Debian the patch version is not specified in the virtual package
```
Provides:
24-2~buster - nginx-plus-r24 ...
24-1~buster - nginx-plus-r24 ...
```
they're both `nginx-plus-r24`. 

I remember reading that this is because `apt` was made with the assumption that people want to always install the latest version of a package 99% of the time .

build/Dockerfile
rafwegv: @pleshakov @lucacome Afair this was pinned to have repeatable builds. I am all good with just installing the latest.

build/Dockerfile
rafwegv: In this (3.2) release only R24 is supported (and this will probably be the case for further releases), so i think installing latest nap supported by plus is a good way to go

build/Dockerfile
lucacome: yeah I think it's better to build with the latest available signatures and threats, outdated ones don't really make sense right?

build/Dockerfile
pleshakov: hm. perhaps it's just a matter of some additional documentation. For example:

```bash
	&& apt-get install --no-install-recommends --no-install-suggests -y \
	## the commands below install the latest patch version of the major NGINX Plus version specified in $NGINX_PLUS_VERSION
	## as well as the latest versions of the modules packages compatible with $NGINX_PLUS_VERSION
	## if you would like to fix the installed version, modify the lines below
	## For example:
	##  nginx-plus=24-2~buster
	nginx-plus=$(apt-cache showpkg nginx-plus-$NGINX_PLUS_VERSION | awk '{ if ($1 == "nginx-plus") {print $2; exit}}') \
	nginx-plus-module-njs=$(apt-cache showpkg nginx-plus-module-njs-$NGINX_PLUS_VERSION | awk '{ if ($1 == "nginx-plus-module-njs") {print $2; exit}}') \
```

I think something like that will give a clear way for customers to fix the versions if that is required. Whether they want reproducible builds or need to roll back to the previous NGINX Plus or any module version

What do you think?

build/Dockerfile
pleshakov: by older versions, you mean R23?

build/Dockerfile
pleshakov: we will need to document what gets installed for app protect. I think here is a good place https://docs.nginx.com/nginx-ingress-controller/app-protect/installation/#build-the-docker-image

pleshakov: > yeah I think it's better to build with the latest available signatures and threats, outdated ones don't really make sense right?

 Can `app-protect-attack-signatures` or `app-protect-threat-campaigns` after being upgraded in the repos require a newer version of the other app protect packages, newer than the one we calculated?  In that case, the install will fail. But not sure if this case can happen at all.

lucacome: No they're not tied to a specific version of the app protect package

build/Dockerfile
lucacome: anything before R24

build/Dockerfile
lucacome: Thinking about this again, we never mentioned before what version of NGINX would be installed. Not sure if we should start now.

build/Dockerfile
lucacome: I've changed it to use `nginx-plus-r24` to match the other builds.

build/Dockerfile
pleshakov: previously, the version was fixed. Not it is floating, which means at different points of time, the `docker build` command can produce an image with different versions. I think we should mention that. This is a change from previous behavior.

pleshakov: Because we will not release a new version of KIC on every patch release, we should explicitly mention that. Something like this:
After a [new NGINX Plus patch version](https://docs.nginx.com/nginx/releases/) is released, make sure to rebuild the image so that you always running the latest version.

build/Dockerfile
pleshakov: what would be a way to fix the versions? could we mention how to do that?

lucacome: Talked to @brianehlert and we shouldn't block this PR because of docs. If a request for docs on how to change the version arises, we'll update them then.

pleshakov: sounds good. could you address this one https://github.com/nginxinc/kubernetes-ingress/pull/1604#discussion_r634786314 ?
we have the same thing for the app protect images as well - because now we will always install the latest

brianehlert: The major version is set in the dockerfile.
The convenience is that the latest matching release of the major version is installed.
I believe this aligns with expectations and provides reasonable stability.

I would expect app protect to operate the same way.  Major version pin.  Latest patch.
This should avoid problems.
I would expect app protect to pin to versions of nginx.  As there are modules involved and those are compiled to the nginx binary.

Adding a note to rebuild your image when there is an update to nginx to get the latest update is prudent.

soneillf5 (approved):

docs-web/installation/building-ingress-controller-image.md
```diff
@@ -55,6 +55,8 @@ We build the image using the make utility and the provided `Makefile`. Let’s c
 
       As a result, the image **myregistry.example.com/nginx-plus-ingress:1.11.2** is built. Note that the tag `1.11.2` comes from the `VERSION` variable, defined in the Makefile.
 
+      **Note**: In the event of a patch version of NGINX Plus being [released](/nginx/releases/), make sure to rebuild your image to get the latest version. If your system is caching the Docker layers, add `DOCKER_BUILD_OPTIONS="--no-cache"` to the `make` command.
```
pleshakov, line 58: could you add a similar note for App Protect? The PR also affects how AppProtect packages are installed 

https://docs.nginx.com/nginx-ingress-controller/app-protect/installation/#build-the-docker-image

lucacome: updated

pleshakov (approved):

PR #1605, opened 2021-05-18
Revert dependabot to daily checks

lucacome:
- Revert back to daily checks now that we went through the initial updates
- Remove minor and patch version from actions to reduce noise


pleshakov (approved):
🙏

PR #1619, opened 2021-05-21
Add AWS Marketplace Entitlement verification

lucacome:
AWS Entitlement verification


cmd/nginx-ingress/aws.go
```diff
@@ -0,0 +1,108 @@
...
+
+	mpm := marketplacemetering.New(marketplacemetering.Options{Region: cfg.Region, Credentials: cfg.Credentials})
+
+	var notEnt *types.CustomerNotEntitledException
+	var invalidRegion *types.InvalidRegionException
+
+	out, err := mpm.RegisterUsage(context.TODO(), &marketplacemetering.RegisterUsageInput{ProductCode: &productCode, PublicKeyVersion: &pubKeyVersion, Nonce: &nonce})
+	if err != nil {
+		if errors.As(err, &notEnt) {
+			log.Fatalf("User not entitled: %v", err)
```
lucacome, line 45: we'll probably need to update the messaging here

cmd/nginx-ingress/aws.go
```diff
@@ -0,0 +1,108 @@
...
+
+var (
+	productCode   string
+	pubKeyVersion int32 = 1
+	pubKeyString  string
+	nonce         string
+)
+
+func init() {
+	rand.Seed(jwt.Now().UnixNano())
```
soneillf5, line 29: This is used for security purposes, so this needs to be secure. Can you double check this is the correct way to securely seed the rand num generator ?

cmd/nginx-ingress/aws.go
```diff
@@ -0,0 +1,108 @@
...
+	nonce = RandStringBytesRmndr(32)
+
+	cfg, err := config.LoadDefaultConfig(context.TODO())
```
soneillf5, line 32: It's probably worth implement the context here correctly. https://blog.golang.org/context

cmd/nginx-ingress/aws.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	mpm := marketplacemetering.New(marketplacemetering.Options{Region: cfg.Region, Credentials: cfg.Credentials})
+
+	var notEnt *types.CustomerNotEntitledException
+	var invalidRegion *types.InvalidRegionException
+
+	out, err := mpm.RegisterUsage(context.TODO(), &marketplacemetering.RegisterUsageInput{ProductCode: &productCode, PublicKeyVersion: &pubKeyVersion, Nonce: &nonce})
+	if err != nil {
+		if errors.As(err, &notEnt) {
```
soneillf5, line 44: I think errors.As is used to convert the error to a specific type, usually to get error details specific to that error type. In this case, just the type is checked so I think it should be errors.Is() ?
WDYT?

soneillf5 (changes requested):
Minor changes and some questions

cmd/nginx-ingress/aws.go
lucacome: Which context should we use? In the docs and examples they always use `context.TODO()` when loading the config.
https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#loading-aws-shared-configuration

cmd/nginx-ingress/aws.go
lucacome: Updated to print details

cmd/nginx-ingress/aws.go
lucacome: I'm pretty sure it was secure enough already, but I've updated it to use `crypto/rand` just in case 👍

cmd/nginx-ingress/aws.go
```diff
@@ -0,0 +1,111 @@
...
+			log.Println("AWS verification successful")
+		} else {
+			log.Fatal("the claims in the JWT token don't match the request")
+		}
+	} else {
+		log.Fatal("something is wrong with the JWT token")
+	}
+}
+
+type Claims struct {
```
pleshakov, line 75: does this type need to be exporter?

cmd/nginx-ingress/aws.go
```diff
@@ -0,0 +1,111 @@
...
+	if c.Nonce == "" {
+		return &jwt.InvalidClaimsError{Message: "the JWT token doesn't include the Nonce"}
+	}
+	if c.ProductCode == "" {
+		return &jwt.InvalidClaimsError{Message: "the JWT token doesn't include the ProductCode"}
+	}
+	if c.PublicKeyVersion == 0 {
+		return &jwt.InvalidClaimsError{Message: "the JWT token doesn't include the PublicKeyVersion"}
+	}
+	if h.Before(c.IssuedAt.Time) {
```
pleshakov, line 92: since `IssuedAt` is a pointer, can it be nil (possible panic here)?

cmd/nginx-ingress/aws.go
pleshakov: can it make a network call? If that is case, it makes sense to add a `context.WithTimeout...` so that we control the timeout.

cmd/nginx-ingress/aws.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+var (
+	productCode   string
+	pubKeyVersion int32 = 1
+	pubKeyString  string
+	nonce         string
+)
+
+func init() {
```
pleshakov, line 27: Right now this function is called every time the binary is executed. However, the binary can be executed without the intention of starting the IC - for example,  a troubleshooting `/nginx-ingress version` command to determine the IC version. With the current `init()` implementation, that troubleshooting command with also register usage.

However, there is a bit bigger concern I see - the function has a lot of logic which we don't test at all. 

To address those concerns I suggest the following:

(1) Let's call this function indirectly. This way we will only call it when the IC is meant to start.
(2) Let's remove all log Fatal statement. Instead, the function will return an error. If the error is not nil, we will exit in main.go

in main.go
```go
var startupCheckFn func() error

func main() {
	flag.Parse()

	err := flag.Lookup("logtostderr").Value.Set("true")
	if err != nil {
		glog.Fatalf("Error setting logtostderr to true: %v", err)
	}

	versionInfo := fmt.Sprintf("Version=%v GitCommit=%v Date=%v", version, commit, date)
	if *versionFlag {
		fmt.Println(versionInfo)
		os.Exit(0)
	}

	if startupCheckFn != nil {
		err := startupCheckFn()
		if err != nil {
			glog.Fatalf("Failed startup check: %v", err)
		}
		glog.Info("Startup check is successful")
	}
	
	. . .
```

in aws.go
```go
func init() {
	startupCheckFn = checkAWSEntitlement
}

func checkAWSEntitlement() error {
	. . .
}
```

this way we will be able to test checkAWSEntitlement. 

Since we only make one external API call, we will be able to test it by wrapping it into an interface like this:
```go
type awsMeter interface {
	RegisterUsage(params *marketplacemetering.RegisterUsageInput) (*marketplacemetering.RegisterUsageOutput, error)
}
```
In unit tests, we will be able to create a fake implementation that will return the right output.

lucacome: What kind of logic would you want to test?

pleshakov: the `if`s in init()

cmd/nginx-ingress/aws.go
soneillf5: Your right, I looked at the docs and its all `context.TODO()` 

As Michael said, it's usually a timeout for network calls.

cmd/nginx-ingress/aws.go
soneillf5: I don't think `init` needs a test. You could pull out and test the logic around the token claims and unit test that.
The question around the binary is interesting. Do we need a version check though? If so  we could just add that inline.

pleshakov: > I don't think init as a whole needs a test as there's so little logic, just ifs for the errors. You could wrap and test the logic around the token claims and unit test that?

we have a happy path when nothing returns an error and we a number places where the error needs to be checked. why would we not want to test it? 🤔 
yeah, some additional refactoring of init can also help

> The question around the binary is interesting. Why would this code need a version check?

this init code is always called, even if the version check command is invoked on the binary.

lucacome: I'm not sure I understand what you're saying, which ifs are you talking about? We're calling an external API and using external libraries to parse the response and a standard library to decode a base64 string.

The only logic that we implement is checking that the strings in the request match the ones in the response and that they're not empty. I don't think a unit test is worth it since we don't generate that response.

> this init code is always called, even if the version check command is invoked on the binary.

If someone starts the container in AWS to check the version, well...they're starting it...I don't see why we shouldn't call the API. If they don't want to run it there, there are other ways to see the version, like looking at the labels in the Docker image

pleshakov: > I'm not sure I understand what you're saying, which ifs are you talking about? We're calling an external API and using external libraries to parse the response and a standard library to decode a base64 string.

we call a number of functions after getting that response. also, the request can fail or return an error -- those cases are also handled. 
I'm not suggesting testing the standard library or other libraries. What I'm suggesting is to test the `init` function --  so that we're sure that all of the error cases and the happy path are handled.

> If someone starts the container in AWS to check the version, well...they're starting it...I don't see why we shouldn't call the API. If they don't want to run it there, there are other ways to see the version, like looking at the labels in the Docker image

the version command would be run via `kubectl exec ... `. 
Yes, there are other ways to see the version.  The new behaviour makes the `version` command billable - I don't think anyone would expect that.

pleshakov: this is a new, but a related question. the IC checks for many prerequisites (After this line https://github.com/nginxinc/kubernetes-ingress/blob/master/cmd/nginx-ingress/main.go#L204  before starting the Controller here https://github.com/nginxinc/kubernetes-ingress/blob/master/cmd/nginx-ingress/main.go#L661)

Do we want to register usage even if the IC will terminate immediately? For example, in case of misconfiguration.

lucacome: >RegisterUsage meters software use per Amazon Elastic Container Service (Amazon ECS) task or per Amazon Elastic Kubernetes Service (Amazon EKS) pod, per hour, with usage prorated to the second. A minimum of 1 minute of usage applies to tasks or pods that are short lived. Continuous metering for software use is automatically handled by the AWS Marketplace Metering Control Plane. Your software is not required to perform any metering specific actions except calling RegisterUsage once for metering of software use to commence.

https://docs.aws.amazon.com/marketplace/latest/userguide/container-metering-registerusage.html

cmd/nginx-ingress/aws.go
lucacome: `LoadDefaultConfig` could make a network call, I think it's the last resource if it can't find a valid config. `RegisterUsage` of course makes a network call, I don't think I see the value of controlling the timeout, but I've added it. It's set to 5 seconds now, no idea what an acceptable timeout is.

cmd/nginx-ingress/aws.go
soneillf5: I'm with Luca on this one, this is a very simple program that makes calls to external APIs and does a minor amount of checking. I don't think the overhead of writing and maintaining the tests is worth it.

soneillf5 (approved):

cmd/nginx-ingress/aws.go
soneillf5: magic thanks!

cmd/nginx-ingress/aws.go
pleshakov: I didn't mention it before - because the original suggestion was to not use log.Fatal and other statements inside of `init`. If we continue to use logs in init, can we use the glog? This is for consistency, so that IC outputs logs in the same format

lucacome: The point of using `init()` is to run it before `main()` automatically, without modifying main.
I don't think it's worth adding complexity and modifying `main()` for the use case of people wanting to check the version, I couldn't even find instructions in the docs. (I would go as far as deleting that parameter and adding docs on how to check the docker labels)

I'm using `log` because `glog` is requiring a `flag.Parse()` before it can be used and I don't think we should parse the flags at that point.

pleshakov: supporting calling the registration from main is easy:

main.go
```go
var startupCheckFn func() error

. . .
       if startupCheckFn != nil {
          startupCheckFn()
       }
```
aws.go
```go
func init() {
	startupCheckFn = checkAWSEntitlement
}

func checkAWSEntitlement() {
	. . .
}
```

However, the result is that (1) it allows  using the same logger (glog) and (2) it will not affect the version check at all.  

> The point of using init() is to run it before main() automatically, without modifying main

the requirement is to run that code as part of the Ingress Controller startup, so that the pod is registered. However, we don't have the requirement that it must be run from the `init()` function. when implementing new features, in general we should not break the existing ones (in this case version check) and make sure that the new functionally fits nicely with the existing one (in this case, use the same logger for consistent format).

lucacome: - The version check still works
- The log is only one line. I just pushed a change to only print on failure, so that would be the only line printed and there won't be any inconsistency.

pleshakov: > The version check still works

yep. but it works with a side effect - it will register a pod

> The log is only one line. I just pushed a change to only print on failure, so that would be the only line printed and there won't be any inconsistency.

Below is an example of inconsistent format:
fatal log in the init
```
2021/05/21 14:58:08 operation error Marketplace Metering: RegisterUsage, failed to resolve service endpoint, an AWS region is required, but was not found
```
fatal log in the main.go
```
I0526 15:28:24.173615   45447 main.go:259] Starting NGINX Ingress controller Version=my-version GitCommit= Date= PlusFlag=false
F0526 15:28:24.173819   45447 main.go:275] error creating client configuration: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
```

bonnyr: I would like to offer some other considerations, for why @pleshakov's points merit consideration:
1. `init` may be called before `main` but there's no ordering guarantee for which `init` is called before another between packages. Not really the case here but it could be if more pakages are using init. This is implementation-dependent and the golang authors are free the change that implementation  leading to silent breaks
2. `init` is magic and we should strive for simplicity rather than things happening automagically. 
3. whilst `init` can be used for setting up structures etc, since it's very early in the lifecycle of the process, using it to interact with state (the registration here) is beyond what would be good practice. IMO it would be better to explicitly initialize and control initialization from `main`
4.  using `log.Fatal` in `init` is a recipe for surprise by developers down the line. I would not use that.
5. Testing should definitely cover the cases in the initialization however simple (or at least create tech debt to record this)
6. I would choose consistency (in this case agreeing on the particular log library) so whichever version allows producing consistent outcome for users is preferrable. You never know how users are going to use the logs for their own purposes



my 2c

lucacome: The best solutions are always the most obvious that don't make you think and obviously this made a lot of people think! Thanks for all the great feedback, I will go with the suggested solution.

pleshakov (approved):

PR #1633, opened 2021-05-26
Support match in TransportServer health checks

pleshakov:
### Proposed changes
Support match in TransportServer health checks
Based on the original work from @soneillf5


internal/configs/transportserver.go
```diff
@@ -164,6 +179,29 @@ func generateTransportServerHealthCheckWithDefaults(up conf_v1alpha1.Upstream) *
...
+	}
+}
+
+func generateHealthCheckMatch(match *conf_v1alpha1.Match, name string) *version2.Match {
+	var modifier string
+	var expect string
+
+	if strings.HasPrefix(match.Expect, "~*") {
+		modifier = "~*"
+		expect = strings.TrimPrefix(match.Expect, "~*")
```
soneillf5, line 192: What does this code do? I'm not sure what `~*` does.

internal/configs/transportserver.go
```diff
@@ -164,6 +179,29 @@ func generateTransportServerHealthCheckWithDefaults(up conf_v1alpha1.Upstream) *
...
+		modifier = "~"
+		expect = strings.TrimPrefix(match.Expect, "~")
+	} else {
+		expect = match.Expect
+	}
+
+	return &version2.Match{
+		Name:                name,
+		Send:                match.Send,
+		ExpectRegexModifier: modifier,
```
soneillf5, line 203: Looks like this field has two values, `~*` and `*` and based on https://serverfault.com/questions/801356/what-is-the-difference-between-nginx-and-regexes it means it's either case insensitive or case sensitive. If that's the case, can we add a field to `Match` called caseSensitiveMatch? Then we can set the `ExpectRegexModifier` based on that.

soneillf5 (changes requested):
Can you add a test to the python test suite that confirms the functionality of this change?

internal/configs/transportserver.go
pleshakov: There are 3 cases - no regex, regex with case sensitive, regex with case insensitive , so just a boolean field will not work. Also see https://nginx.org/en/docs/stream/ngx_stream_upstream_hc_module.html#match 

The current approach is consistent with the existing configuration:

```yaml
  - path: ~ ^/decaf/.*\\.jpg$
    action:
      pass: coffee
```

```yaml
  - path: /coffee
    matches:
    - conditions:
      - cookie: user
        value: ~^m
      action:
        pass: coffee-v2
    action:
      pass: coffee-v1
```

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -407,3 +408,121 @@ def test_tcp_request_load_balanced_method(
...
+
+        match = f"match_ts_{transport_server_setup.namespace}_transport-server_tcp-app"
+
+        assert "health_check interval=5s port=3333" in result_conf
+        assert f"passes=1 jitter=0s fails=1 match={match}" in result_conf
+        assert "health_check_timeout 3s;"
+        assert 'send "health"' in result_conf
+        assert 'expect  "healthy"' in result_conf
+
+        # Step 2 - confirm load balancing works
```
soneillf5, line 448: I would recommend not checking the load balancing here. I think the test should confirm the healthcheck functionality not the load balancing. If a test fails, it will be more accurate and the test easier to diagnose.

internal/configs/transportserver.go
soneillf5: Can we add a new, optional, field called 'regexModifier' with default of empty? 
So it looks like:
```
match:
   expect:
   send:
   regexModifier:
```
I think we shouldn't compress values and modifiers into the same field. In terms of design, it's easier to change and easier to reason about.

tests/suite/test_transport_server_tcp_load_balance.py
pleshakov: In this test we check if  health checks that suppose to pass actually pass. For NGINX Plus, this means it will load balance to all 3 upstream servers, which NGINX Plus will consider healthy. If the health checks don't pass for any of upstream servers, NGINX Plus will not load balance to it. That's how we know there is a problem. Hope this clarifies.

internal/configs/transportserver.go
pleshakov: As mentioned here https://github.com/nginxinc/kubernetes-ingress/pull/1633#discussion_r640920775, the choice was made to towards consistency with the existing configuration:
```yaml
  - path: ~ ^/decaf/.*\\.jpg$ # regex
    action:
      pass: coffee
  - path: /coffee
    matches:
    - conditions:
      - cookie: user
        value: ~^m # regex
      action:
        pass: coffee-v2
    action:
      pass: coffee-v1
```

Note that YAML configuration is independent of how we represent the NGINX config in the structs we use for templates. And In those structs, the regexModifier field is present, as it makes it easier to generate the config.

tests/suite/test_transport_server_tcp_load_balance.py
vepatel: @soneillf5 @pleshakov looks like another "setting up clear boundaries" to mw, however I'm happy enough with just checking the contents in conf and making sure that reload passes.

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -407,3 +408,121 @@ def test_tcp_request_load_balanced_method(
             client.close()
 
         assert len(endpoints) is 3
+
+    @pytest.mark.skip_for_nginx_oss
+    def test_tcp_passing_healthcheck_with_match(
```
vepatel, line 413: @pleshakov these tests also looks like a decent candidates to use parameterization to pass various ts/cases, see: https://github.com/nginxinc/kubernetes-ingress/blob/master/tests/suite/test_ingress_mtls.py#L82

internal/configs/transportserver_test.go
```diff
@@ -526,10 +526,13 @@ func TestGenerateUnixSocket(t *testing.T) {
 
 func TestGenerateTransportServerHealthChecks(t *testing.T) {
 	upstreamName := "dns-tcp"
+	generatedUpsteamName := "ts_namespace_name_dns-tcp"
```
vepatel, line 529: is there one for dns-udp too?

tests/suite/test_transport_server_tcp_load_balance.py
soneillf5: @pleshakov Ah I see now, the functionality is the change in load balancing behaviour. In which, yeah this is a great test and is fine as is.

pleshakov: @vepatel @soneillf5 
I'm  happy to remove the load balancing part and leave out only the config check.

There is another way to check that health check pass or fail -- get the status information about the upstream servers from the API. Take a look at https://demo.nginx.com/#tcp_upstreams -- the health monitors group. That information is available via API.

tests/suite/test_transport_server_tcp_load_balance.py
pleshakov: makes sense. let me know about this https://github.com/nginxinc/kubernetes-ingress/pull/1633/files#r643518158

soneillf5 (approved):

tests/suite/test_transport_server_tcp_load_balance.py
vepatel: Yeah that's fine for now that its added and working but we don't need to verify all the attribs as nginx plus is doing lb and the UI is provided by the same.

vepatel (approved):

PR #1652, opened 2021-06-08
Add HTTPS option to prometheus endpoint

soneillf5:
### Proposed changes
This change adds https as an option of the prometheus endpoint.

### Checklist


pleshakov:
Hi @soneillf5 

The implementation looks good.

However, I think we can make it more integrated with Kubernetes and add some usability. The issue I see is how the admin will propagate the cert and key to the file system of  the pod? Also consider our three installation methods: manifests, helm chart and the operator.

A good approach that would work well for all installation methods is for the admin to create a Secret with a TLS cert and key and pass its name as an Ingress Controller parameter. This would be similar to the `-default-server-tls-secret` or `-wildcard-tls-secret` command-line flags. 

Could we have a similar `-prometheus-tls-secret` flag?

If it is set, the TLS will be enabled for Prometheus metrics and the IC will use the TLS cert and key from the referenced Secret. If the secret doesn't exist or it is invalid, the IC will fail to start with an error.

An extra feature here - supporting updating that secret - if the Secret is updated, the IC will start using the new version. This is something that we support for the default and the wildcard secrets. However, supporting updating will be more involved, and we can start with a simple implementation without it.

Regarding different installation methods:
* For the manifests installation, the admin will need to create a TLS Secret and update the Ingress Controller deployment/daemonset to use it (the cli args). There are also annotations that are needed to be added as described here https://docs.nginx.com/nginx-ingress-controller/logging-and-monitoring/prometheus/#enabling-metrics 
* For Helm chart, we can add parameters similar to the parameters for the default server secret and the wildcard secret ( see `controller.wildcardTLS.*`). There is no need to provide the default value of the cert and key. Note that we already have `prometheus.*` parameter, so it makes sense to extend them. For example, `prometheus.tls.cert`, `prometheus.tls.key` and `prometheus.tls.secret`. If the `prometheus.tls.secret`, helm will configure the IC args to use the referenced secret. If ``prometheus.tls.cert` and `prometheus.tls.key` are set, helm will create a Secret with those values and configure the IC args to use it.
* For the Operator, we can add  a field:
```
   prometheus:
     enable: true
     port: 9114
     tlsSecret: ns/name
 ```

Regarding documentation - we will need to document all new cli args, helm parameters, etc. It makes sense to mention that you can secure the Prometheus endpoint with TLS here https://docs.nginx.com/nginx-ingress-controller/logging-and-monitoring/prometheus/ 

Not tested, but we might need to add `prometheus.io/scheme: https` annotation to the pod template if TLS is enabled. For manifests installation, we can simply add and comment it out. For helm and operator, we will need to add it if the TLS is enabled for Prometheus.

While the operator lives in a separate repo and requires a separate PR, I mentioned it here for completeness.

**Comment by soneillf5** [comment]
@pleshakov I've updated to use a Kubernetes Secret resource for the TLS cert data, I still need to add an automated test for it, but I've confirmed it works locally. I'll update the operator/helm/docs in the next few days too.

File: cmd/nginx-ingress/main.go [hunk 1]
```diff
@@ -590,14 +601,14 @@ func main() {
...
 				streamUpstreamServerVariableLabels, streamServerZoneVariableLabels, streamUpstreamServerPeerVariableLabelNames)
 			plusCollector = nginxCollector.NewNginxPlusCollector(plusClient, "nginx_ingress_nginxplus", variableLabelNames, constLabels)
-			go metrics.RunPrometheusListenerForNginxPlus(*prometheusMetricsListenPort, plusCollector, registry)
+			go metrics.RunPrometheusListenerForNginxPlus(*prometheusMetricsListenPort, plusCollector, registry, prometheusSecret)
 		} else {
 			httpClient := getSocketClient("/var/lib/nginx/nginx-status.sock")
 			client, err := metrics.NewNginxMetricsClient(httpClient)
 			if err != nil {
-				glog.Fatalf("Error creating the Nginx client for Prometheus metrics: %v", err)
+				glog.Warningf("Error creating the Nginx client for Prometheus metrics: %v", err)
```
soneillf5, line 609: Ignore this, will remove.

internal/metrics/listener.go
```diff
@@ -49,5 +53,47 @@ func runServer(port string, registry prometheus.Gatherer) {
...
+			glog.Errorf("failed to create key file for prometheus: %w", err)
+		}
+		defer removeTemporaryFile(keyFile)
+
+		glog.Fatal("Error in Prometheus listener server: ", http.ListenAndServeTLS(address, certFile.Name(), keyFile.Name(), nil))
+	}
+}
+
+func writeTempFile(data []byte, name string) (*os.File, error) {
+	f, err := ioutil.TempFile("", name)
```
pleshakov, line 80: at this point, the name is already unique, because it includes the date prefix before the `nginx-prometheus.cert` and `nginx-prometheus.key`.   In `ioutil.TempFile("", name)` "the filename is generated by taking pattern and adding a random string to the end". So we basically we're doing the uniqueness work twice. Is it necessary? Do we need to have the date prefix?

Also note that it makes sense to store it with `600` permission so only the process under `nginx` user has access to it - this user is used to run both the IC and NGINX.  This is similar to the permission of the regular TLS secrets the IC writes:
```
	TLSSecretFileMode            = 0600  // TLSSecretFileMode defines the default file mode for files with TLS Secrets.
```

internal/metrics/listener.go
```diff
@@ -49,5 +53,47 @@ func runServer(port string, registry prometheus.Gatherer) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temp file: %w", err)
+	}
+	fmt.Println("Temp file name:", f.Name())
```
pleshakov, line 84: temp debug statement?

cmd/nginx-ingress/main.go
```diff
@@ -154,6 +154,9 @@ var (
 	enablePrometheusMetrics = flag.Bool("enable-prometheus-metrics", false,
 		"Enable exposing NGINX or NGINX Plus metrics in the Prometheus format")
 
+	prometheusTLSSecretName = flag.String("prometheus-tls-secret", "",
+		`A Secret with a TLS certificate and key for TLS termination of the prometheus endpoint.`)
```
pleshakov, line 158: neat: for consistent naming. we used Prometheus before. could you capitalize other new ocurrences?
```suggestion
		`A Secret with a TLS certificate and key for TLS termination of the Prometheus endpoint.`)
```

internal/metrics/listener.go
```diff
@@ -49,5 +53,47 @@ func runServer(port string, registry prometheus.Gatherer) {
...
+	} else {
+		// Unfortunately, http.ListenAndServeTLS() takes a filename instead of cert/key data, so we
+		// Write the cert and key to a temporary file. We create a unique file name to prevent collisions.
+		certFileName := fmt.Sprintf("%v-%s", time.Now().Unix(), "nginx-prometheus.cert")
+		keyFileName := fmt.Sprintf("%v-%s", time.Now().Unix(), "nginx-prometheus.key")
+		certFile, err := writeTempFile(prometheusSecret.Data[api_v1.TLSCertKey], certFileName)
+		if err != nil {
+			glog.Errorf("failed to create cert file for prometheus: %w", err)
+		}
+		defer removeTemporaryFile(certFile)
```
pleshakov, line 67: not sure why we need to remove those files. If the IC process terminates or exists, Docker runtime will remove the IC container along with all its files.

internal/metrics/listener.go
```diff
@@ -49,5 +53,47 @@ func runServer(port string, registry prometheus.Gatherer) {
...
+	if prometheusSecret == nil {
+		glog.Fatal("Error in Prometheus listener server: ", http.ListenAndServe(address, nil))
+	} else {
+		// Unfortunately, http.ListenAndServeTLS() takes a filename instead of cert/key data, so we
+		// Write the cert and key to a temporary file. We create a unique file name to prevent collisions.
+		certFileName := fmt.Sprintf("%v-%s", time.Now().Unix(), "nginx-prometheus.cert")
+		keyFileName := fmt.Sprintf("%v-%s", time.Now().Unix(), "nginx-prometheus.key")
+		certFile, err := writeTempFile(prometheusSecret.Data[api_v1.TLSCertKey], certFileName)
+		if err != nil {
+			glog.Errorf("failed to create cert file for prometheus: %w", err)
```
pleshakov, line 65: I wonder if it makes sense to Fatal here (and in line 71). Otherwise, the error will occur in `http.ListenAndServeTLS`, which might be less clear that the err here.

internal/metrics/listener.go
soneillf5: removed

cmd/nginx-ingress/main.go
soneillf5: I think it's fine as it is.

internal/metrics/listener.go
soneillf5: I'd rather not leave certs hanging around on the file system unless they are needed again.

internal/metrics/listener.go
soneillf5: fixed

internal/metrics/listener.go
pleshakov: it is a good practice! 

however, note that `removeTemporaryFile()` will never be executed.  That's because ` http.ListenAndServeTLS(address, certFile.Name(), keyFile.Name(), nil)` blocks forever. Even if it doesn't block, but returns an error, glog.Fatal will print it and terminate the IC process, so that `removeTemporaryFile()` will not be called.

 If the intention is to remove the files immediately after `http.ListenAndServeTLS` is called , then the current approach doesn't work :(

cmd/nginx-ingress/main.go
pleshakov: is `Warningf` still needed?

deployments/helm-chart/values.yaml
```diff
@@ -310,3 +310,6 @@ prometheus:
 
   ## Configures the port to scrape the metrics.
   port: 9113
+
+  ## Specifies the namespace/name of a Kubernetes TLS Secret which will be used to protect the Prometheus endpoint.
+  secret: ""
```
pleshakov, line 315: could we add the ability to configure the secret values? this is handy, as it allows the admins to define every part of the Ingress Controller installation through a helm chart. 

> For Helm chart, we can add parameters similar to the parameters for the default server secret and the wildcard secret ( see `controller.wildcardTLS.*`). There is no need to provide the default value of the cert and key. Note that we already have `prometheus.*` parameter, so it makes sense to extend them. For example, prometheus.tls.cert, prometheus.tls.key and prometheus.tls.secret. If the prometheus.tls.secret, helm will configure the IC args to use the referenced secret. If `prometheus.tls.cert` and `prometheus.tls.key` are set, helm will create a Secret with those values and configure the IC args to use it.
From https://github.com/nginxinc/kubernetes-ingress/pull/1652#issuecomment-857284938

pleshakov, line 315: there are two docs for helm chart that we need to update to include the new parameters:
- https://github.com/nginxinc/kubernetes-ingress/blob/master/deployments/helm-chart/README.md
- https://deploy-preview-1652--nginx-kubernetes-ingress.netlify.app/nginx-ingress-controller/installation/installation-with-helm/

internal/metrics/listener.go
```diff
@@ -2,14 +2,17 @@ package metrics
 
 import (
 	"fmt"
```
pleshakov, line 4: Is it possible to address the linting errors?

deployments/helm-chart/templates/controller-daemonset.yaml
```diff
@@ -142,6 +142,7 @@ spec:
 {{- end }}
           - -enable-prometheus-metrics={{ .Values.prometheus.create }}
           - -prometheus-metrics-listen-port={{ .Values.prometheus.port }}
+          - -prometheus-tls-secret={{ .Values.prometheus.secret }}
```
pleshakov, line 145: I haven't tested it (it needs to be checked with a deployed Prometheus) but the pod template is probably needed to be updated with `prometheus.io/scheme: https`  (both for Deployment and DS) if TLS is enabled for Prometheus endpoint

tests/suite/test_prometheus_endpoint.py
```diff
@@ -0,0 +1,148 @@
...
+            kube_apis, f"{TEST_DATA}/smoke/standard/smoke-ingress.yaml", test_namespace
+        )
+        delete_secret(kube_apis.v1, secret_name, test_namespace)
+
+    request.addfinalizer(fin)
+
+    return IngressSetup(ingress_host)
+
+
+@pytest.mark.demo
```
pleshakov, line 104: is it a temp mark?

tests/suite/test_prometheus_endpoint.py
```diff
@@ -0,0 +1,148 @@
...
+            enable_exporter_port,
+            expected_metrics,
+            ingress_setup,
+    ):
+        resp = {}
+
+        # assert http fails
+        req_url = f"http://{ingress_controller_endpoint.public_ip}:{ingress_controller_endpoint.metrics_port}/metrics"
+        try:
+            resp = requests.get(req_url, verify=False)
```
pleshakov, line 133: do we need to fail after the resp line when no exception is thrown?

tests/suite/test_prometheus_endpoint.py
```diff
@@ -0,0 +1,148 @@
...
+        except:
+            print("request fails as expected")
+
+        assert resp.status_code == 400, f"Expected 400 code for http request to /metrics but got {resp.status_code}"
+
+        # assert https succeeds
+        req_url = f"https://{ingress_controller_endpoint.public_ip}:{ingress_controller_endpoint.metrics_port}/metrics"
```
pleshakov, line 140: will not work for both NodePort and Loadbalancer service?

tests/suite/test_prometheus_endpoint.py
```diff
@@ -0,0 +1,148 @@
...
+
+    if cli_arguments["deployment-type"] == "deployment":
+        kube_apis.apps_v1_api.patch_namespaced_deployment(ingress_controller, namespace, body)
+    else:
+        kube_apis.apps_v1_api.patch_namespaced_daemon_set(ingress_controller, namespace, body)
+    wait_until_all_pods_are_ready(kube_apis.v1, namespace)
+
+
+@pytest.fixture(scope="class")
+def ingress_setup(request, kube_apis, ingress_controller_endpoint, test_namespace) -> IngressSetup:
```
pleshakov, line 72: what is the rationale behind testing an Ingress resource in this test?

pleshakov (commented):
Hi @soneillf5 

Could we also update this part https://docs.nginx.com/nginx-ingress-controller/logging-and-monitoring/prometheus/#enabling-metrics ? For example, you can make the second step there like this:
2. To enable TLS for the Prometheus endpoint, configure the `-prometheus-tls-secret` cli argument with the namespace and name of a [TLS Secret](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets).

Also, the `prometheus.io/scheme: https` annotation is required by the Prometheus to use TLS, we need to mention it  the step 3 there.

tests/suite/test_prometheus_endpoint.py
soneillf5: merged with previous test.

tests/suite/test_prometheus_endpoint.py
soneillf5: I'm not sure what your comment means, can you rephrase it?

tests/suite/test_prometheus_endpoint.py
soneillf5: removed

internal/metrics/listener.go
soneillf5: I thought I had it setup correctly, I've added 'go imports' to the file watcher so it won't happen again I think.

tests/suite/test_prometheus_endpoint.py
soneillf5: updated.

deployments/helm-chart/values.yaml
soneillf5: I've updated the first link but I couldn't find the second in the codebase.

deployments/helm-chart/templates/controller-daemonset.yaml
soneillf5: You're correct, I didn't realise it but prometheus uses the annotations as a marker to scrape. Fixed.

tests/suite/test_prometheus_metrics.py
```diff
@@ -173,3 +186,48 @@ def test_latency_metrics(
...
+                {"extra_args": ["-enable-prometheus-metrics", "-enable-latency-metrics", "-prometheus-tls-secret=nginx-ingress/prometheus-test-secret"]},
+                [
+                    'nginx_ingress_controller_ingress_resources_total{class="nginx",type="master"} 0',
+                    'nginx_ingress_controller_ingress_resources_total{class="nginx",type="minion"} 0',
+                ],
+            )
+        ],
+        indirect=["ingress_controller"],
+    )
+    @pytest.mark.sean
```
ciarams87, line 202: I'm guessing this was left in by accident?

ciarams87 (approved):

tests/suite/test_prometheus_metrics.py
soneillf5: fixed and added a local githook to prevent further issues.

tests/suite/test_prometheus_endpoint.py
pleshakov: typo. "will it work ... " , my bad!

I was concerned whether accessing the Prometheus endpoint will work in the test when the IC is exposed through a service of the type LoadBalancer.

This answers my question https://github.com/nginxinc/kubernetes-ingress/blob/79c98c311223fec1c1d956e17e40dbc58f0d68da/tests/suite/fixtures.py#L129 -- port 9113 is exposed through a service and it will work whether the service type is LoadBalancer and NodePort.

deployments/helm-chart/values.yaml
pleshakov: there are a few ways I see to keep those values protected:
- a plugin to encrypt secret values https://itnext.io/helm-3-secrets-management-4f23041f05c3 
- have a separate values files only for secrets (helm install can take multiple values , for examle, `helm install .. -f a.yaml -f b.yaml -f c.yaml`) and not store it in the source control, but generate it during deployment 
- you can also configure all or some helm parameters via cli args, so helm can take the values for those parameters from env variables.

Allowing configuring secret values in the helm chart allows the admin to use the chart as a single source of truth -- so to install the Ingress Controller, you only need to run `helm install ... ` and nothing else.

I think it is important to provide an alternative, so that secrets could be managed separately. That's why the `secret` parameter makes sense.

pleshakov: here is the link https://github.com/nginxinc/kubernetes-ingress/blob/master/docs-web/installation/installation-with-helm.md

cmd/nginx-ingress/main.go
pleshakov: @soneillf5 it appears to be a temp change. Shall we revert back to  Fatalf ?

pleshakov:
Hi @soneillf5  the changes look good 

There are few remaining comments:
https://github.com/nginxinc/kubernetes-ingress/pull/1652#discussion_r652977978
https://github.com/nginxinc/kubernetes-ingress/pull/1652#discussion_r652977540
https://github.com/nginxinc/kubernetes-ingress/pull/1652#discussion_r652978934

cmd/nginx-ingress/main.go
soneillf5: fixed, thanks for catching this.

deployments/helm-chart/values.yaml
soneillf5: thanks, updated.

soneillf5: I understand there are mitigations, such as additional values files and encryption and cli-args. 

I still think it's not a good idea and I won't be adding secret values to this PR. If you'd like to block this PR based on this, let me know so we can do something to resolve it.

pleshakov (approved):
🚀

Issue #1448, opened 2021-03-11
Help needed - tls secret took long to discover(ingress version:  1.10, kubernetes version: 1.20)


kongdewen:
we are seeing a weird problem when upgrade to latest ingress, 

when the k8s ingress pods start, it takes couple minutes for actually `find` the tls for the ingresses. It would first report secret missing or invalid type(we did have to switch tls type to kubernetes.io/tls lately to prepare of the upgrade) until sometime later came back to normal:

error sample:
```
W0311 14:30:31.734196       7 controller.go:1983] Error trying to get the secret  for Ingress xxx: secret doesn't exist or of an unsupported type
W0311 14:30:31.734212       7 controller.go:1983] Error trying to get the secret tls-wildcard-xxx for Ingress xxx: secret doesn't exist or of an unsupported type
```
secret sample:
```
 kubectl -n xxx describe secret/tls-wildcard-xxx
Name:         tls-wildcard-xxx
Namespace:    xxx
Labels:     
Annotations:  <none>

Type:  kubernetes.io/tls

Data
====
tls.crt:  3420 bytes
tls.key:  1704 bytes
``` 
ingress sample: 
```
#
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: "xxx"
  namespace: "xxx"
  annotations:
    ingress.kubernetes.io/rewrite-target: "/"
    nginx.org/redirect-to-https: "true"
    kubernetes.io/ingress.class: "nginx"
  labels:
    app: "xxx"
    track: "prod"
spec:
  tls:
    - hosts:
        - "xxx.default"
    - hosts:
        - "xxx"
      secretName: tls-wildcard-xxx
  rules:
    - host: "xxx.default"
      http:
        paths:
          - path: /
            backend:
              serviceName: xxx
              servicePort: xx
    - host: xxx
      http:
        paths:
          - backend:
              serviceName: xxx
              servicePort: xx
            path: /
```

Any ideas what might happened?

pleshakov:
Hi @kongdewen 

This seems like a timing issue in the IC - it processed Ingresses before it had processed the referenced Secrets.

How many Ingress and secrets resources are handled by the Ingress Controller in your cluster?

kongdewen:
Hey @pleshakov, for this particular ingress.class, it has 183 ingress, total tls secrets in the cluster is 46.

pleshakov:
@kongdewen 
[...]

PR #1654, opened 2021-06-09
Fix config and warnings about secrets at IC start

pleshakov:
### Proposed changes
This PR ensures that during the Ingress Controller start, Secret resources are processed before any of the resources that reference those Secrets. This prevents the IC from generating incorrect config and reporting warnings about missing Secrets.

Note that the incorrect config and warnings were temporary and only occurred at the Ingress Controller start before the IC pod was considered ready (passed the readiness probe).

Fixes https://github.com/nginxinc/kubernetes-ingress/issues/1448

### DIFF
--- a/internal/k8s/secrets/store.go
+++ b/internal/k8s/secrets/store.go
@@ -117,7 +117,7 @@ func NewFakeSecretsStore(secrets map[string]*SecretReference) *FakeSecretStore {
 	}
 }
 
-// NewFakeSecretsStore creates a new empty FakeSecretStore.
+// NewEmptyFakeSecretsStore creates a new empty FakeSecretStore.
 func NewEmptyFakeSecretsStore() *FakeSecretStore {
 	return &FakeSecretStore{
 		secrets: make(map[string]*SecretReference),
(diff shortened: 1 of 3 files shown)


soneillf5 (approved):
Looks good to me.

pleshakov:
rebased against master to fix the failing pipeline

internal/k8s/secrets/store.go
```diff
@@ -111,14 +111,28 @@ type FakeSecretStore struct {
...
 
 // NewFakeSecretsStore creates a new FakeSecretStore.
-func NewFakeSecretsStore(secrets map[string]*SecretReference) SecretStore {
+func NewFakeSecretsStore(secrets map[string]*SecretReference) *FakeSecretStore {
 	return &FakeSecretStore{
 		secrets: secrets,
 	}
 }
 
+// NewFakeSecretsStore creates a new empty FakeSecretStore.
```
ciarams87, line 120: nit: incorrect function name in the comment
```suggestion
// NewEmptyFakeSecretsStore creates a new empty FakeSecretStore.
```

ciarams87 (approved):

PR #1686, opened 2021-06-21
Use latest Docker image for golangci-lint

lucacome:
After some thought (and finding a fast way to use the docker container), I don't think we need to pin the version of golangci-lint, we can just use the latest version available, what do you think?


pleshakov:
> After some thought (and finding a fast way to use the docker container), I don't think we need to pin the version of golangci-lint, we can just use the latest version available, what do you think?

does it mean we will have a mismatch between the version here and here https://github.com/nginxinc/kubernetes-ingress/blob/master/.github/workflows/lint.yml#L34 ? can this create problems?  like the linting succeeds locally but fails on GitHub or vice-versa.

lucacome:
> does it mean we will have a mismatch between the version here and here https://github.com/nginxinc/kubernetes-ingress/blob/master/.github/workflows/lint.yml#L34 ?

@pleshakov no that action will use the latest binary available if it can't find a "pin" anywhere. Now it's using the version that we have in `go.mod`. With the change in this PR everything will use "latest".

pleshakov (approved):

soneillf5 (approved):

PR #1687, opened 2021-06-21
Enable default linters and gosec

lucacome:
- Added the rest of the default linters, seems like a good idea to enable them. https://golangci-lint.run/usage/linters/#enabled-by-default-linters
- Added `gosec` linter

(More to come in future PRs 😬)

### DIFF
--- a/internal/nginx/manager.go
+++ b/internal/nginx/manager.go
@@ -19,18 +19,24 @@ import (
 const (
 	ReloadForEndpointsUpdate     = true  // ReloadForEndpointsUpdate means that is caused by an endpoints update.
 	ReloadForOtherUpdate         = false // ReloadForOtherUpdate means that a reload is caused by an update for a resource(s) other than endpoints.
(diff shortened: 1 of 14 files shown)


lucacome:
@nginxinc/kic any thoughts on the remaining `gosec` errors?

soneillf5 (approved):

ciarams87 (approved):

internal/nginx/manager.go
```diff
@@ -270,7 +266,11 @@ func (lm *LocalManager) ClearAppProtectFolder(name string) {
 func (lm *LocalManager) Start(done chan error) {
 	glog.V(3).Info("Starting nginx")
 
-	cmd := exec.Command(lm.binaryFilename)
+	binaryFilename := nginxBinaryPath
```
pleshakov, line 269: this repeats 4 times in the code - can it be a function?

pleshakov (approved):

internal/nginx/manager.go
lucacome: makes sense 👍

PR #1688, opened 2021-06-21
Add errorlint

lucacome:
Adding errorlint https://github.com/polyfloyd/go-errorlint

> go-errorlint is a source code linter for Go software that can be used to find code that will cause problems with the error wrapping scheme introduced in Go 1.13.


soneillf5 (approved):
Great change, thanks!

ciarams87 (approved):

PR #1709, opened 2021-06-30
Add cli argument to optionally disable snippet annotations

soneillf5:
### Proposed changes
This PR adds a cli argument to optionally disable snippet annotations for Ingress Resources.

### Checklist


vepatel (approved):

internal/configs/annotations.go
```diff
@@ -66,7 +66,7 @@ var minionInheritanceList = map[string]bool{
 	"nginx.org/fail-timeout":             true,
 }
 
-func parseAnnotations(ingEx *IngressEx, baseCfgParams *ConfigParams, isPlus bool, hasAppProtect bool, enableInternalRoutes bool) ConfigParams {
+func parseAnnotations(ingEx *IngressEx, baseCfgParams *ConfigParams, isPlus bool, hasAppProtect bool, enableInternalRoutes bool, enableIngressSnippets bool) ConfigParams {
```
vepatel, line 69: should have unit tests for the same scenarios as well, like functional tests? It's be much quicker and will increase our existing coverage too, we should ideally have more unit tests than functional ones.

pleshakov:
Hi @soneillf5 

The code looks good to me. However, I wonder if the user experience can be improved here.

Imagine a situation when snippets are disabled and a user creates an Ingress with snippets assuming they are enabled. As a result, the IC will silently ignore the snippets without letting the user knowing about it.

We have the same problem with VirtualServers/VirtualServerRoutes: the IC silently ignores snippets, which causes confusion and could lead to problems. For example, users are confident that some NGINX features is configured via snippets, but in fact it doesn't configure :( 

However, when implementing the TransportServer, we addressed that problem and the IC rejects TransportServers with snippets when snippets are disabled:
```
Status:
  Message:  TransportServer default/secure-app was rejected with error: spec.serverSnippets: Forbidden: snippet specified but snippets feature is not enabled
  Reason:   Rejected
  State:    Invalid
Events:
  Type     Reason    Age              From                      Message
  ----     ------    ----             ----                      -------
  Warning  Rejected  4s (x2 over 4s)  nginx-ingress-controller  TransportServer default/secure-app was rejected with error: spec.serverSnippets: Forbidden: snippet specified but snippets feature is not enabled
```

I think if we implement the same behavior for Ingress snippets, that will lead to better user experience.  Note that implementing that is just a matter of adding validation functions for snippets annotation that fail when snippets are not enabled -- https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/k8s/validation.go#L117-L118 

Including @brianehlert here to provide guidance around the requirements 

Additionally, are you planning to document the new cli arg https://docs.nginx.com/nginx-ingress-controller/configuration/global-configuration/command-line-arguments/ ?  
Shall we also have a helm chart parameter, similar to `controller.enableSnippets` ? 

I also suggest we add a sentence here https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-snippets/#disadvantages-of-using-snippets  after the note:
You can disable snippets for Ingress resources via [-enable-ingress-snippets](#) cli argument.

brianehlert:
I like the example of what has been done for TransportServer.
We should do the same for VS/VSR.  That is a nice experience.
And yes, it needs to be complete across Helm, manifest, docs, etc.
Also, this should not be disabled by default until we have adequate deprecation notification  (reminder)

PR #1714, opened 2021-07-01
Increase timeout for golangci-lint and add more linters

lucacome:
Almost aligned with the other repos, the last two left are `revive` and `noctx` but they have a lot of issues.

The increased timeout should be useful mostly on the first run when the cache is still not present.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,7 @@ linters-settings:
 
 linters:
     enable:
+    - asciicheck
     - deadcode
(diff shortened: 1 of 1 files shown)


lucacome:
@nginxinc/kic If you think there are more linter that we should enable the list is here https://golangci-lint.run/usage/linters/

pleshakov (approved):

soneillf5 (approved):

PR #1720, opened 2021-07-08
Add GoReleaser config for AWS

lucacome:
- Added configuration for building the binary for AWS with GoReleaser
- Added labels in Dockerfile to identify how the image was built
- Improved error messages for AWS RegisterUsage
- Moved the print of the binary version at the start

### DIFF
--- a/.github/workflows/edge.yml
+++ b/.github/workflows/edge.yml
@@ -69,7 +69,7 @@ jobs:
         uses: goreleaser/goreleaser-action@v2
         with:
           version: latest
-          args: --rm-dist --debug --skip-publish --snapshot
+          args: build --snapshot --rm-dist --id kubernetes-ingress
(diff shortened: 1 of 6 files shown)


pleshakov (approved):
👍 

Would it make sense to add an enhancement label, since  `cmd/nginx-ingress/aws.go` and `cmd/nginx-ingress/main.go` include enhancements? In the future, it'd be better to have those in separate PRs or commits. cc @vepatel for awareness

lucacome:
@pleshakov I thought about moving them in separate commits and then forgot somehow, doing it now. 👍 


I didn't see it as a feature enhancement since it was just about changing log format or rearranging it (for main) but I'm fine with adding the label.

soneillf5 (approved):

PR #1732, opened 2021-07-13
Use Exponential Backoff to get correct NGINX version

lucacome:
Use [Exponential Backoff](https://github.com/cenkalti/backoff) when trying to get the correct version of NGINX.

I've set the initial interval at 25ms (like before) and the max interval to 500ms, but we can tweak those numbers.

### DIFF
--- a/internal/nginx/verify.go
+++ b/internal/nginx/verify.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/golang/glog"
 )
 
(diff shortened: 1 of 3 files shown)


internal/nginx/verify.go
```diff
@@ -61,24 +62,32 @@ func (c *verifyClient) GetConfigVersion() (int, error) {
 // WaitForCorrectVersion calls the config version endpoint until it gets the expectedVersion,
 // which ensures that a new worker process has been started for that config version.
 func (c *verifyClient) WaitForCorrectVersion(expectedVersion int) error {
-	interval := 25 * time.Millisecond
-	startTime := time.Now()
-	endTime := startTime.Add(c.timeout)
+	b := backoff.NewExponentialBackOff()
+	b.InitialInterval = 25 * time.Millisecond
+	b.MaxInterval = 500 * time.Millisecond
```
soneillf5, line 67: based on these number we have:
25, 38, 57, 86, 129, 194, 291, 436, 500, 500, 500... 

I'm not sure about pulling a third party library for this.

pleshakov: I'd agree with @soneillf5 about not pulling a third party library for that. Also note that we're not exposing the algorithm parameters to the admin  (interval, maxInterval) to the admin, so we can just use a pre-calculated slice.

Please find a quick prototype below so we can compare with and without the library:
```diff
diff --git a/internal/nginx/verify.go b/internal/nginx/verify.go
index 33033bbb..94b45b3f 100644
--- a/internal/nginx/verify.go
+++ b/internal/nginx/verify.go
@@ -58,12 +58,14 @@ func (c *verifyClient) GetConfigVersion() (int, error) {
        return v, nil
 }

+var waitPeriods = []int{25, 38, 57, 86, 129, 194, 291, 436, 500}
+
 // WaitForCorrectVersion calls the config version endpoint until it gets the expectedVersion,
 // which ensures that a new worker process has been started for that config version.
 func (c *verifyClient) WaitForCorrectVersion(expectedVersion int) error {
-       interval := 25 * time.Millisecond
        startTime := time.Now()
        endTime := startTime.Add(c.timeout)
+       periodIdx := 0

        glog.V(3).Infof("Starting poll for updated nginx config")
        for time.Now().Before(endTime) {
@@ -76,7 +78,13 @@ func (c *verifyClient) WaitForCorrectVersion(expectedVersion int) error {
                        glog.V(3).Infof("success, version %v ensured. took: %v", expectedVersion, time.Since(startTime))
                        return nil
                }
+
+               interval := time.Duration(waitPeriods[periodIdx]) * time.Millisecond
                time.Sleep(interval)
+
+               if periodIdx < len(waitPeriods) - 1 {
+                       periodIdx++
+               }
        }
        return fmt.Errorf("could not get expected version: %v after %v", expectedVersion, c.timeout)
 }
```

PR #1744, opened 2021-07-14
Create separate workflows for PRs and master in release-drafter

ciarams87:
### Proposed changes
We are running into this issue: https://github.com/release-drafter/release-drafter/issues/680 when we have "filter-by-commitish" and the PR "autolabeler" set. As far as I can tell, running the release-drafter on every PR creates a different draft release when we have "filter-by-commitish" set. This attempts to rectify by creating separate release drafter workflows so we can run the auto-labeller only on PRs. I think we can support releasing from release branches going forward by setting the 'commitish' to the name of the release branch - this should support patch releases.

### Checklist

### DIFF
--- a/.github/workflows/release-drafter-pr.yml
+++ b/.github/workflows/release-drafter-pr.yml
@@ -1,0 +1,15 @@
+name: Release Drafter
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  update_release_draft:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: release-drafter/release-drafter@v5
+        with:
+          disable-releaser: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


vepatel (approved):

pleshakov (approved):

.github/workflows/release-drafter-pr.yml
```diff
@@ -12,5 +9,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: release-drafter/release-drafter@v5
+        with:
+          disable-releaser: true
```
lucacome, line 13: Nice! How did you find it? It's not documented 😞

lucacome (approved):

.github/workflows/release-drafter-pr.yml
ciarams87: I know, I had to go read the code! https://github.com/release-drafter/release-drafter/blob/master/action.yml#L46

PR #1746, opened 2021-07-14
Add ingressClassName to Policy

soneillf5:
### Proposed changes
This change adds a filed `ingressClassName` to a Policy resource. The ingress class is used by the controller to filter out resources that it should not handle. If the ingress class is empty or matches the ingress class of the IngressController then the Policy resource is processed as normal, otherwise it is ignored with no warning or error.

### Checklist


tests/suite/test_policy_ingress_class.py
```diff
@@ -0,0 +1,162 @@
...
+
+vs_src = f"{TEST_DATA}/policy-ingress-class/virtual-server.yaml"
+vs_policy_src = f"{TEST_DATA}/policy-ingress-class/virtual-server-policy.yaml"
+
+policy_src = f"{TEST_DATA}/policy-ingress-class/policy.yaml"
+policy_ingress_class_src = f"{TEST_DATA}/policy-ingress-class/policy-ingress-class.yaml"
+policy_other_ingress_class_src = f"{TEST_DATA}/policy-ingress-class/policy-other-ingress-class.yaml"
+
+
+@pytest.mark.sean
```
pleshakov, line 25: did you remove your git hook? ;)

internal/k8s/controller.go
```diff
@@ -792,6 +792,11 @@ func (lbc *LoadBalancerController) syncPolicy(task task) {
 		return
 	}
 
+	if !lbc.HasCorrectIngressClass(obj) {
```
pleshakov, line 795: Unfortunately, we can't simply ignore a policy here. Consider a case if the policy IngressClassName is updated:
a) IngressClassName is updated to the one that doesn't match the class of the IC. In this case, the IC needs to regenerate the configuration for all affected VSes 
b) IngressClassName is updated to the one that matches the class of the IC. In this case, the IC needs also to regenerate the configuration for all affected VSes 


I think adding the IngressClass check in `getPolicies()` will solve the problem - if the Policy IngressClass doesn't match, getPolicies returns an error. As a result, the IC will generate the correct config (with `return 500`) and report the error in the VS/VSR status/events.

Note that we will still need to keep not reporting status/events for policies of a different IngressClass.

tests/suite/test_policy_ingress_class.py
```diff
@@ -0,0 +1,162 @@
...
+                and policy_info["status"]["state"] == "Valid"
+        )
+        assert occur.count(200) <= 1
+
+    @pytest.mark.parametrize("src", [vs_policy_src])
+    def test_policy_non_matching_ingress_class(
+            self, kube_apis, crd_ingress_controller, virtual_server_setup, test_namespace, src,
+    ):
+        """
+        Test if rate-limiting policy is working with 1 rps
```
pleshakov, line 142: it looks like this test case wasn't fully implemented.
the expected behavior here is for NGINX to return 500 and for the IC to report   
 Reason:   AddedOrUpdatedWithWarning
  State:    Warning
 and the message that the policy is missing or invalid 
 in the VS status and an event similar to this one
 ```
   Warning  AddedOrUpdatedWithWarning  55s   nginx-ingress-controller  Configuration for default/virtual-server was added or updated ; with warning(s): Policy default/rate-limit-primary is missing or invalid
```

File: internal/k8s/controller.go (hunk 2 above)
Comment (pleshakov) [inline internal/k8s/controller.go:795]: another corner case for the existing code:
```
kubectl apply -f policy-other-ingress-class.yaml
kubectl apply -f virtual-server-policy.yaml
kubectl describe vs
  Message:  Configuration for default/virtual-server was added or updated
  Reason:   AddedOrUpdated
  State:    Valid
. . .
Events:
  Type    Reason          Age   From                      Message
  ----    ------          ----  ----                      -------
  Normal  AddedOrUpdated  9s    nginx-ingress-controller  Configuration for default/virtual-server was added or updated
```
Expected:
```
  Message:  Configuration for default/virtual-server was added or updated ; with warning(s): Policy default/rate-limit-primary is missing or invalid
  Reason:   AddedOrUpdatedWithWarning
  State:    Warning
Events:
  Type     Reason                     Age   From                      Message
  ----     ------                     ----  ----                      -------
  Warning  AddedOrUpdatedWithWarning  5s    nginx-ingress-controller  Configuration for default/virtual-server was added or updated ; with warning(s): Policy default/rate-limit-primary is missing or invalid
  ```

pleshakov (changes requested):
Hi @soneillf5 

Please see my review. There is a bug and a few suggestions.

cmd/nginx-ingress/main.go
```diff
@@ -79,7 +79,7 @@ var (
 
 	For Kubernetes < 1.18, the Ingress Controller only processes resources that belong to its class -
 	i.e have the annotation "kubernetes.io/ingress.class" (for Ingress resources)
-	or field "ingressClassName" (for VirtualServer/VirtualServerRoute/TransportServer resources) equal to the class.
+	or field "ingressClassName" (for VirtualServer/VirtualServerRoute/TransportServer/Policy resources) equal to the class.
```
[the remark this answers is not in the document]
soneillf5: it's annoying to update docs randomly when adding a minor change to a resource. I've removed the reference to specific Resources instead, so it won't have to be updated again.

tests/suite/test_policy_ingress_class.py
soneillf5: must have!!! heck

internal/k8s/controller.go
soneillf5: updated

tests/suite/test_policy_ingress_class.py
soneillf5: updated

internal/k8s/controller.go
soneillf5: updated to use `getPolicies()`

pleshakov: > Note that we will still need to keep not reporting status/events for policies of a different IngressClass.


https://github.com/nginxinc/kubernetes-ingress/blob/f1ebca1e3524346c6840e9256be208f2638f525e/internal/k8s/controller.go#L797-L821

because an IC can report status for a Policy of a different IngressClass, this could cause problems. Consider an case of two ICs:
(1) with class `one` with enabled preview policies
(2) twi class `two` with disabled preview policies. 

If a user creates a preview policy with IngressClassName `one`, the (2) IC will still update the status of the Policy. If that happens after the (1) IC updated the status, the user will see a confusing error (the (2) IC will update the status with an error because preview policies are disabled) 

Could you make sure the IC only reports status for policies of its IngressClass?

cmd/nginx-ingress/main.go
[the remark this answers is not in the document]
pleshakov: oh no, we have one more place -- https://docs.nginx.com/nginx-ingress-controller/installation/running-multiple-ingress-controllers/ , I think it's important to keep the list there though. Could you update it?

ciarams87 (approved):

cmd/nginx-ingress/main.go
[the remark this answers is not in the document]
soneillf5: updated

internal/k8s/controller.go
soneillf5: updated

pleshakov: Hi @soneillf5  could you also disable reporting events for policies of a different class? sorry, I mentioned the events in my first comment, but didn't mention them in the second one.

cmd/nginx-ingress/main.go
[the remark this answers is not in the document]
pleshakov: > updated

Did you commit it? Can't find the change (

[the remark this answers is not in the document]
soneillf5: I removed the reference to the resources that I missed, it was mentioned in your last comment. https://github.com/nginxinc/kubernetes-ingress/pull/1746/files#diff-a296ef9c8b7e3b2b10ba91854e200adde65dbd7ed35fc1d57285c20525da71e7

I don't want the resources mentioned like this in the docs. If I change the resource, i.e. add policy, it's difficult to find where in the docs to change. Now that I've removed the references, it's safer and easier to add Ingress class to another Resource without making breaking changes to the docs. The Resource itself should document the field.

internal/k8s/controller.go
soneillf5: Can you clarify this? I was looking at the status updating logic and didn't see any other exclusion logic for TransportServer or VirtualServer. It's very painful to develop a feature via pull request.

cmd/nginx-ingress/main.go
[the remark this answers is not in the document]
pleshakov: those changes make sense - we had that information duplicated in many places for no reason and it was difficult to maintain 

however, we have a doc that describes the concept of IngressClass -- https://docs.nginx.com/nginx-ingress-controller/installation/running-multiple-ingress-controllers/ That doc basically connects everything else together. However, it doesn't describe that the IngressClass applies to the Policy resource.

internal/k8s/controller.go
pleshakov: Sure. VirtualServer and TransportServer are processed through the Configuration type.  The Configuration type checks for the resource class and if it doesn't match - it will ignore it -- https://github.com/nginxinc/kubernetes-ingress/blob/master/internal/k8s/configuration.go#L465  At the end, no change or problem for it will be generated so no status will be reported or event emitted.

In contrast, the Policy resource isn't processes through the Configuration type, so that's why we need to add that check.  If you condition this code https://github.com/nginxinc/kubernetes-ingress/blob/dfad4850b02cedc3b08deb72643677b475eee91d/internal/k8s/controller.go#L798-L822 so that it doesn't get invoked when the policy is of a different class, that will solve the problem of both reporting an event and the status for the policy of a different class.

soneillf5: > If you condition this code ... so that it doesn't get invoked when the policy is of a different class
I'm not following you again sorry, you mean add a check for the ingress class and if it doesn't match, don't invoke this code?

soneillf5: Out of interest, I've updated the tests to capture the change around the status. The test no confirms that if the ingress class does not match, the status of the policy is not updated. `tests/suite/test_policy_ingress_class.py:143`

pleshakov: > if polExists && !hasCorrectIngressClass(obj) {

👍 

>  The test no confirms that if the ingress class does not match, the status of the policy is not updated.

what about the event for the Policy resource?

vepatel (approved):

tests/suite/test_policy_ingress_class.py
```diff
@@ -0,0 +1,164 @@
...
+
+        delete_policy(kube_apis.custom_objects, pol_name, test_namespace)
+        self.restore_default_vs(kube_apis, virtual_server_setup)
+
+    @pytest.mark.parametrize("src", [vs_policy_src])
+    def test_policy_non_matching_ingress_class(
+            self, kube_apis, crd_ingress_controller, virtual_server_setup, test_namespace, src,
+    ):
+        """
+        Test if non matching policy gets caught by vc validation
```
vepatel, line 135: vc?

tests/suite/test_policy_ingress_class.py
```diff
@@ -0,0 +1,164 @@
...
+                vs_info["status"]
+                and vs_info["status"]["reason"] == "AddedOrUpdated"
+                and vs_info["status"]["state"] == "Valid"
+        )
+
+        delete_policy(kube_apis.custom_objects, pol_name, test_namespace)
+        self.restore_default_vs(kube_apis, virtual_server_setup)
+
+    @pytest.mark.parametrize("src", [vs_policy_src])
+    def test_policy_non_matching_ingress_class(
```
vepatel, line 131: what do you think of a negative case where VS `ingressClass` is `other` policy `ingressClass` is `nginx`, like VS won't be managed by IC but policy will?

internal/k8s/controller.go
soneillf5: sorted

pleshakov (approved):

PR #1747, opened 2021-07-14
Refactor controller.go

soneillf5:
### Proposed changes
To reduce the size of the controller I've made the following changes:

1. Group all resources api calls into one package, called `apis`. If you need to access the k8s API to get a resource, it should be in this package. The files are organised by resource type.
2. Group all resource calls into one package, resources. If you need to need to get some resources but it depends on some state or logic, use the resources package, i.e. `GetHealthChecksForIngressBackend(backend *networking.IngressBackend, namespace string)`
3. Reduce the size of the listers interface from cache.Store to client specific functions under `apis` ([solid](https://en.wikipedia.org/wiki/SOLID) specifically [Interface segregation](https://en.wikipedia.org/wiki/Interface_segregation_principle))

Moving the resource logic into resources allowed us to make several of the functions private, reducing cognitive load.

I have not made any logical changes ( I sincerely hope! ) just moved things around. 

### Checklist


soneillf5:
https://github.com/nginxinc/kubernetes-ingress/pull/1747 moved to public repo

PR #1748, opened 2021-07-14
Allow ingress to use a PSP

wilsonge:
### Proposed changes
Allows the cluster role to use a non-default pod security policy if they are enabled in the repository. Currently just allows using an existing pod security policy (`existingPsp`). But by having a separate `enabled` flag we setup the helm chart params in such a way that we can include a home baked pod security policy in the future if desired with minimal permissions required.

### Checklist
I have read all the items below - as far as I can tell there are no tests for the helm chart (and I'm not modifying anything else) but feel free to point me in the right direction if I've missed some!

### DIFF
--- a/deployments/helm-chart/README.md
+++ b/deployments/helm-chart/README.md
@@ -220,6 +220,8 @@ Parameter | Description | Default
 `controller.readyStatus.port` | The HTTP port for the readiness endpoint. | 8081
 `controller.enableLatencyMetrics` |  Enable collection of latency metrics for upstreams. Requires `prometheus.create`. | false
 `rbac.create` | Configures RBAC. | true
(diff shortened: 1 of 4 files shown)


soneillf5:
Hi @wilsonge 

Thanks for taking the time to create this MR, we appreciate the effort. 

Looking at the changes themselves, it seems good. However, the feature itself is now deprecated https://kubernetes.io/docs/concepts/policy/pod-security-policy/

Based on this, I'm afraid we won't be approving this change.

wilsonge:
@soneillf5 I'm well aware that PSP's are deprecated - but the successor project https://github.com/kubernetes/enhancements/pull/2582/files will be behind a flag and alpha level in the the still unreleased kubernetes 1.22 - this means that there's still no practical alternatives to PSP's yet for relatively simplistic use cases (as acknowledged in that blog post that you linked to)

PR #1766, opened 2021-07-19
refactor controller.go

soneillf5:
moved k8s api access to api package. moved handler setup to handler file.

### Proposed changes

To reduce the size of the controller I've made the following changes:

* Group all k8s api calls into one package, called apis. If you need to access the k8s API to get a resource, it should be in this package. The files are organised by resource type.
* Group all calls to pods, endpoints and services into one package, upstreams. If you need to need to get some of these resources but it depends on some state or logic, use the upstreams package, i.e. GetHealthChecksForIngressBackend(backend *networking.IngressBackend, namespace string)
* Reduce the size of the listers interface from cache.Store to client specific functions under apis (solid specifically Interface segregation)
* Moving the upstream logic into resources allowed us to make several of the functions private, reducing cognitive load.
* Moved the handlers setup code into handlers file.

### Checklist


internal/k8s/api/api.go
```diff
@@ -0,0 +1,23 @@
+package api
+
+type Apis struct {
```
pleshakov, line 3: could you add the comments to the new exported types and their exported methods? this will follow the code style

internal/k8s/status.go
```diff
@@ -638,12 +639,11 @@ func (su *statusUpdater) UpdatePolicyStatus(pol *v1.Policy, state string, reason
...
 	}
 
-	polCopy := polLatest.(*v1.Policy)
-
-	if !hasPolicyStatusChanged(polCopy, state, reason, message) {
+	if !hasPolicyStatusChanged(polLatest, state, reason, message) {
 		return nil
 	}
 
+	polCopy := polLatest.DeepCopy()
```
pleshakov, line 646: this looks like a bug fix. would it make sense to put it separately?

internal/k8s/upstreams/upstreams_test.go
```diff
@@ -0,0 +1,397 @@
...
+				},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			actualType, actualName := getPodOwnerTypeAndName(test.pod)
+			if actualType != test.expType {
+				t.Errorf("getPodOwnerTypeAndName() returned %s for owner type but expected %s", actualType, test.expType)
```
pleshakov, line 163: neat picking: the new methods and functions are exported, but they are not exported (start with a lowercase letter) in error messages in the tests

pleshakov (commented):
looks good! just some style and neat picking suggestions

internal/k8s/api/api.go
soneillf5: I'll put a comment or two where I thinks it's needed, if that's okay. I removed quite a few comments like:
```
\\ GetPods gets the pods
fun GetPods() {
```

soneillf5: Some methods are self documenting, i.e.
```
func NewApis() *Apis {
	return &Apis{}
}
```

soneillf5: updated.

internal/k8s/status.go
soneillf5: I can fix it now as part of this, or I can log a bug, put it on the backlog, discuss it in sprint planning + backlog review, create a PR, wait a few days and merge it.

internal/k8s/upstreams/upstreams_test.go
soneillf5: I don't follow you I'm afraid, can you explain the issue?

lucacome:
Looks like there's a panic? https://github.com/nginxinc/kubernetes-ingress/pull/1766/checks?check_run_id=3113018039#step:11:61

internal/k8s/status.go
pleshakov: Well, I suggest looking at it from the reviewer perspective. As a submitter you create a refactor PR, along the way there is an unrelated code change. For the reviewer, it appears as a bug and  brings different questions: what is that change about? why is it part of the refactor? is it a bug? Is anyone affected? Do we need to mention it in the release notes. That change came out of the blue.

in any case, when making a change like that, better to do it separately - a PR or at least a commit. and include its own rationale. Such changes can also introduce bugs, so this way it will be easier to track it down.

internal/k8s/upstreams/upstreams_test.go
pleshakov: my bad! I thought `getPodOwnerTypeAndName()` was `GetPodOwnerTypeAndName()` for some reason :(

internal/k8s/api/api.go
pleshakov: > As the comment for this would just be the signature of the method in English.

some will probably do
the existing code across the packages follows that convention - exported types with their exported methods are documented. the new code - doesn't, so it's inconsistent

soneillf5: If consistency is an issue, I'm more than happy to remove the redundant comments in the rest of the codebase.

internal/k8s/status.go
soneillf5: grand so, removed.

soneillf5:
@lucacome I took a look at the panics, there was an issue around the order of when the 'api' struct gets set so the pointer was nil, hence the panic.

It showed there was a more fundamental issue around the dependencies and the order of the initialisation. I've split out the api reference and passed it around as needed, to status updater and the handlers. 

As Michael suggested in another PR, there's scope here for refactoring the construction of the Api object.

ciarams87 (approved):
Looks good! I'm rerunning the gitlab pipeline now as well, the tests are passing here though so we should be good.

internal/k8s/api/api.go
lucacome: I agree that they should have a comment, it's a common best practice https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

tomasohaodha:
Genuine question - did the reviewers go through each and every of the 1,702 lines added and 1,334 lines removed? I see that reviewers were already approving the PR in principle a day after it was submitted, which seems very fast indeed given the extent of the changes.

Given that this is a refactoring, would it help if it was split up into smaller reviews, each covering one particular area being refactored? This would also help people more new to the code use the review as a learning aid, which is also a goal of code reviews.

Please feel free to put me right by saying that everyone else on this review is so intimately familiar with the code that they were able to review 3000 line changes (all mixed up in the same area) without breaking a sweat. But I would ask that if that is not true then we be transparent about it and see if we can approach this from a different direction. Smaller reviews done thoroughly are better imho that large reviews done more shallowly.

soneillf5:
Closing for now, as will need to rebase on master and redo some of the work.

PR #1780, opened 2021-07-26
App Protect Dos

pasmant:
Adding App Protect Dos inside KIC

Whats new:
RBAC: new yaml file for app-protect-dos - apdos-rbac.yaml
CRDs: new crds - appprotectdos.f5.com_apdoslogconfs.yaml + appprotectdos.f5.com_apdospolicies.yaml

Annotations: app-protect-dos-enable, app-protect-dos-monitor, app-protect-dos-name, app-protect-dos-policy, app-protect-dos-security-log, app-protect-dos-security-log-destination, app-protect-dos-security-log-enable
Policy: new policy - bados: enable, name , apDosMonitor, apDosPolicy, dossecurityLog.enable, dossecurityLog.apLogConf, dossecurityLog.logDest.

Makefile + Dockerfile:
Add options for app-protect-dos stand alone or app-protect+app-protect-dos
Add arbitrator option

Installations:
helm+Manifest: add option for app-protect-dos


internal/configs/version1/nginx-plus.tmpl
```diff
@@ -70,6 +79,13 @@ http {
     {{if .AppProtectPhysicalMemoryThresholds}}app_protect_physical_memory_util_thresholds {{.AppProtectPhysicalMemoryThresholds}};{{end}}
     include /etc/nginx/waf/nac-usersigs/index.conf;
     {{- end}}
+
+    {{- if .AppProtectDosLoadModule}}
+        {{- if .AppProtectDosLivenessEnable}}
+    app_protect_dos_liveness {{.AppProtectDosLivenessEnable}} uri:/{{.AppProtectDosLivenessURI}} port:{{.AppProtectDosLivenessPort}};
```
soneillf5, line 85: Are these values (`AppProtectDosLivenessURI` and `AppProtectDosLivenessPort`) optional or required? 
If they are required, we can help the user by validating they are not empty when parsing the config map in `configmaps.go`

deployments/helm-chart/crds/appprotectdos.f5.com_apdoslogconfs.yaml
```diff
@@ -0,0 +1,70 @@
...
+  group: appprotectdos.f5.com
+  names:
+    kind: APDosLogConf
+    listKind: APDosLogConfList
+    plural: apdoslogconfs
+    singular: apdoslogconf
+  preserveUnknownFields: false
+  scope: Namespaced
+  versions:
+    - name: v1beta1
```
soneillf5, line 18: is this v1 now with k8s v1.22 ? https://kubernetes.io/docs/reference/kubernetes-api/extend-resources/custom-resource-definition-v1/

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -50,7 +50,11 @@ This document is the reference documentation for the resources. To see additiona
 
 ## VirtualServer Specification
 
-The VirtualServer resource defines load balancing configuration for a domain name, such as `example.com`. Below is an example of such configuration:
+The VirtualServer resource defines load ba
```
soneillf5, line 53: minor type here

deployments/helm-chart/README.md
```diff
@@ -215,6 +243,10 @@ Parameter | Description | Default
 `controller.reportIngressStatus.annotations` | The annotations of the leader election configmap. | {}
 `controller.pod.annotations` | The annotations of the Ingress Controller pod. | {}
 `controller.appprotect.enable` | Enables the App Protect module in the Ingress Controller. | false
+`controller.appprotectdos.enable` | Enables the App Protect Dos module in the Ingress Controller. | false
+`controller.appprotectdos.debug` | Enables debug logs for App Protect Dos | false
+`controller.appprotectdos.arbitrator.name` | The name of the App Protect Dos Arbitrator deployment. | Autogenerated
```
lucacome, line 248: what does `Autogenerated` mean? if it's autogenerated why do we need a parameter?

docs/content/troubleshooting/troubleshooting-with-app-protect-dos.md
```diff
@@ -0,0 +1,102 @@
...
+
+When using the Ingress Controller with the App Protect Dos module, the following issues have been reported. The occurrence of these issues is commonly related to a higher number of Ingress Resources with App Protect Dos being enabled in a cluster.
+
+When you make a change that requires NGINX to apply a new configuration, the Ingress Controller reloads NGINX automatically. Without the App Protect module enabled, usual reload times are around 150ms. If App Protect Dos module is enabled and is being used by any number of Ingress Resources, these reloads might take a few seconds instead.
+
+### NGINX Configuration Skew
+
+If you are running more than one instance of the Ingress Controller, the extended reload time may cause the NGINX configuration of your instances to be out of sync. This can occur because there is no order imposed on how the Ingress Controller processes the Kubernetes Resources. The configurations will be the same after all instances have completed the reload.
+
+In order to reduce these inconsistencies, we advise that you do not apply changes to multiple resources handled by the Ingress Controller at the same time.
```
lucacome, line 102: Is this true for DOS? Or is it just for NAP+DOS?

internal/configs/annotations.go
```diff
@@ -16,6 +16,15 @@ const AppProtectLogConfAnnotation = "appprotect.f5.com/app-protect-security-log"
...
 const AppProtectLogConfDstAnnotation = "appprotect.f5.com/app-protect-security-log-destination"
 
+// AppProtectDosPolicyAnnotation is where the NGINX App Protect Dos policy is specified
+const AppProtectDosPolicyAnnotation = "appprotectdos.f5.com/app-protect-dos-policy"
+
+// AppProtectDosLogConfAnnotation is where the NGINX AppProtect Dos Log Configuration is specified
+const AppProtectDosLogConfAnnotation = "appprotectdos.f5.com/app-protect-dos-security-log"
+
+// AppProtectDosLogConfDstAnnotation is where the NGINX AppProtect Dos Log Configuration is specified
+const AppProtectDosLogConfDstAnnotation = "appprotectdos.f5.com/app-protect-dos-security-log-destination"
```
lucacome, line 26: these two have the same comment 🤔

internal/k8s/appprotectdos/app_protect_dos_resources.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+var (
+	logDstEx     = regexp.MustCompile(`(?:syslog:server=((?:\d{1,3}\.){3}\d{1,3}|localhost):\d{1,5})|stderr|(?:\/[\S]+)+`)
+	logDstFileEx = regexp.MustCompile(`(?:\/[\S]+)+`)
+)
+
+// ValidateAppProtectDosLogDestination validates destination for log configuration
+func ValidateAppProtectDosLogDestination(dstAntn string) error {
+	errormsg := "Error parsing App Protect Dos Log config: Destination must follow format: syslog:server=<ip-address | localhost>:<port> or stderr or absolute path to file"
```
lucacome, line 70: > or absolute path to file

what's the use case for saving log to file in a container?

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -2,14 +2,16 @@
 worker_processes  {{.WorkerProcesses}};
 {{- if .WorkerRlimitNofile}}
 worker_rlimit_nofile {{.WorkerRlimitNofile}};{{end}}
+{{- if .WorkerRlimitCore}}
```
lucacome, line 5: why is this needed?

internal/configs/annotations.go
```diff
@@ -372,6 +389,43 @@ func parseAnnotations(ingEx *IngressEx, baseCfgParams *ConfigParams, isPlus bool
 			}
 		}
 
+	}
+	if hasAppProtectDos {
+		if appProtectDosEnable, exists, err := GetMapKeyAsBool(ingEx.Ingress.Annotations, "appprotectdos.f5.com/app-protect-dos-enable", ingEx.Ingress); exists {
```
soneillf5, line 394: should we replace these string values with the const, i.e. `AppProtectDosPolicyAnnotation` ?

internal/configs/configmaps.go
```diff
@@ -501,6 +509,33 @@ func ParseConfigMap(cfgm *v1.ConfigMap, nginxPlus bool, hasAppProtect bool) *Con
...
 	}
 
+	if hasAppProtectDos {
+
+		appProtectDosLivenessEnable, livenessExistsEnable, errEnable := GetMapKeyAsBool(cfgm.Data, "app-protect-dos-liveness-enable", cfgm)
+		appProtectDosLivenessURI, livenessExistsURI := cfgm.Data["app-protect-dos-liveness-uri"]
+		appProtectDosLivenessPort, livenessExistsPort, errPort := GetMapKeyAsInt(cfgm.Data, "app-protect-dos-liveness-port", cfgm)
+
+		if livenessExistsEnable {
+			if errEnable != nil {
```
soneillf5, line 519: I'm not 100% about this logic, I think it's worth checking it works. I normally see the following pattern in Golang:
```
value, err := getValue()
if err != nil {
   ... handle error
}
...use value
```

What happens in this case if there's an `errEnable`, will livenessExistsEnable ever be true in which case the error won't be logged and found ?

internal/configs/configmaps.go
```diff
@@ -501,6 +509,33 @@ func ParseConfigMap(cfgm *v1.ConfigMap, nginxPlus bool, hasAppProtect bool) *Con
...
+				glog.Error(errEnable)
+			} else if errPort != nil {
+				glog.Error(errPort)
+			} else {
+				if appProtectDosLivenessEnable {
+					cfgParams.MainAppProtectDosLivenessEnable = "on"
```
soneillf5, line 525: I'm not sure about use a variable to mean two different things. It appears in places where variables are set to 'on' or 'off'.

It's probably not a good thing as the variable has two different meanings, one for enabling/disabling the value and one for the value itself 'on/off'. 

Can we simplify the logic and and set the default value to 'off' ? Then, if `appProtectDosLivenessEnable` is true, we can set it to 'on' ?

internal/configs/configurator.go
```diff
@@ -1484,6 +1518,95 @@ func (cnf *Configurator) RefreshAppProtectUserSigs(
...
+	return fmt.Sprintf("%s%s_%s.json", appProtectDosPolicyFolder, unst.GetNamespace(), unst.GetName())
+}
+
+func appProtectDosLogConfFileNameFromUnstruct(unst *unstructured.Unstructured) string {
+	return fmt.Sprintf("%s%s_%s.json", appProtectDosLogConfFolder, unst.GetNamespace(), unst.GetName())
+}
+
+// DeleteAppProtectDosPolicy updates Ingresses and VirtualServers that use AP Dos Policy after that policy is deleted
+func (cnf *Configurator) DeleteAppProtectDosPolicy(polNamespaceName string, ingExes []*IngressEx, mergeableIngresses []*MergeableIngresses, vsExes []*VirtualServerEx) (Warnings, error) {
+	if len(ingExes)+len(mergeableIngresses)+len(vsExes) > 0 {
```
soneillf5, line 1531: Would it be a good idea to create a function that generates the name of the file? We can use this new function in both places we need the file name, where the file is created and where it is deleted. In this way, they will always match?

internal/configs/virtualserver.go
```diff
@@ -988,6 +992,72 @@ func (p *policiesCfg) addWAFConfig(
...
 }
 
+func (p *policiesCfg) addBadosConfig(
+	Bados *conf_v1.Bados,
+	polKey string,
+	polNamespace string,
+	apResources map[string]string,
+) *validationResults {
+	res := newValidationResults()
+	if p.Bados != nil {
```
soneillf5, line 1002: I might be wrong but it looks like `p.Bados` is never reset to nil, so once it's set, it can never be changed?

It also looks like `addBadsConfig()` is called frequently, as it is downstream from `GenerateVirtualServerConfig`. Does that mean it will generate a warning every time a VirtualServer is added/changed ?

internal/configs/virtualserver.go
```diff
@@ -988,6 +992,72 @@ func (p *policiesCfg) addWAFConfig(
...
+		return res
+	}
+
+	if Bados.Enable {
+		p.Bados = &version2.Bados{Enable: "on"}
+	} else {
+		p.Bados = &version2.Bados{Enable: "off"}
+	}
+
+	if Bados.Name != "" {
```
soneillf5, line 1013: Why do we check this? isn't the default value of a string "" ?

internal/configs/virtualserver.go
```diff
@@ -988,6 +992,72 @@ func (p *policiesCfg) addWAFConfig(
...
+	apResources map[string]string,
+) *validationResults {
+	res := newValidationResults()
+	if p.Bados != nil {
+		res.addWarningf("Multiple Bados policies in the same context is not valid. Bados policy %s will be ignored", polKey)
+		return res
+	}
+
+	if Bados.Enable {
+		p.Bados = &version2.Bados{Enable: "on"}
```
soneillf5, line 1008: Why not use a bool?

internal/configs/virtualserver.go
```diff
@@ -988,6 +992,72 @@ func (p *policiesCfg) addWAFConfig(
...
+		}
+
+		if logConfPath, ok := apResources[logConfKey]; ok {
+			logDest := generateString(Bados.DosSecurityLog.DosLogDest, "stderr")
+			p.Bados.ApDosLogConf = fmt.Sprintf("%s %s", logConfPath, logDest)
+		} else {
+			res.addWarningf("Bados policy %s references an invalid or non-existing log config %s", polKey, logConfKey)
+			res.isError = true
+		}
+		if Bados.DosSecurityLog.DosAccessLogDest != "" {
```
soneillf5, line 1049: another default value string

internal/configs/virtualserver.go
```diff
@@ -988,6 +992,72 @@ func (p *policiesCfg) addWAFConfig(
...
+			p.Bados.ApDosAccessLogDest = Bados.DosSecurityLog.DosAccessLogDest
+		}
+	}
+
+	if Bados.ApDosMonitor != "" {
```
soneillf5, line 1054: another default value string. Am I missing somewhere with default values ?

internal/k8s/controller.go
```diff
@@ -1239,6 +1277,86 @@ func (lbc *LoadBalancerController) processAppProtectProblems(problems []appprote
...
 }
 
+func (lbc *LoadBalancerController) processAppProtectDosChanges(changes []appprotectdos.Change) {
+	glog.V(3).Infof("Processing %v App Protect Dos changes", len(changes))
+
+	for _, c := range changes {
+		if c.Op == appprotectdos.AddOrUpdate {
+			switch impl := c.Resource.(type) {
+			case *appprotectdos.DosPolicyEx:
+				namespace := impl.Obj.GetNamespace()
```
soneillf5, line 1287: I think we can reduce the duplication in this function, particularly 

```
namespace := impl.Obj.GetNamespace()
 				name := impl.Obj.GetName()
 				resources := lbc.configuration.FindResourcesForAppProtectDosLogConfAnnotation(namespace, name)

 				for _, badosPol := range getBadosPoliciesForAppProtectLogConf(lbc.getAllPolicies(), namespace+"/"+name) {
 					resources = append(resources, lbc.configuration.FindResourcesForPolicy(badosPol.Namespace, badosPol.Name)...)
 				}

 				resourceExes := lbc.createExtendedResources(resources)
```

internal/k8s/controller.go
```diff
@@ -2099,6 +2217,27 @@ func (lbc *LoadBalancerController) createIngressEx(ing *networking.Ingress, vali
 				}
 			}
 		}
+
+		if lbc.appProtectDosEnabled {
+			if apDosPolicyAntn, exists := ingEx.Ingress.Annotations[configs.AppProtectDosPolicyAnnotation]; exists {
+				policy, err := lbc.getAppProtectDosPolicy(ing)
+				if err != nil {
+					glog.Warningf("Error Getting App Protect Dos policy %v for Ingress %v/%v: %v", apDosPolicyAntn, ing.Namespace, ing.Name, err)
```
soneillf5, line 2225: Any idea why this is a warning and not an error?

internal/k8s/controller.go
```diff
@@ -2649,6 +2840,46 @@ func (lbc *LoadBalancerController) addWAFPolicyRefs(
...
+
+		if pol.Spec.Bados.DosSecurityLog != nil && pol.Spec.Bados.DosSecurityLog.ApDosLogConf != "" {
+			logConfKey := pol.Spec.Bados.DosSecurityLog.ApDosLogConf
+			if !strings.Contains(pol.Spec.Bados.DosSecurityLog.ApDosLogConf, "/") {
+				logConfKey = fmt.Sprintf("%v/%v", pol.Namespace, logConfKey)
+			}
+
+			logConf, err := lbc.appProtectDosConfiguration.GetAppDosResource(appprotectdos.DosLogConfGVK.Kind, logConfKey)
+			if err != nil {
+				return fmt.Errorf("Bados policy %q is invalid: %w", logConfKey, err)
```
soneillf5, line 2874: should this be `%q` ?

deployments/helm-chart/README.md
pasmant: I took it from nginx-ingress name:

controller:
  ## The name of the Ingress controller daemonset or deployment.
  # name: nginx-ingress

this parameter change the autogenerated arbitrator pod name to the name that given.
arbitrator:
  name: appprotect-dos-arb
Pod name will be:
from:
my-release-nginx-ingress-appprotect-dos-arb-86f5c89894-rsqxh
to:
appprotect-dos-arb-598f6647b8-gxzzh

build/Dockerfile
```diff
@@ -79,6 +79,42 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
...
+RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode=0644 \
+    --mount=type=secret,id=nginx-repo.key,dst=/etc/ssl/nginx/nginx-repo.key,mode=0644 \
+    set -x \
+    && apt-get update \
+    && apt-get -y install ca-certificates \
+    && printf "deb https://pkgs.nginx.com/app-protect-dos/debian buster nginx-plus\n" | tee /etc/apt/sources.list.d/nginx-app-protect-dos.list \
+    && apt-get update \
+    && apt-get clean all \
+    && apt-get -o DPkg::Options::="--force-confold" -y install nginx-plus-module-appprotectdos \
+    && apt-get -o DPkg::Options::="--force-confold" -y install app-protect-dos
```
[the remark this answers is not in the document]
pasmant: Change the docker file without --force-confold

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
pasmant: Fix typo

build/app-protect-dos/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode=0644 \
+    --mount=type=secret,id=nginx-repo.key,dst=/etc/ssl/nginx/nginx-repo.key,mode=0644 \
+    wget --certificate=/etc/ssl/nginx/nginx-repo.crt --private-key=/etc/ssl/nginx/nginx-repo.key https://pkgs.nginx.com/app-protect-dos/app-protect-dos-arb
+
+# Remove nginx repository key/cert from docker
+RUN rm -rf /etc/ssl/nginx
+
+
+RUN chmod 755 /app/app-protect-dos-arb
+CMD ["/app/app-protect-dos-arb"]
```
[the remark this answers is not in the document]
pasmant: No specific case to build it, the Dockerfile was added because before the arbitrator was in the registry.

Remove the Dcokergfile

docs/content/troubleshooting/troubleshooting-with-app-protect-dos.md
pasmant: Copy paste issue not true for DOS, will be remove

internal/configs/annotations.go
pasmant: Will fix it (also will be fix in nap)

internal/k8s/appprotectdos/app_protect_dos_resources.go
pasmant: Those are all the option that can be in the directive of log destination.

I just copy that from nap:
// ValidateAppProtectLogDestination validates destination for log configuration
func ValidateAppProtectLogDestination(dstAntn string) error {
	errormsg := "Error parsing App Protect Log config: Destination must follow format: syslog:server=<ip-address | localhost>:<port> or stderr or absolute path to file"

So I thought to make the same error msg.

internal/configs/annotations.go
pasmant: Its just copy past from nap, I can change it both for nap and dos.

If u approve will make the changes for both

cmd/nginx-ingress/main.go
```diff
@@ -70,6 +70,10 @@ var (
 
 	appProtect = flag.Bool("enable-app-protect", false, "Enable support for NGINX App Protect. Requires -nginx-plus.")
 
+	appProtectDos = flag.Bool("enable-app-protect-dos", false, "Enable support for NGINX App Protect dos. Requires -nginx-plus.")
+
+	appProtectDosDebug = flag.Bool("app-protect-dos-debug", false, "Enable debugging for App Protect Dos. Uses the log debug flag")
```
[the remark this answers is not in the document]
pasmant: This will help to find bugs in admd module that run with nginx that not in debug mode.

internal/configs/version1/nginx.tmpl
```diff
@@ -2,13 +2,15 @@
...
+{{- if .WorkerRlimitCore}}
+worker_rlimit_core {{.WorkerRlimitCore}};{{end}}
 {{- if .WorkerCPUAffinity}}
 worker_cpu_affinity {{.WorkerCPUAffinity}};{{end}}
 {{- if .WorkerShutdownTimeout}}
 worker_shutdown_timeout {{.WorkerShutdownTimeout}};{{end}}
 daemon off;
 
-error_log  stderr {{.ErrorLogLevel}};
+error_log  {{.ErrorLogLevel}} {{.ErrorLogLevel}};
```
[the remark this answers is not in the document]
pasmant: ErrorLogLevel need to be ErrorLogDst - typo.
This option is for the automation tests that read the file inside the container.

deployments/helm-chart/crds/appprotectdos.f5.com_apdospolicy.yaml
```diff
@@ -0,0 +1,66 @@
...
+            kind:
+              description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+              type: string
+            metadata:
+              type: object
+            spec:
+              type: object
+              description: APDosPolicySpec defines the desired state of APDosPolicy
+              properties:
+                mitigation_mode:
```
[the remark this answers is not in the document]
pasmant: Took the style from the nap
example file: appprotect.f5.com_aplogconfs.yaml
its like APLogConfSpec where there are args: format_string, max_message_size ....

Do we wants to change all of them?

internal/configs/version1/nginx-plus.tmpl
pasmant: No needed used for internal, will removed.

internal/configs/annotations.go
pasmant: Copy from nap:
if hasAppProtect {
		if appProtectEnable, exists, err := GetMapKeyAsBool(ingEx.Ingress.Annotations, "appprotect.f5.com/app-protect-enable", ingEx.Ingress); exists {

Do we want to make changes both in nap + napdos?

internal/configs/configmaps.go
pasmant: Change logic, handle first err and then the usecase

internal/configs/configmaps.go
pasmant: Make enable default value "off", and change the logic.

cmd/nginx-ingress/main.go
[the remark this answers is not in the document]
pasmant: We want to be able the option of debug log only for app protect dos process (admd) while nginx not in debug mode.

internal/nginx/manager.go
```diff
@@ -512,6 +519,42 @@ func (lm *LocalManager) AppProtectPluginQuit() {
...
+	// Perform installtion by adminstall
+	cmdInstall := exec.Command(appProtectDosAgentInstallCmd)
+
+	if err := cmdInstall.Run(); err != nil {
+		glog.Fatalf("Failed to install AppProtectDos: %v", err)
+	}
+
+	// case debug add debug flag to admd
+	appProtectDosAgentCmd := appProtectDosAgentStartCmd
+	if debug {
```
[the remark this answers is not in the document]
pasmant: I want to make debug option when the debug flag is true, why you think need to check the log level?

docs/content/app-protect-dos/installation.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. [Configure role-based access control (RBAC)](/nginx-ingress-controller/installation/installation-with-manifests/#configure-rbac).
+
+   > **Important**: You must have an admin role to configure RBAC in your Kubernetes cluster.
+
+2. [Create the common Kubernetes resources](/nginx-ingress-controller/installation/installation-with-manifests/#create-common-resources).
+3. Enable the App Protect Dos module by adding the `enable-app-protect-dos` [cli argument](/nginx-ingress-controller/configuration/global-configuration/command-line-arguments/#cmdoption-enable-app-protect-dos) to your Deployment or DaemonSet file.
+5. [Deploy the Ingress Controller](/nginx-ingress-controller/installation/installation-with-manifests/#deploy-the-ingress-controller).
+
+For more information, see the [Configuration guide](/nginx-ingress-controller/app-protect-dos/configuration) and the [NGINX Ingress Controller with App Protect Dos examples on GitHub](https://github.com/nginxinc/kubernetes-ingress/tree/v1.11.0/examples/appprotect-dos).
```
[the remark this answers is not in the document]
pasmant: will be changed

internal/configs/annotations.go
soneillf5: yeah that would be great, thanks.

internal/nginx/manager.go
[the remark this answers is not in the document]
lucacome: I don't think we need a separate debug flag for DOS in the Ingress Controller

internal/configs/version1/nginx.tmpl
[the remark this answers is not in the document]
lucacome: Maybe we should change the tests?

deployments/helm-chart/crds/appprotectdos.f5.com_apdospolicy.yaml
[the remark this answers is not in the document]
lucacome: I think we should pick just one for Dos and stick to it, doesn't matter If App protect is not doing it right

internal/k8s/appprotectdos/app_protect_dos_resources.go
lucacome: I don't think it makes sense to save logs inside the container, but I'm not familiar with NAP/DOS use cases

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -73,7 +73,28 @@ server {
...
+	app_protect_dos_security_log_enable {{$server.AppProtectDosLogEnable}};
+	{{if $server.AppProtectDosLogConf}}app_protect_dos_security_log {{$server.AppProtectDosLogConf}};{{end}}
+	    {{- if $server.AppProtectDosAccessLogDst}}
+        set $loggable '0';
+        access_log syslog:server={{ .AppProtectDosAccessLogDst }} log_dos if=$loggable;
+        {{- end}}
+	{{- end}}
+
+	{{- if $server.AppProtectDosMonitor}}
+	app_protect_dos_monitor "{{$server.AppProtectDosMonitor}}";
```
pleshakov, line 91: could you possibly explain how this directive works?

for example, let's say we have `app_protect_dos_monitor example.com`

Does this mean the DOS component will resolve this DNS name and send requests to it? 
What resolver will be used?
If multiple IPs are resolved, what IP will be used?
What will happen if the resolved DNS record expires?

Is it correct to assume that the expectation that a monitor request will go through an Ingress Controller pod? 

I'm trying to understand if this directive can be a source of any failures. For example:
- What will happen if the DNS cannot be resolved (1) no DNS record (2) the DNS server is unavailable?
- What will happen if the resolved IP address (1) doesn't respond, (2) responds with an error?

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -73,7 +73,28 @@ server {
...
+	{{- end}}
+
+    {{- if $server.AppProtectDosName}}
+    app_protect_dos_name "{{$server.AppProtectDosName}}";
```
pleshakov, line 95: What will happen if the configured name isn't unique? Let's say there is two Ingress resources and they both configure the same DOS name.

pkg/apis/configuration/v1/types.go
```diff
@@ -351,6 +351,7 @@ type PolicySpec struct {
 	EgressMTLS    *EgressMTLS    `json:"egressMTLS"`
 	OIDC          *OIDC          `json:"oidc"`
 	WAF           *WAF           `json:"waf"`
+	Bados         *Bados         `json:"bados"`
```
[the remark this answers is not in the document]
pleshakov: what is BADOS? is it an internal abbreviation?

pleshakov (commented):
Hi @pasmant 

Please see my comments and questions about DOS configuration

pkg/apis/configuration/v1/types.go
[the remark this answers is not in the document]
pasmant: Bados name will be changed

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -113,6 +129,23 @@ http {
     include oidc/oidc_common.conf;
     {{- end}}
 
+    {{- if .AppProtectDosLoadModule}}
+        {{- if .AppProtectDosLivenessEnable}}
```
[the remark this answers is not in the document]
pasmant: If want to add need manually add to the deployment the args inside ports:
        - name: liveness-port
          containerPort: 8090
 livenessProbe:
          httpGet:
            path: app_protect_dos_liveness
            port: liveness-port
          initialDelaySeconds: 60
          periodSeconds: 10

Also add to config map:
app-protect-dos-liveness-enable: "True"

I saw that there is an implementation of readiness probe but not for liveness, so I add the implementation for liveness that we add in the app-protect-dos moudle to KIC

internal/configs/configmaps.go
```diff
@@ -326,6 +330,10 @@ func ParseConfigMap(cfgm *v1.ConfigMap, nginxPlus bool, hasAppProtect bool) *Con
 		cfgParams.MainWorkerRlimitNofile = workerRlimitNofile
 	}
 
+	if workerRlimitCore, exists := cfgm.Data["worker-rlimit-core"]; exists {
```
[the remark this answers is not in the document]
pasmant: Remove from the PR

pkg/apis/configuration/v1/types.go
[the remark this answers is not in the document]
pasmant: Bados is Behavioral-Analytics-based Denial of Service mitigation
We sometimes use it as synonym to app-protect-dos

internal/configs/virtualserver.go
```diff
@@ -988,6 +992,72 @@ func (p *policiesCfg) addWAFConfig(
 	return res
 }
 
+func (p *policiesCfg) addBadosConfig(
```
pleshakov, line 995: Could we have a test similar to TestAddWafConfig?

internal/k8s/appprotectdos/app_protect_dos_configuration.go
```diff
@@ -0,0 +1,271 @@
...
+	Object *unstructured.Unstructured
+	// Reason tells the reason. It matches the reason in the events of our configuration objects.
+	Reason string
+	// Message gives the details about the problem. It matches the message in the events of our configuration objects.
+	Message string
+}
+
+// Configuration configures App Protect Dos resources that the Ingress Controller uses.
+type Configuration interface {
+	AddOrUpdateDosPolicy(policyObj *unstructured.Unstructured) (changes []Change, problems []Problem)
```
pleshakov, line 74: Because the dos part is already in the name of the package, I suggest removing the Dos part from the names of all methods of this interface.

internal/k8s/appprotectdos/app_protect_dos_configuration.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// FakeConfiguration holds representations of fake App Protect Dos cluster resources
+type FakeConfiguration struct {
+	DosPolicies map[string]*DosPolicyEx
+	DosLogConfs map[string]*DosLogConfEx
+}
+
+// NewFakeConfiguration creates a new App Protect Dos Configuration
+func NewFakeConfiguration() Configuration {
```
pleshakov, line 217: it doesn't look like FakeConfiguration is used anywhere in the code. If it is the case, than better to remove it

internal/k8s/appprotectdos/app_protect_dos_resources.go
```diff
@@ -0,0 +1,151 @@
...
+	{"spec", "signatures"},
+	{"spec", "bad_actors"},
+}
+
+var appProtectDosLogConfRequiredFields = [][]string{
+	{"spec", "content"},
+	{"spec", "filter"},
+}
+
+func validateRequiredFields(policy *unstructured.Unstructured, fieldsList [][]string) error {
```
pleshakov, line 26: in this file we have multiple functions similar to the functions in app_protect_configuration.go
- validateRequiredFields
- ValidateAppProtectDosLogDestination (similar to ValidateAppProtectLogDestination)
- GetNsName 
- ParseResourceReferenceAnnotation

I suggest adding another package - `appprotect_common` and move common functions there and reference them from the `approtect` and `approtectdos` packages.

Later, after this PR, we can further refactor the package structure - for example, have `appprotect`, `approtect/waf`, `approtect/dos`, `appprotect/common` packages. However, we don't need to do it now, so that we don't increase the amount of changes in this PR.

internal/k8s/appprotectdos/app_protect_dos_resources.go
```diff
@@ -0,0 +1,151 @@
...
+			return fmt.Errorf("Error checking for required field %v: %w", field, err)
+		}
+		if !found {
+			return fmt.Errorf("Required field %v not found", field)
+		}
+	}
+	return nil
+}
+
+func validateRequiredFieldsNoCopy(policy *unstructured.Unstructured, fieldsList [][]string) error {
```
pleshakov, line 39: can we have a unit test for this function?

internal/k8s/appprotectdos/app_protect_dos_resources.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+		if !found {
+			return fmt.Errorf("Required field %v not found", field)
+		}
+	}
+	return nil
+}
+
+// ValidateAppDosProtectLogConf validates LogConfiguration resource
+func ValidateAppProtectDosLogConf(logConf *unstructured.Unstructured) error {
```
pleshakov, line 53: can we have a unit test for this function?

internal/k8s/appprotectdos/app_protect_dos_resources.go
```diff
@@ -0,0 +1,151 @@
...
+		return fmt.Errorf("Error parsing host: %v is not a valid ip address", ipstr)
+	}
+
+	return nil
+}
+
+var accessLog = regexp.MustCompile(`^(((\d{1,3}\.){3}\d{1,3}):\d{1,5})$`)
+
+// ValidateAppProtectDosAccessLog validates destination for access log configuration
+func ValidateAppProtectDosAccessLogDest(accessLogDest string) error {
```
pleshakov, line 106: can we have a unit test for this function?

internal/k8s/appprotectdos/app_protect_dos_resources.go
```diff
@@ -0,0 +1,151 @@
...
+
+	if net.ParseIP(dstchunks[0]) == nil {
+		return fmt.Errorf("Error parsing host: %v is not a valid ip address", dstchunks[0])
+	}
+
+	return nil
+}
+
+// ValidateAppProtectDosPolicy validates Policy resource
+func ValidateAppProtectDosPolicy(policy *unstructured.Unstructured) error {
```
pleshakov, line 129: can we have a unit test for this function?

internal/k8s/appprotectdos/app_protect_dos_resources.go
```diff
@@ -0,0 +1,151 @@
...
+	err := validateRequiredFieldsNoCopy(policy, appProtectDosPolicyRequiredFields)
+	if err != nil {
+		return fmt.Errorf("Error validating App Protect Dos Policy %v: %w", polName, err)
+	}
+
+	return nil
+}
+
+// ParseResourceReferenceAnnotation returns a namespace/name string
+func ParseResourceReferenceAnnotation(ns, antn string) string {
```
pleshakov, line 141: this one is not used in the code

internal/k8s/controller.go
```diff
@@ -358,6 +374,24 @@ func (lbc *LoadBalancerController) addAppProtectUserSigHandler(handlers cache.Re
 	lbc.cacheSyncs = append(lbc.cacheSyncs, informer.HasSynced)
 }
 
+// addappProtectDosPolicyHandler creates dynamic informers for custom appprotectdos policy resource
```
pleshakov, line 377:
```suggestion
// addAppProtectDosPolicyHandler creates dynamic informers for custom appprotectdos policy resource
```

internal/k8s/controller.go
```diff
@@ -358,6 +374,24 @@ func (lbc *LoadBalancerController) addAppProtectUserSigHandler(handlers cache.Re
...
+func (lbc *LoadBalancerController) addAppProtectDosPolicyHandler(handlers cache.ResourceEventHandlerFuncs) {
+	informer := lbc.dynInformerFactory.ForResource(appprotectdos.DosPolicyGVR).Informer()
+	informer.AddEventHandler(handlers)
+	lbc.appProtectDosPolicyLister = informer.GetStore()
+
+	lbc.cacheSyncs = append(lbc.cacheSyncs, informer.HasSynced)
+}
+
+// addappProtectDosLogConfHandler creates dynamic informer for custom appprotectdos logging config resource
```
pleshakov, line 386:
```suggestion
// addAppProtectDosLogConfHandler creates dynamic informer for custom appprotectdos logging config resource
```

internal/k8s/controller.go
```diff
@@ -2099,6 +2217,27 @@ func (lbc *LoadBalancerController) createIngressEx(ing *networking.Ingress, vali
...
+					glog.Warningf("Error Getting App Protect Dos policy %v for Ingress %v/%v: %v", apDosPolicyAntn, ing.Namespace, ing.Name, err)
+				} else {
+					ingEx.AppProtectDosPolicy = policy
+				}
+			}
+
+			if apDosLogConfAntn, exists := ingEx.Ingress.Annotations[configs.AppProtectDosLogConfAnnotation]; exists {
+				logConf, logDst, err := lbc.getAppProtectDosLogConfAndDst(ing)
+				if err != nil {
+					glog.Warningf("Error Getting App Protect policy %v for Ingress %v/%v: %v", apDosLogConfAntn, ing.Namespace, ing.Name, err)
```
pleshakov, line 2234:
```suggestion
					glog.Warningf("Error Getting App Protect Dos Log Config %v for Ingress %v/%v: %v", apDosLogConfAntn, ing.Namespace, ing.Name, err)
```

internal/k8s/controller.go
```diff
@@ -2649,6 +2840,46 @@ func (lbc *LoadBalancerController) addWAFPolicyRefs(
 	return nil
 }
 
+// addBadosPolicyRefs ensures the app protect dos resources that are referenced in policies exist.
+func (lbc *LoadBalancerController) addBadosPolicyRefs(
```
pleshakov, line 2844: Could we have a unit test similar to TestAddWAFPolicyRefs?

internal/k8s/controller.go
```diff
@@ -2697,6 +2928,30 @@ func getWAFPoliciesForAppProtectLogConf(pols []*conf_v1.Policy, key string) []*c
 	return policies
 }
 
+func getBadosPoliciesForAppProtectDosPolicy(pols []*conf_v1.Policy, key string) []*conf_v1.Policy {
```
pleshakov, line 2931: Could we have  a unit test similar to TestGetWAFPoliciesForAppProtectPolicy?

internal/k8s/controller.go
```diff
@@ -2697,6 +2928,30 @@ func getWAFPoliciesForAppProtectLogConf(pols []*conf_v1.Policy, key string) []*c
...
+	for _, pol := range pols {
+		if pol.Spec.Bados != nil && isMatchingResourceRef(pol.Namespace, pol.Spec.Bados.ApDosPolicy, key) {
+			policies = append(policies, pol)
+		}
+	}
+
+	return policies
+}
+
+func getBadosPoliciesForAppProtectLogConf(pols []*conf_v1.Policy, key string) []*conf_v1.Policy {
```
pleshakov, line 2943: Could we have a unit test similar to TestGetWAFPoliciesForAppProtectLogConf?

internal/k8s/validation.go
```diff
@@ -227,6 +231,20 @@ var (
 			validateRequiredAnnotation,
 			validateBoolAnnotation,
 		},
+		appProtectDosEnableAnnotation: {
```
pleshakov, line 234: There are more Dos annotations:
- appprotectdos.f5.com/app-protect-dos-policy
- appprotectdos.f5.com/app-protect-dos-security-log
-  appprotectdos.f5.com/app-protect-dos-monitor
- appprotectdos.f5.com/app-protect-dos-name

Could you add validation for them? Both that they (1) require Dos and also  (2) validating their values? Note that validating the values here will make validation in `getAppProtectDosLogConfAndDst` unnecessary, because an Ingress with invalid annotations values will not be processed.
	
Note that In AppProtect annotations we have a validation gap which we will address separately later.

internal/nginx/manager.go
```diff
@@ -37,6 +37,10 @@ const (
 
 	// appProtectLogConfigFileName is the location of the NGINX App Protect logging configuration file
 	appProtectLogConfigFileName = "/etc/app_protect/bd/logger.cfg"
+
+	appProtectDosAgentInstallCmd    = "/usr/bin/adminstall"
```
pleshakov, line 41: What does the install command do? Just curious

internal/nginx/manager.go
```diff
@@ -37,6 +37,10 @@ const (
...
+	appProtectDosAgentStartCmd      = "/usr/bin/admd -d --standalone > /var/log/adm/admd.log 2>&1"
```
pleshakov, line 42: how big the log file can grow?
if it can grow big, is there a log rotation mechanism?
usually, we expose logs through stdout/stderr of the container

pkg/apis/configuration/validation/policy.go
```diff
@@ -284,6 +302,51 @@ func validateLogConf(logConf, logDest string, fieldPath *field.Path) field.Error
 	return allErrs
 }
 
+func validateBados(bados *v1.Bados, fieldPath *field.Path) field.ErrorList {
```
pleshakov, line 305: Could we have tests similar to TestValidateWAF and TestValidateWAFInvalid?

pkg/apis/configuration/validation/policy.go
```diff
@@ -284,6 +302,51 @@ func validateLogConf(logConf, logDest string, fieldPath *field.Path) field.Error
...
+	}
+	return allErrs
+}
+
+func validateDosLogAccessLogDest(accessLogDest string, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	err := appprotectdos.ValidateAppProtectDosAccessLogDest(accessLogDest)
+	if err != nil {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("dosAccessLogDest"), accessLogDest, err.Error()))
```
pleshakov, line 345: fieldPath is already dosAccessLogDest here, no need to create another child

pleshakov (commented):
Hi @pasmant 

Please see the comments and suggestions about the code.
A lot of them are about adding missing unit tests.

If `Bados` word is removed from the configuration, could we also remove (rename) it from any identifiers in the code and error messages?

internal/configs/version1/nginx-plus.tmpl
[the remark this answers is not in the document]
pleshakov: Yep, currently we don't have a liveness probe (my bad, I mentioned the readiness probe instead of liveness)

However, the IC process monitors NGINX master process and the DOS process. If they crash, the IC will terminate. I wonder if it is enough?

Ultimately, if we want to have a liveness probe, it should probe the IC, not DOS component, because the DOS component is only part of the story. So the IC needs to be responsible for monitoring the processes it starts.

Also, one potential problem I see with monitoring liveness via an HTTP request to NGINX, is that if NGINX is under pressure (let's say a lot of RPS from clients), it might temporarily not be able to respond. If you consider that typically there are multiple IC pods running and let's say Kubernetes restarts one IC pod because of the failed liveness probe (which is based on making an HTTP request to NGINX), the other pods, which are also probably already under pressure, will immediately get the share of the killed pod's traffic and as a result, they might get killed as well...

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -73,7 +73,28 @@ server {
...
+	{{- if $server.AppProtectDosEnable}}
+	app_protect_dos_enable {{$server.AppProtectDosEnable}};
+	{{if $server.AppProtectDosPolicy}}app_protect_dos_policy_file {{$server.AppProtectDosPolicy}};{{end}}
+	{{- end}}
+	{{- if $server.AppProtectDosLogEnable}}
+	app_protect_dos_security_log_enable {{$server.AppProtectDosLogEnable}};
+	{{if $server.AppProtectDosLogConf}}app_protect_dos_security_log {{$server.AppProtectDosLogConf}};{{end}}
+	    {{- if $server.AppProtectDosAccessLogDst}}
+        set $loggable '0';
+        access_log syslog:server={{ .AppProtectDosAccessLogDst }} log_dos if=$loggable;
```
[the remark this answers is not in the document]
pasmant: We set loggable to 0, but inside the .so of app-protect-dos we change it to 1 if we not pass the rate limit.

internal/configs/version1/nginx-plus.ingress.tmpl
pasmant: Lets take this example:
server {
   server_name coffee.com;
   listen 80;
   location /abc {
    app_protect_dos_monitor coffee.com/abc;
    proxy_pass be;
   }
}

Result: curl -H "Host: coffee.com" http://localhost/abc

So, the request go to localhost and get to the nginx process which send it to the proxy_pass.
There is no DNS resolved.
In case there is no respond or error respond it will be handle by the admd

internal/configs/version1/nginx-plus.ingress.tmpl
pasmant: Lets look at this example:

server {​​​​​​​
   server_name coffee.com;
   listen 80;
   app_protect_dos_monitor coffee.com/abc;
   app_protect_dos_name example;
   location /abc {​​​​​​​
        app_protect_dos_name example;
        proxy_pass app1;
   }​​​​​​​


   location / {​​​​​​​
        app_protect_dos_name example;
        proxy_pass app1;
   }​​​​​​​
}​​​​​​​

We see that there is same name to different locations.
What will happen is that the admd will handle to request from both / and /abc and like one napdos service.
So, if there will be mitigation on the napdos service, both the location perform the mitigation.

deployments/helm-chart/crds/appprotectdos.f5.com_apdospolicy.yaml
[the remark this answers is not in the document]
pasmant: The policy file that create look like:
{
"automation_tools_detection":"on",
"bad_actors":"on",
"mitigation_mode":"standard",
"signatures":"on",
"tls_fingerprint":"on"
}

So,  _ is to described the fields inside the json file.

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -113,6 +129,23 @@ http {
...
+    server {
+        server_name probe;
+        listen {{.AppProtectDosLivenessPort}};
+        location /{{.AppProtectDosLivenessURI}} {
+            app_protect_dos_enable on;
+            proxy_pass http://localhost:8091;
+        }
+    }
+
+    server {
```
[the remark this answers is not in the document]
pasmant: The server is use when the app-protect-dos failed the server will return 503

internal/nginx/manager.go
pasmant: For IPC between admd and nginx

internal/k8s/appprotectdos/app_protect_dos_resources.go
pasmant: Its optional, so its give more flexibility for the log dest options

internal/nginx/manager.go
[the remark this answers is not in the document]
pasmant: The main reasons is that the debug of nginx is traffic affective while for admd it doesn't affect to the traffic.

internal/configs/version1/nginx-plus.tmpl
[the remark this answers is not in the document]
pasmant: However, the IC process monitors NGINX master process and the DOS process. If they crash, the IC will terminate. I wonder if it is enough?
There can be an option of admd deadlock for example ,so, the liveness probe will find it.
I agree that it should probe IC, but for now until liveness will be to IC there is an option to use dos liveness.
potential problem - maybe need to add threshold for the liveness probe?

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -51,6 +56,10 @@ http {
                       '"$http_user_agent" "$http_x_forwarded_for"';
     {{- end}}
 
+    {{- if .AppProtectDosLoadModule}}
+    log_format log_dos ', vs_name_al=$app_protect_dos_vs_name, ip=$remote_addr, tls_fp=$app_protect_dos_tls_fp, outcome=$app_protect_dos_outcome, reason=$app_protect_dos_outcome_reason, policy_name=$app_protect_dos_policy_name, dos_version=$app_protect_dos_version, ip_tls=$remote_addr:$app_protect_dos_tls_fp, ';
```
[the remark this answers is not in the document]
pasmant: Make the format customizable.
The reason for separate format in the dos format there are some var that only known by the app-protect-dos module.

internal/configs/version1/nginx.tmpl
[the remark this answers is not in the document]
pasmant: Will move from the PR

internal/configs/annotations.go
pasmant: I look on the const value all of them are used by another file and this is the reason for the const.

internal/configs/virtualserver.go
pasmant: Remove check

internal/configs/virtualserver.go
pasmant: Remove check

internal/configs/virtualserver.go
pasmant: Remove check

internal/k8s/controller.go
pasmant: Change to %v

internal/configs/version1/nginx-plus.ingress.tmpl
pleshakov: thanks for the clarification!

Although, I still see cases when it might not work. But it should work in most of the cases.

Most of Ingress or VirtualServer resources have TLS termination enabled with a HTTPs-redirect if a request comes to an HTTP port. So I think the user will need to monitor `https://`.

If you monitor `https://`, I don't think it will work in the following cases:
- TLS passthrough is enabled (https://docs.nginx.com/nginx-ingress-controller/configuration/transportserver-resource/ ), because in that case, NGINX config will require that client HTTPs connections include the [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication), which the shown curl command doesn't set 
- when client authentication is enabled. For example, using https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#ingressmtls policy.

However, those cases are not as common.

internal/configs/version1/nginx-plus.ingress.tmpl
pleshakov: I see a potential problem here.

Let's say we have two Ingress resources:

first one, managed by team A:

```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: team-a-ingress
  annotations:
    appprotectdos.f5.com/app-protect-dos-enable: "True"
    . . .
    appprotectdos.f5.com/app-protect-dos-name: "teamA.example.com"
spec:
  rules:
    - host: teamA.example.com 
```   

second one, managed by team B:

```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: team-a-ingress
  annotations:
    appprotectdos.f5.com/app-protect-dos-enable: "True"
    . . .
    appprotectdos.f5.com/app-protect-dos-name: "teamA.example.com"
spec:
  rules:
    - host: teamB.example.com 
```

as you can see, team B can put `teamA` into the name annotation. This mean that team B can affect DOS configuration of team A, because admd will treat them as one service now. 

I wonder if we should allow that?

deployments/deployment/appprotect-dos-arb.yaml
```diff
@@ -0,0 +1,40 @@
...
+    matchLabels:
+      app: appprotect-dos-arb
+  template:
+    metadata:
+      labels:
+        app: appprotect-dos-arb
+    spec:
+      containers:
+        - name: appprotect-dos-arb
+          image: docker-registry.nginx.com/nap-dos/app_protect_dos_arb:latest
```
pleshakov, line 18: do we need to reference a version here rather than the latest?

The IC support policy is described here -- https://docs.nginx.com/nginx-ingress-controller/technical-specifications/#supported-nginx-ingress-controller-versions . We support a release for 2 years.

Let's imagine a situation that somebody deploys this file as part of 1.13 release half year of a year after the release 1.13. Will the DOS software for NGINX Plus R25  will still work with the latest arbitrator?

deployments/helm-chart/templates/controller-appprotect-dos-arb.yaml
```diff
@@ -0,0 +1,44 @@
...
+            limits:
+              memory: "128Mi"
+              cpu: "500m"
+          ports:
+            - containerPort: 3000
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: svc-appprotect-dos-arb
```
pleshakov, line 32: because this name is static, it creates a problem with multiple Ingress Controller deployments (helm releases) are installed in the same namespace, which is supported installation type. https://docs.nginx.com/nginx-ingress-controller/installation/running-multiple-ingress-controllers/  this way, we can have multiple Ingress Controllers where each Ingress Controller handles resources only of its IngressClass.  The static name will prevent helm from installing more than one helm release in one namespace :( 

It also looks like the DOS module doesn't support configuring the DNS name of the Arbitrator service :( 

can we support the described installation type?

deployments/helm-chart/templates/controller-appprotect-dos-arb.yaml
```diff
@@ -0,0 +1,44 @@
...
+  selector:
+    matchLabels:
+      app: {{ include "nginx-ingress.appProtectDosArbName" . }}
+  template:
+    metadata:
+      labels:
+        app: {{ include "nginx-ingress.appProtectDosArbName" . }}
+    spec:
+      containers:
+        - name: {{ include "nginx-ingress.appProtectDosArbName" . }}
```
pleshakov, line 19: does the arbitrator support any cli arguments? for example, log verbosity. If it supports something that the admin might want to configure, we should expose them through the helm chart parameters.

deployments/helm-chart/templates/controller-appprotect-dos-arb.yaml
```diff
@@ -0,0 +1,44 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{ include "nginx-ingress.appProtectDosArbName" . }}
+  template:
+    metadata:
+      labels:
+        app: {{ include "nginx-ingress.appProtectDosArbName" . }}
+    spec:
+      containers:
```
pleshakov, line 18: the IC runs as a non-root user and drop all Linux capabilities expect the needed ones, which is a security best practice. Could we also support that  for the arbitrator pod?

deployments/helm-chart/templates/controller-appprotect-dos-arb.yaml
```diff
@@ -0,0 +1,44 @@
...
+        - name: {{ include "nginx-ingress.appProtectDosArbName" . }}
+          image: "{{ .Values.controller.appprotectdos.arbitrator.repository }}"
```
pleshakov, line 20: do we need to expose any readiness and liveliness probes for the arbitrator?

pleshakov, line 20: could we support the image related parameters in a consistent way with the IC?


```
  image:
    ## The image repository of the Ingress controller.
    repository: nginx/nginx-ingress

    ## The tag of the Ingress controller image.
    tag: "1.12.0"

    ## The pull policy for the Ingress controller image.
    pullPolicy: IfNotPresent
```

For example:
arbitrator.image.repository
arbitrator.image.tag
arbitrator.image.pullPolicy

deployments/helm-chart/templates/controller-appprotect-dos-arb.yaml
```diff
@@ -0,0 +1,44 @@
...
+          imagePullPolicy: IfNotPresent
+          resources:
```
pleshakov, line 22: could we make the resources configurable? In case the admin needs to change the values. we can support it via `arbitrator.resources` parameter, similarly to the IC container.

pleshakov (commented):
Hi @pasmant

Please see comments and suggestion related to the installation manifests and the helm chart.

internal/configs/configurator.go
pasmant: Done both

internal/configs/version1/nginx-plus.ingress.tmpl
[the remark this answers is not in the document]
pasmant: add the comment

internal/configs/ingress.go
```diff
@@ -153,6 +165,11 @@ func generateNginxCfg(ingEx *IngressEx, apResources AppProtectResources, isMinio
 			server.AppProtectLogConfs = apResources.AppProtectLogconfs
 		}
 
+		if hasAppProtectDos {
```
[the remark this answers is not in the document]
pasmant: Done

internal/configs/virtualserver.go
pasmant: Add uni test TestAddBadosConfig

internal/configs/configurator.go
```diff
@@ -1277,6 +1279,20 @@ func (cnf *Configurator) updateApResources(ingEx *IngressEx) (apRes AppProtectRe
 		apRes.AppProtectLogconfs = append(apRes.AppProtectLogconfs, logConfFileName+" "+logConf.Dest)
 	}
 
+	if ingEx.AppProtectDosPolicy != nil {
```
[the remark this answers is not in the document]
pasmant: Create new unit test TestUpdateApDosResourcesForVs

internal/configs/configurator.go
```diff
@@ -1301,6 +1317,24 @@ func (cnf *Configurator) updateApResourcesForVs(vsEx *VirtualServerEx) map[strin
 		}
 	}
 
+	if vsEx.ApDosPolRefs != nil {
```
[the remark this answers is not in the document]
pasmant: Create new unit test TestUpdateApDosResourcesForVs

examples/appprotect-dos/README.md
```diff
@@ -0,0 +1,69 @@
...
+    ```
+    $ kubectl create -f apdos-policy.yaml
+    $ kubectl create -f apdos-logconf.yaml
+    ```
+4. Create an Ingress Resource:
+
+    Update the `appprotectdos.f5.com/app-protect-dos-security-log-destination` annotation from `webapp-ingress.yaml` with the ClusterIP of the syslog service. For example, if the IP is `10.101.21.110`:
+    ```yaml
+    . . .
+    appprotect.f5.com/app-protect-security-log-destination: "syslog:server=10.101.21.110:514"
```
soneillf5, line 46: I think we've talked about this in another channel, but I think it's good to add it here too so it's not lost. 

at the moment this annotation uses the cluster IP of the service to write log files. This is fine for normal nginx, but in Kubernetes, the cluster IP can change between different clusters or on service remove/add. 

The suggested fix is that the annotation specifies the 'namespace/name:port' of the service instead and a lookup is performed to get the cluster IP.

something like:
```
svcObj, svcExists, err := lbc.svcLister.GetByKey(svcKey)
ip := svcObj.Spec.ClusterIP
```

internal/k8s/appprotectdos/app_protect_dos_resources.go
pasmant: Remove this function, used validateRequiredFields instead

internal/k8s/appprotectdos/app_protect_dos_resources.go
pasmant: added unit test

internal/k8s/appprotectdos/app_protect_dos_resources.go
pasmant: Add unit test for ValidateAppProtectDosLogConf

internal/k8s/appprotectdos/app_protect_dos_resources.go
pasmant: Add unit test for ValidateAppProtectDosPolicy

internal/k8s/controller.go
pasmant: Add unit test  for AddBadosPolicyRefs

examples/appprotect-dos/README.md
soneillf5: Just a further thought, if we have four Ingress controllers, will they all point at the same logging service? Will that service have multiple replicas ? i.e 4 ingresses and 4 loggers ? In which case how do we control which logs go where? How do we correlate which log is from which ingress?

cmd/nginx-ingress/main.go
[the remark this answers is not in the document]
tomasohaodha: Why not put this in the ConfigMap for DOS along with its other configuration?

internal/k8s/appprotectdos/app_protect_dos_configuration.go
pasmant: Used now in the new unit test: TestAddBadosPolicyRefs

internal/k8s/controller.go
pasmant: done

internal/k8s/controller.go
pasmant: Done

internal/k8s/controller.go
pasmant: Done

internal/k8s/appprotectdos/app_protect_dos_configuration.go
pasmant: Done

internal/k8s/controller.go
pasmant: done

internal/k8s/controller.go
pasmant: done

pkg/apis/configuration/validation/policy.go
pasmant: Done

pkg/apis/configuration/validation/policy.go
pasmant: Done

deployments/helm-chart/templates/controller-appprotect-dos-arb.yaml
pasmant: No need to expose any probes for the arbitrator

deployments/helm-chart/templates/controller-appprotect-dos-arb.yaml
pasmant: No cli args

deployments/deployment/appprotect-dos-arb.yaml
```diff
@@ -0,0 +1,40 @@
...
+  namespace: nginx-ingress
+spec:
+  selector:
+    app: appprotect-dos-arb
+  ports:
+    - name: arb
+      port: 3000
+      protocol: TCP
+      targetPort: 3000
+  clusterIP: None
```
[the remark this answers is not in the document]
pasmant: We perform request by the SVC name. 
We use headless service with selector https://kubernetes.io/docs/concepts/services-networking/service/#with-selectors.
So, I think in this case, the DNS name will be resolved into a VIP of the new arbitrator pod.

cmd/nginx-ingress/main.go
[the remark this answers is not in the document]
pasmant: Will move to config map

internal/nginx/manager.go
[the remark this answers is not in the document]
pasmant: move to config map

deployments/helm-chart/README.md
pasmant: Remove the name option and autogenerated will be only option

pasmant: Done

internal/nginx/manager.go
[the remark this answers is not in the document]
pasmant: AppProtectDosAgentStart called before parsing config map. should I move the function AppProtectDosAgentStart after ParseConfigMap is called?

internal/k8s/appprotectdos/app_protect_dos_resources.go
pasmant: Will remove this option

pasmant: Done

internal/nginx/manager.go
pasmant: Will expose through stdout of the container

internal/k8s/appprotectdos/app_protect_dos_resources.go
pasmant: Add package appprotect_common

internal/k8s/appprotectdos/app_protect_dos_resources.go
pasmant: Removed (used from appprotect_common package)

cmd/nginx-ingress/main.go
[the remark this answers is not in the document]
pasmant: Done

internal/nginx/manager.go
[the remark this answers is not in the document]
pasmant: Move the AppProtectDosAgentStart after configmap was parse

deployments/helm-chart/templates/controller-appprotect-dos-arb.yaml
pasmant: done

deployments/helm-chart/templates/controller-appprotect-dos-arb.yaml
pasmant: done

deployments/helm-chart/templates/controller-appprotect-dos-arb.yaml
pasmant: Adding new helm chart for arbitrator

internal/k8s/validation.go
pasmant: Done, add validation to all new app protect dos annotation.
Remove the validation from the `getAppProtectDosLogConfAndDst`

examples/appprotect-dos/webapp.yaml
```diff
@@ -0,0 +1,32 @@
...
+      containers:
+      - name: webapp
+        image: nginxdemos/hello:plain-text
+        ports:
+        - containerPort: 80
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: webapp-svc
```
tomasohaodha, line 24: Other NIC examples, including WAF, use the cafe app. Is there a good reason to break the pattern and introduce a new example application here?

tomasohaodha (changes requested):

deployments/deployment/appprotect-dos-arb.yaml
tomasohaodha: "Not change a lot at all" - we cannot tell the future, and specifying a version protects us against that. If best design is to specify a version then we should reference a version. Good design protects us against our own incorrect assumptions.

pleshakov:
closing. the work was merged into `feature/bados` branch

Issue #1655, opened 2021-06-09
Improve the IC start time


pleshakov:
**Is your feature request related to a problem? Please describe.**

When the IC starts, it processes resources in the cluster (Ingresses, VirtualServers, policies, …) one by one and generates configuration for them. Once it finishes the processing, it becomes ready to accept the traffic.

If you have many resources, the one by one processing will take significant time. For example, for 100 hundred Ingresses, it can take a minute for KIC to process all of them. 
```
I0312 19:48:15.053181       1 main.go:258] Starting NGINX Ingress controller Version=1.11.0-test-c-1 GitCommit=b99525a1
. . .
I0312 19:49:03.985287       1 controller.go:771] NGINX is ready
```

**Describe the solution you'd like**
Reduce the start time 

**Describe alternatives you've considered**
Split the Ingress, VirtualServer and other resources among multiple NGINX Ingress Controllers, each with a different IngressClass -- https://docs.nginx.com/nginx-ingress-controller/installation/running-multiple-ingress-controllers/#running-multiple-nginx-ingress-controllers

**Additional context**
See also  https://github.com/nginxinc/kubernetes-ingress/issues/1113 and https://github.com/nginxinc/kubernetes-ingress/issues/1448

Aha! Link: https://nginx.aha.io/features/IC-98

PR #1788, opened 2021-07-29
Reload only once during the start

pleshakov:
### Proposed changes

During the start, the IC processes resources in the cluster one by one
Each processing might lead to a new generated config or an updated config,
which requires an NGINX reload. After all resources are processed,
the IC pod is considered Ready (its readiness probe succeeds).

For clusters with big number of resources the combined time
of NGINX reloads during the start can be significant (minutes).

This PR ensures that the IC only reloads NGINX once during
the start, which reduces the start up time for clusters
with big number of resources.

Fixes https://github.com/nginxinc/kubernetes-ingress/issues/1655

### DIFF
--- a/internal/configs/configurator.go
+++ b/internal/configs/configurator.go
@@ -94,6 +94,10 @@ type metricLabelsIndex struct {
 }
 
 // Configurator configures NGINX.
+// Until reloads are enabled via EnableReloads(), the Configurator will not reload NGINX and update NGINX Plus
+// upstream servers via NGINX Plus API for configuration changes.
+// This allows the Ingress Controller to incrementally build the NGINX configuration during the IC start and
+// then apply it at the end of the start.
 type Configurator struct {
 	nginxManager            nginx.Manager
 	staticCfgParams         *StaticConfigParams
@@ -111,6 +115,7 @@ type Configurator struct {
 	isPrometheusEnabled     bool
 	latencyCollector        latCollector.LatencyCollector
 	isLatencyMetricsEnabled bool
+	isReloadsEnabled        bool
 }
 
 // NewConfigurator creates a new Configurator.
@@ -146,6 +151,7 @@ func NewConfigurator(nginxManager nginx.Manager, staticCfgParams *StaticConfigPa
 		isPrometheusEnabled:     isPrometheusEnabled,
 		latencyCollector:        latencyCollector,
 		isLatencyMetricsEnabled: isLatencyMetricsEnabled,
+		isReloadsEnabled:        false,
 	}
 	return &cnf
 }
@@ -248,7 +254,7 @@ func (cnf *Configurator) AddOrUpdateIngress(ingEx *IngressEx) (Warnings, error) 
 		return warnings, fmt.Errorf("Error adding or updating ingress %v/%v: %w", ingEx.Ingress.Namespace, ingEx.Ingress.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(nginx.ReloadForOtherUpdate); err != nil {
+	if err := cnf.reload(nginx.ReloadForOtherUpdate); err != nil {
 		return warnings, fmt.Errorf("Error reloading NGINX for %v/%v: %w", ingEx.Ingress.Namespace, ingEx.Ingress.Name, err)
 	}
 
@@ -289,7 +295,7 @@ func (cnf *Configurator) AddOrUpdateMergeableIngress(mergeableIngs *MergeableIng
 		return warnings, fmt.Errorf("Error when adding or updating ingress %v/%v: %w", mergeableIngs.Master.Ingress.Namespace, mergeableIngs.Master.Ingress.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(nginx.ReloadForOtherUpdate); err != nil {
+	if err := cnf.reload(nginx.ReloadForOtherUpdate); err != nil {
 		return warnings, fmt.Errorf("Error reloading NGINX for %v/%v: %w", mergeableIngs.Master.Ingress.Namespace, mergeableIngs.Master.Ingress.Name, err)
 	}
 
@@ -420,7 +426,7 @@ func (cnf *Configurator) AddOrUpdateVirtualServer(virtualServerEx *VirtualServer
 		return warnings, fmt.Errorf("Error adding or updating VirtualServer %v/%v: %w", virtualServerEx.VirtualServer.Namespace, virtualServerEx.VirtualServer.Name, err)
 	}
 
-	if err := cnf.nginxManager.Reload(nginx.ReloadForOtherUpdate); err != nil {
+	if err := cnf.reload(nginx.ReloadForOtherUpdate); err != nil {
 		return warnings, fmt.Errorf("Error reloading NGINX for VirtualServer %v/%v: %w", virtualServerEx.VirtualServer.Namespace, virtualServerEx.VirtualServer.Name, err)
 	}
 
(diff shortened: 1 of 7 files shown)


internal/k8s/configuration.go
```diff
@@ -765,14 +773,16 @@ func (c *Configuration) buildListenersAndTSConfigurations() (newListeners map[st
 // GetResources returns all configuration resources.
 func (c *Configuration) GetResources() []Resource {
 	return c.GetResourcesWithFilter(resourceFilter{
-		Ingresses:      true,
```
tomasohaodha, line 768: There are changes related to TransportServers in the following files. Are transport servers intimately connected to the reload functionality?

internal/k8s/configuration_test.go
```diff
@@ -2332,6 +2334,10 @@ func TestAddGlobalConfiguration(t *testing.T) {
 	if diff := cmp.Diff(expectedProblems, problems); diff != "" {
 		t.Errorf("AddOrUpdateTransportServer() returned unexpected result (-want +got):\n%s", diff)
 	}
+	storedGC := configuration.GetGlobalConfiguration()
```
tomasohaodha, line 2337: A very close variation on this code is repeated 5 times. Worth pulling it out to avoid duplicate code?

internal/configs/configurator.go
```diff
@@ -996,8 +1006,21 @@ func (cnf *Configurator) updatePlusEndpoints(ingEx *IngressEx) error {
...
+	if !cnf.isReloadsEnabled {
+		return nil
+	}
+
+	return cnf.nginxManager.Reload(isEndpointsUpdate)
+}
+
 // UpdateConfig updates NGINX configuration parameters.
-func (cnf *Configurator) UpdateConfig(cfgParams *ConfigParams, ingExes []*IngressEx, mergeableIngs []*MergeableIngresses, virtualServerExes []*VirtualServerEx) (Warnings, error) {
+func (cnf *Configurator) UpdateConfig(cfgParams *ConfigParams, resources ExtendedResources) (Warnings, error) {
```
tomasohaodha, line 1023: Should we address this lint failure now while we're in this part of the code?

pleshakov: better to do it separately

internal/k8s/configuration_test.go
pleshakov: while that test code is verbose, it is easy to read and follow. I don't think it's worth spending time reducing the duplication.

internal/k8s/configuration.go
pleshakov: good question. this is needed for this new function https://github.com/nginxinc/kubernetes-ingress/blob/18b5aba08cffa5734ad316c560d1977dd93402cf/internal/k8s/controller.go#L640 
the function regenerates all the config, reloads NGINX and then reports events and updates statuses

if the first reload finishes with an error, that would be reflected in the statuses and events of the resources

internal/configs/configurator.go
tomasohaodha: Isn't that just creating technical debt? Was this lint error here before the changes?

internal/k8s/configuration.go
tomasohaodha: Understood - but why does it seem trasnport servers are a special case? The added lines all have to do with the transport server - the ingress and virtual server were there previously, but the transport server was added.

internal/configs/configurator.go
pleshakov: yep, it was there before. the linter is configured so that it doesn't show the errors for the existing code, only for the new code. the new code in the PR changed the existing code with the existing problem, so that's why the linter complained.
better to stay focus in this PR and not increase its scope any further, and acknowledge and address the problematic method (there are more than one in Configurator with high complexity) separately

internal/k8s/configuration.go
pleshakov: Let me clarify that. hope it's not confusing.

The PR ensures that after all the initial config is generated, the IC reloads NGINX and emits events and updates statuses of the resources with the reload status (and any additional warnings that could be generated for resources). To accomplish that, the PR reuses the existing code that was previously in `syncConfigMap` (processes changes to the ConfigMap). That code would gather all resources,  re-generate all configs and reload NGINX  (via Configurator) and emit events and update statuses. The PR put that code in the new `updateAllConfigs()` method of the Controller. So that it can still be called from `syncConfigMap`. And we also call in from `sync()` method when the queue is drained - this is what makes that first reload.

Note that we need to re-generate the Config, because we want to report events and update statuses with warnings and errors, and some of the reported warnings are only discovered in the config-generation phase in the Configurator. (Note that UpdateConfig of Configurator returns warnings).

However, the code previously lacked reporting statuses and events for TransportServer (omission, not a bug),  so that's why I added it.

Technically, we don't need to regenerate the TransportServer configs yet in Configurator, because: https://github.com/nginxinc/kubernetes-ingress/blob/219d2479be3d38665efdf92fa2dbfc774eb01cbe/internal/configs/configurator.go#L1085:
- changes to the ConfigMap don't affect the TransportServer configs directly
- Configurator doesn't generate Warnings for TransportServers -- https://github.com/nginxinc/kubernetes-ingress/blob/219d2479be3d38665efdf92fa2dbfc774eb01cbe/internal/configs/configurator.go#L1086

I removed that code that I initially added in the PR. Note that in the future, we will need to regenerate the config. For example, we will soon support custom templates for the TransportServers in the ConfigMap, which will make it necessary. I added an explanatory comment there, because it's not obvious why we don't re-generate TransportServer configs.

tests/suite/test_smoke.py
```diff
@@ -77,20 +81,20 @@ def fin():
...
-        pytest.param({"extra_args": ["-enable-prometheus-metrics"]}, id="one-additional-cli-args"),
-        pytest.param(
-            {"extra_args": ["-nginx-debug", "-health-status=true", "-enable-prometheus-metrics"]},
-            id="some-additional-cli-args",
-        ),
-    ],
-    indirect=True,
-)
 class TestSmoke:
+    @pytest.mark.ingresses
```
vepatel, line 86: `ingresses` marker and `ingress_controller` fixture can and should be called at class level as ic one is common to both tests and `ingresses` marker will only cause first test to run in its current form

internal/k8s/configuration.go
tomasohaodha: Gotcha, thanks for the explanation.

internal/configs/configurator.go
```diff
@@ -859,7 +867,8 @@ func (cnf *Configurator) UpdateEndpointsForVirtualServers(virtualServerExes []*V
 			return fmt.Errorf("Error adding or updating VirtualServer %v/%v: %w", vs.VirtualServer.Namespace, vs.VirtualServer.Name, err)
 		}
 
-		if cnf.isPlus {
+		// if reloads are not enabled, the upstream will not exist in the NGINX Plus, so we need to skip the update.
+		if cnf.isPlus && cnf.isReloadsEnabled {
```
soneillf5, line 871: It would be worth checking if we can unify the decision making here with `cnf.reload()`

pleshakov: let me know what you think of https://github.com/nginxinc/kubernetes-ingress/pull/1788/commits/db061cb5b55bc53f97e072e92e031467c526e8d7

soneillf5 (approved):

internal/configs/configurator.go
soneillf5: all good thanks

lucacome (approved):

PR #1846, opened 2021-08-11
Update CodeQL workflow

lucacome:
- Added check for javascript (files in the docs folder)
- Added the cancel for the workflow inside the workflow itself (I think it makes it more clear to understand what happens)

### DIFF
--- a/.github/workflows/cancel.yml
+++ b/.github/workflows/cancel.yml
@@ -29,5 +29,5 @@ jobs:
     steps:
       - uses: styfle/cancel-workflow-action@0.9.1
         with:
-          workflow_id: 2012221,5339701,7611535
+          workflow_id: 2012221,5339701
           access_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


soneillf5 (approved):

pleshakov:
@lucacome I wonder if we really need JavaScript checking? that code is not developed in our repo, just copied to our repo. So even if a problem is found, it will have to be addressed in a different repo.
cc @jputrino

jputrino:
> @lucacome I wonder if we really need JavaScript checking? that code is not developed in our repo, just copied to our repo. So even if a problem is found, it will have to be addressed in a different repo.
> cc @jputrino

Agreed, @pleshakov -- it would be great to have this type of check in the source repo, but I don't think it's needed here.

lucacome:
@pleshakov I thought it wouldn't hurt to have a check in this repo, since the code is here, but I've removed it.

pleshakov (approved):

PR #1850, opened 2021-08-11
Support Ingress and Ingress class v1

pleshakov:
### Proposed changes
- Use Ingress v1 in the code
- Reject Ingresses with resource backends, because the IC doesn't support such backends:
```yaml
          - path: /icons
            backend:
              resource:
                apiGroup: k8s.example.com
                kind: StorageBucket
                name: icon-assets
```
- Use IngressClass v1 in the code
- Remove k8s < 1.19 from nightly

Note: the existing tests work for k8s 1.19-1.21
in a separate PR we will update the tests so that they work for k8s 1.19-1.22

The PR doesn't change the minimum supported k8s version  for the IC - that will be done separately.
The PR doesn't update any docs - that will be done separately.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -121,7 +121,7 @@ jobs:
     needs: [build, binary, unit-tests]
     strategy:
       matrix:
-        k8s: [1.21.1, 1.20.7, 1.19.11, 1.18.19, 1.17.17, 1.16.15]
+        k8s: [1.21.1, 1.20.7, 1.19.11]
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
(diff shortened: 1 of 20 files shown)


pleshakov:
(forced pushed to remove linting errors)

.github/workflows/nightly.yml
```diff
@@ -121,7 +121,7 @@ jobs:
     needs: [build, binary, unit-tests]
     strategy:
       matrix:
-        k8s: [1.21.1, 1.20.7, 1.19.11, 1.18.19, 1.17.17, 1.16.15]
+        k8s: [1.21.1, 1.20.7, 1.19.11]
```
lucacome, line 124:
```suggestion
        k8s: [1.21.2, 1.20.7, 1.19.11]
```
new minor version of 1.21

lucacome (approved):

.github/workflows/nightly.yml
vepatel: no 1.22?

soneillf5 (approved):

.github/workflows/nightly.yml
pleshakov: @lucacome I'll send a separate PR. it will also change the version in the edge version

pleshakov: @vepatel yep, no 1.22 for now

PR #1851, opened 2021-08-12
Update docs and examples to use networking.k8s.io/v1

lucacome:
- Updated docs, examples and tests to use `networking.k8s.io/v1`
- Update edge pipeline to use `1.22.0`


examples/externalname-services/README.md
```diff
@@ -53,9 +53,11 @@ spec:
       paths:
       - path: /
```
tomasohaodha, line 54: Do we need pathType here?

tomasohaodha (changes requested):

docs/content/configuration/ingress-resources/basic-configuration.md
```diff
@@ -63,17 +69,21 @@ Starting from Kubernetes 1.18, you can use the following new features:
...
       pathType: Exact
       backend:
-        serviceName: tea-svc
-        servicePort: 80
+          service:
+            name: tea-svc
+            port:
+              number: 80
     - path: /coffee
       pathType: ImplementationSpecific # default
```
pleshakov, line 77: it is mandatory, not default in v1, right? shall we remove the default comment?

I think we need to rewrite this doc a bit and merge `New Features Available in Kubernetes 1.18 and Above` into the rest of the doc, because with NIC 1.13, there will be only v1. I will create a PR for that

pleshakov (approved):

docs/content/configuration/ingress-resources/basic-configuration.md
lucacome, line 77:
```suggestion
      pathType: ImplementationSpecific
```

ciarams87 (approved):

PR #1855, opened 2021-08-13
Update minimum k8s version to 1.19

soneillf5:
### Proposed changes
Update the minimum k8s version to 1.19 and removing the redundant removing the flag `-use-ingress-class-only argument` 

### Checklist


CHANGELOG.md
```diff
@@ -210,14 +210,14 @@ UPGRADE:
 * For NGINX Plus, please build your own image using the 1.9.0 source code.
 * For Helm, use version 0.7.0 of the chart.
```
pleshakov, line 212: no need to remove the -use-ingress-class-only mentions from the changelog

cmd/nginx-ingress/main.go
```diff
@@ -80,16 +80,10 @@ var (
 	For Kubernetes < 1.18, the Ingress Controller only processes resources that belong to its class -
```
pleshakov, line 80: Since support for K8s < 1.19 was also removed, we can make additional editing:

A class of the Ingress controller.
A corresponding IngressClass resource with the name equal to the class must be deployed. Otherwise,
	the Ingress Controller will fail to start.
The Ingress controller only processes resources that belong to its class - i.e. have the "ingressClassName" field resource equal to the class.
The Ingress Controller processes all the VirtualServer/VirtualServerRoute/TransportServer resources that do not have the "ingressClassName" field.

What do you think?

(the corresponding cli arg doc, helm values file and two docs will have to be updated as well)

deployments/helm-chart/README.md
```diff
@@ -172,8 +172,7 @@ Parameter | Description | Default
 `controller.volumeMounts` | The volumeMounts of the Ingress controller pods. | []
```
pleshakov, line 172: this doc https://github.com/nginxinc/kubernetes-ingress/blob/master/docs/content/installation/installation-with-helm.md also needs to be updated

pleshakov (commented):
@soneillf5 since PR addresses two things - min version and -use-ingress-class-only argument, could we update the name and the description of it to include -use-ingress-class-only?

note that this `/installation/running-multiple-ingress-controllers.md:` doc also mentions `-use-ingress-class-only` argument

soneillf5:
@pleshakov I can't find either the file `/installation/running-multiple-ingress-controllers.md` or any mentions of `-use-ingress-class-only` in this project. This PR is only able to address this codebase.

CHANGELOG.md
soneillf5: done

cmd/nginx-ingress/main.go
soneillf5: updated

deployments/helm-chart/README.md
soneillf5: updated. I realised during the docs change I had excluded that folder to keep the project functional. It's fixed now but searching and other IDE features were ignoring that folder. Fixed now!

pleshakov (approved):

lucacome (approved):

PR #1877, opened 2021-08-17
Use release branch for k8s python client

lucacome:
Seems like that with the branch the source folder is not created.


ciarams87 (approved):

dreesf5 (approved):
Looks good.

Issue #1893, opened 2021-08-21
Enable setting gzip support for OIDC policy


Binsabbar:
**Auth0: OIDC authorization code sent but token response is not JSON**

**Describe the bug** When using OIDC policy in VirtualServer, I faced an issue with token exchange. In the logs I get `OIDC authorization code sent but token response is not JSON`

[...]

PR #1894, opened 2021-08-21
Add accept-encoding header for oidc flow

Binsabbar:
add gzip as accept-encoding header to fix OIDC authorization code sent but token response is not JSON

### Proposed changes
Add accept encoding to resolve code exchange for a token. this is linked to #1893

### Checklist

### DIFF
--- a/internal/configs/oidc/oidc.conf
+++ b/internal/configs/oidc/oidc.conf
@@ -15,6 +15,7 @@
         proxy_method GET;                             # In case client request was non-GET
         proxy_set_header Content-Length "";           # ''
         proxy_pass $oidc_jwt_keyfile;                 # Expecting to find a URI here
+        proxy_set_header    Accept-Encoding "gzip";
         proxy_ignore_headers Cache-Control Expires Set-Cookie; # Does not influence caching
     }
 
@@ -39,6 +40,7 @@
         internal;
         proxy_ssl_server_name on; # For SNI to the IdP
         proxy_set_header      Content-Type "application/x-www-form-urlencoded";
+        proxy_set_header      Accept-Encoding "gzip";
         proxy_set_body        "grant_type=authorization_code&client_id=$oidc_client&$args&redirect_uri=$redirect_base$redir_location";
         proxy_method          POST;
         proxy_pass            $oidc_token_endpoint;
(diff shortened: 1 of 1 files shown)


internal/configs/oidc/oidc.conf
```diff
@@ -15,6 +15,7 @@
         proxy_method GET;                             # In case client request was non-GET
         proxy_set_header Content-Length "";           # ''
         proxy_pass $oidc_jwt_keyfile;                 # Expecting to find a URI here
+        proxy_set_header    Accept-Encoding "gzip";
```
brianehlert, line 18: From my understanding this has the downstream impact of setting this as the default for all customers instead of implement this as an option that can be enabled or disabled as needed by any customer.

Binsabbar:
would it be better to add headers via Policy itself? though I think it is making it more complicated this way. Why I think gzip will not affect other providers? because Nginx can handle gzip (which is the client of the request to Auth Identity)

Binsabbar:
Hi @lucacome , is this planned to be merged?

lucacome:
Hi @Binsabbar

we use https://github.com/nginxinc/nginx-openid-connect under the hood. I think it would make more sense to open an issue/PR there, since they're the experts on this and also to avoid diverging from that repo.

Let me know what you think about this.

Binsabbar:
> Hi @Binsabbar
> 
> we use https://github.com/nginxinc/nginx-openid-connect under the hood. I think it would make more sense to open an issue/PR there, since they're the experts on this and also to avoid diverging from that repo.
> 
> Let me know what you think about this.

Do you mean the code in nginx is actually referencing the implementation in https://github.com/nginxinc/nginx-openid-connect or it auto-generate the current config from there somehow?

Binsabbar:
> > Hi @Binsabbar
> > we use https://github.com/nginxinc/nginx-openid-connect under the hood. I think it would make more sense to open an issue/PR there, since they're the experts on this and also to avoid diverging from that repo.
> > Let me know what you think about this.
> 
> Do you mean the code in nginx is actually referencing the implementation in https://github.com/nginxinc/nginx-openid-connect or it auto-generate the current config from there somehow?

Just checked there, they have it as something I need to add: https://github.com/nginxinc/nginx-openid-connect#troubleshooting 

>_**Check for OIDC authorization code sent but token response is not JSON. messages in the error log. This is common when NGINX Plus cannot decompress the IdP's response. Add the following configuration snippet to the /_jwks_uri and /_token locations to openid_connect.server_conf:**_

```
    proxy_set_header Accept-Encoding "gzip";
```

PR #1913, opened 2021-08-31
Use composite actions for Smoke Tests

lucacome:
- Moved smoke tests to composite action to avoid duplication in workflows
- Removed the need to pass NGINX_VERSION to the build


vepatel (approved):

ciarams87 (approved):
🎉  looks great! 🚀

PR #1926, opened 2021-09-02
Fix increased IC pod startup time when hundreds VSRs are used

pleshakov:
### Proposed changes

Previously, the IC would update the status of a VSR (via an API call to the k8s API) even if the status of the VSR wasn't changed. The unnecessary API calls lead to the increased IC pod startup time for cases when a VS had multiple (hundreds) VSRs.

The PR fixes the problem - the IC will skip the update if the status of a VSR isn't changed.

The PR includes a test that creates a VS with multiple VSRs and then checks the startup time of the pod.

The output is for 500 VSRs:

Before: gave up after 15 mins 

Now: 201 seconds 
```
All pods are ContainersReady
All pods came up in 201 seconds
Scale a deployment 'nginx-ingress': complete
```

### DIFF
--- a/tests/suite/custom_resources_utils.py
+++ b/tests/suite/custom_resources_utils.py
@@ -260,15 +260,31 @@ def create_virtual_server_from_yaml(
     print("Create a VirtualServer:")
     with open(yaml_manifest) as f:
         dep = yaml.safe_load(f)
+
+    return create_virtual_server(custom_objects, dep, namespace)
+
+
+def create_virtual_server(
+    custom_objects: CustomObjectsApi, vs, namespace
+) -> str:
+    """
+    Create a VirtualServer.
+
+    :param custom_objects: CustomObjectsApi
+    :param vs: a VirtualServer
+    :param namespace:
+    :return: str
+    """
+    print("Create a VirtualServer:")
     try:
         custom_objects.create_namespaced_custom_object(
-            "k8s.nginx.org", "v1", namespace, "virtualservers", dep
+            "k8s.nginx.org", "v1", namespace, "virtualservers", vs
         )
-        print(f"VirtualServer created with name '{dep['metadata']['name']}'")
-        return dep["metadata"]["name"]
+        print(f"VirtualServer created with name '{vs['metadata']['name']}'")
+        return vs["metadata"]["name"]
     except ApiException as ex:
         logging.exception(
-            f"Exception: {ex} occurred while creating VirtualServer: {dep['metadata']['name']}"
+            f"Exception: {ex} occurred while creating VirtualServer: {vs['metadata']['name']}"
         )
         raise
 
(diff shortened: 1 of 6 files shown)


tests/suite/test_batch_startup_times.py
```diff
@@ -503,3 +507,88 @@ def test_ap_waf_policy_vs_batch_start(
...
 
         assert num is None
+
+
+##############################################################################################################
+
+VSR_COUNT=500
+
+@pytest.fixture(scope="class")
+def vs_vsr_setup(
```
pleshakov, line 517: this function doesn't remove any created resources because k8s will remove them when the test_namespace is deleted.

tests/suite/test_batch_startup_times.py
```diff
@@ -503,3 +507,88 @@ def test_ap_waf_policy_vs_batch_start(
...
+
+        metrics_url = f"http://{ingress_controller_endpoint.public_ip}:{ingress_controller_endpoint.metrics_port}/metrics"
+
+        assert (
+            get_total_vs(metrics_url, "nginx") == "1"
+            and get_total_vsr(metrics_url, "nginx") == str(VSR_COUNT)
+            and get_last_reload_status(metrics_url, "nginx") == "1"
+        )
+
+        assert num is None
```
pleshakov, line 594: I took this assert from the other tests - not fully sure though about it

vepatel: this is to make sure that it scaled down to 0 before starting to scale up but not necessary anymore since we've better checks now.

tests/suite/test_batch_startup_times.py
vepatel: `test_namespace` fixture is class scoped not method so if there are other tests in same class then those resources will remain until all the tests are executed.

tests/suite/custom_resources_utils.py
```diff
@@ -697,11 +713,23 @@ def create_v_s_route_from_yaml(custom_objects: CustomObjectsApi, yaml_manifest,
     with open(yaml_manifest) as f:
         dep = yaml.safe_load(f)
 
+    return create_v_s_route(custom_objects, dep, namespace)
+
+def create_v_s_route(custom_objects: CustomObjectsApi, vsr, namespace) -> str:
```
vepatel, line 718: there's already a function called `create_v_s_route_from_yaml` in this file, which so no need to read a yaml in test and pass contents to a new function.

tests/suite/custom_resources_utils.py
```diff
@@ -260,15 +260,31 @@ def create_virtual_server_from_yaml(
     print("Create a VirtualServer:")
     with open(yaml_manifest) as f:
         dep = yaml.safe_load(f)
+
+    return create_virtual_server(custom_objects, dep, namespace)
+
+
+def create_virtual_server(
```
vepatel, line 267: there's already a function called `create_virtual_server_from_yaml` so no need to read yaml and pass it to a new fucntion.

vepatel (changes requested):
The two new functions in `custom_resource_utils.py` are not needed as there are pre-existing functions to perform those tasks.

vepatel (approved):

soneillf5 (approved):

PR #1939, opened 2021-09-08
AP: enable FQDN as syslog destination

rafwegv:
### Proposed changes
Add checks that enable an FQDN be specified as the destination for security logs

### Checklist

### DIFF
--- a/examples/appprotect/README.md
+++ b/examples/appprotect/README.md
@@ -40,13 +40,6 @@ $ kubectl create -f cafe.yaml
     $ kubectl create -f ap-apple-uds.yaml
     ```
 4. Create an Ingress Resource:
-
-    Update the `appprotect.f5.com/app-protect-security-log-destination` annotation from `cafe-ingress.yaml` with the ClusterIP of the syslog service. For example, if the IP is `10.101.21.110`:
-    ```yaml
-    . . .
-    appprotect.f5.com/app-protect-security-log-destination: "syslog:server=10.101.21.110:514"
-    ```
-    Create the Ingress Resource:
     ```
     $ kubectl create -f cafe-ingress.yaml
     ```
--- a/internal/k8s/appprotect/app_protect_resources.go
+++ b/internal/k8s/appprotect/app_protect_resources.go
@@ -73,13 +73,14 @@ func validateAppProtectLogConf(logConf *unstructured.Unstructured) error {
 }
 
 var (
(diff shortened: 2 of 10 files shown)


soneillf5 (approved):
Hi @rafwegv The changes look good

examples/appprotect/README.md
```diff
@@ -41,10 +41,11 @@ $ kubectl create -f cafe.yaml
     ```
 4. Create an Ingress Resource:
 
-    Update the `appprotect.f5.com/app-protect-security-log-destination` annotation from `cafe-ingress.yaml` with the ClusterIP of the syslog service. For example, if the IP is `10.101.21.110`:
+    Update the `appprotect.f5.com/app-protect-security-log-destination` annotation from `cafe-ingress.yaml` with the FQDN of the syslog service. 
+    For example, if the service name is `syslog-svc` and it is in the `default` namespace (You can also use the service ClusterIP):
     ```yaml
     . . .
-    appprotect.f5.com/app-protect-security-log-destination: "syslog:server=10.101.21.110:514"
+    appprotect.f5.com/app-protect-security-log-destination: "syslog:server=syslog-svc.default:514"
```
pleshakov, line 48: Can we add this directly to the Ingress resource so that we don't need the reader to ask to do that? Previously, the IP was something that would be different from env to env, but now we have a "static" DNS name, so we can just use it in the Ingress annotation.
can we also do a similar update to the waf policy example?  https://github.com/nginxinc/kubernetes-ingress/tree/master/examples-of-custom-resources/waf

internal/k8s/appprotect/app_protect_resources.go
```diff
@@ -105,6 +106,10 @@ func ValidateAppProtectLogDestination(dstAntn string) error {
 		return nil
 	}
 
+	if logDstFQDNEx.MatchString(ipstr) {
+		return nil
+	}
+
 	if net.ParseIP(ipstr) == nil {
 		return fmt.Errorf("Error parsing host: %v is not a valid ip address", ipstr)
```
pleshakov, line 114: do we need to update the message here, "is not a valid ip address or host" ?

pleshakov (commented):
Hi @rafwegv 

Please see my comments

Additionally:

(1) Shall we update the doc for  the `logDest` field of the WAF Policy? (https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#waf). the annotation doc seems fines.

(2) Would it be possible to update and simplify the AppProtect tests so that they no longer rely on fetching the svc IP address, but rather use the DNS name?

PR #1968, opened 2021-09-15
Use k8s version in the filename for nightly tests

lucacome:
The nightly doesn't have a marker and the action to upload results was having issues with that, so we can use k8s instead.


pleshakov (approved):

soneillf5:
Hi Luca, 
Can you go into more detail on this change? What requires the marker and why do we use the k8s version?

vepatel (approved):

vepatel:
@soneillf5 I think it's a fix for 
Screenshot 2021-09-16 at 09 30 01

ciarams87 (approved):

PR #2009, opened 2021-09-24
Increase default upstream zone size for NGINX Plus

pleshakov:
### Proposed changes
NGINX Plus R25 allocates more memory for storing upstream server (peer) data: +720 bytes per peer. This means upstream server zones will use more memory to accommodate that data.

If a zone is full, NGINX Plus will fail to reload and fail to add more upstream servers via the API.

To prevent reload failures after an upgrade to R25, this commit increases the default upstream zone size for NGINX Plus from 256K to 512K.

### DIFF
--- a/tests/suite/test_annotations.py
+++ b/tests/suite/test_annotations.py
@@ -165,7 +165,7 @@ def annotations_grpc_setup(request,
 @pytest.mark.ingresses
 @pytest.mark.parametrize('annotations_setup', ["standard", "mergeable"], indirect=True)
 class TestAnnotations:
-    def test_nginx_config_defaults(self, kube_apis, annotations_setup, ingress_controller_prerequisites):
+    def test_nginx_config_defaults(self, kube_apis, annotations_setup, ingress_controller_prerequisites, cli_arguments):
         print("Case 1: no ConfigMap keys, no annotations in Ingress")
         result_conf = get_ingress_nginx_template_conf(kube_apis.v1,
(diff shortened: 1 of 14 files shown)


tests/suite/test_annotations.py
```diff
@@ -318,7 +322,7 @@ def test_upstream_zone_size_0(self, cli_arguments, kube_apis,
         if cli_arguments["ic-type"] == "nginx-plus-ingress":
             print("Run assertions for Nginx Plus case")
             assert "zone " in result_conf
-            assert " 256k;" in result_conf
+            assert " 512k;" in result_conf
```
ciarams87, line 325: This is causing failures in the pipeline, although for some reason it is not showing up in the PR: https://gitlab.com/f5/nginx/kic/kubernetes-ingress/-/jobs/1623726013

pleshakov: thanks for pointing out. forgot to update the template for N+ for Ingress, it still uses 256k default:
```
	zone {{$upstream.Name}} {{if ne $upstream.UpstreamZoneSize "0"}}{{$upstream.UpstreamZoneSize}}{{else}}256k{{end}};
```

lucacome (approved):
I don't think we need to have two different sizes for OSS and Plus (since it's just an increase of 256k) and the added logic that comes with it, but I'm approving the PR since it's time-sensitive.

ciarams87 (approved):

PR #1994, opened 2021-09-22
Release 2.0.0

ciarams87:
### Proposed changes
Release 2.0.0

### Checklist
Before creating a PR, run through this checklist and mark each as complete.

- [ ] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/master/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that all unit tests pass after adding my changes
- [ ] I have updated necessary documentation
- [ ] I have rebased my branch onto master
- [ ] I will ensure my PR is targeting the master branch and pulling from my branch from my own fork

PR #2029, opened 2021-09-30
Release 2.0.0 (#1994)

ciarams87:
Cherry-pick Release 2.0.0 commit to master.

### Proposed changes

### Checklist


lucacome (changes requested):
I think the other one was good, there's not much we can do about the cherry-picks, we have to do it better next time. 
The only thing is that we can't squash or rebase, because it will not keep the tag. You'll just have to do the merge commit.
I think we should reopen the other one.

lucacome:
oh I just noticed your comment
>There was merge conflicts so I couldn't do a clean merge with the tag (as the tag is immutable, I couldn't change it). I'll try a different tactic, but we still may not get the tag in master.

I'm gonna try and open one.

PR #2047, opened 2021-10-06
Python test

soneillf5:
### Proposed changes
This change adds a python test for the DOS feature. There are two tests, one to confirm the config generated in nginx conf is correct and the other checks the functionality by making a request and seeing the dos log in the syslog server.


tests/suite/fixtures.py
```diff
@@ -670,6 +671,140 @@ def fin():
...
+        print("Remove ap-rbac")
+        cleanup_rbac(kube_apis.rbac_v1, rbac)
+        print("Remove the IC:")
+        delete_ingress_controller(
+            kube_apis.apps_v1_api, name, cli_arguments["deployment-type"], namespace
+        )
+        pytest.fail("IC setup failed")
+
+    def fin():
+        print("--------------Cleanup----------------")
```
pleshakov, line 789: should we also delete non-DOS crds?

tests/suite/test_dos.py
```diff
@@ -0,0 +1,192 @@
...
+        )
+        ingress_host = get_first_ingress_host_from_yaml(src_ing_yaml)
+
+        print("--------- Run test while DOS module is enabled with correct policy ---------")
+
+        ensure_response_from_backend(dos_setup.req_url, ingress_host, check404=True)
+
+        print("----------------------- Send request ----------------------")
+        response_block = requests.get(
+            dos_setup.req_url, headers={"host": "dos.example.com"}, verify=False
```
pleshakov, line 173: what is the difference between this request and the request in line 181? they look the same. why do we need two?

tests/suite/fixtures.py
```diff
@@ -670,6 +671,140 @@ def fin():
     request.addfinalizer(fin)
 
 
+@pytest.fixture(scope="class")
+def crd_ingress_controller_with_dos(
```
pleshakov, line 675: this function `crd_ingress_controller_with_dos` is very similar to `crd_ingress_controller_with_ap` and `crd_ingress_controller` - they all deploy and delete IC CRDs (vs,ts,gc...)

would it make sense to have a separate fixture that is responsible for that? `crd_ingress_controller_with_dos`, `crd_ingress_controller_with_ap` and `crd_ingress_controller` can all depend on it. that will reduce the duplication

tests/suite/test_dos.py
soneillf5: well spotted, de-duped

tests/suite/fixtures.py
soneillf5: this is moot now that the crds are handled by a common fixture

PR #2060, opened 2021-10-07
Release 1.12.2

lucacome:
Docs changes for 1.12.2


docs-web/releases.md
```diff
@@ -1,5 +1,20 @@
 # Releases
 
+## NGINX Ingress Controller 1.12.2
+
+7 October 2021
+
+FIXES:
+* [2048](https://github.com/nginxinc/kubernetes-ingress/pull/2048) Use release specific repo for NGINX Plus on Debian.
```
pleshakov, line 8: I wonder if this is clear enough for users to understand what this release tries to fix.

How about this?
```suggestion
* [2048](https://github.com/nginxinc/kubernetes-ingress/pull/2048) Use release specific repo for NGINX Plus on Debian. This fixes an error when building Debian-based images with NGINX Plus: previously, building an image would fail with the error `Package 'nginx-plus-r24' has no installation candidate`. The bug first appeared when NGINX Plus R25 was released on 28 September 2021.
```suggestion

PR #2062, opened 2021-10-07
Release 2.0.1

lucacome:
Docs changes for 2.0.1


docs/content/releases.md
```diff
@@ -1,11 +1,45 @@
...
 doctypes: ["concept"]
 toc: true
 ---
 
+## NGINX Ingress Controller 2.0.1
+
+07 Oct 2021
+
+FIXES:
+* [2051](https://github.com/nginxinc/kubernetes-ingress/pull/2051) Use release specific repo for NGINX Plus on Debian. This fixes an error when building Debian-based images with NGINX Plus: previously, building an image would fail with the error `Package 'nginx-plus-r24' has no installation candidate`. The bug first appeared when NGINX Plus R25 was released on 28 September 2021.
```
pleshakov, line 14: clarified because it only affects the NAP image and there is no need for first appeared, because 2.0.0 was released with the bug - the build was always failing
```suggestion
* [2051](https://github.com/nginxinc/kubernetes-ingress/pull/2051) Use release specific repo for NGINX Plus on Debian. This fixes an error when building the Debian-based image with NGINX Plus and App Protect: previously, building the image would fail with the error `Package 'nginx-plus-r24' has no installation candidate`.
```

PR #2067, opened 2021-10-08
use dns name for syslog destination service

soneillf5:
### Proposed changes
As the IP address of the syslog service can change, we would like to use a reference that des not change.
In this change, I've used the k8s dns name of the service. 

I've confirmed the change works by updating the python test and getting that to pass.

I've also updated the validation, separating it from the app protect validation.

I've also removed the 'syslog:server=' value from the annotation. The annotation has to have this value, so instead of validating it has the value, we just inject it.

### Checklist


internal/k8s/validation.go
```diff
@@ -499,7 +499,7 @@ func validateAppProtectDosOnlyAnnotation(context *annotationValidationContext) f
 func validateAppProtectDosLogDestAnnotation(context *annotationValidationContext) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	err := api_validation.ValidateAppProtectLogDestination(context.value)
+	err := api_validation.ValidateAppProtectDosAccessLogDest(context.value)
```
pasmant, line 502: The `ValidateAppProtectLogDestination` was for `app-protect-dos-security-log-destination` and `ValidateAppProtectDosAccessLogDest` was for `app-protect-dos-access-log-destination
Now u joined them together so I think change the name for the function for example to `ValidateAppProtectDosLogDest` will be more correct

pkg/apis/configuration/validation/dos.go
```diff
@@ -33,26 +32,26 @@ func ValidateAppProtectDosLogConf(logConf *unstructured.Unstructured) error {
 	return nil
 }
 
-var accessLog = regexp.MustCompile(`^(((\d{1,3}\.){3}\d{1,3}):\d{1,5})$`)
+var (
+	dosLogDstEx = regexp.MustCompile(`(\S+:\d{1,5})|stderr`)
+)
```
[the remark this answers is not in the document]
soneillf5: yes, if we allow dns names, they can almost be any string. Is there some validation you think we can do?

pkg/apis/configuration/validation/policy.go
```diff
@@ -329,7 +329,7 @@ func validateDosLogConf(logConf, logDest string, fieldPath *field.Path) field.Er
 		}
 	}
 
-	err := ValidateAppProtectLogDestination(logDest)
+	err := ValidateAppProtectDosAccessLogDest(logDest)
```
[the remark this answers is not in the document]
soneillf5: fixed I think

pkg/apis/configuration/validation/dos.go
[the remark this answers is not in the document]
pasmant: Suggestion:
use the format of ip=191.2.2.2 or dns=dns.bla.bla
in dos.go:

> var (
> 	dosLogDstEx = regexp.MustCompile(`(\S+:\d{1,5})|stderr`)
> )
> 
> // ValidateAppProtectDosLogDest validates destination for log configuration
> func ValidateAppProtectDosLogDest(dstAntn string) error {
> 	errormsg := "Error parsing App Protect Dos Log config: Destination must follow format: <ip=ip-address | localhost | dns=dns name>:<port> or stderr"
> 	if !dosLogDstEx.MatchString(dstAntn) {
> 		return fmt.Errorf("%s Log Destination did not follow format", errormsg)
> 	}
> 	if dstAntn == "stderr" {
> 		return nil
> 	}
> 
> 	dstchunks := strings.Split(dstAntn, ":")
> 
> 	// // This error can be ignored since the regex check ensures this string will be parsable
> 	port, _ := strconv.Atoi(dstchunks[1])
> 	if port > 65535 || port < 1 {
> 		return fmt.Errorf("error parsing port: %v not a valid port number", port)
> 	}
> 
> 	dstchunks = strings.Split(dstchunks[0], "=")
> 
>     if dstchunks[0] == "ip" {
>         if net.ParseIP(dstchunks[1]) == nil {
>             return fmt.Errorf("Error parsing host: %v is not a valid ip address", dstchunks[0])
>         }
>     }
> 
> 	return nil
> }

in configuration.go:

> func generateDosLogDest(dest string) string {
> 	if dest == "stderr" {
> 		return dest
> 	}
> 	if strings.HasPrefix(dest, "ip=") || strings.HasPrefix(dest, "dns=") {
> 		dstchunks := strings.Split(dest, "=")
> 		return dstchunks[1]
> 	}
> 	return "syslog:server=" + dest
> }

internal/configs/virtualserver.go
```diff
@@ -1078,7 +1078,7 @@ func (p *policiesCfg) addDosConfig(
 		}
 	}
 
-	p.Dos.ApDosAccessLogDest = dos.DosAccessLogDest
+	p.Dos.ApDosAccessLogDest = generateDosLogDest(dos.DosAccessLogDest)
```
pasmant, line 1081: need to change to:
```
	if dos.DosAccessLogDest != "" {
		p.Dos.ApDosAccessLogDest = generateDosLogDest(dos.DosAccessLogDest)
	}
```

pasmant: Also in annotations.go in line 426:
`cfgParams.AppProtectDosAccessLogDst = appProtectDosAccessLogDst`
Need to change:
`cfgParams.AppProtectDosAccessLogDst = generateDosLogDest(appProtectDosAccessLogDst)
`

soneillf5: I think these are fixed now.

internal/k8s/validation.go
soneillf5: fixed

internal/configs/virtualserver.go
```diff
@@ -1071,14 +1071,16 @@ func (p *policiesCfg) addDosConfig(
...
-			p.Dos.ApDosLogConf = fmt.Sprintf("%s %s", logConfPath, logDest)
+			p.Dos.ApDosLogConf = fmt.Sprintf("%s %s", logConfPath, generateDosLogDest(logDest))
 		} else {
 			res.addWarningf("Dos policy %s references an invalid or non-existing log config %s", polKey, logConfKey)
 			res.isError = true
 		}
 	}
 
-	p.Dos.ApDosAccessLogDest = dos.DosAccessLogDest
+	if p.Dos.ApDosAccessLogDest != "" {
```
pasmant, line 1081: `if p.Dos.ApDosAccessLogDest != "" {`
need to be 
`if dos.DosAccessLogDest != "" {`

PR #2074, opened 2021-10-12
Update JWT library to golang-jwt/jwt

lucacome:
https://github.com/dgrijalva/jwt-go is no longer maintained, the development has moved to https://github.com/golang-jwt/jwt. 

This PR replaces the old library with the new one. Also added some simple unit tests for claims validation.

### DIFF
--- a/cmd/nginx-ingress/aws.go
+++ b/cmd/nginx-ingress/aws.go
@@ -1,4 +1,4 @@
-// +build aws
+//go:build aws
 
 package main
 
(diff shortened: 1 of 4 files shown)


ciarams87 (approved):

cmd/nginx-ingress/aws.go
```diff
@@ -1,4 +1,4 @@
-// +build aws
+//go:build aws
```
soneillf5, line 1: which tool uses this tag? I'm not familiar with it

lucacome: So you can do something like `go build -tags=aws github.com/nginxinc/kubernetes-ingress/cmd/nginx-ingress` (and GoReleaser does the same)

PR #2076, opened 2021-10-12
Add new func validateEscapedString

pasmant:
use the func for dos monitor and name validation

### Proposed changes

### Checklist

### DIFF
--- a/pkg/apis/configuration/validation/common.go
+++ b/pkg/apis/configuration/validation/common.go
@@ -17,6 +17,14 @@ const (
 
 var escapedStringsFmtRegexp = regexp.MustCompile("^" + escapedStringsFmt + "$")
 
+func validateEscapedString(body string, examples ...string) error {
+	if !escapedStringsFmtRegexp.MatchString(body) {
+		msg := validation.RegexError(escapedStringsErrMsg, escapedStringsFmt, examples...)
+		return fmt.Errorf(msg)
+	}
+	return nil
+}
+
 func validateVariable(nVar string, validVars map[string]bool, fieldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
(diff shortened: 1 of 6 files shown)


soneillf5 (approved):

pkg/apis/configuration/validation/transportserver.go
```diff
@@ -249,8 +249,8 @@ func validateHashLoadBalancingMethod(method string, fieldPath *field.Path, isPlu
 		allErrs = append(allErrs, varErrs...)
 	}
 
-	if !escapedStringsFmtRegexp.MatchString(method) {
-		msg := fmt.Sprintf("invalid value for hash: %v", validation.RegexError(escapedStringsErrMsg, escapedStringsFmt))
+	if err := validateEscapedString(method); err != nil {
+		msg := fmt.Sprintf("invalid value for hash: %v",err.Error())
```
pleshakov, line 253: there is no need to call Error() here
```suggestion
		msg := fmt.Sprintf("invalid value for hash: %v", err)
```

pkg/apis/configuration/validation/common.go
```diff
@@ -219,3 +219,11 @@ func validateParameter(nPar string, validParams map[string]bool, fieldPath *fiel
 	}
 	return allErrs
 }
+
+func validateEscapedString(body string, examples ...string) error {
```
pleshakov, line 223: better to move this function right below `var escapedStringsFmtRegexp = regexp.MustCompile("^" + escapedStringsFmt + "$")` so that var is close to the function where it is used. this will be consistent with the rest of the validation code

pleshakov (approved):
looks good. but before merging, (1) could you address the comments? (2) could you make sure the code is formatted using gofmt?

pkg/apis/configuration/validation/dos.go
```diff
@@ -64,6 +64,10 @@ func ValidateAppProtectDosName(name string) error {
 		return fmt.Errorf("App Protect Dos Name max length is %v", MaxNameLength)
 	}
 
+	if err := validateEscapedString(name, "http://www.example.com"); err != nil {
```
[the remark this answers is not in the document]
pasmant: change to protected-object-one

PR #2107, opened 2021-10-19
Add option to download IC binary

lucacome:
- Added option to extract the binary from the Docker image
- Updated docs
- Small fixes to docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,11 +58,11 @@ endif
 download-binary-docker: ## Download Docker image from which to extract Ingress Controller binary
 	@docker -v || (code=$$?; printf "\033[0;31mError\033[0m: there was a problem with Docker\n"; exit $$code)
 ifeq (${TARGET},download)
-DOWLOAD_TAG := $(shell ./hack/docker.sh $(GIT_COMMIT) $(GIT_TAG))
-ifeq ($(DOWLOAD_TAG),fail)
+DOWNLOAD_TAG := $(shell ./hack/docker.sh $(GIT_COMMIT) $(GIT_TAG))
+ifeq ($(DOWNLOAD_TAG),fail)
 $(error unable to build with TARGET=download, this function is only available when building from a git tag or from the latest commit matching the edge image)
 endif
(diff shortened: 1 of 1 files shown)


hack/docker.sh
```diff
@@ -0,0 +1,26 @@
+#! /bin/bash
```
pleshakov, line 1: to double check, does this script work on Linux?

hack/docker.sh
```diff
@@ -0,0 +1,26 @@
...
+
+if [[ ${commit_tag} == ${git_tag} ]]; then
+    # we're on the tag, use the docker image for the tag
+    docker_tag=${git_tag//v/}
+    echo ${docker_tag}
+    exit 0
+
+else
+    # we're on master or a branch, try to download the latest edge and see if sha matches
+    docker pull nginx/nginx-ingress:${docker_tag} >/dev/null 2>&1
```
pleshakov, line 17: do we rebuild the edge images on every commit in the master or we optimize (for example, when no code changes were in the commit?)? if latter is the case, then we might not have an image that matches the commit

hack/docker.sh
lucacome: yep, I've tested it in Debian

hack/docker.sh
lucacome: We rebuild the images for every change except for docs and examples. In that case, you're right, it won't be possible to download the binary. Unfortunately, I don't see a workaround for this unless we want to always use the latest available binary, but I think that could cause more issues.

Similarly, if somebody is on an old commit, there's no way to get the corresponding `edge` image that was built in the past (unless we start tagging the image with the sha).

pleshakov: if this can cause problems, perhaps we shall only support downloading for release tags then?

lucacome: if it's not the latest commit or a tag it will print this message:
> unable to build with TARGET=download, this function is only available when building from a git tag or from the latest commit matching the edge image

do you it's not enough?

pleshakov: > do you it's not enough?

yeah, I think it is enough

PR #2124, opened 2021-10-25
Apply -enable-snippets cli arg to Ingresses

pleshakov:
### Proposed changes

This PR extends the existing `-enable-snippets` cli argument to apply to Ingress resources.  If snippets are not enabled, the Ingress Controller will reject any Ingress resources with snippets annotations.

Previously, the  argument only applied to VirtualServer, VirtualServerRoute and TransportServer resources.

**Note**: this is a breaking change. Because the default value of `-enable-snippets` is `false`,  If you're using snippets in Ingress resources, make sure to explicitly set the `-enable-snippets` to `true` before upgrading the Ingress Controller, so that the new version of the Ingress Controller doesn't reject Ingresses with snippets annotations.

TO-DO:
* [x] Code
* [x] Docs
* [x] Python tests

### DIFF
--- a/internal/configs/annotations.go
+++ b/internal/configs/annotations.go
@@ -66,7 +66,7 @@ var minionInheritanceList = map[string]bool{
 	"nginx.org/fail-timeout":             true,
 }
 
-func parseAnnotations(ingEx *IngressEx, baseCfgParams *ConfigParams, isPlus bool, hasAppProtect bool, enableInternalRoutes bool, enableSnippets bool) ConfigParams {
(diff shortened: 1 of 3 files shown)


adamdecaf:
FYI pushing breaking changes within a bugfix release is a bad practice. We had a production outage due to this. 

From https://semver.org/

> 3. PATCH version when you make backwards compatible bug fixes.

adamdecaf:
Also, can you please update the docs to reflect this CLI flags larger impact in 2.0.3? 

https://docs.nginx.com/nginx-ingress-controller/configuration/global-configuration/command-line-arguments#-enable-snippets

PR #2156, opened 2021-11-03
Add more badges to README

lucacome:
Adding a few badges


tomasohaodha:
For something like this, I have to ask:
What are we trying to communicate with the badges? For example, why is it important to show how large our image is?

lucacome:
Added Artifact Hub and Slack Channel
Preview https://github.com/nginxinc/kubernetes-ingress/blob/chore/add-more-badges/README.md

ciarams87 (approved):

PR #2157, opened 2021-11-03
Move custom-resources with other examples

lucacome:
I think it makes it cleaner to have all the examples inside one folder.


tomasohaodha:
I agree with this change in principle, but would appreciate @brianehlert especially chiming in.

pleshakov:
custom resources is a big group of examples that covers a lot of use cases. however, with the current layout, they are somewhat hidden among Ingress resource examples:
image

imho it would be better if we have the following structure:
- examples
  -- ingresses
  -- custom-resources
  
 Also, perhaps we can come up with a better name than just "custom resources"?  @brianehlert

brianehlert:
I agree that it is valuable.  And that "custom-resources" is too generic / obscure.

consumers look for answers based on the use case first and how to implement is the choice.
I don't think a large 'custom-resources' folder follows this pattern, but is rather an anti-pattern.

tomasohaodha:
I would agree with all previous comments, however I would recommend approving this PR and merging the basic location change. Further changes can come in further PRs after more discussion. Perfect enemy of the good and all that.

tomasohaodha (approved):

pleshakov:
I don't think we should merge the current state of the PR. As I mentioned previously, the new layout hides custom resource related examples. Note that the regular and custom resource examples are supposed to be equal alternatives to each other -- that's why there were two separate folders in the root directory. With the new layout, the custom resource examples are hidden away.

brianehlert (approved):

PR #2160, opened 2021-11-04
Dos/update resource design

soneillf5:
### Proposed changes
Update the design of the DOS resources so the IC resources use a reference rather than config directly.

### Checklist


deployments/common/crds/appprotectdos.f5.com_dosprotectedresources.yaml
```diff
@@ -0,0 +1,63 @@
...
+    plural: dosprotectedresources
+    shortNames:
+      - pr
+    singular: dosprotectedresource
+  scope: Namespaced
+  versions:
+    - name: v1beta1
+      schema:
+        openAPIV3Schema:
+          description: 'DosProtectedResource defines a collection of Dos protected resources. status: preview'
```
tomasohaodha, line 22: Why does DosProtectedResource (singular) define a collection (plural) of protected resources?
Why the preview flag?

deployments/common/crds/appprotectdos.f5.com_dosprotectedresources.yaml
```diff
@@ -0,0 +1,63 @@
...
+              type: object
+              properties:
+                apDosMonitor:
+                  type: string
+                apDosPolicy:
+                  type: string
+                dosAccessLogDest:
+                  type: string
+                dosSecurityLog:
+                  description: DosSecurityLog defines the security log of a Dos policy.
```
tomasohaodha, line 43: Why don't the other properties have their own descriptions?

deployments/helm-chart/crds/appprotectdos.f5.com_dosprotectedresources.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: apiextensions.k8s.io/v1
```
tomasohaodha, line 1: This file is the same as the one above, is there any way to avoid the duplication?

internal/k8s/appprotectdos/app_protect_dos_configuration.go
```diff
@@ -269,7 +356,12 @@ func (fc *FakeConfiguration) DeletePolicy(key string) (changes []Change, problem
 	return changes, problems
 }
 
-// DeleteLogConf deletes an App Protect Dos Log Configuration from App Protect Dos Configuration
+// DeleteLogConf deletes an App Protect Dos Portected resource from App Protect Dos Configuration
```
tomasohaodha, line 359: Typo and I'm not sure it's correct? (DeleteLogConf delete a protected resource?)

deployments/common/crds/appprotectdos.f5.com_dosprotectedresources.yaml
soneillf5: Updated to plural and preview removed.

deployments/common/crds/appprotectdos.f5.com_dosprotectedresources.yaml
soneillf5: Added descriptions, @pasmant can you confirm the descriptions are correct? I have no idea.

deployments/helm-chart/crds/appprotectdos.f5.com_dosprotectedresources.yaml
soneillf5: looks like all the other crds are in here. If there is a change to make in this regard, this PR is not the place for it.

internal/k8s/appprotectdos/app_protect_dos_configuration.go
soneillf5: fixed

deployments/common/crds/appprotectdos.f5.com_dosprotectedresources.yaml
soneillf5: updated from docs

examples-of-custom-resources/dos/apdos-protected.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: apps/v1
```
pasmant, line 1: need to change:
apiVersion: appprotectdos.f5.com/v1beta1

internal/configs/annotations.go
```diff
@@ -40,29 +34,22 @@ var masterBlacklist = map[string]bool{
...
+	"nginx.org/hsts-include-subdomains":                 true,
+	"nginx.org/server-tokens":                           true,
+	"nginx.org/listen-ports":                            true,
+	"nginx.org/listen-ports-ssl":                        true,
+	"nginx.org/server-snippets":                         true,
+	"appprotect.f5.com/app_protect_enable":              true,
+	"appprotect.f5.com/app_protect_policy":              true,
+	"appprotect.f5.com/app_protect_security_log_enable": true,
+	"appprotect.f5.com/app_protect_security_log":        true,
+	"appprotectdos.f5.com/app-protect-dos-resource":     true,
```
soneillf5, line 52: should be plural

pleshakov (commented):
Hi @soneillf5 

Please see my comments

Note: the review doesn't include the python tests.

Additionally:

(1)
Could you validation for the new annotation in `internal/k8s/validation.go` ? The validation needs to check that the value is not empty and is valid name or namespace/name.

(2)
Could you remove the validation for the existing DOS-related annotations from `internal/k8s/validation.go`?


(3) 
I didn't fully test this PR. but still run the IC in a cluster. I noticed a few bugs:
(3.1)
Deleting a Dos Policy doesn't regenerate the config for the VS
(3.2)
Deleting  a DosProtectectedResource results into a failed configuration reload for the VS (`app_protect_dos_enable ;`  causes that) 

(4) Edge cases
The removed Dos policy had good UX for edge cases -- https://github.com/nginxinc/kubernetes-ingress/blob/feature/bados/internal/configs/virtualserver.go#L1027 : if  the policy or the referenced dos policy or log conf were invalid or missing, the IC would generate NGINX configuration with `return 500` and will emit event and update status of the affected VS/VSR. 
This doesn't happen for the DosProtectectedResource - all the errors are silent (only admin can see them in the logs, but not users who create VS/VSRs) and the error-case behaviour is also just to not enable Dos rather than `return 500`.
Could you handle edge cases for VS/VSR for DosProtectectedResource similarly to how we handle them for Policies?

(5) readmes of the dos examples (Ingress and VirtualServer) need to be updated according to the changes in the PR.  Were that not done intentionally? 

(6) What about docs? Are you going to prepare additional PRs?

soneillf5:
@pleshakov Thanks for the review, unfortunately I can't reply inline to the comments, not sure why. I'll add more comments and quote the original below.

soneillf5:
```
could you explain the benefits of using *unstructured.Unstructured instead of directly using the type DosProtectedResources from `pkg/apis/configuration/v1beta1/?

The benefits of using DosProtectedResources I see:

no need to write conversation logic from instruct to appProtectDosResources
no need to have this intermediate type appProtectDosResources
we can reuse the validation that was there for the dos policy, just change it slightly so it works with the DosProtectedResources resource.
```

At the moment, the custom Dos resources are 'ApDosLogConf, ApDosPolicy and DosProtectedResources' they are unstructured and we don't hold their type definition in our codebase. We do this as we would like the Dos team to be responsible for them. I had hoped, in the future, these resources would be moved out of our codebase into a Dos controller. At the moment, the type is present in our codebase in this PR as I'm using the `make update-crds` to build the resource. You mentioned several reasons why it's easier to just use the type, all of which I agree with, but the larger concern is about storing Dos resource definitions and code for handling them. Does that make sense?

soneillf5:
```
config generation for DOS in VSRs is missing. Is it intentional?
```
It wasn't implemented yet. Adding it now.

pleshakov:
Hi @soneillf5 

> At the moment, the custom Dos resources are 'ApDosLogConf, ApDosPolicy and DosProtectedResources' they are unstructured and we don't hold their type definition in our codebase. We do this as we would like the Dos team to be responsible for them. I had hoped, in the future, these resources would be moved out of our codebase into a Dos controller. At the moment, the type is present in our codebase in this PR as I'm using the make update-crds to build the resource. You mentioned several reasons why it's easier to just use the type, all of which I agree with, but the larger concern is about storing Dos resource definitions and code for handling them. Does that make sense?

I see a big difference between in ApDosLogConf, ApDosPolicy and DosProtectedResources:  ApDosLogConf, ApDosPolicy are "containers" of JSON files that the IC just writes to the file systems - it doesn't really care about fields of those resources. In contrast, the IC parses every field DosProtectedResources and also needs to validate those fields. This creates coupling, and I'm not sure such coupling is better than just simply using the go type, which allows removing all the parsing logic. So my point is that we still coupled to that resource, even though we use `*unstructured.Unstructured`

pasmant:
Some issues that I found:

(1)
When creating ingress from yaml inside namespace no directives added to conf file, example
k apply -n test -f ../appprotect-dos/
inside file:
/etc/nginx/conf.d/example-webapp-ingress.conf
No directives that related to dos.

(3)
In virtual server when creating another virtual server but in different namespace, no file was created for the virtual server.

(4)
When changing information for APDosLogConf or APDosPolicy after already ingress yaml was apply, the changes doesnt reflect on the conf file inside the nginx pod

(5)
If apDosPolicy or apDosLogConf are not valid (there is no object of them) the DosProtectedResources will be added also the ingress will be add with success but inside the conf file of the nginx dos related directives will not appear.

(6)
in virtual server yaml file, in case put the dos field inside the path field, example:
  routes:
    - path: /
      dos: dos-protected
      action:
        pass: webapp
The directives of dos will be appear inside the server block and not inside the location block

pleshakov (commented):
Hi @soneillf5  Please see my comments and suggestions. Overall looks good, with some non-major issues pointed out👍

pasmant:
Need to change the readme file of the examples inside:
kubernetes-ingress/examples/appprotect-dos
kubernetes-ingress/examples-of-custom-resources/dos

PR #2161, opened 2021-11-05
Tests improvements

lucacome:
Updating containers, python syntax and default retries, increased verbosity.


soneillf5 (approved):

tests/data/ap-waf/syslog.yaml
```diff
@@ -29,4 +37,34 @@ spec:
...
\ No newline at end of file
+    app: syslog
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: syslog-config
+data:
+  syslog-ng.conf: |-
+    @version: 3.29
```
pleshakov, line 48: just curious, what are the reasons for adding the config?

lucacome: the logs were saying that the connection was refused because the max number of connections was reached. So that is basically the default config from the container and I added max_connections

pleshakov (approved):

PR #2174, opened 2021-11-11
Test Trivy fix to ignore unfixed vulnerabilities

lucacome:
Testing the fix mentioned here https://github.com/aquasecurity/trivy-action/issues/77


lucacome:
Looks like it's working, we can update to the new release once it comes out.

PR #2188, opened 2021-11-18
Reject VS with empty TLS secret reference

pleshakov:
This PR ensures that if the tls section is configured in a VirtualServer, than the secret field must not be empty, unless the wildcard TLS secret is enabled via the -wildcard-tls-secret cli argument.

### DIFF
--- a/cmd/nginx-ingress/main.go
+++ b/cmd/nginx-ingress/main.go
@@ -605,7 +605,7 @@ func main() {
 	controllerNamespace := os.Getenv("POD_NAMESPACE")
 
 	transportServerValidator := cr_validation.NewTransportServerValidator(*enableTLSPassthrough, *enableSnippets, *nginxPlus)
-	virtualServerValidator := cr_validation.NewVirtualServerValidator(*nginxPlus)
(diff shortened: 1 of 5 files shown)


pleshakov:
Closing this, because the fix makes the following supported configuration not supported:

```yaml
apiVersion: k8s.nginx.org/v1
kind: VirtualServer
metadata:
  name: virtual-server-tls
spec:
  host: virtual-server-tls.example.com
  tls:
    secret: "" # no TLS termination
    redirect:
      enable: True
      code: 308
      basedOn: x-forwarded-proto
  upstreams:
  - name: backend2
    service: backend2-svc
    port: 80
  - name: backend1
    service: backend1-svc
    port: 80
  routes:
  - path: "/backend1"
    action:
      pass: backend1
  - path: "/backend2"
    action:
      pass: backend2
```

In the configuration above, the secret is not configured (no TLS termination is needed), but TLS redirect is configured. This is necessary when an load balancer in front of the IC is doing TLS termination (like ELB or F5 BIG-IP) and it is necessary configure a TLS redirect in NGINX.

PR #2189, opened 2021-11-18
Run tests for Plus images

lucacome:
Test the NGINX Plus images in the pipeline

### DIFF
--- a/.github/actions/smoke-tests/action.yaml
+++ b/.github/actions/smoke-tests/action.yaml
@@ -17,6 +17,12 @@ inputs:
   marker:
     description: Marker to use
     required: false
+  nginx-key:
+    description: Nginx key to use
+    required: false
+  nginx-crt:
+    description: Nginx cert to use
+    required: false
 
 outputs:
   test-results-name:
@@ -32,6 +38,16 @@ runs:
         path: ${{ github.workspace }}/dist
         key: nginx-ingress-${{ github.run_id }}-${{ github.run_number }}-single
 
+    - name: Ingress type
+      id: ingress-type
+      run: |
+        plus=""
+        if [[ ${{ inputs.image }} == *plus* ]]; then
+          plus="-plus"
+        fi
+        echo ::set-output name=name::nginx$plus-ingress
+      shell: bash
+
     - name: Docker Buildx
       uses: docker/setup-buildx-action@v1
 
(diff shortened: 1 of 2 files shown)


.github/actions/smoke-tests/action.yaml
```diff
@@ -43,10 +59,13 @@ runs:
         cache-from: type=gha,scope=${{ inputs.image }}
         cache-to: type=gha,scope=${{ inputs.image }},mode=max
         target: goreleaser
-        tags: 'docker.io/nginx/nginx-ingress:${{ inputs.image }}-${{ github.sha }}'
+        tags: 'docker.io/nginx/${{ steps.ingress-type.outputs.name }}:${{ inputs.image }}-${{ github.sha }}'
         load: true
         pull: true
         build-args: BUILD_OS=${{ inputs.image }}
+        secrets: |
+          "nginx-repo.crt=${{ inputs.nginx-crt }}"
```
soneillf5, line 67: where do these certs come from ?

.github/actions/smoke-tests/action.yaml
```diff
@@ -32,6 +38,16 @@ runs:
         path: ${{ github.workspace }}/dist
         key: nginx-ingress-${{ github.run_id }}-${{ github.run_number }}-single
 
+    - name: Ingress type
+      id: ingress-type
+      run: |
+        plus=""
+        if [[ ${{ inputs.image }} == *plus* ]]; then
```
soneillf5, line 45: I'm still not 100% sure about making decisions about an image based on if the string 'plus' is in the name. @ciarams87 ?

soneillf5 (approved):

.github/actions/smoke-tests/action.yaml
lucacome: `inputs.nginx-crt` and key are taking the input value from https://github.com/nginxinc/kubernetes-ingress/pull/2189/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR178 and there are stored as secrets in GitHub like all the other secrets. Does that answer your question? I wasn't really sure which part you were asking about

.github/actions/smoke-tests/action.yaml
lucacome: Just to clarify this is just to figure out which type of tests to run based on the name of the image. All the plus images have `plus` in the name.

ciarams87: Other than hardcoding it elsewhere, there isn't many other good options. The test suite literally does the same thing when judging whether or not it should run the AP tests. I think this is fine.

ciarams87 (approved):

PR #2190, opened 2021-11-18
Add go caching in CI

lucacome:
Shaving off a couple of minutes in every step with caching for golang.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,15 @@ jobs:
           access_token: ${{ secrets.GITHUB_TOKEN }}
       - name: Checkout Repository
         uses: actions/checkout@v2
+      - name: Cache Go controller tools
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-tools-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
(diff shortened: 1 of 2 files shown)


soneillf5:
how have you tested this?

lucacome:
> how have you tested this?

if you look at the logs in the old jobs you can see that go is downloading all the dependencies to run the tests or check for CRDs changes. With this change when there's a cache hit it won't have to download them. You can also compare the times for the tests/build form old runs with this one.

.github/workflows/ci.yml
```diff
@@ -50,6 +50,15 @@ jobs:
...
         uses: actions/checkout@v2
+      - name: Cache Go controller tools
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-tools-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-tools-
```
pleshakov, line 61: what is the reason for configuring `restore-keys`?

lucacome: it's used to search for partial matches if there's a cache miss for the key.

I just used the example from here btw https://github.com/actions/cache/blob/main/examples.md#go---modules

pleshakov (approved):

ciarams87 (approved):

PR #2230, opened 2021-11-30
Add vsr support

soneillf5:
### Proposed changes
Adds support for VSR Dos config.

### Checklist

### DIFF
--- a/internal/configs/virtualserver.go
+++ b/internal/configs/virtualserver.go
@@ -326,6 +326,7 @@ func (vsc *virtualServerConfigurator) GenerateVirtualServerConfig(
 		vsName:         vsEx.VirtualServer.Name,
 	}
 	policiesCfg := vsc.generatePolicies(ownerDetails, vsEx.VirtualServer.Spec.Policies, vsEx.Policies, specContext, policyOpts)
+
 	dosCfg := generateDosCfg(dosResources[""])
 
 	// crUpstreams maps an UpstreamName to its conf_v1.Upstream as they are generated
@@ -572,6 +573,9 @@ func (vsc *virtualServerConfigurator) GenerateVirtualServerConfig(
 				routePoliciesCfg.OIDC = policiesCfg.OIDC
 			}
 			limitReqZones = append(limitReqZones, routePoliciesCfg.LimitReqZones...)
+
+			dosRouteCfg := generateDosCfg(dosResources[r.Path])
+
 			if len(r.Matches) > 0 {
 				cfg := generateMatchesConfig(
 					r,
@@ -591,6 +595,7 @@ func (vsc *virtualServerConfigurator) GenerateVirtualServerConfig(
 					vsc.warnings,
 				)
 				addPoliciesCfgToLocations(routePoliciesCfg, cfg.Locations)
+				addDosConfigToLocations(dosRouteCfg, cfg.Locations)
 
 				maps = append(maps, cfg.Maps...)
 				locations = append(locations, cfg.Locations...)
@@ -602,6 +607,7 @@ func (vsc *virtualServerConfigurator) GenerateVirtualServerConfig(
 				cfg := generateDefaultSplitsConfig(r, upstreamNamer, crUpstreams, variableNamer, len(splitClients), vsc.cfgParams,
 					errorPages, r.Path, locSnippets, vsc.enableSnippets, len(returnLocations), isVSR, vsr.Name, vsr.Namespace, vsc.warnings)
 				addPoliciesCfgToLocations(routePoliciesCfg, cfg.Locations)
+				addDosConfigToLocations(dosRouteCfg, cfg.Locations)
 
 				splitClients = append(splitClients, cfg.SplitClients...)
 				locations = append(locations, cfg.Locations...)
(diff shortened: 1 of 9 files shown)


internal/configs/virtualserver_test.go
```diff
@@ -1361,6 +1361,357 @@ func TestGenerateVirtualServerConfigForVirtualServerWithMatches(t *testing.T) {
...
+							Dos: "test_ns/dos_protected",
+							Action: &conf_v1.Action{
+								Pass: "coffee-v1",
+							},
+						},
+					},
+				},
+			},
+		},
+		//DosProtectedEx: map[string]*DosEx{
```
pleshakov, line 1485: 👀

internal/k8s/controller.go
```diff
@@ -2530,6 +2530,14 @@ func (lbc *LoadBalancerController) createVirtualServerEx(virtualServer *conf_v1.
 			if err != nil {
 				glog.Warningf("Error getting WAF policies for VirtualServerRoute %v/%v: %v", vsr.Namespace, vsr.Name, err)
 			}
+
+			if sr.Dos != "" {
```
pleshakov, line 2534: currently this looks like a bug (I haven't tested it by running the IC in the cluster). 
what will happen if the DOS is configured for NGINX OSS IC or NGINX Plus without DOS enabled? this is related to this comment https://github.com/nginxinc/kubernetes-ingress/pull/2160#discussion_r753627593

internal/configs/virtualserver_test.go
```diff
@@ -1361,6 +1361,357 @@ func TestGenerateVirtualServerConfigForVirtualServerWithMatches(t *testing.T) {
...
+						},
+						{
+							Name:    "coffee-v2",
+							Service: "coffee-svc-v2",
+							Port:    80,
+						},
+					},
+					Subroutes: []conf_v1.Route{
+						{
+							Path: "/coffee",
```
pleshakov, line 1462: because the new code calls `addDosConfigToLocations` for 3 different cases, I suggest testing them all. For that, we can have a VSR with 3 subroutes:
- one a regular around with a `pass` or `proxy` action 
- one subroute with splits
- one subroute with matches (the one that we have now)

internal/configs/virtualserver_test.go
```diff
@@ -1361,6 +1361,357 @@ func TestGenerateVirtualServerConfigForVirtualServerWithMatches(t *testing.T) {
...
+					{
+						Name:    "tea-v2",
+						Service: "tea-svc-v2",
+						Port:    80,
+					},
+				},
+				Routes: []conf_v1.Route{
+					{
+						Path: "/tea",
+						Matches: []conf_v1.Match{
```
pleshakov, line 1402: because we don't test DOS generation for VS in this test, we can remove the upstreams, the tea endpoints and the `/tea` route from this test - this will greatly reduce the amount of the code.

pleshakov (commented):
Hi @soneillf5! Please see my comments and suggestions

internal/k8s/controller.go
soneillf5: fixed

soneillf5:
@pleshakov Hello, thanks for your review. I've updated the VSR test for DOS to target the location additions for `match` `splits` and normal case.

internal/configs/virtualserver.go
```diff
@@ -2098,6 +2106,14 @@ func (vsc *virtualServerConfigurator) generateSSLConfig(owner runtime.Object, tl
 	return &ssl
 }
 
+func (vsc *virtualServerConfigurator) generateDos(vsr runtime.Object, resource *appProtectDosResource) *version2.Dos {
+	if !vsc.isDosEnabled && resource != nil {
```
pleshakov, line 2110: if we start an Ingress Controller without enabled DOS, and have a VSR with something like this:
```yaml
  subroutes:
  - action:
      pass: coffee
    dos: dos-resource-2
    path: /coffee
```
`generateDos` will be called with `resource` nil, because in the Controller the following code:
```
		if r.Dos != "" {
			routeDosEx, err := lbc.dosConfiguration.GetValidDosEx(virtualServer.Namespace, r.Dos
```
will return nil.

This means the warning reporting code will never be called, and users will not see any warnings.

pleshakov (commented):
Hi @soneillf5  Thanks. The tests look good!
I noticed a problem with the reported warnings in the new code.

internal/configs/virtualserver.go
soneillf5: I've updated the PR with a check during validation to see if dos is enabled.

pleshakov: After the validation was added in the validation package, I can't think of a case when this check `!vsc.isDosEnabled && resource != nil` will evaluated to `true`. If it is dead code, could you remove it?

pkg/apis/configuration/validation/virtualserver.go
```diff
@@ -116,14 +118,18 @@ func validateTLS(tls *v1.TLS, fieldPath *field.Path) field.ErrorList {
...
+func validateDos(isDosEnabled bool, dos string, fieldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	if dos == "" {
 		// valid, dos is not required
 		return allErrs
 	}
 
+	if !isDosEnabled {
+		allErrs = append(allErrs, field.Invalid(fieldPath, dos, "field: 'dos' is used but DOS feature is not enabled"))
```
pleshakov, line 130: we usually use `field.Forbidden` for the fields with valid values that configured disabled features. Could you update it for consistency?

pleshakov (approved):
looks good! with some minor comments

internal/configs/virtualserver.go
soneillf5: removed

PR #2260, opened 2021-12-08
Check if refresh token is `undefined` and do not store it in this case

tippexs:
### Proposed changes
In some cases the token response from the Identity Provider (IdP) does not contain a refresh token. In this case the variable `tokenset.refresh_token` will be `undefined`. As `undefined` is unequal to the current refresh-token undefined is stored as the tokens value in the key-value store. This results in a deletion of the token from the key-value-store.

### Checklist

- [X ] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/master/CONTRIBUTING.md) doc
- [X ] I have added tests that prove my fix is effective or that my feature works
- [ X] I have checked that all unit tests pass after adding my changes
- [ X] I have updated necessary documentation
- [ X ] I have rebased my branch onto master
- [ X] I will ensure my PR is targeting the master branch and pulling from my branch from my own fork

### DIFF
--- a/internal/configs/oidc/openid_connect.js
+++ b/internal/configs/oidc/openid_connect.js
@@ -83,7 +83,8 @@ function auth(r) {
                         r.variables.session_jwt = tokenset.id_token; // Update key-value store
 
                         // Update refresh token (if we got a new one)
-                        if (r.variables.refresh_token != tokenset.refresh_token) {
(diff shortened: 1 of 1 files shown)


lucacome (approved):

pleshakov (approved):

pleshakov:
@lucacome @tippexs is it a bug fix? shall we tag it as a bug fix?

lucacome:
yeah I think it makes sense to add the bug label

PR #2267, opened 2021-12-11
Fix URI rewrite in VirtualServers and VirtualServerRoutes

### LINKED ISSUE #1993: VirtualServer with conditional routes creates wrong rewrite rules
**Describe the bug**
When creating a VirtualServer with a route that contains a match condition and a rewritePath, it seems to create wrong rewrite rules. On the proxy destination i can observere request parameters duplicated.

**To Reproduce**

1. Create the following VirtualServer
```
apiVersion: k8s.nginx.org/v1
kind: VirtualServer
metadata:
  name: myvserver
spec:
  host: myvserver.example.org
  server-snippets: |
    rewrite_log on;
  upstreams:
    - name: app
      service: myservice
      port: 8666
  routes:
    - path: /
      action:
        pass: app
    - path: /demo
      matches:
      - conditions:
        - cookie: user
          value: john
        action:
          proxy:
            upstream: app
            rewritePath: /user/john
      action:
        proxy:
          upstream: app
          rewritePath: /
```
2. Do a curl request to the created vhost. It doesn't matter it if the conditions is matched or not.
```
$ curl http://myvserver.example.org/demo?hello=world
...
```
3. Observe in the access log of app that the request is proxied in an unexpected way
```
10.142.184.178 - - [21/Sep/2021:16:21:49 +0000] "GET /%3Fhello=world?hello=world HTTP/1.1" 404 153 "-" "curl/7.64.1"
```

The location block in the generated config seems to be like this:
```
...
   location /internal_location_matches_0_default {
        set $service "myservice";
        internal;
        set $default_connection_header close;
        rewrite ^ $request_uri;
        rewrite "^/demo(.*)$" "/$1" break;
...
```

**Expected behavior**
The requests are proxied to the destination as /?hello=world.

**Your environment**
* 1.12.1 (installed via nginx-ingress-0.0.0-edge helm chart)
* 1.20.7
* Azure Kubernetes (AKS)
* Nginx + NginxPlus (on eval license)

**Additional context**
Without the matches: block the request is correctly passed to the upstream.

Aha! Link: https://nginx.aha.io/features/IC-280

pleshakov:
### Proposed changes

Previously, if a rewrite was configured in a VS/VSR route, and the route had a split action, match action or a regex path, NGINX would mess up the arguments in the resulting rewrite. For example, the result would be `/test%3Fhello=world?hello=world` instead of the correct `/test?hello=world`

Fixes #1993

Tests:
- Combine the VS and VSR tests into a single one.
- Extend the tests to cover the bug fix.


vepatel (approved):

tests/suite/test_rewrites.py
```diff
@@ -0,0 +1,110 @@
...
+             ("/match/", {"arg": "value"}, {}, "/?arg=value"),
+             ("/match/abc", {"arg": "value"}, {},  "/abc?arg=value"),
+             ("/match/", {"arg": "value"}, {"user": "john"}, "/user/john/?arg=value"),
+             ("/match/abc", {"arg": "value"}, {"user": "john"}, "/user/john/abc?arg=value"),
+             ("/regex1/", {"arg": "value"}, {}, "/?arg=value"),
+             ("/regex1//", {"arg": "value"}, {}, "/?arg=value"),
+             ("/regex2/abc", {"arg": "value"}, {}, "/abc?arg=value")]
+
+
+@pytest.mark.rewrite
```
vepatel, line 87: this marker is redundant as tests will run under vs and vsr

ciarams87 (approved):

PR #2270, opened 2021-12-14
Update-codegen needs a specific path

lucacome:
The script only works if the path is exactly the same as the API, this moves the folder to the right place so we can catch changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,9 @@ jobs:
           make update-crds && git diff --name-only --exit-code deployments/common/crds* deployments/helm-chart/crds*
       - name: Check if Codegen changed
         run: |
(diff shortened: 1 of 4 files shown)


ciarams87 (approved):

pleshakov:
@lucacome looks like the CI is failing

lucacome:
> @lucacome looks like the CI is failing

@pleshakov there was an extra `.` that made it fail 😭, fixed!

pleshakov:
@lucacome foo https://github.com/nginxinc/kubernetes-ingress/blob/b27777eaed9c4232b5d9bae9f6cad8cafb3c03f3/foo ? 😄

lucacome:
lol nice catch @pleshakov

pleshakov (approved):

PR #2271, opened 2021-12-14
Update codegen

lucacome:
Looks like it's been a while since `make update-codegen` was run and the pipeline wasn't catching it because of how the `update-codegen` script works (there's a fix for that in https://github.com/nginxinc/kubernetes-ingress/pull/2270)


ciarams87 (approved):
Good catch! Thanks for updating 👍🏼

pleshakov (approved):

PR #2288, opened 2021-12-17
Extend docs and example about wildcard TLS secret

pleshakov:
### Proposed changes
Extend docs and the example about wildcard TLS secret


ciarams87 (approved):

examples/wildcard-tls-certificate/README.md
```diff
@@ -45,31 +45,26 @@ spec:
...
+    service: bar-service
+    port: 80
+  routes:
+  - path: /
+    action:
+      pass: bar
 ```
 
-Because we don't reference any TLS secret in the `tls` section (there is no `secretName` field) in both Ingress resources, NGINX will use the wildcard secret specified in the `-wildcard-tls-secret` command-line argument.
+Because we don't reference any TLS secret in the resources above -- there is no `secret` field  in the `tls` section of the Ingress resource and the `secret` field is empty in the VirtualServer -- NGINX will use the wildcard secret specified in the `-wildcard-tls-secret` command-line argument.
```
lucacome, line 70:
```suggestion
Because we don't reference any TLS secret in the resources above -- there is no `secret` field in the `tls` section of the Ingress resource and the `secret` field is empty in the VirtualServer -- NGINX will use the wildcard secret specified in the `-wildcard-tls-secret` command-line argument.
```

lucacome (approved):

PR #2360, opened 2022-01-18
Add app protect reconnect period directive

galitskiy:
### Proposed changes
Add _app_protect_reconnect_period_seconds_ directive, which determines the period of time between reconnect retries of the app-protect module to the web application firewall (WAF) engine.

### Checklist


internal/nginx/manager.go
```diff
@@ -35,7 +35,7 @@ const (
 	appProtectAgentStartCmd  = "/opt/app_protect/bin/bd_agent"
 
 	// appPluginParams is the configuration of App-Protect plugin
-	appPluginParams = "tmm_count 4 proc_cpuinfo_cpu_mhz 2000000 total_xml_memory 307200000 total_umu_max_size 3129344 sys_max_account_id 1024 no_static_config"
+	appPluginParams = "tmm_count 4 proc_cpuinfo_cpu_mhz 2000000 total_xml_memory 471859200 total_umu_max_size 3129344 sys_max_account_id 1024 no_static_config"
```
ciarams87, line 38: What's the purpose of this change?

galitskiy: This is to give enough memory available for xml parser in order to avoid issuers with big policies. https://docs.nginx.com/nginx-app-protect/admin-guide/install/#:~:text=file.%20The%20number-,471859200,-should%20be%20enough

internal/configs/configmaps.go
```diff
@@ -499,6 +499,15 @@ func ParseConfigMap(cfgm *v1.ConfigMap, nginxPlus bool, hasAppProtect bool, hasA
 				glog.Error("ConfigMap Key 'app-protect-physical-memory-thresholds' must follow pattern: 'high=<0 - 100> low=<0 - 100>'. Ignoring.")
 			}
 		}
+		if appProtectReconnectPeriod, exists := cfgm.Data["app-protect-reconnect-period-seconds"]; exists {
+			if period, err := ParseFloat64(appProtectReconnectPeriod); err == nil {
```
lucacome, line 503: Is a Float really needed? I would assume that seconds are int, float seems a little bit overkill 🤷‍♂️

internal/configs/configmaps.go
```diff
@@ -499,6 +499,15 @@ func ParseConfigMap(cfgm *v1.ConfigMap, nginxPlus bool, hasAppProtect bool, hasA
...
+				if period > 0 && period <= 60 {
+					cfgParams.MainAppProtectReconnectPeriod = appProtectReconnectPeriod
+				}
+			} else {
+				glog.Error("ConfigMap Key 'app-protect-reconnect-period-second' must have value between '0' and '60'. '0' is illegal. Ignoring.")
```
lucacome, line 508: I find this a little bit confusing. If the value is between `0` and `60` I would assume that `0` and `60` are included. Is `60` a valid value?
Can you also add the valid values to the configmap key description?

lucacome (changes requested):
Left a couple of comments, mostly about clarifying the allowed values and adding them to the docs.

internal/nginx/manager.go
lucacome: it's not really relevant to the reconnect period, is it?

internal/configs/configmaps.go
galitskiy: Hi. according to docs "Value type: number with decimal fraction" (https://docs.nginx.com/nginx-app-protect/configuration-guide/configuration/)

internal/configs/configmaps.go
galitskiy: Hi. the same here, its up to doc "Value Range: 0-60. 0 is illegal". so 60 is valid and 0 is not.

internal/nginx/manager.go
galitskiy: agree. create separate PR?

internal/configs/configmaps.go
lucacome: I don't think it hurts to specify the values in the description and to make it more clear even if it doesn't match the one on the website.

internal/nginx/manager.go
lucacome: yeah I think it makese sense

galitskiy: created PR https://github.com/nginxinc/kubernetes-ingress/pull/2455

internal/configs/configmaps.go
galitskiy: added some wording, please reopen if needed

ashutoshpradhan-f5 (approved):

docs/content/configuration/global-configuration/configmap-resource.md
```diff
@@ -185,6 +185,7 @@ See the doc about [VirtualServer and VirtualServerRoute resources](/nginx-ingres
 |``app-protect-failure-mode-action`` | Sets the ``app_protect_failure_mode_action`` [global directive](/nginx-app-protect/configuration/#global-directives). | ``pass`` |  |
 |``app-protect-cpu-thresholds`` | Sets the ``app_protect_cpu_thresholds`` [global directive](/nginx-app-protect/configuration/#global-directives). | ``high=100 low=100`` |  |
 |``app-protect-physical-memory-util-thresholds`` | Sets the ``app_protect_physical_memory_util_thresholds`` [global directive](/nginx-app-protect/configuration/#global-directives). | ``high=100 low=100`` |  |
+|``app-protect-reconnect-period-seconds`` | Sets the ``app_protect_reconnect_period_seconds`` [global directive](/nginx-app-protect/configuration/#global-directives). Value type: number with decimal fraction. Value Range: 0-60. 0 is illegal. | ``5`` |  |
```
tomasohaodha, line 188: I think this wording is strange, and may point to an underlying change that can make it more natural.
- "Decimal fraction" is not commonly used.
- "Range 0-60. 0 is illegal" is strange. Ranges are generally inclusive, or alternatively include the first number but exclude the second. I've never seen a range that excluded the first number.

Generally when time measurements require precision, I've seen milliseconds used, not seconds. This is the equivalent of seconds with three decimal places. It also makes the range description more readable, without the need for '0 is illegal'. We instead say "Range: 1-60000"

Can we either
a) Change unit to milliseconds to keep precision and eliminate awkward wording, or
b) Keep as seconds but remove the precision (value always rounded to the next whole second)

tomasohaodha (changes requested):

docs/content/configuration/global-configuration/configmap-resource.md
jputrino: Agreed, this is unclear as written. However, as this is verbatim how the directive is described in the App Protect docs, this PR may not be the appropriate forum for requesting changes. It's sufficient to link to the directive in the NAP WAF docs here instead of duplicating the information.

```suggestion
|`app-protect-reconnect-period-seconds` | Sets the `app_protect_reconnect_period_seconds` [global directive](/nginx-app-protect/configuration/#global-directives). | `5` |  |
```

jputrino (approved):
Approved pending application of suggestion

docs/content/configuration/global-configuration/configmap-resource.md
tomasohaodha: Thanks @jputrino , that's a good way around it.

lucacome (approved):

tomasohaodha (approved):

internal/configs/configmaps.go
```diff
@@ -499,6 +499,15 @@ func ParseConfigMap(cfgm *v1.ConfigMap, nginxPlus bool, hasAppProtect bool, hasA
 				glog.Error("ConfigMap Key 'app-protect-physical-memory-thresholds' must follow pattern: 'high=<0 - 100> low=<0 - 100>'. Ignoring.")
 			}
 		}
+		if appProtectReconnectPeriod, exists := cfgm.Data["app-protect-reconnect-period-seconds"]; exists {
+			if period, err := ParseFloat64(appProtectReconnectPeriod); err == nil {
+				if period > 0 && period <= 60 {
```
lucacome, line 504: @galitskiy do we want to throw an error if the value is not between the accepted values? I think it would be easier to catch if we output something when the value is incorrect

internal/configs/configmaps.go
```diff
@@ -499,6 +499,15 @@ func ParseConfigMap(cfgm *v1.ConfigMap, nginxPlus bool, hasAppProtect bool, hasA
...
+					cfgParams.MainAppProtectReconnectPeriod = appProtectReconnectPeriod
+				}
+			} else {
+				glog.Error("ConfigMap Key 'app-protect-reconnect-period-second' must have value between '0' and '60'. '0' is illegal. Ignoring.")
+			}
+		}
 	}
```
galitskiy, line 511: @lucacome Thx for review. Here is suggestion to validate period. Please let me know if its' ugly or I misunderstood smth.
```suggestion
		if appProtectReconnectPeriod, exists, err := ParseFloat64(cfgm.Data["app-protect-reconnect-period-seconds"]); exists {
			if err != nil || appProtectReconnectPeriod <= 0 || appProtectReconnectPeriod > 60 {
				glog.Error("ConfigMap Key 'app-protect-reconnect-period-second' must have value between '0' and '60'. '0' is illegal. Ignoring.")
			} else {
				cfgParams.MainAppProtectReconnectPeriod = appProtectReconnectPeriod
			}
		}
	}
```

lucacome: Unfortunately that doesn't work, I would do something lilke
```suggestion
		if appProtectReconnectPeriod, exists := cfgm.Data["app-protect-reconnect-period-seconds"]; exists {
			period, err := ParseFloat64(appProtectReconnectPeriod)
			if err == nil && period > 0 && period <= 60 {
				cfgParams.MainAppProtectReconnectPeriod = appProtectReconnectPeriod
			} else {
				glog.Error("ConfigMap Key 'app-protect-reconnect-period-second' must have value between '0' and '60'. '0' is illegal. Ignoring.")
			}
		}
	}
```

galitskiy: yeah, you right. Thank you!

Jcahilltorre (approved):

lucacome (changes requested):
The PR looks good, but I just noticed that the pipeline is not triggered. I'm going to block the merge until we sort that out.

lucacome (approved):

PR #2362, opened 2022-01-19
Remove UBI 7 and condense WAF/DOS builds

lucacome:
- NAP WAF and DoS now support UBI 8, we don't need UBI 7 anymore
- Reduced the duplication for installing NAP modules


ciarams87 (approved):
👏🏼

pleshakov (approved):
```+58 −126```  nice 👍

PR #2367, opened 2022-01-19
Remove extra kind nodes, add ipv6 (dual stack)

lucacome:
- removes unnecessary nodes
- enables dual stack


pleshakov (approved):

f5-eugenio (approved):

ciarams87:
This is failing in Gitlab but I don't think we should block testing in GitHub, so can we instead create a second kind config file and update the file used in the workflow here: https://github.com/nginxinc/kubernetes-ingress/blob/master/.github/actions/smoke-tests/action.yaml#L86

ciarams87 (approved):

PR #2378, opened 2022-01-24
Fix healthcheck ports

ciarams87:
### Proposed changes
Don't set the upstream port as the default port for healthchecks - healthchecks will default to the port of the server instead (see https://nginx.org/en/docs/http/ngx_http_upstream_hc_module.html#health_check). 

### Checklist


docs/content/configuration/transportserver-resource.md
```diff
@@ -168,7 +168,7 @@ Note: This feature is supported only in NGINX Plus.
 |``jitter`` | The time within which each health check will be randomly delayed. By default, there is no delay. | ``string`` | No |
 |``fails`` | The number of consecutive failed health checks of a particular upstream server after which this server will be considered unhealthy. The default is ``1``. | ``integer`` | No |
 |``passes`` | The number of consecutive passed health checks of a particular upstream server after which the server will be considered healthy. The default is ``1``. | ``integer`` | No |
-|``port`` | The port used for health check requests. By default, the port of the upstream is used. Note: in contrast with the port of the upstream, this port is not a service port, but a port of a pod. | ``integer`` | No |
+|``port`` | The port used for health check requests. By default, the [server port is used](https://nginx.org/en/docs/http/ngx_http_upstream_hc_module.html?&_ga=2.177145440.496664158.1625655456-810965074.1624792158#health_check_port). Note: in contrast with the port of the upstream, this port is not a service port, but a port of a pod. | ``integer`` | No |
```
pleshakov, line 171:
```suggestion
|``port`` | The port used for health check requests. By default, the [server port is used](https://nginx.org/en/docs/stream/ngx_stream_upstream_hc_module.html#health_check_port). Note: in contrast with the port of the upstream, this port is not a service port, but a port of a pod. | ``integer`` | No |
```

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -367,7 +367,7 @@ Note: This feature is supported only in NGINX Plus.
 |``jitter`` | The time within which each health check will be randomly delayed. By default, there is no delay. | ``string`` | No |
 |``fails`` | The number of consecutive failed health checks of a particular upstream server after which this server will be considered unhealthy. The default is ``1``. | ``integer`` | No |
 |``passes`` | The number of consecutive passed health checks of a particular upstream server after which the server will be considered healthy. The default is ``1``. | ``integer`` | No |
-|``port`` | The port used for health check requests. By default, the port of the upstream is used. Note: in contrast with the port of the upstream, this port is not a service port, but a port of a pod. | ``integer`` | No |
+|``port`` | The port used for health check requests. By default, the [server port is used](https://nginx.org/en/docs/http/ngx_http_upstream_hc_module.html?&_ga=2.177145440.496664158.1625655456-810965074.1624792158#health_check_port). Note: in contrast with the port of the upstream, this port is not a service port, but a port of a pod. | ``integer`` | No |
```
pleshakov, line 370:
```suggestion
|``port`` | The port used for health check requests. By default, the [server port is used](https://nginx.org/en/docs/http/ngx_http_upstream_hc_module.html#health_check_port). Note: in contrast with the port of the upstream, this port is not a service port, but a port of a pod. | ``integer`` | No |
```

internal/configs/version2/http.go
```diff
@@ -241,7 +241,7 @@ type HealthCheck struct {
 	Jitter              string
 	Fails               int
 	Passes              int
-	Port                int
+	Port                *int
```
pleshakov, line 244: shall we `StreamHealthCheck` from  `steam.go` be updated similarly?

internal/configs/transportserver.go
```diff
@@ -154,7 +154,7 @@ func generateTransportServerHealthCheck(upstreamName string, generatedUpstreamNa
 			}
 
 			if u.HealthCheck.Port > 0 {
-				hc.Port = u.HealthCheck.Port
+				hc.Port = &u.HealthCheck.Port
```
pleshakov, line 157: Assigning a pointer from a loop variable can be a source of subtle bugs.  Please see an example here https://go.dev/play/p/5pKPotijvmX 

In this case, because `HealthCheck` itself is a pointer, it doesn't look like we have a bug here. However, it is not clear from the code, and as a result, it can make the reader confused.

Considering that, I wonder if it makes sense to get rid of the pointer all together - a use `int` both in `StreamHealthCheck` in `stream.go`  and `HealthCheck` in `http.go`.

For that case, the statement `{{ if $s.HealthCheck.Port }} port={{ $s.HealthCheck.Port }}` will never evaluate if the value of `HealthCheck.Port` is `0`. So the existing logic would still work. This applies to the VirtualServer too. 

Also, we will be able to simplify conditions like below:
```
if u.HealthCheck.Port > 0 {
	hc.Port = u.HealthCheck.Port
	hc.Port = &u.HealthCheck.Port
```
->
```
hc.Port = u.HealthCheck.Port 
```
and
```
	if upstream.HealthCheck.Port > 0 {
		hc.Port = &upstream.HealthCheck.Port
	}
```
->
```
hc.Port = upstream.HealthCheck.Port
```

I apologies - my comment here https://github.com/nginxinc/kubernetes-ingress/pull/2378#discussion_r792134992 was a suggestion to convert int to a pointer for stream (to be consistent with http). However, now that it is implemented, I noticed a potential problem with that  :(

ciarams87: Ah, good catch! That makes sense - thanks Michael! I'll update this 👍🏼

pleshakov (approved):

PR #2406, opened 2022-02-03
Fix notification and add alpine-opentracing

lucacome:
- Fixes quotes in notification message (if the commit message contained `"` it would fail to send the notification)
- Adds build for Alpine with opentracing
- Adds build for s390x for UBI


ciarams87 (approved):
👍🏼

pleshakov (approved):

PR #2413, opened 2022-02-04
Update syslog image tag

pleshakov:
The old tag no longer exists


ciarams87 (approved):

ciarams87:
I was so confused when I first looked at this until I realised I was looking at docs 🤦🏼 Good catch!

PR #2479, opened 2022-03-10
Add cli argument to configure NAP log level

galitskiy:
### Proposed changes
implement NAP log level configuration via cli argument. Replacing manual preparing of logger configuration file (in Dockerfile and internally when using nginxdebug) with invoking perl tool (installed with app-protect).

### Checklist

### DIFF
--- a/cmd/nginx-ingress/main.go
+++ b/cmd/nginx-ingress/main.go
@@ -71,6 +71,9 @@ var (
 
 	appProtect = flag.Bool("enable-app-protect", false, "Enable support for NGINX App Protect. Requires -nginx-plus.")
 
+	appProtectLogLevel = flag.String("app-protect-log-level", appProtectLogLevelDefault,
+		`Sets log level for App Protect. Allowed values: fatal, error, warn, info, debug, trace. Requires -nginx-plus and -enable-app-protect.`)
+
 	appProtectDos = flag.Bool("enable-app-protect-dos", false, "Enable support for NGINX App Protect dos. Requires -nginx-plus.")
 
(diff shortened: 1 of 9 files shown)


lucacome:
@galitskiy looks like there's a security error

galitskiy:
HI @lucacome . not sure ho to avoid this warning... prbl have a harcoded loglevel instead of using variable value directly (smth like if var=x, fullCmd = cmd +x)

cmd/nginx-ingress/main.go
```diff
@@ -248,6 +251,17 @@ func main() {
 		glog.Fatal("NGINX App Protect support is for NGINX Plus only")
 	}
 
+	if *appProtectLogLevel != appProtectLogLevelDefault && !*appProtect && !*nginxPlus {
```
lucacome, line 254: I know that DOS is implemented the same way, but I find it a bit redundant and I think it also increases the complexity of the code

I feel something like this would be better
```
if *appProtect {
	if *appProtectLogLevel != appProtectLogLevelDefault {
		...
	}
	// all the other future checks
}
```
but maybe we can refactor both later. Do you think it adds much value to check those other flags if AP it's not enabled, instead of just ignoring it?

Anyway at this point in the code you also already checked that it's nginxPlus so there's no need to check it again

lucacome (approved):

ciarams87 (approved):

cmd/nginx-ingress/main.go
galitskiy: Yeah, frankly speaking there is not much of a value in checking all flags if AP is not enabled. But It also sounds reasonable to me to refactor later in order to have the same approach for AP and DOS.

PR #2484, opened 2022-03-12
Add container resources

lucacome:
Adds default container resources for the Ingress Controller


jasonwilliams14:
Just for my own curiosity, is this a OpenShift requirement that we have to set 'requests'? 
is this something that is now being required for Kubernetes as well?
This is interesting because putting resource limits on pods can cause some major problems, if not scoped properly.

ciarams87 (approved):

PR #2556, opened 2022-03-29
Add build for AWS Marketplace

lucacome:
Adds build for AWS


ciarams87 (approved):

lucacome:
I've updated it to always have the `-mktpl` tag. I think it will be better to just avoid confusion.

jjngx (approved):

PR #2566, opened 2022-03-30
Unbind policy from preview policies

haywoodsh:
### Proposed changes
Policy settings are no longer gated through the use of -enable-preview-policies

Affected policies:

- RateLimit
- JWT
- IngressMTLS
- EgressMTLS

### Checklist


internal/configs/version1/nginx-plus.tmpl
```diff
@@ -26,9 +26,7 @@ load_module modules/ngx_http_app_protect_dos_module.so;
 {{$value}}{{end}}
 {{- end}}
 
-{{if .PreviewPolicies}}
```
lucacome, line 29: This was a "workaround" to avoid loading the module all the time when it's not actually needed (only OIDC requires it). I'm not sure if want to always load it now and how else we can easily control it.

docs/content/configuration/policy-resource.md
```diff
@@ -94,7 +94,7 @@ policies:
 
 ### RateLimit
 
-> **Feature Status**: Rate-Limiting is available as a preview feature[^1]: We might introduce some backward-incompatible changes to the resource definition. The feature is disabled by default. To enable it, set the [enable-preview-policies](/nginx-ingress-controller/configuration/global-configuration/command-line-arguments/#cmdoption-enable-preview-policies) command-line argument of the Ingress Controller.
+> **Feature Status**: Rate-Limiting is in preview status until release 2.1.2.[^1]
```
lucacome, line 97:
```suggestion
> **Feature Status**: Rate-Limiting was in preview status up to release 2.1.2.[^1]
```
I think this would be more clear but not 100% sure. @nginxinc/nginx-docs maybe can suggest?

brianehlert: I would remove the past tense reference to what was.  It should **not** be noted with each and every previous case that required the parameter.  Because the statement refers to what was and not what is.

I would have a top level release note statement: '"preview status" has been removed and is no longer necessary to unlock any capabilities.'

brianehlert: yes.
OIDC is the only one that should mention that the preview flag is necessary.
And a new PR should introduce new wording as it would be a deprecation statement of the parameter.  Specific to the new parameter for OIDC Policy enablement.

The preview language in the Policy (this) document should be removed from all Policies except OIDC in this PR.

jputrino: Agreed @brianehlert

jputrino (approved):
This looks good to me, but I'd like to make sure we review the deploy preview as well. That should get kicked off the next time a commit is made to this feature branch.

docs/content/configuration/policy-resource.md
haywoodsh: Documentation updated as suggested

lucacome (approved):

jputrino (approved):

ciarams87 (approved):

PR #2572, opened 2022-04-01
Implement cert-manager for support for VirtualServer resources

ciarams87:
### Proposed changes
This PR implements a cert-manager controller to automate the provision and management of Certificate resources for VirtualServer resources using cert-manager. This mimics the functionality available for Ingress and Gateway resources (see https://cert-manager.io/docs/usage/gateway/).

### Checklist


examples/custom-resources/certmanager/README.md
```diff
@@ -0,0 +1,67 @@
+# Example
+
+In this example, we deploy [cert-manager](https://cert-manager.io/docs/installation/#default-static-install) and a [self signed certificate issuer](https://cert-manager.io/docs/configuration/selfsigned/#bootstrapping-ca-issuers).
+Then, we deploy the NGINX or NGINX Plus Ingress controller, a simple web application and then configure load balancing for that application using the VirtualServer resource.
+
+## Deploying the Certmanager and the self signed authority
```
tomasohaodha, line 6: self-signed?

deployments/common/crds/k8s.nginx.org_virtualservers.yaml
```diff
@@ -479,6 +479,26 @@ spec:
                   description: TLS defines TLS configuration for a VirtualServer.
                   type: object
                   properties:
+                    cert-manager:
+                      description: CertManager defines a cert manager config for a TLS.
+                      type: object
+                      properties:
```
f5yacobucci, line 485: Are any of these fields required? Can they all be empty string?

This will validate:
```
cert-manager: {}
```

But will it configure anything?

deployments/common/crds/k8s.nginx.org_virtualservers.yaml
```diff
@@ -479,6 +479,26 @@ spec:
                   description: TLS defines TLS configuration for a VirtualServer.
                   type: object
                   properties:
+                    cert-manager:
+                      description: CertManager defines a cert manager config for a TLS.
```
f5yacobucci, line 483: I think descriptions should only refer to themselves as they are named in the schema. `cert-manager` defines yada yada.

Looks like a pattern across the schema definition, has anyone brought this up as an issue?

deployments/helm-chart/README.md
```diff
@@ -184,6 +184,7 @@ Parameter | Description | Default
 `controller.enableCustomResources` | Enable the custom resources. | true
 `controller.enablePreviewPolicies` | Enable preview policies. | false
 `controller.enableTLSPassthrough` | Enable TLS Passthrough on port 443. Requires `controller.enableCustomResources`. | false
+`controller.enableCertManager` | Enable cert manager for Virtual Server resources. Requires controller.enableCustomResources. | false
```
f5yacobucci, line 187: Is this enough information for someone to know why they'd enable this?

Perhaps a suggestion: `Enable x509 automated certificate management for VirtualServer resources using cert-manager (cert-manager.io).`

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -76,6 +76,7 @@ redirect:
 | ---| ---| ---| --- |
 |``secret`` | The name of a secret with a TLS certificate and key. The secret must belong to the same namespace as the VirtualServer. The secret must be of the type ``kubernetes.io/tls`` and contain keys named ``tls.crt`` and ``tls.key`` that contain the certificate and private key as described [here](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls). If the secret doesn't exist or is invalid, NGINX will break any attempt to establish a TLS connection to the host of the VirtualServer. If the secret is not specified but [wildcard TLS secret](/nginx-ingress-controller/configuration/global-configuration/command-line-arguments#cmdoption-wildcard-tls-secret) is configured, NGINX will use the wildcard secret for TLS termination. | ``string`` | No |
 |``redirect`` | The redirect configuration of the TLS for a VirtualServer. | [tls.redirect](#virtualservertlsredirect) | No | ### VirtualServer.TLS.Redirect |
+|``certmanager`` | The certmanager configuration of the TLS for a VirtualServer. | [tls.certmanager](#virtualservertlscertmanager) | No | ### VirtualServer.TLS.CertManager |
```
f5yacobucci, line 79: Should this be `cert-manager` to match the field in the schema?

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -95,6 +96,28 @@ basedOn: scheme
 |``basedOn`` | The attribute of a request that NGINX will evaluate to send a redirect. The allowed values are ``scheme`` (the scheme of the request) or ``x-forwarded-proto`` (the ``X-Forwarded-Proto`` header of the request). The default is ``scheme``. | ``string`` | No | ### VirtualServer.Policy |
 {{% /table %}}
 
+### VirtualServer.TLS.CertManager
+
+The certmanager field configures the provisioning of automated certificate resources using certmanager for a VirtualServer (Please note that ACME Issuers are not yet supported):
```
f5yacobucci, line 101: If we're talking about the API objects we should capitalize Certificate. It helps disambiguate the API entity from the concept it represents.

The project consistently refers to itself as `cert-manager`. Staying consistent with how they refer to themselves should help users.

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -95,6 +96,28 @@ basedOn: scheme
...
+enable: true
+code: 301
+basedOn: scheme
+```
+
+{{% table %}}
+|Field | Description | Type | Required |
+| ---| ---| ---| --- |
+|``issuer`` |  the name of an Issuer to acquire the certificate required for this VirtualServer. The Issuer must be in the same namespace as the VirtualServer resource. | ``string`` | No |
+|``cluster-issuer`` | The name of a ClusterIssuer to acquire the Certificate required for this VirtualServer. It does not matter which namespace your VirtualServer resides, as ClusterIssuers are non-namespaced resources. | ``string`` | No |
```
f5yacobucci, line 112: Are there precedence rules if `issuer` and `clusterissuer` are provided?

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -95,6 +96,28 @@ basedOn: scheme
...
+```yaml
+enable: true
+code: 301
+basedOn: scheme
+```
+
+{{% table %}}
+|Field | Description | Type | Required |
+| ---| ---| ---| --- |
+|``issuer`` |  the name of an Issuer to acquire the certificate required for this VirtualServer. The Issuer must be in the same namespace as the VirtualServer resource. | ``string`` | No |
```
f5yacobucci, line 111: Is there another document which describes more about what this is and why it's needed?

I think we could use a tutorial that goes more into helping people setup; either in our docs or one we contribute to cert-manager (probably the former I'd guess).

What do you think about linking to their concepts in the introduction?

f5yacobucci: What about:

> the name of an Issuer. An Issuer is a cert-manager resource which describes the certificate authority capable of signing certificates. The Issuer must be in the same namespace as the VirtualServer resource.

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -95,6 +96,28 @@ basedOn: scheme
...
+|``cluster-issuer`` | The name of a ClusterIssuer to acquire the Certificate required for this VirtualServer. It does not matter which namespace your VirtualServer resides, as ClusterIssuers are non-namespaced resources. | ``string`` | No |
+|``issuer-kind`` | The kind of the external issuer resource, for example AWSPCACIssuer. This is only necessary for out-of-tree issuers. | ``string`` | No |
```
f5yacobucci, line 113: AWSPCACIssuer -> AWSPCAIssuer

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -95,6 +96,28 @@ basedOn: scheme
...
+|``issuer-group`` | The API group of the external issuer controller, for example awspca.cert-manager.io. This is only necessary for out-of-tree issuers. | ``string`` | No |
+|``common-name`` | This field allows you to configure spec.commonName for the Certificate to be generated. | ``string`` | No |
```
f5yacobucci, line 115: I think it's an important, however pedantic point, that this in turn adds a CN to the x509 certificate.

I say this because I've seen errors from golang stdlib when using CN over SAN.

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
brianehlert: There should be a tutorial document.  But not until phase 2 is complete.

examples/custom-resources/certmanager/self-signed.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: sandbox
```
lucacome, line 4: we never used `sandbox` before, should we use `nginx-ingress`?

internal/k8s/controller.go
```diff
@@ -95,6 +97,7 @@ type LoadBalancerController struct {
 	client                        kubernetes.Interface
 	confClient                    k8s_nginx.Interface
 	dynClient                     dynamic.Interface
+	restConfig                    *rest.Config
```
lucacome, line 100: Wouldn't be possible to use the `client` that we are already passing? That's what the config is used for in the end right?

lucacome (approved):

deployments/common/crds/k8s.nginx.org_virtualservers.yaml
ciarams87: Keeping with the pattern for the ingress and gateway controller shims, the validation happens within the cert-manager controller - see:

https://github.com/nginxinc/kubernetes-ingress/pull/2572/files/9a4f092c3121bd7fe591418830148a3426e38519#diff-5aceb73e31c7cb433bbd6cdb4664756a743795cea4714c2fa9773583dde5f8e7R257

and

https://github.com/nginxinc/kubernetes-ingress/pull/2572/files/9a4f092c3121bd7fe591418830148a3426e38519#diff-cd557a292871ac838e3b39960721fcb038ab8a639834fd70e404518d63a67549R71

None of the fields are required by themselves, but for example one or the other of Issuer and ClusterIssuer must be defined. Defining both gives the following error:

`I0406 10:10:19.694289       1 event.go:285] Event(v1.ObjectReference{Kind:"VirtualServer", Namespace:"default", Name:"cafe", UID:"cbadb574-dee7-4271-9d23-86045a7cee27", APIVersion:"k8s.nginx.org/v1", ResourceVersion:"42066", FieldPath:""}): type: 'Warning' reason: 'BadConfig' Could not determine issuer for virtual server due to bad config: both "tls.cert-manager.issuer" and "tls.cert-manager.cluster-issuer" may not be set`

Defining nothing (or everything as an empty string) gives:

I0406 10:13:05.598696       1 event.go:285] Event(v1.ObjectReference{Kind:"VirtualServer", Namespace:"default", Name:"cafe", UID:"cbadb574-dee7-4271-9d23-86045a7cee27", APIVersion:"k8s.nginx.org/v1", ResourceVersion:"42415", FieldPath:""}): type: 'Warning' reason: 'BadConfig' Could not determine issuer for virtual server due to bad config: failed to determine issuer name to be used for virtualserver resource

deployments/common/crds/k8s.nginx.org_virtualservers.yaml
ciarams87: Good point, I think for now I'd rather leave it consistent with what is there, and raise this as a discussion with the team.

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
ciarams87: Good catch, thank you!

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
ciarams87: If both are provided there will be a validation error - they are mutually exclusive. I'll add that.

internal/k8s/controller.go
ciarams87: We need it to create a cert-manager api client in the cm_controller also.

examples/custom-resources/certmanager/self-signed.yaml
ciarams87: I left it as `sandbox` as per the cert-manager standard, but I'm open to changing it. I don't think issuers should be deployed in the `nginx-ingress` namespace though, as they are independent from NIC. What do you think?

deployments/common/crds/k8s.nginx.org_virtualservers.yaml
f5yacobucci: I see, 'issuer' and 'cluster-issuer' are mutually exclusive. Exclusion is a pattern I think we can error closer to the user when they are setting the VirtualServer.

For instance:
```
kc apply -f lock-fail.yaml 
The Lock "lock-test" is invalid: <nil>: Invalid value: "": "spec" must validate one and only one schema (oneOf). Found 2 valid alternatives
```

You can do this with `oneOf`, then the case issuer and cluster-issuer are set, k8s API server rejects the VirtualServer and the user gets immediate feedback it's broken.

E.g.)
```
# CRD snippet:
    schema:
      openAPIV3Schema:
        properties:
          spec:
            type: object
            description: "test"
            properties:
              locked:
                type: string
              unlocked:
                type: string
            oneOf:
              - properties:
                  unlocked:
                    maximum: 0
                required: ["locked"]
              - properties:
                  locked:
                    maximum: 0
                required: ["unlocked"]
        type: object
---
# succeeds
apiVersion: lock.example.com/v1alpha1
kind: Lock
metadata:
  name: locked-test
spec:
  locked: "yes"
---
# succeeds
apiVersion: lock.example.com/v1alpha1
kind: Lock
metadata:
  name: unlocked-test
spec:
  unlocked: "yes"
---
# fails
apiVersion: lock.example.com/v1alpha1
kind: Lock
metadata:
  name: lock-test
spec:
  locked: "yes"
  unlocked: "yes"
```

deployments/rbac/rbac.yaml
```diff
@@ -94,6 +94,16 @@ rules:
     - list
     - watch
     - get
+- apiGroups:
+    - cert-manager.io
+  resources:
+    - certificates
+  verbs:
```
f5yacobucci, line 101: Does cert-manager handled the lifecycle of Certificates? Without a `delete` will we orphan or leak them over time?

internal/certmanager/cm_controller.go
```diff
@@ -0,0 +1,246 @@
...
+	"github.com/golang/glog"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+
+	cmapi "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
```
f5yacobucci, line 32: Any reason not to update these to the new cert-manager location now?

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -95,6 +96,27 @@ basedOn: scheme
...
+|``cluster-issuer`` | the name of a ClusterIssuer. A ClusterIssuer is a cert-manager resource which describes the certificate authority capable of signing certificates. It does not matter which namespace your VirtualServer resides, as ClusterIssuers are non-namespaced resources. Please note that one of `issuer` and `cluster-issuer` are required, but they are mutually exclusive - one and only one must be defined. | ``string`` | No |
+|``issuer-kind`` | The kind of the external issuer resource, for example AWSPCAIssuer. This is only necessary for out-of-tree issuers. This cannot be defined if `cluster-issuer` is also defined. | ``string`` | No |
+|``issuer-group`` | The API group of the external issuer controller, for example awspca.cert-manager.io. This is only necessary for out-of-tree issuers. This cannot be defined if `cluster-issuer` is also defined. | ``string`` | No |
+|``common-name`` | This field allows you to configure spec.commonName for the Certificate to be generated. This configuration adds a CN to the x509 certificate. | ``string`` | No |
+|``duration`` | This field allows you to configure spec.duration field for the Certificate to be generated. | ``string`` | No |
```
f5yacobucci, line 115: `duration` and `renew-before` are governed by ParseDuration rules in the time package. We must document that is the expected format.

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -95,6 +96,27 @@ basedOn: scheme
...
+|``renew-before`` |  this annotation allows you to configure spec.renewBefore field for the Certificate to be generated. | ``string`` | No |
+|``usages`` |  This field allows you to configure spec.usages field for the Certificate to be generated. Pass a string with comma-separated values i.e. ``key agreement,digital signature, server auth``. | ``string`` | No |
```
f5yacobucci, line 117: Cert-manager documentation lists the valid inputs, we should refer the to full set.

Is there a test that misconfigures these? What's that error and how is it communicated?

pkg/apis/configuration/validation/virtualserver.go
```diff
@@ -15,15 +15,17 @@ import (
...
-	isPlus       bool
-	isDosEnabled bool
+	isPlus               bool
+	isDosEnabled         bool
+	isCertManagerEnabled bool
 }
 
 // NewVirtualServerValidator creates a new VirtualServerValidator.
-func NewVirtualServerValidator(isPlus bool, isDosEnabled bool) *VirtualServerValidator {
+func NewVirtualServerValidator(isPlus bool, isDosEnabled bool, isCertManagerEnabled bool) *VirtualServerValidator {
```
f5yacobucci, line 24: This signature doesn't need to change.

It should use the Options pattern:
```
type Option func(*Object)
type Object struct {
  optionField bool
}
func OptionField(option bool) Option {
  return func(o *Object) {
    o.optionField = option
  }
}
func NewObect(Option...) *Object {
  obj := Object{}
  for _, o := range opts {
    o(&obj)
  }
  return &obj
}
```

pkg/apis/configuration/validation/virtualserver.go
```diff
@@ -103,7 +105,7 @@ func validatePolicies(policies []v1.PolicyReference, fieldPath *field.Path, name
 	return allErrs
 }
 
-func validateTLS(tls *v1.TLS, fieldPath *field.Path) field.ErrorList {
+func validateTLS(isCertManagerEnabled bool, tls *v1.TLS, fieldPath *field.Path) field.ErrorList {
```
f5yacobucci, line 108: If these were methods then they'd have access to the Validator state and not require SDK signature changes.

tests/data/virtual-server-certmanager/self-signed.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
```
f5yacobucci, line 1: Are these copies of the other example objects? Can we store these in a single place and render them in docs and include from in tests from a single store?

examples/custom-resources/certmanager/self-signed.yaml
lucacome: yeah maybe not `nginx-ingress` you're right

internal/certmanager/cm_controller.go
ciarams87: The only reason was that 1.8.0 (the first release to use the new repository location) was not released yet. I see it was released 2 days ago, I'll update to the latest version and update the module paths accordingly 👍🏼

deployments/rbac/rbac.yaml
ciarams87: They are deleted by cert-manager I think when we delete a VirtualServer, but there could be scenarios where we need to delete them (and indeed we have a function to do so). Good catch, I've added the delete verb here.

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
ciarams87: Yes, we have a test that misconfigures these (in `helper_test.go`). I'll add a reference to the full set.

If misconfigured, it emits a warning and an event, e.g.:

`E0407 15:05:24.206276       1 sync.go:77] Incorrect cert-manager configuration for VirtualServer resource: invalid cert manager field "tls.cert-manager.usages": "something", invalid cert manager field "tls.cert-manager.usages": "funnythat", invalid cert manager field "tls.cert-manager.usages": "nogood"
I0407 15:05:24.206928       1 event.go:285] Event(v1.ObjectReference{Kind:"VirtualServer", Namespace:"default", Name:"cafe", UID:"63d1bd9f-eb3e-46d5-918f-420e3b6d9625", APIVersion:"k8s.nginx.org/v1", ResourceVersion:"86529", FieldPath:""}): type: 'Warning' reason: 'BadConfig' Incorrect cert-manager configuration for VirtualServer resource: invalid cert manager field "tls.cert-manager.usages": "something", invalid cert manager field "tls.cert-manager.usages": "funnythat", invalid cert manager field "tls.cert-manager.usages": "nogood"`

tests/data/virtual-server-certmanager/self-signed.yaml
ciarams87: This particular file is, but the rest are a different. Having a single store is a great suggestion and is something that we implement, but probably outside the scope of this PR.

f5yacobucci (approved):

PR #2576, opened 2022-04-04
Add support for IPv6 

jjngx:
### Proposed changes

Add support for IPv6

### Checklist

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -50,7 +50,7 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
 	&& apt-get install -y zlib1g \
 	&& curl -fsSL https://cs.nginx.com/static/keys/nginx_signing.key | gpg --dearmor > /etc/apt/trusted.gpg.d/nginx_signing.gpg \
 	&& curl -fsSL -o /etc/apt/apt.conf.d/90pkgs-nginx https://cs.nginx.com/static/files/90pkgs-nginx \
-	&& DEBIAN_VERSION=bullseye \
+	&& DEBIAN_VERSION=$(awk -F '=' '/^VERSION_CODENAME=/ {print $2}' /etc/os-release) \
 	&& printf "%s\n" "Acquire::https::pkgs.nginx.com::User-Agent \"k8s-ic-$IC_VERSION${BUILD_OS##debian-plus}-apt\";" >> /etc/apt/apt.conf.d/90pkgs-nginx \
 	&& printf "%s\n" "deb https://pkgs.nginx.com/plus/${NGINX_PLUS_VERSION^^}/debian ${DEBIAN_VERSION} nginx-plus" > /etc/apt/sources.list.d/nginx-plus.list \
 	&& apt-get update \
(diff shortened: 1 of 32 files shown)


eufinco (commented):
No unit tests with IPv6 addresses?

build/Dockerfile
```diff
@@ -50,7 +50,7 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
 	&& apt-get install -y zlib1g \
 	&& curl -fsSL https://cs.nginx.com/static/keys/nginx_signing.key | gpg --dearmor > /etc/apt/trusted.gpg.d/nginx_signing.gpg \
 	&& curl -fsSL -o /etc/apt/apt.conf.d/90pkgs-nginx https://cs.nginx.com/static/files/90pkgs-nginx \
-	&& DEBIAN_VERSION=$(awk -F '=' '/^VERSION_CODENAME=/ {print $2}' /etc/os-release) \
+	&& DEBIAN_VERSION=bullseye \
```
pleshakov, line 53: Can we revert this change?

build/Dockerfile
```diff
@@ -59,7 +59,16 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
 	&& rm -rf /var/lib/apt/lists/*
 
 
-############################################# Base image for Debian with NGINX Plus and App Protect WAF/DoS #############################################
```
pleshakov, line 62: Do we need this commented out part?

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -197,6 +199,7 @@ http {
     # NGINX Plus APIs
     server {
         listen {{.NginxStatusPort}};
+        listen [::]:{{.NginxStatusPort}};
```
pleshakov, line 202: I think it makes sense to change the default value of https://github.com/nginxinc/kubernetes-ingress/blob/main/cmd/nginx-ingress/main.go#L134 to include the IPV6 localhost.  That value is used in this server in  the allow directive:
```
        {{range $value := .NginxStatusAllowCIDRs}}
        allow {{$value}};{{end}}
        deny all;
        location /api {
            api write=off;
        }
```

currently, the default value is `127.0.0.1`.

this endpoint is used  to access the dashboard https://docs.nginx.com/nginx-ingress-controller/logging-and-monitoring/status-page/#accessing-live-activity-monitoring-dashboard 

same for NGINX OSS.

internal/configs/version1/nginx.tmpl
```diff
@@ -107,13 +107,15 @@ http {
...
         {{if .TLSPassthrough}}
         listen unix:/var/lib/nginx/passthrough-https.sock ssl default_server{{if .HTTP2}} http2{{end}} proxy_protocol;
         set_real_ip_from unix:;
         real_ip_header proxy_protocol;
         {{else}}
         listen 443 ssl default_server{{if .HTTP2}} http2{{end}}{{if .ProxyProtocol}} proxy_protocol{{end}};
+        listen [::]:443 ssl default_server{{if .HTTP2}} http2{{end}}{{if .ProxyProtocol}} proxy_protocol{{end}};
         {{end}}
 
         {{if .SSLRejectHandshake}}
```
pleshakov, line 121: listen directive in ` {{- if .NginxStatus}}` needs to be updated similarly to NGINX Plus.

tests/kind/kind-config-template.yaml
```diff
@@ -0,0 +1,8 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+networking:
+  ipFamily: {{.IPFamily}}
+  apiServerAddress: {{.APIServerAddress}}
+  #apiServerPort: 6443
```
pleshakov, line 6: do we need this parameter commented out? perhaps better to remove all together?

pleshakov (changes requested):
Hi @jjngx 

Please see my comments and suggestions.

Note that the review doesn't cover the changes added to the branch and covered in this PR https://github.com/nginxinc/kubernetes-ingress/pull/2577 

Could we also double check that the examples that provision NGINX backends or configure NGINX using snippets (`grep -R "listen" examples`) also work in IPv6 mode?

build/Dockerfile
jjngx: @pleshakov reverted

build/Dockerfile
lucacome: I think there's still something left in line 62

lucacome (approved):
I think you can revert `build/Dockerfile`

pleshakov (approved):
👍

PR #2583, opened 2022-04-07
Add runAsNonRoot in deployments

lucacome:
Indicates that the container must run as a non-root user. If true, the Kubelet will validate the image at runtime to ensure that it does not run as UID 0 (root) and fail to start the container if it does. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#securitycontext-v1-core


ciarams87 (approved):
Nice 🚀

jjngx (approved):

PR #2654, opened 2022-05-05
Sync changes from OIDC repo, add field in policy

lucacome:
Syncs changes from https://github.com/nginxinc/nginx-openid-connect fixing a couple of bugs.


internal/configs/oidc/oidc.conf
```diff
@@ -8,7 +8,7 @@
 
     location = /_jwks_uri {
         internal;
-        proxy_cache jwk;                              # Cache the JWK Set received from IdP
+        proxy_cache jwk;                              # Cache the JWK Set recieved from IdP
```
ciarams87, line 11: This line is just bring in a misspelling - it should be removed.

lucacome: lol I just glanced over it and I thought it was correcting it 😅

ciarams87 (approved):

docs/content/configuration/policy-resource.md
```diff
@@ -314,8 +314,9 @@ The OIDC policy defines a few internal locations that can't be customized: `/_jw
...
 |``tokenEndpoint`` | URL for the token endpoint provided by your OpenID Connect provider. | ``string`` | Yes |
 |``jwksURI`` | URL for the JSON Web Key Set (JWK) document provided by your OpenID Connect provider. | ``string`` | Yes |
-|``scope`` | List of OpenID Connect scopes. Possible values are ``openid``, ``profile``, ``email``, ``address` and ``phone``. The scope ``openid`` always needs to be present and others can be added concatenating them with a ``+`` sign, for example ``openid+profile+email``. The default is ``openid``. | ``string`` | No |
+|``scope`` | List of OpenID Connect scopes. Possible values are ``openid``, ``profile``, ``email``, ``address`` and ``phone``. The scope ``openid`` always needs to be present and others can be added concatenating them with a ``+`` sign, for example ``openid+profile+email``. The default is ``openid``. | ``string`` | No |
 |``redirectURI`` | Allows overriding the default redirect URI. The default is ``/_codexch``. | ``string`` | No |
+|``zoneSyncLeeway`` | Specifies the maximum timeout for synchronizing ID tokens between multiple Ingress Controllers. The default is ``0s``. | ``string`` | No |
```
pleshakov, line 319: There is a mismatch here. The documentation says the type is `string`, but the golang type is `ZoneSyncLeeway *int`

Additionally, the doc says `0s` is the default, but this is an illegal value for an int.

I suggest using `string` as a golang type, as most of the other time-related fields use it (please see timeout-related fields in https://docs.nginx.com/nginx-ingress-controller/configuration/virtualserver-and-virtualserverroute-resources/ ). This way the new field will be consistent with the existing fields. For example:
```yaml
zoneSyncLeeway: 500ms
```

pleshakov, line 319: From the documentation it is not clear why this field is needed, what problem it solves? Could we add that? For example, the doc could answer the following questions:
- when it is necessary to enable this field (for example, when what errors/conditions clients see  or admins see an error in the logs)
- what values to use, how to pick the value
- what the downsides of enabling that field

Additionally, based on the description here https://github.com/nginxinc/nginx-openid-connect/commit/bd4e9cbccbac94f63666b7711d63cf9b4d7a05ce#diff-512d27cf176962c2b1cffd370bbd1adba80368558ad4aaa2a6255eea9ff453bdR47-R53 , would it make sense to make the default non-zero? Because with the Ingress Controller, it is usually at least 2 replicas of it running in a cluster with a load balancer in front them.

docs/content/configuration/policy-resource.md
```diff
@@ -297,7 +297,7 @@ NGINX Plus will pass the ID of an authenticated user to the backend in the HTTP
 
 #### Prerequisites
 
-In order to use OIDC, you need to enable [zone synchronization](https://docs.nginx.com/nginx/admin-guide/high-availability/zone_sync/). If you don't set up zone synchronization, NGINX Plus will fail to reload. 
+In order to use OIDC, you need to enable [zone synchronization](https://docs.nginx.com/nginx/admixn-guide/high-availability/zone_sync/). If you don't set up zone synchronization, NGINX Plus will fail to reload.
```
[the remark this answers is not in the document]
lucacome: hmm not sure how that happened

docs/content/configuration/policy-resource.md
lucacome: I made a mistake in the doc here. The value is not time in `ms` like in the nginx config, it's an integer that is passed to the javascript function to calculate the time.

docs/content/configuration/policy-resource.md
```diff
@@ -314,8 +314,9 @@ The OIDC policy defines a few internal locations that can't be customized: `/_jw
...
 |``tokenEndpoint`` | URL for the token endpoint provided by your OpenID Connect provider. | ``string`` | Yes |
 |``jwksURI`` | URL for the JSON Web Key Set (JWK) document provided by your OpenID Connect provider. | ``string`` | Yes |
-|``scope`` | List of OpenID Connect scopes. Possible values are ``openid``, ``profile``, ``email``, ``address` and ``phone``. The scope ``openid`` always needs to be present and others can be added concatenating them with a ``+`` sign, for example ``openid+profile+email``. The default is ``openid``. | ``string`` | No |
+|``scope`` | List of OpenID Connect scopes. Possible values are ``openid``, ``profile``, ``email``, ``address`` and ``phone``. The scope ``openid`` always needs to be present and others can be added concatenating them with a ``+`` sign, for example ``openid+profile+email``. The default is ``openid``. | ``string`` | No |
 |``redirectURI`` | Allows overriding the default redirect URI. The default is ``/_codexch``. | ``string`` | No |
+|``zoneSyncLeeway`` | Specifies the maximum timeout in milliseconds for synchronizing ID tokens between multiple Ingress Controllers. The default is ``0``. | ``int`` | No |
```
lucacome, line 319:
```suggestion
|``zoneSyncLeeway`` | Specifies the maximum timeout in milliseconds for synchronizing ID tokens and shared values between Ingress Controller pods. The default is ``0``. | ``int`` | No |
```

pleshakov (approved):

PR #2657, opened 2022-05-05
Use new format for images metadata

lucacome:
https://github.com/docker/metadata-action/issues/159 was implemented, so we can use the new cleaner format.


jjngx (approved):
👍🏻

PR #2745, opened 2022-06-09
Support new directive for dos: app_protect_dos_arb_fqdn

pasmant:
Added  - app-protect-dos-arb-fqdn -  new option in Config Map

Arbitrator FQDN has one argument which is the FQDN/IP.
The argument is the FQDN to the desired Arbitrator.

Examples for ConfigMap data:

FQDN:
```
data:
  app-protect-dos-arb-fqdn svc-appprotect-dos-arb.arb.svc.cluster.local;
```

IP address:
```
data:
  app-protect-dos-arb-fqdn 192.168.1.10;
```


tests/suite/test_dos.py
```diff
@@ -430,8 +434,8 @@ def test_dos_arbitrator(
             wait_before_test()
 
         print("------------------------- Check if new pod receive info from arbitrator -----------------------------")
-        print("Wait for 30 seconds")
-        wait_before_test(30)
+        print("Wait for 60 seconds")
```
lucacome, line 437: What's the reason for increasing this wait?

tests/suite/test_dos.py
```diff
@@ -449,3 +453,107 @@ def test_dos_arbitrator(
         assert (
                 len(learning_units_hostname) == 2
         )
+
+    @pytest.mark.skip
```
lucacome, line 457: Did you mean to leave the `skip` here?

lucacome (commented):
@pasmant left some comments. Would also mind adding a description for the changes in this PR? Thanks!

tests/suite/test_dos.py
pasmant: When I perform the tests, I saw that in some tests the wait for 30 sec isn't enough, when changed it to 60 the test pass all the time

docs/content/configuration/global-configuration/configmap-resource.md
```diff
@@ -188,4 +188,5 @@ See the doc about [VirtualServer and VirtualServerRoute resources](/nginx-ingres
...
 |``app-protect-dos-log-format`` | Sets the custom [log format](https://nginx.org/en/docs/http/ngx_http_log_module.html#log_format) for Dos Access log traffic. For convenience, it is possible to define the log format across multiple lines (each line separated by ``\n``). In that case, the Ingress Controller will replace every ``\n`` character with a space character. All ``'`` characters must be escaped. | `, vs_name_al=$app_protect_dos_vs_name, ip=$remote_addr, tls_fp=$app_protect_dos_tls_fp, outcome=$app_protect_dos_outcome, reason=$app_protect_dos_outcome_reason, policy_name=$app_protect_dos_policy_name, dos_version=$app_protect_dos_version, ip_tls=$remote_addr:$app_protect_dos_tls_fp,` | |
 |``app-protect-dos-log-format-escaping`` | Sets the characters escaping for the variables of the stream log format. Supported values: ``json`` (JSON escaping), ``default`` (the default escaping) ``none`` (disables escaping). | ``default`` |  |
+|``app-protect-dos-arb-fqdn`` | Sets the ``app-protect-dos-arb-fqdn`` [directive](/nginx-app-protect-dos/directives-and-policy/learn-about-directives-and-policy/#readiness-probe-directive-app_protect_dos_readiness). | ``svc-appprotect-dos-arb`` |  |
```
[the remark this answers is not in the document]
pasmant: FIx to:
arbitrator-fqdn-directive-app_protect_dos_arb_fqdn

tests/suite/test_dos.py
pasmant: The skip added by Sean O'Neill, I don't know for sure why it changed to skip.

tests/suite/test_virtual_server_dos.py
```diff
@@ -463,3 +468,102 @@ def test_dos_arbitrator(
         assert (
                 len(learning_units_hostname) == 2
         )
+
+    @pytest.mark.skip
```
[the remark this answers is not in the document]
pasmant: The skip added by Sean O'Neill, I don't know for sure why it changed to skip.

tests/suite/test_dos.py
lucacome: What do you mean added by Sean?
If you mark a test with `skip` it will not run

lucacome: I think there were concerns about creating an actual DoS attack since these tests would run in GKE. Do you think it would be a problem for these specific tests?

I also feel like we should find an alternative solution, adding automation tests but marking them with `skip` is like not adding them at all, since they will never run.

cc @ciarams87

pasmant: I run those tests in my minikube environment and they are all passed.
If there is no problem to rum them on GKE, need to run them there.
otherwise, need to understand why we cannot run those tests and try to solve the problems.
Case we cannot solve the problems and rum the tests in GKE, maybe we can run them on minikube.

lucacome: I think that if they can be run in `minikube`/`kind` successfully, then there's no need to run them in GKE

lucacome (approved):

ciarams87 (approved):

PR #2781, opened 2022-06-21
feat: Report Hostname in ExternalEndpoint for VS and VSR resources

ciarams87:
### Proposed changes
When we report the External Endpoints as part of the VS and VSR status, we only report the external LoadBalancer Ingress IP, and we don't report the Hostname (some external load balancers assign a Hostname instead of a static IP, for example AWS ELB LoadBalancers). This adds a Hostname field to the VS and VSR which will be populated when present in the LoadBalancer service. This field won't be visible by default when running a get on a VS, but the `-o wide` option will display the field.

### Checklist


jjngx (approved):
👍🏻

deployments/common/crds/k8s.nginx.org_virtualserverroutes.yaml
```diff
@@ -608,9 +612,11 @@ spec:
                 externalEndpoints:
                   type: array
                   items:
-                    description: ExternalEndpoint defines the IP and ports used to connect to this resource.
+                    description: ExternalEndpoint defines the IP/ Hostname and ports used to connect to this resource.
```
shaun-nx, line 615: Minor detail here. We can remove the space between the `/` and `Hostname` in this description.

shaun-nx (approved):

deployments/common/crds/k8s.nginx.org_virtualserverroutes.yaml
ciarams87: There is likely only going to be either IP or Hostname reported in the status (there may be cases where both are reported for a LoadBalancer service but I think it would be rare?) which is why I included the slash.

PR #2800, opened 2022-06-28
feat: Integrate ExternalDNS with VirtualServer resources

ciarams87:
### Proposed changes
This PR creates an `ExternalDNS` block in the VirtualServer resource and uses it to create and configure DNSEndpoint objects which will be picked up an ExternalDNS deployment. ExternalDNS can then use the information in DNSEndpoint object to create, update, or delete, as appropriate, DNS records with the configured provider. See https://kubernetes-sigs.github.io/external-dns/v0.12.0/.

### Checklist


internal/externaldns/handlers.go
```diff
@@ -0,0 +1,153 @@
...
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	logf "github.com/cert-manager/cert-manager/pkg/logs"
```
shaun-nx, line 15: Why are we using a different log library instead of glog here?

internal/externaldns/sync.go
```diff
@@ -0,0 +1,238 @@
...
+		glog.V(3).Infof("DNDEndpoint already exist for this object, ensuring it is up to date")
+		if metav1.GetControllerOf(existingDNSEndpoint) == nil {
+			glog.V(3).Infof("DNSEndpoint has no owner. refusing to update non-owned resource")
+			return nil, nil, nil
+		}
+		if !metav1.IsControlledBy(existingDNSEndpoint, vs) {
+			glog.V(3).Infof("external DNS endpoint resource isnot owned by this object. refusing to update non-owned resource")
+			return nil, nil, nil
+		}
+		if !extdnsendpointNeedsUpdate(existingDNSEndpoint, dnsEndpoint) {
```
shaun-nx, line 175: I would change this function name to `isUpdateRequired` since this function is already in the context of ExternalDNS

jjngx (approved):

PR #2837, opened 2022-07-11
Remove OpenTracing images in Updater

lucacome:
Removes OpentTracing images as they are on longer separate images.

To be merged after v2.3.0


.github/workflows/update-docker-images.yml
```diff
@@ -138,17 +138,9 @@ jobs:
             platforms: linux/arm,linux/arm64,linux/amd64,linux/ppc64le,linux/s390x
             needs-updating: ${{ needs.check.outputs.needs-updating-debian }}
           - image: alpine
-            marker: vsr
```
pleshakov, line 141: we're loosing a vsr marker? can we add it to any of the images?

lucacome: Ah good call, I've added them back in other images 👍

pleshakov (approved):

PR #2845, opened 2022-07-13
Use `nginxcontrib/nginx` as base image for UBI

lucacome:
`nginxcontrib/nginx-ubi:1.23.1` mimics `nginx:1.23.1` and `nginx:1.23.1-alpine` but for UBI. 

- Removes the need to manually add the `nginx` user
- Removes the need to manually add the `nginx` repo
- Adds builds for `linux/ppc64le`
- The image adds `nginx-module-xslt`, `nginx-module-image-filter` and `nginx-module-njs` like for `debian` and `alpine`


vepatel (approved):

hafe:
This is not a good idea since users would not be able to (re)build the final image when RH has released a new version of the base image (e.g. with CVE fixes).

lucacome:
> This is not a good idea since users would not be able to (re)build the final image when RH has released a new version of the base image (e.g. with CVE fixes).

@hafe why wouldn’t users be able to rebuild the final image?

hafe:
> > This is not a good idea since users would not be able to (re)build the final image when RH has released a new version of the base image (e.g. with CVE fixes).
> 
> @hafe why wouldn’t users be able to rebuild the final image?

Sorry I misread the diff! Now I see redhat/ubi8 is still the base image!

jjngx (approved):

PR #2846, opened 2022-07-13
Update path for CRD generation to include all

lucacome:
Updates the path to the command that generates the CRDs to include all directories.


shaun-nx (approved):
👍

jjngx (approved):

PR #2884, opened 2022-07-25
Feature/flag include year

shaun-nx:
### Proposed changes
This change introduces a new flag to the ingress controller `-include-year` which will add the year to the log header when set. The default value of this flag is `false`

### Deprecation notice
This flag is planned to be remove from the ingress controller by release 2.7 at which point all logs from the ingress controller will display the year by default.

### Checklist


jjngx (approved):
👍🏻 great job!

deployments/deployment/nginx-plus-ingress.yaml
```diff
@@ -65,6 +65,7 @@ spec:
           - -nginx-plus
           - -nginx-configmaps=$(POD_NAMESPACE)/nginx-config
           - -default-server-tls-secret=$(POD_NAMESPACE)/default-server-secret
+         #- -include-year
```
vepatel, line 68: can specify the same in `deployments/daemon-set/nginx-plus-ingress.yaml` as well

shaun-nx: Good catch. I will also add it to the OSS version of the deployments

deployments/helm-chart/values.yaml
```diff
@@ -178,6 +178,9 @@ controller:
   ## Enable OIDC policies.
   enableOIDC: false
 
+  ## Include year in log header. This parameter will be removed in release 2.7 and the year will be included by default.
+  includeYear: false
```
vepatel, line 182: a corresponding entry in https://github.com/nginxinc/kubernetes-ingress/blob/feature/flag-include-year/docs/content/installation/installation-with-helm.md  as `controller.includeYear`

shaun-nx: Not sure I understand this. None of the values in this file start with `controller.` why would the `includeYear` value need to start with that?

vepatel: it refers to https://github.com/nginxinc/kubernetes-ingress/blob/feature/flag-include-year/deployments/helm-chart/values.yaml#L1 when you're installing using ` helm install nginx-ingress nginx-stable/nginx-ingress --set controller.kind=daemonset --set controller.includeYear=true ......`

vepatel (approved):

PR #2902, opened 2022-08-03
Add check for go.mod and go.sum changes

lucacome:
Adds check to prevent committing changes without running `go mod tidy`


ciarams87 (approved):

shaun-nx (approved):
👍

PR #2929, opened 2022-08-13
Add isort and Black

lucacome:
- Adds isort and Black
- Formats all the tests

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -38,3 +38,10 @@ jobs:
         uses: golangci/golangci-lint-action@v3
         with:
           only-new-issues: true
+
+  lint-python:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: isort/isort-action@v1.0.0
+      - uses: psf/black@stable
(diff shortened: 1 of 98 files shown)


lucacome:
@vepatel I noticed this warning 🤔 
```
.../tests/suite/test_prometheus_metrics.py:191: SyntaxWarning: assertion is always true, perhaps remove parentheses?
```

vepatel:
@lucacome `.../tests/suite/test_prometheus_metrics.py:191: SyntaxWarning: assertion is always true, perhaps remove parentheses?` is a legitimate warning because interpreter is treating it as a tuple which is equivalent to assert True, removing parentheses will fix it.

vepatel:
@lucacome seems like linter is failing to execute, should we include a`pyproject.toml` with black and flake8 config?

lucacome:
the `pyproject.toml` already has black, but unfortunately flake8 doesn't support it at least for now https://github.com/PyCQA/flake8/issues/234#issuecomment-1206730688

did you run `make lint-python`? you would need to run it manually (the workflow in the CI is just a check)...at least until this goes in https://github.com/nginxinc/kubernetes-ingress/pull/2154 😅

vepatel (approved):

tomasohaodha (approved):

PR #2775, opened 2022-06-20
Add validation to Ingress path

shaun-nx:
### Proposed changes
- Add string validation to the path spec for the Ingress resource
- Update unit tests to validate the type of error returned

### Checklist
Before creating a PR, run through this checklist and mark each as complete.

- [x] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that all unit tests pass after adding my changes
- [x] I have updated necessary documentation
- [x] I have rebased my branch onto main
- [x] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

PR #2971, opened 2022-08-25
fix: Correct error message on missing path in path validation

zachomedia:
### Proposed changes

In #2775 some path validation was added, including making the `path` field within the rules required (Kubernetes does not require it). The error message we were receiving was `spec: Required field`, which didn't point obviously to the issue.

The PR fixes the error message to point to the correct place. With this fix, you'll now see `spec.rules[0].http.path[0].path: Required value`. Additionally, I have noted the required value in the `Restrictions` section of the documentation.

### Checklist


lucacome (commented):
Hi @zachomedia 

Thanks for the PR! I'm just trying to understand your config, did you use an empty path before `v2.3.0` with the `pathType` as `ImplementationSpecific`? Because I think that's the only way to not get an error form k8s when applying the Ingress with recent versions of k8s. Are you using an old version? Any additional info would be appreciated 😄

zachomedia:
@lucacome Your guess would be correct :) We had some old ingresses which were using `pathType: ImplementationSpecific` and not path. Most recently, I had tested it with v2.2.2 and the ingresses were accepted by it. The upgrade to v2.3.0 caused those old ingresses to not be accepted.

lucacome (approved):
I feel like we might revisit the enforcement when the `pathType` is `ImplementationSpecific` since technically the `path` shouldn't be required in that case, but for now we should merge this so at least the error is clear.

Thanks again @zachomedia !

haywoodsh (approved):

jjngx (approved):

Issue #2999, opened 2022-09-01
Helm Chart values are not sorted alphabetically


aknot242:
**Describe the bug**
Helm Chart values are not sorted alphabetically, making it easy to miss specific values when scanning the page.

brianehlert:
Through internal conversation we learned that the current scheme has functionality groupings, that would be broken by a blanket alphabetical sort.
We are looking into how to improve the usability and maintain some level of logical grouping without introducing breaking changes.

aknot242:
Thank you @brianehlert!

PR #3000, opened 2022-09-01
Sort helm chart lines ascending alphabetically

aknot242:
### Proposed changes
Addresses #2999 

### Checklist


lucacome (changes requested):
Hi @aknot242 thanks for the PR!

The docs are (mostly) aligned with the `values.yaml` so I think we would need to change the order there too if we want to change in the docs, and also in the `README.md`

aknot242:
@lucacome I respectfully disagree. While the order in the helm chart would ideally match the docs, that is beyond the scope of the issue created. If someone wants to contribute it, that's fine but it would represent quite a bit of manual work to sort it and I'll let someone else contribute that (ideally the dev team, as IMO, seems like it was earned tech debt). 

I can meet you in the middle by sorting the README, but it seems strange that the values exist there separately from in the hugo docs - seems needlessly duplicative. I'll re-order the README in a commit here shortly.

jasonwilliams14:
@aknot242 We took a good look at this and discussed internally. We came the conclusion that we are going to stay with the way things currently are. 
Thank you.

aknot242:
Would be good to document the reasons behind the decision. Doing it in the linked issue (#2999) might be ideal. What is the sorting scheme that is currently in place? Thx.

PR #3004, opened 2022-09-01
added custom annotation to ingress for CORS

jmozd:
### Proposed changes

support CORS by providing a custom annotation to Ingress', which then is reflected in the NGINX ingress configuration.

By adding the custom annotation "custom.nginx.org/cors-allow-origin" to an ingress, with the value of the target server (i.e. "https://sample.server.org/"), the according NGINX ingress configuration is created in the "server {}" section:

    # replace Access-Control-Allow-Origin header 
    proxy_hide_header "Access-Control-Allow-Origin"; 
    add_header Access-Control-Allow-Origin "https://sample.server.org";

Hiding the original header values proved to be required, as else the browser might complain about multiple set values.
 
### Checklist


jasonwilliams14:
We are reviewing this internally and are planning this for a future release.

PR #3029, opened 2022-09-09
Remove deprecated kubernetes.io/ingress.class from examples

lucacome:
Replaces `kubernetes.io/ingress.class` with  `ingressClassName`.

Opening it as a draft as I'm not sure if we need to add something to the examples to call out the need to deploy the ingressClass. It's already called out in the https://docs.nginx.com/nginx-ingress-controller/installation/installation-with-manifests/ and we deploy it automatically with `helm`

### DIFF
--- a/docs/content/configuration/handling-host-and-listener-collisions.md
+++ b/docs/content/configuration/handling-host-and-listener-collisions.md
@@ -31,9 +31,8 @@ Consider the following two resources:
     kind: Ingress
     metadata:
       name: cafe-ingress
-      annotations:
-        kubernetes.io/ingress.class: "nginx"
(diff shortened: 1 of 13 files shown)


jjngx (approved):

vepatel:
@lucacome can we do it for docs only first? some tests are breaking because of this change.

tomasohaodha (changes requested):
Are the first two files really part of this PR? If not can they go into their own PR?

tomasohaodha:
Hi Luca are you going to remove the non-relevant files (as discussed last week)?

tomasohaodha (approved):

PR #3034, opened 2022-09-13
Allow extra args to be provided to the OIDC auth endpoint

alanwilkie-finocomp:
### Proposed changes
Some OIDC Identity Providers provide extended capabilities by adding extra query string arguments to the authentication request. This change allows the OIDC policy to specify the extra arguments.

Specifically, Keycloak allows a default identity provider to be specified by adding a "kc_idp_hint" parameter to the authentication request (see https://www.keycloak.org/docs/latest/server_admin/#_client_suggested_idp).

### Checklist


alanwilkie-finocomp:
I meant this PR to have the "enhancement" label, but I'm not sure how that gets done. Is it possible to add the label after creation?

brianehlert:
Could you please open and link an issue and outline the problem there?

jjngx (approved):

jasonwilliams14:
@alanwilkie-finocomp Our project depends an another repo for OIDC. I would recommend you open this on that specific project, which would allow us to bring over the features that are would be implemented.
Here is the repo:

https://github.com/nginxinc/nginx-openid-connect

Thank you.

brianehlert:
To clarify.
The specific changes to internal/configs/oidc/openid_connect.js would need to be contributed to https://github.com/nginxinc/nginx-openid-connect

We don't maintain a fork of this project, but rather sync with it.

The other changes are specific to this project.

alanwilkie-finocomp:
Thanks @jasonwilliams14 and @brianehlert, I'll have a look at nginx-openid-connect and see if I can get the JS changes into that codebase.

alanwilkie-finocomp:
The JS change has now been merged into nginx-openid-connect (https://github.com/nginxinc/nginx-openid-connect/pull/67).

jasonwilliams14:
> The JS change has now been merged into nginx-openid-connect ([nginxinc/nginx-openid-connect#67](https://github.com/nginxinc/nginx-openid-connect/pull/67)).

Thank you. We will review on our side now that the changes have been merged into the OIDC project.

lucacome (changes requested):
Hi @alanwilkie-finocomp 

Just one small change
> multiple arguments may be provided by concatenating them with ``&``, for example ``arg1=value1&arg2=value2``

I think we want to use something like a list separated by `,` `arg1=value1,arg2=value2` instead of `&` in the YAML file. That would be more consistent with other parameters and YAML in general.
Then you can just replace `,` wit `&` when you pass the string to the javascript.

Let me know what you think about it and if you have any questions.

alanwilkie-finocomp:
Hi @lucacome 

Would it be better to make the arguments a YAML list? Like:
```
authExtraArgs:
  - arg1=value1
  - arg2=value2
```
or:
```
authExtraArgs: [ arg1=value1, arg2=value2 ]
```

lucacome:
Yeah, that's probably even better @alanwilkie-finocomp !

alanwilkie-finocomp:
I've updated the PR with the new policy item as a list of strings that are concatenated with `&` to form the final value.

pkg/apis/configuration/validation/policy_test.go
```diff
@@ -1097,6 +1130,27 @@ func TestValidateURL(t *testing.T) {
 	}
 }
 
+func TestValidateQueryStringt(t *testing.T) {
+	t.Parallel()
+	validInput := []string{"foo=bar", "foo", "foo=bar&baz=zot", "foo=bar&foo=baz", "foo=bar%3Bbaz"}
```
lucacome, line 1135: something like `foo=bar&baz=zot` is a valid input?

alanwilkie-finocomp: The validation function is called after the query string is constructed, so that's the kind of string it would see if multiple args with different names were listed in the policy.

lucacome: ah right! It makes sense 🙂

lucacome (approved):

PR #3041, opened 2022-09-15
Feat/transport server external name

jjngx:
### Proposed changes

Initial support for Transport Server external name including examples.

### Checklist


shaun-nx (commented):
Can we add a README for the new `externalname-services` example files?

internal/configs/configurator.go
```diff
@@ -1132,20 +1125,23 @@ func (cnf *Configurator) UpdateConfig(cfgParams *ConfigParams, resources Extende
...
 	}
 
-	// we don't need to regenerate config for TransportServers, because:
-	// (1) Changes to the ConfigMap don't affect TransportServer configs directly
-	// (2) addOrUpdateTransportServer doesn't return any warnings that we need to propagate to the caller.
-	// if (1) and (2) is no longer the case, we need to generate the config for TransportServers
+	for _, tsEx := range resources.TransportServerExes {
+		warnings, err := cnf.addOrUpdateTransportServer(tsEx)
+		if err != nil {
+			return allWarnings, err
```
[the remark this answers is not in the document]
jjngx: variable ```warnings``` (line 1129) is irrelevant when ```err``` is returned, therefore we return ```allWarnings``` and the ```err``` on the line 1131. Please note that although it works it is not entirely correct. The outer func at this point should return ```nil``` and the error. Currently, the implementation pattern (applied in few more places in this package) can lead into thinking that returned ```warnings``` ALONG with not nil ```err``` can be used, which is not correct.

In short, lines 1128, 1132 should read:
```go
for _, tsEx := range resources.TransportServerExes {
    warnings, err := cnf.addOrUpdateTransportServer(tsEx)
    if err != nil {
        return nil, err
    }
    allWarnings.Add(warnings)
```
If ```cnf.addOrUpdateTransportServer(tsEx)``` returns warnings they are added to allWarnings and returned by the outer function at the end:
```go
return allWarnings, nil
```

Correcting of the returning values will be done in a separate PR as it will require small refactoring of functions/methods and corresponding unit tests.

internal/configs/configurator.go
```diff
@@ -1132,20 +1125,23 @@ func (cnf *Configurator) UpdateConfig(cfgParams *ConfigParams, resources Extende
...
+		}
+		allWarnings.Add(warnings)
```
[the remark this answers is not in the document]
jjngx: In short, not. First we handle err and add warnings ONLY when the err is nil. Addressed this in the comment above.

internal/configs/configurator.go
```diff
@@ -1154,7 +1150,7 @@ func (cnf *Configurator) UpdateConfig(cfgParams *ConfigParams, resources Extende
 // UpdateTransportServers updates TransportServers.
 func (cnf *Configurator) UpdateTransportServers(updatedTSExes []*TransportServerEx, deletedKeys []string) error {
 	for _, tsEx := range updatedTSExes {
-		err := cnf.addOrUpdateTransportServer(tsEx)
+		_, err := cnf.addOrUpdateTransportServer(tsEx)
```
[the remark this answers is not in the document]
jjngx: Applied the same pattern as we have for Virtual Server. Please see Michael's comment in the code (line 888)

internal/configs/configurator.go
```diff
@@ -697,16 +691,16 @@ func (cnf *Configurator) AddOrUpdateResources(resources ExtendedResources) (Warn
 	}
 
 	for _, tsEx := range resources.TransportServerExes {
-		err := cnf.addOrUpdateTransportServer(tsEx)
+		warnings, err := cnf.addOrUpdateTransportServer(tsEx)
 		if err != nil {
-			return allWarnings, fmt.Errorf("Error adding or updating TransportServer %v/%v: %w", tsEx.TransportServer.Namespace, tsEx.TransportServer.Name, err)
+			return allWarnings, fmt.Errorf("error adding or updating TransportServer %v/%v: %w", tsEx.TransportServer.Namespace, tsEx.TransportServer.Name, err)
 		}
+		allWarnings.Add(warnings)
```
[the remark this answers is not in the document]
jjngx: If there is an err we should handle the error first and do not assume anything about value of warnings. If the err is nil it means that the func worked correctly and returned warnings (that could mean warnings are `empty`).

jjngx:
> Can we add a README for the new `externalname-services` example files?

It's WIP (work in progress)

examples/custom-resources/externalname-services/transport-server/README.md
```diff
@@ -0,0 +1,101 @@
...
+    ```
+
+4. Apply the change
+    ```bash
+    $ kubectl apply -f nginx-config.yaml
+    ```
+
+## 4. Change the TS to point to the ExternalName and verify if it is working correctly
+
+1. Navigate to the example [tls-passthrough](../../../custom-resources/tls-passthrough/README.md) and open the ```transport-server.yaml``` file.
```
ciarams87, line 69: Typo:

```suggestion
1. Navigate to the example [tls-passthrough](../../../custom-resources/tls-passthrough/README.md) and open the ```transport-server-passthrough.yaml``` file.
```
Although honestly I think it would be better to not send users to another directory to modify a file from another example. It's a bit hard to follow.

examples/custom-resources/externalname-services/externalname-svc.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: externalname-service
+spec:
+  type: ExternalName
+  externalName: external-backend-svc.external-ns.svc.cluster.local
```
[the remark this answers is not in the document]
ciarams87: Sorry - I realise now that updating this is called out in the README. Again I think this is confusing and we are better off duplicating the files with the correct service name in each (one for TS and one for VS); or just having the same service name for both of the external name services

examples/custom-resources/externalname-services/transport-server/README.md
jjngx: Applied

examples/custom-resources/externalname-services/transport-server/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+1. Deploy the backend application as described in the [tls-passthrough example](../../tls-passthrough/README.md), and make sure it is working as described.
+
+## 2. Deploy external service to external namespace
+
+1. Navigate to the [external-name example](../../../custom-resources/externalname-services/README.md)
+
+2. Deploy external namespace (```external-ns```) and the backend application. Note that the namespace is not being watched by ```NIC```
+    ```
+    $ kubectl apply -f transport-server/secure-app-external.yaml
```
shaun-nx, line 30: I think we should remove the `$` from the start of these code blocks.
If a user is reading these docs and clicks the button to copy the full code block, they will also copy that `$` which will mess up the command

vepatel: @shaun-nx we included it to keep inline with other examples

vepatel (approved):

examples/custom-resources/externalname-services/externalname-svc.yaml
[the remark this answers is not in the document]
jjngx: Updated and moved all yaml files required for this example to transport server folder.

examples/custom-resources/externalname-services/transport-server/externalname-svc.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: externalname-service
+spec:
+  type: ExternalName
+  externalName: external-backend-svc.external-ns.svc.cluster.local
```
ciarams87, line 7:
```suggestion
  externalName: secure-app-external-backend-svc.external-ns.svc.cluster.local
```

examples/custom-resources/externalname-services/virtual-server/external-svc-dep.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: v1
```
ciarams87, line 1: Can we either remove this folder for the virtual server external name service example, or complete the example with a readme and the required yaml files?

vepatel: @jjngx  I'd say we can  remove it for now, can do a separate PR for this later on.

jjngx: done, removed the VS folder

examples/custom-resources/externalname-services/transport-server/externalname-svc.yaml
ciarams87: This still needs to be changed - I ran through the example again just now and I hit the same issue where I had to change this line myself locally

jjngx: Fixed

ciarams87 (approved):

PR #3065, opened 2022-09-20
Helm Chart: Add annotations to the service account in the Helm chart

0m1xa:
Add annotations to pass with helm chart values.

### Proposed changes
The main idea is to have the ability to provide annotations to the service account. 
It may be useful in many cases. 
For example, according to this [documentation](https://docs.nginx.com/nginx-ingress-controller/installation/using-aws-marketplace-image/), the service account should have the proper annotation when using the ingress controller with AWS Marketplace. 
Before this PR it was not possible to specify this annotation with the helm chart.

### Checklist


deployments/helm-chart/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: nginx-ingress
-version: 0.14.0
+version: 0.14.1
```
jasonwilliams14, line 2: It is not necessary to update our chart.yaml file. We update this with releases.

deployments/helm-chart/README.md
```diff
@@ -146,111 +146,112 @@ See [running multiple ingress controllers](https://docs.nginx.com/nginx-ingress-
...
-`controller.serviceMonitor.endpoints` | A list of endpoints allowed as part of this ServiceMonitor. | ""
+Parameter | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              | Default
+--- |----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------| ---
```
jasonwilliams14, line 150: This line appears to have changed the entire formatting for the README.md file. Can correct the formatting for this file.

jasonwilliams14 (changes requested):
Please see comments.

deployments/helm-chart/Chart.yaml
0m1xa: I reverted the original version

deployments/helm-chart/README.md
0m1xa: Also, I've fixed that line

0m1xa:
@jasonwilliams14 
Please see another update. I mentioned that `values.yml` uses the unnecessary parameter `includeYear`, preventing PODs from successfully starting. 
I've tested it with `2.3.0` and `2.3.1`

deployments/helm-chart/README.md
lucacome: @0m1xa seems like there's still some weird formatting to the whole file, instead of just a change with your addition. I feel like it might be your IDE.

lucacome:
@0m1xa if you test it with `edge` it shouldn't complain about `includeYear`

deployments/helm-chart/README.md
0m1xa: @lucacome you're right. It was my IDE's formatting. Sorry for that.
Now it is fixed.

deployments/helm-chart/values.yaml
```diff
@@ -185,9 +185,6 @@ controller:
   ## Enable OIDC policies.
   enableOIDC: false
 
-  ## Include year in log header. This parameter will be removed in release 2.7 and the year will be included by default.
-  includeYear: false
-
```
lucacome, line 190: This shouldn't be removed

0m1xa: @lucacome 
Please take a look at [this](https://github.com/nginxinc/helm-charts/blob/master/stable/nginx-ingress-0.14.0.tgz) (and later) version of the helm chart. 
The `values.yaml` doesn't contain `includeYear` and no one of the templates doesn't use this parameter. 
If you'll pass this parameter to `args` the application will raise an error. Why? Because this flag was removed.
So, I don't see any reason to keep this parameter.
You can try it on your own.

lucacome: @0m1xa the flag wasn't removed, there hasn't been a version release with it yet. That's why I suggested you try with `edge` which is a container built from the `main` branch instead of a release. Or you can build the container yourself from your branch and see that it works with `includeYear`.

jasonwilliams14: @0m1xa as @lucacome mentioned, there is not a version released yet with that specific argument. It will be in a future release which is why it needs to remain in the code.
`edge` should work or building the container from your branch are options.

0m1xa: Ok. I've reverted this

deployments/helm-chart/values.yaml
```diff
@@ -184,10 +184,10 @@ controller:
 
   ## Enable OIDC policies.
   enableOIDC: false
-
+
```
lucacome, line 187:
```suggestion
 
```

deployments/helm-chart/values.yaml
```diff
@@ -184,10 +184,10 @@ controller:
 
   ## Enable OIDC policies.
   enableOIDC: false
-
+  
   ## Include year in log header. This parameter will be removed in release 2.7 and the year will be included by default.
   includeYear: false
-
+
```
lucacome, line 190:
```suggestion
 
```

deployments/helm-chart/values.yaml
```diff
@@ -302,6 +302,9 @@ controller:
     customPorts: []
 
   serviceAccount:
+    ## The annotations of the Ingress Controller service.
```
lucacome, line 305:
```suggestion
    ## The annotations of the service account of the Ingress Controller pods.
```

jasonwilliams14 (approved):
Looks good.

lucacome (approved):

PR #3088, opened 2022-09-22
Feat/ignore helm secrets

shaun-nx:
### Proposed changes
This change will filter out secrets of type `helm.sh/release.v1` from the k8s api.

### Checklist


vepatel (changes requested):
@shaun-nx can you please mark these test as `@pytest.mark.ingresses` as secret isn't valid selector in pipeline so these won't be triggered

ciarams87 (approved):

vepatel (approved):

PR #3094, opened 2022-09-23
Optimise path validation

haywoodsh:
### Proposed changes
Optimize path validation and add test cases.

### Checklist

### DIFF
--- a/internal/k8s/validation.go
+++ b/internal/k8s/validation.go
@@ -72,18 +72,15 @@ const (
 const (
 	commaDelimiter     = ","
 	annotationValueFmt = `([^"$\\]|\\[^$])*`
-	pathFmt            = `/[^\s{};\\]*`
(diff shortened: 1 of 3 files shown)


shaun-nx (approved):
Great work! 👍

vepatel (approved):

ciarams87 (approved):
🚀

luanmm:
Why consider paths like "/root", "/etc/nginx/secrets", etc... as invalids? This makes this ingress unusable in a project that has a "/root" path on it (that can't change the name, because it is a big project running for several years).

I was using normally before this PR and now can't update the cluster just because of this strange validation. If a path passes the official RFC, it should be valid. My scenario have a "/root" that has nothing to do with Linux "/root" folder (if that is the intention of the validation).

Other validations seems useful, but this one is just nonsense (with all the respect).

PR #3104, opened 2022-09-27
Eliminate dead code

jjngx:
### Proposed changes

This PR eliminates few dead code paths.

Note that the value of ```err``` returned by this func is always ```nil```. This PR is a result of small refactoring that simplifies logic by removing dead code from the ```configs``` package.
```go
// GetMapKeyAsStringSlice tries to find and parse a key in the map as string slice splitting it on delimiter.
func GetMapKeyAsStringSlice(m map[string]string, key string, _ apiObject, delimiter string) ([]string, bool, error) {
    if str, exists := m[key]; exists {
	slice := strings.Split(str, delimiter)
	return slice, exists, nil
    }
    return nil, false, nil
}
```

Refactored func:
```go
// GetMapKeyAsStringSlice tries to find and parse a key in the map as string slice splitting it on delimiter.
func GetMapKeyAsStringSlice(m map[string]string, key string, _ apiObject, delimiter string) ([]string, bool) {
    if str, exists := m[key]; exists {
	slice := strings.Split(str, delimiter)
	return slice, exists
    }
    return nil, false
}
```

### Checklist

### DIFF
--- a/internal/configs/annotations.go
+++ b/internal/configs/annotations.go
@@ -148,20 +148,12 @@ func parseAnnotations(ingEx *IngressEx, baseCfgParams *ConfigParams, isPlus bool
 		}
 	}
 
-	if serverSnippets, exists, err := GetMapKeyAsStringSlice(ingEx.Ingress.Annotations, "nginx.org/server-snippets", ingEx.Ingress, "\n"); exists {
-		if err != nil {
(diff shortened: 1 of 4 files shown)


shaun-nx (approved):
👍

ciarams87 (approved):

PR #3113, opened 2022-09-29
Add JSON Schema for Helm Chart

lucacome:
Adds Schema file for Helm Chart https://helm.sh/docs/topics/charts/#schema-files

It will help validate that the correct values are used for example during `install`. This is the first pass at it, the second pass will be adding the required values for the chart and other enforcements.


jjngx (approved):

haywoodsh (approved):

ciarams87 (approved):
Nice! 🚀

PR #3142, opened 2022-10-12
Support multiple loadbalancers

orkhanM:
### Proposed changes
In some scenarios a user might chose to have multiple loadbalancers terminating connections for nginx / nginx+. 

In our case we need to be able to provision an external load balancer and an internal loadbalancer respectively with their own distinct annotations.

### Checklist


brianehlert:
If I followed correctly, you want some hostnames to be linked to an external loadbalancer and other hostnames to be linked to an internal loadbalancer.
And you want this with one Ingress Controller deployment?  

Not an internal deployment and an external deployment.  But rather one ingress controller deployment that is linked to both external and internal loadbalancers at the same time.

Is that what your intention is?

orkhanM:
@brianehlert Yes, is the recommended approach for this use case to deploy two separate ingress controllers?

jasonwilliams14:
@orkhanM Yes, it is very easy and recommended to use multiple deployments per @brianehlert recomendation.

PR #3170, opened 2022-10-18
Watch subset of namespaces for secrets

### LINKED ISSUE #3154: Allow configuration of namespaces watched for secrets

Either all (by default) or a subset (new 2.4 feature) of namespaces are watched for ingress related resources BUT ALSO SECRETS..

This means all found secrets are cached in the ingress controller even if not needed/used. This of course is a security concern.

With multi tenancy and namespace isolation as described by:

 "Enabling Multi-Tenancy and Namespace Isolation in Kubernetes with NGINX - NGINX" https://www.nginx.com/blog/enabling-multi-tenancy-namespace-isolation-in-kubernetes-with-nginx/

There is no requirement/need for the controller to read secrets from the application namespaces.

Thus security can be improved by allowing to configure what namespaces should be watched for secrets.

Using custom RBAC roles and bindings the cluster admin can fine tune the controller privileges to the minimum for it to work.

One possible and simple solution is to provide an additional flag similar to "watch-namespace" but for secrets. Let's call it "watch-secret-namespaces".

For this I have a patch I will submit for review.

Other implementations are more complex using for example annotations/labels on namespaces

**Additional context**

hafe:
Add new command line option "watch-secret-namespace" that can be used to configure namespaces watched for secrets.

Closes #3154

### Proposed changes

### Checklist


hafe:
Needs some fixing/testing

hafe:
Tested and works as expected!

cmd/nginx-ingress/flags.go
```diff
@@ -200,6 +205,16 @@ func parseFlags(versionInfo string, binaryInfo string) {
 	glog.Info(binaryInfo)
 
 	watchNamespaces = strings.Split(*watchNamespace, ",")
+	glog.Infof("Namespaces watched: %v", watchNamespaces)
+
+	if len(*watchNamespace) > 0 {
```
ciarams87, line 210: Shouldn't this be 
```suggestion
	if len(*watchSecretNamespaces) > 0 {
```

deployments/helm-chart/values.yaml
```diff
@@ -183,6 +183,9 @@ controller:
   ## Comma separated list of namespaces to watch for Ingress resources. By default the Ingress Controller watches all namespaces.
   watchNamespace: ""
 
+  ## Comma separated list of namespaces to watch for Sercet resources. By default the Ingress Controller watches all namespaces.
```
ciarams87, line 186: Nit: typo
```suggestion
  ## Comma separated list of namespaces to watch for Secret resources. By default the Ingress Controller watches all namespaces.
```

cmd/nginx-ingress/flags.go
hafe: Yes seems more correct!

cmd/nginx-ingress/flags.go
```diff
@@ -40,6 +40,11 @@ var (
 
 	watchNamespaces []string
 
+	watchSecretNamespace = flag.String("watch-secret-namespace", "",
+		`Comma separated list of namespaces the Ingress Controller should watch for secrets. By default the Ingress Controller watches all namespaces`)
```
ciarams87, line 44: Maybe something like this would be better? This makes it seem like all namespaces will be watched for secrets even if `watch-namespace` is set.
```suggestion
		`Comma separated list of namespaces the Ingress Controller should watch for secrets. If this arg is not configured, the Ingress Controller watches the same namespaces for all resources. See `watch-namespace`. `)
```

docs/content/installation/installation-with-helm.md
```diff
@@ -185,6 +185,7 @@ The following tables lists the configurable parameters of the NGINX Ingress Cont
...
 |``controller.setAsDefaultIngress`` | New Ingresses without an ingressClassName field specified will be assigned the class specified in `controller.ingressClass`. | false |
 |``controller.watchNamespace`` | Comma separated list of namespaces the Ingress Controller should watch for resources. By default the Ingress Controller watches all namespaces. Please note that if configuring multiple namespaces using the Helm cli `--set` option, the string needs to wrapped in double quotes and the commas escaped using a backslash - e.g. ``--set controller.watchNamespace="default\,nginx-ingress"``. | "" |
+|``controller.watchSecretNamespace`` | Comma separated list of namespaces the Ingress Controller should watch for resources of type Secret. By default the Ingress Controller watches all namespaces. Please note that if configuring multiple namespaces using the Helm cli `--set` option, the string needs to wrapped in double quotes and the commas escaped using a backslash - e.g. ``--set controller.watchSecretNamespace="default\,nginx-ingress"``. | "" |
```
ciarams87, line 188: Ditto.
```suggestion
|``controller.watchSecretNamespace`` | Comma separated list of namespaces the Ingress Controller should watch for resources of type Secret. If this arg is not configured, the Ingress Controller watches the same namespaces for all resources. See `watch-namespace`. Please note that if configuring multiple namespaces using the Helm cli `--set` option, the string needs to wrapped in double quotes and the commas escaped using a backslash - e.g. ``--set controller.watchSecretNamespace="default\,nginx-ingress"``. | "" |
```

ciarams87:
Looks good @hafe! Just a couple of other docs improvements I noticed, and a rebase to main, and we should be good to merge! Thanks so much for the contribution 😃

ciarams87 (approved):
LGTM! Thank you!

cmd/nginx-ingress/flags.go
```diff
@@ -40,6 +40,11 @@ var (
 
 	watchNamespaces []string
 
+	watchSecretNamespace = flag.String("watch-secret-namespace", "",
+		`Comma separated list of namespaces the Ingress Controller should watch for secrets. If this arg is not configured, the Ingress Controller watches the same namespaces for all resources. See `watch-namespace`. `)
```
ciarams87, line 44: Apologies - this is the issue with suggestions committed from the UI 🤦🏼‍♀️ 
```suggestion
		`Comma separated list of namespaces the Ingress Controller should watch for secrets. If this arg is not configured, the Ingress Controller watches the same namespaces for all resources. See "watch-namespace". `)
```

jjngx (approved):

PR #3260, opened 2022-11-16
Add support for EndpointSlices

shaun-nx:
### Proposed changes
This PR updates the Ingress Controller to support [EndpointSlices](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)

### Checklist


deployments/rbac/rbac.yaml
```diff
@@ -3,11 +3,18 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: nginx-ingress
 rules:
```
coolbry95, line 5: Helm will need to be updated too.

shaun-nx: Nice catch!

internal/k8s/controller_test.go
```diff
@@ -470,79 +472,516 @@ func TestFormatWarningsMessages(t *testing.T) {
...
+				},
+			},
+		},
+		Status: api_v1.ServiceStatus{},
+	}
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			gotEndpoints, _ := lbc.getEndpointsForPortFromEndpointSlices(test.svcEndpointSlices, backendServicePort, &svc)
+
+			if result := unorderedEqual(gotEndpoints, test.expectedEndpoints); !result {
```
jjngx, line 640: would the `cmp`  package be helpful here?
```go
if !cmp.Equal(gotEndpoints, test.expectedEndpoints) {
    t.Error(cmp.Diff(gotEndpoints, test.expectedEndpoints))
}
```

internal/k8s/controller_test.go
```diff
@@ -470,79 +472,516 @@ func TestFormatWarningsMessages(t *testing.T) {
...
+			gotEndpoints := getEndpointsFromEndpointSlicesForSubselectedPods(test.targetPort, test.pods, test.svcEndpointSlices)
+
+			if result := unorderedEqual(gotEndpoints, test.expectedEndpoints); !result {
+				t.Errorf("getEndpointsFromEndpointSlicesForSubselectedPods() = got %v, want %v", gotEndpoints, test.expectedEndpoints)
 			}
 		})
 	}
 }
 
+func unorderedEqual(got, want []podEndpoint) bool {
```
jjngx, line 969: what if ```!cmp.Equal(got, want)``` ?

internal/k8s/controller_test.go
shaun-nx: I tried using `cmp.Diff` and noticed that the test would fail if the value of `gotEndpoint` was not in the same order. On some runs I would get a result like this:

```
lbc.getEndpointsForPortFromEndpointSlices() 
 got [{10.0.0.1:8080  { }} {1.2.3.4:8080  { }} {5.6.7.8:8080  { }}], 
 want [{1.2.3.4:8080  { }} {5.6.7.8:8080  { }} {10.0.0.1:8080  { }}]
```

internal/k8s/controller.go
```diff
@@ -3409,55 +3421,63 @@ func (lbc *LoadBalancerController) getEndpointsForIngressBackend(backend *networ
 		return nil, false, err
 	}
 
-	result, err = lbc.getEndpointsForPort(endps, backend.Service.Port, svc)
+	result, err = lbc.getEndpointsForPortFromEndpointSlices(endpointSlices, backend.Service.Port, svc)
 	if err != nil {
-		glog.V(3).Infof("Error getting endpoints for service %s port %v: %v", svc.Name, configs.GetBackendPortAsString(backend.Service.Port), err)
+		glog.V(3).Infof("Error getting endpoitnslices for service %s port %v: %v", svc.Name, configs.GetBackendPortAsString(backend.Service.Port), err)
```
ciarams87, line 3426: Nit: typo
```suggestion
		glog.V(3).Infof("Error getting endpointslices for service %s port %v: %v", svc.Name, configs.GetBackendPortAsString(backend.Service.Port), err)
```

internal/k8s/controller.go
```diff
@@ -3409,55 +3421,63 @@ func (lbc *LoadBalancerController) getEndpointsForIngressBackend(backend *networ
...
 			if err != nil {
-				return nil, fmt.Errorf("error determining target port for port %v in Ingress: %w", backendPort, err)
+				return 0, fmt.Errorf("error determining target port for port %v in Ingress: %w", backendPort, err)
 			}
 			break
 		}
 	}
 
-	if targetPort == 0 {
+	return targetPort, nil
```
ciarams87, line 3443: Do we not need to return an error if targetPort is 0 any more?

shaun-nx: We do still return an error. It happens within the `getEndpointsForProtFromEndpointSlices` function. I moved code that validated the target port into its own function called `validateTargetPort`. The error we return now will happen if we get an error back from that function.

internal/k8s/controller.go
```diff
@@ -3409,55 +3421,63 @@ func (lbc *LoadBalancerController) getEndpointsForIngressBackend(backend *networ
...
 		return nil, false, err
 	}
 	return result, false, nil
 }
 
-func (lbc *LoadBalancerController) getEndpointsForPort(endps api_v1.Endpoints, backendPort networking.ServiceBackendPort, svc *api_v1.Service) ([]podEndpoint, error) {
-	var targetPort int32
-	var err error
-
+func (lbc *LoadBalancerController) validateTargetPort(backendPort networking.ServiceBackendPort, svc *api_v1.Service) (targetPort int32, err error) {
```
ciarams87, line 3432: We shouldn't use named return parameters here - they are really only for very short utility functions, and require a bare return statement (i.e. just the keyword return). To improve the readability of the code, we should opt for explicitly returning the result parameters (i.e. not use named parameters). This means removing the names from the return parameter, and initialising them in the function (i.e. returning lines 3421 and 3422)

internal/k8s/controller.go
ciarams87: This is the validatePort function 😄.  I see we are returning an error, above on line 3437, but we aren't checking if targetPort is 0 as far as I can see before returning. We could have a case where no targetPort has been found for the backendPort/service combinations but no errors have ben noted. TargetPort will have been initialised to 0, which is not a valid port and will cause errors later in the code where it will be harder to find, so we need to check this somewhere.

shaun-nx: Ah, I see what you mean. Thanks for catching that. I'll add a check and a unit test for that case.

internal/k8s/controller_test.go
```diff
@@ -470,79 +472,516 @@ func TestFormatWarningsMessages(t *testing.T) {
...
+					Port:       80,
+					TargetPort: intstr.FromInt(8080),
+				},
+			},
+		},
+		Status: api_v1.ServiceStatus{},
+	}
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			gotEndpoints, _ := lbc.getEndpointsForPortFromEndpointSlices(test.svcEndpointSlices, backendServicePort, &svc)
```
jjngx, line 638: Does this func return error? If so, what happens if the err is not nil?

shaun-nx: Whenever and error is returned I also return `nil` for `[]podEndpoints` and that I set that as my expected value of `gotEndpoints`.

jjngx (commented):
how about:

```go
func TestGetEndpointsFromEndpointSlieces_ErrorsOnInvalidInput(t *testing.T) {
// content goes here
}
```

Since we are interested to test if the method errors on invalid inputs (sourced from "table tests") how about:
```go
_, err := lbc.getEndpointsForPortFromEndpointSlices(....)
if err == nil {
    t.Error("want error, got nil")
}
```
Since we are interested in the fact that the func returns an error, lines 743...745 seem to be of no use.

In general, it may be worth to leverage ```cmp.Equal``` method since we use the ```cmp``` package anyway. How about:
```go
if !cmp.Equal(want, got) {
    t.Error(cmp.Diff(want, got))
}
```
Note, that we defer calculating Diff, and do so only if the test fails. Checking for equality is cheeper comparing to calculating ```Diff```.

Just food for thought.

On line 745... it is worth to ask a question: "Do we really care what message we get?" Do we run any business logic (decision) based on this error? If yes, this may be a candidate for using sentinel errors, and would require to check what type of the error we get (```errors.Is()``` method from the errors package.

ciarams87 (approved):
🚀

PR #3284, opened 2022-11-19
Add label for Helm Chart

lucacome:
This adds configuration for automatically labeling PRs with `helm_chart` when there's a change to any of the files in the helm chart directory.
We use those labels to create the sections in the [releases](https://github.com/nginxinc/kubernetes-ingress/releases), so this change will add a new `Helm Chart` section if we had any PRs modifying the helm chart.

Since we're having quite a few PRs modifying the helm chart, this will group them together and "highlight" them.


tomasohaodha (changes requested):
Can you please add more detail to the PR description for those not as familiar as you with github actions? "Adds label for Helm Chart changes." is factual but not helpful - we want PRs to be a learning tool not just a review tool. Thanks

ciarams87 (approved):

tomasohaodha (approved):

PR #3299, opened 2022-11-23
feature: Support Dynamic namespaces using Labels

ciarams87:
### Proposed changes
Provide a new CLI arg `watch-namespace-label` which supports watching for a namespace label instead of for discrete namespace names - this allows the list of watched namespaces to change dynamically based on a selector. The IC will watch for namespaces with the label selector and will dynamically add or remove watched namespaces as namespaces have this label added or removed.

### Checklist


cmd/nginx-ingress/main.go
```diff
@@ -243,6 +242,25 @@ func validateIngressClass(kubeClient kubernetes.Interface) {
...
 }
 
+func checkNamespaces(kubeClient kubernetes.Interface) {
+	if *watchNamespaceLabel != "" {
+		// bootstrap the watched namespace list
+		var newWatchNamespaces []string
+		nsList, err := kubeClient.CoreV1().Namespaces().List(context.TODO(), meta_v1.ListOptions{LabelSelector: *watchNamespaceLabel})
+		if err != nil {
+			glog.Warningf("Error when getting Namespaces with the label selector %v: %v", watchNamespaceLabel, err)
+		}
```
shaun-nx, line 252: I think we can remove the word "Error" from this log since it's a warning log. Could be changed to "unable to get Namespaces with the label selector"

ciarams87: Changed it to an error message

hafe (commented):
Why not have a fixed name namespace label? Reduces complexity of the change

deployments/helm-chart/values.yaml
```diff
@@ -52,10 +52,10 @@ controller:
 
   image:
     ## The image repository of the Ingress Controller.
-    repository: nginx/nginx-ingress
+    repository: gcr.io/f5-gcs-7899-ptg-ingrss-ctlr/dev/nginx-ic/nginx-plus-ingress
 
     ## The tag of the Ingress Controller image.
-    tag: "2.4.1"
+    tag: "2.4.1-labels"
```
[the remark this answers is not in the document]
ciarams87: Good catch, thanks!

ciarams87:
> Why not have a fixed name namespace label? Reduces complexity of the change

@hafe I'm not sure what you mean here?

hafe:
> > Why not have a fixed name namespace label? Reduces complexity of the change
> 
> @hafe I'm not sure what you mean here?

That you decide and use a hardcoded label name.

What is the purpose of being able to change the name on the command line?

ciarams87:
> > > Why not have a fixed name namespace label? Reduces complexity of the change
> > 
> > 
> > @hafe I'm not sure what you mean here?
> 
> That you decide and use a hardcoded label name.
> 
> What is the purpose of being able to change the name on the command line?

Thanks for the feedback @hafe. The purpose here is so that the user can configure the Ingress Controller to watch whatever label makes sense for their environment. It cannot be changed during runtime, so it is effectively hardcoded in the deployment spec. If the label needs to be changed, a redeployment will be required to do so.

hafe:
I was thinking that freedom to change the name is perhaps not needed

hafe:
> I think that one thing we were cognizant of when we were designing this, is that a customer could have multiple NGINX Ingress Controller deployments running within any cluster at the same time, and thus we could not have only one hard coded label as the two deployments would clash.
> This way each deployment can have a unique label, just like they might have unique namespaces.

Of course, thanks for the explanation!

shaun-nx (approved):
👏

vepatel (approved):
Great stuff 👏🏼

PR #3309, opened 2022-11-24
Use Debian 11 with NAP WAF

lucacome:
Updates the `Dockerfile` and `Makefile` to use Debian 11 with NAP WAF. 

The builds will fail until R28 comes out.


ciarams87:
These changes were included with https://github.com/nginxinc/kubernetes-ingress/pull/3320 - Debian 10 support was removed, so the changes were required in the Plus version bump PR for the pipeline to succeed

PR #3337, opened 2022-12-06
feat: Fix minor issues, add docs contribution quickstart.

ADubhlaoich:
This commit addresses two issues flagged as problems in documentation. One is a small typo, the other is an incorrect link. Additionally, it adds a README to the docs subfolder to explain how to run a local Hugo instance for testing and the broad strokes of contributing.

### Checklist


docs/README.md
```diff
@@ -0,0 +1,29 @@
...
+git clone git@github.com:nginxinc/kubernetes-ingress.git
+cd docs/
+hugo server -e production -b "http://127.0.0.1"
+```
+## Style Guide
+NGINX (and F5) has its own style guide, but broadly speaking, you can stick to the [Microsoft Style Guide](https://learn.microsoft.com/en-us/style-guide/welcome/).
+
+We prefer the active voice, and plain language. When possible, be concise and avoid jargon - don't assume the reader knows as much as you.
+
+If an acronym or abbreviation could cause confusion to someone, explicitly define it during its first use, e.g *Security Operations Centor (SOC)*.
```
ADubhlaoich, line 29:
```suggestion
If an acronym or abbreviation could cause confusion to someone, explicitly define it during its first use, e.g *Security Operations Center (SOC)*.
```

jjngx (approved):

docs/README.md
```diff
@@ -0,0 +1,29 @@
...
+## Set-up
+**Requirements**
+* [git](https://git-scm.com/downloads)
+* [hugo](https://gohugo.io/installation/)
+
+**Quick Start**
+```
+git clone git@github.com:nginxinc/kubernetes-ingress.git
+cd docs/
+hugo server -e production -b "http://127.0.0.1"
```
jputrino, line 22: It shouldn't be necessary to use the production environment when running hugo locally (needs verification, though). You should be able to just run `hugo server`. (Using the production environment inserts `docs.nginx.com` as the baseUrl instead of just using localhost.)

docs/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+## Style Guide
+NGINX (and F5) has its own style guide, but broadly speaking, you can stick to the [Microsoft Style Guide](https://learn.microsoft.com/en-us/style-guide/welcome/).
+
+We prefer the active voice, and plain language. When possible, be concise and avoid jargon - don't assume the reader knows as much as you.
```
jputrino, line 27: As opposed to "As much as you", I think we shouldn't assume the reader is familiar with the same internal terms or acronyms that we use. A good guideline is to write with empathy and to try to keep the language at about 9th grade or below (the online app Hemingway lets you check for this.)

docs/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+If an acronym or abbreviation could cause confusion to someone, explicitly define it during its first use, e.g *Security Operations Center (SOC)*.
```
jputrino, line 29: We also avoid using abbreviations like "e.g.", "i.e.", and "etc.", as these are not translation friendly. This is a good rule to call out.

docs/content/installation/installation-with-operator.md
```diff
@@ -17,7 +17,7 @@ This document describes how to install the NGINX Ingress Controller in your Kube
     * For NGINX Plus Ingress Controller, see [here](/nginx-ingress-controller/installation/pulling-ingress-controller-image) for details on how to pull the image from the F5 Docker registry.
     * To pull from the F5 Container registry, configure a docker registry secret using your JWT token from the MyF5 portal by following the instructions from [here](/nginx-ingress-controller/installation/using-the-jwt-token-docker-secret).
     * It is also possible to build your own image and push it to your private Docker registry by following the instructions from [here](/nginx-ingress-controller/installation/building-ingress-controller-image).
-2. Install the NGINX Ingress Operator following the [instructions](https://github.com/nginxinc/nginx-ingress-helm-operator/blob/v1.2.1/docs/installation.md).
+2. Install the NGINX Ingress Operator following the [instructions](https://github.com/nginxinc/nginx-ingress-helm-operator/blob/main/docs/manual-installation.md).
```
jputrino, line 20: Should this link to the latest release branch?

jputrino (approved):

docs/README.md
```diff
@@ -0,0 +1,19 @@
+# How To Contribute
+## Introduction
+The NGINX Ingress Controller makes use of the [Hugo](https://gohugo.io/) static site generator.
+
+Documentation is stored within the `docs/`subfolder of the main repository, and is where `hugo` commands should be run.
```
lucacome, line 5:
```suggestion
Documentation is stored within the `docs/` subfolder of the main repository, and is where `hugo` commands should be run.
```

docs/content/installation/installation-with-operator.md
lucacome: I wouldn't change this, I think we want to point to the latest release (tag)

lucacome (approved):

PR #3392, opened 2022-12-22
Add feature: Overwrite default proxy_http_version

sass1997:
Overwrite the default proxy_http_version to v1.1

### Proposed changes
If an OIDC Provider forces you to use HTTP 1.1 nginx will receive a http response 426 which is at the moment not handled. To make this compatiblty I suggest to increase it by default. If there is the need to still have 1.0 I'm happy to configure it as configuration parameter

### Checklist

- [x ] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc
- [ x] I have checked that all unit tests pass after adding my changes
- [ x] I have updated necessary documentation
- [ x] I have rebased my branch onto main
- [ x] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

### DIFF
--- a/internal/configs/oidc/oidc.conf
+++ b/internal/configs/oidc/oidc.conf
@@ -14,6 +14,7 @@
         proxy_ssl_server_name on;                     # For SNI to the IdP
         proxy_method GET;                             # In case client request was non-GET
         proxy_set_header Content-Length "";           # ''
+        proxy_http_version    1.1;
         proxy_pass $oidc_jwt_keyfile;                 # Expecting to find a URI here
         proxy_ignore_headers Cache-Control Expires Set-Cookie; # Does not influence caching
     }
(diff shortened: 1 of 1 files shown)


shaun-nx:
Let's take the time to make this configurable. Keep the default as 1.0 and allow it to be set to 1.1 if needed.
Do you have an example of OIDC providers that force HTTP1.1?

jasonwilliams14:
@sass1997 You will need to open up a PR with our openid project repo. We "pull" in that project into our NGINX Ingress project. Any changes you would propose, need to be requested in that repo. here is the link:

https://github.com/nginxinc/nginx-openid-connect

sass1997:
@shaun-nx -> okey will do this configurable. Not yet I'm mainly working on OIDC Implementations which aren't public used like google, linkedin or similiar.
@jasonwilliams14 -> thank your for the link. Then I will create a PR overthere and mention it here as soon as I have it

brianehlert:
This change needs to be submitted to the NGINX OIDC reference implementation repo.
After which an update PR can be opened in this project to update with those changes.
This project does not intend to be a fork of the reference implementation, while still supporting the customization that customers require.

PR #3440, opened 2023-01-13
Add dependabot config for release branch

lucacome:
Adds dependabot config to keep go dependencies up to date in the latest release branch

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,38 +1,38 @@
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


jjngx (approved):

vepatel (approved):

ciarams87 (changes requested):
I'm not sure we should be making these automatic updates to the release branch - can I ask what's the motivation behind doing so? When we make patch releases, I don't know that we should always be bringing in all these unrelated dependency bumps.

PR #3451, opened 2023-01-18
feat: Enable keepalive-time for healthchecks in VS and VSR

ciarams87:
### Proposed changes
[Keepalive connections for health checks](https://www.nginx.com/blog/nginx-plus-r27-released/#keepalive-health-checks) – The new keepalive_time parameter to the HTTP health_check directive enables keepalive connections for health checks and sets their lifetime. Establishing new connections is computationally expensive, so reusing connections can reduce CPU usage considerably when there are large numbers of upstream servers.

### Checklist


tests/data/virtual-server-upstream-options/plus-virtual-server-with-invalid-keys-openapi.yaml
```diff
@@ -19,7 +19,8 @@ spec:
       enable: "yes"
       path: 200
       interval: 1.5
-      jitter: 2.0
+      jitter: 2.
```
shaun-nx, line 22: Zero looks to have been removed here by mistake.

jjngx (approved):

tests/data/virtual-server-upstream-options/plus-virtual-server-with-invalid-keys-openapi.yaml
ciarams87: Good catch!

shaun-nx (approved):

jasonwilliams14 (commented):
I think it looks good. We can set a default between 30-60 seconds.
we can also give customers to disable entirely if they choose to.
it is a great new feature that will improve performance and HC effectiveness.

Issue #2635, opened 2022-04-27
OIDC: option to pass access token as authorization header to upstream


suedschwede:
I would like to configure OIDC in NGINX and pass the access token as authorization header to a backend application
At the moment it is only possible to pass jwt claims

lucacome:
Hi @suedschwede 

we use https://github.com/nginxinc/nginx-openid-connect under the hood. I would suggest opening an issue there.

nixx:
I have the same issue. This is a blocker for using Nginx Ingress controller.

My case is slightly different. I need to pass the ID token to upstream. This is already supported in nginx-openid-connect.

I tried to add $session_jwt to a header through policies, but are denied as there is a whitelist at pkg/apis/validation/virtualserver.go:1094 (on the v2.2.2 tag). 

Username is also hard coded in the template (nginx-plus.virtualserver.tmpl:363).

I´m able to add a workaround if I patch the code, but I would rather not.

In the longer run, I assume we also need to pass access token to upstream. Exposing tokens to upstream is a functionality most other similar software has, and this is a blocker for using Nginx Plus now.

PR #3474, opened 2023-01-23
feat: Add access token support in the OIDC

shawnhankim:
### Proposed changes
- The current OIDC only support ID token for user authentication and does not support access token for the API authorization yet. The feature of access token has been recently added in the [NGINX Plus OIDC repo](https://github.com/nginxinc/nginx-openid-connect), and this PR is to synchronize the feature into NGINX Ingress Controller.
- This change is to add `$access_token` support for authorizing protected resources in the OpenID Connect (OIDC).
- This PR is also to resolve the issue #2635 to enable OIDC option (`accessTokenEnable`) to pass `$access_token` as authorization header to upstream.

### Checklist


docs/content/configuration/policy-resource.md
```diff
@@ -384,7 +385,8 @@ The OIDC policy defines a few internal locations that can't be customized: `/_jw
 |``jwksURI`` | URL for the JSON Web Key Set (JWK) document provided by your OpenID Connect provider. | ``string`` | Yes |
 |``scope`` | List of OpenID Connect scopes. Possible values are ``openid``, ``profile``, ``email``, ``address`` and ``phone``. The scope ``openid`` always needs to be present and others can be added concatenating them with a ``+`` sign, for example ``openid+profile+email``. The default is ``openid``. | ``string`` | No |
 |``redirectURI`` | Allows overriding the default redirect URI. The default is ``/_codexch``. | ``string`` | No |
-|``zoneSyncLeeway`` | Specifies the maximum timeout in milliseconds for synchronizing ID tokens and shared values between Ingress Controller pods. The default is ``200``. | ``int`` | No |
+|``zoneSyncLeeway`` | Specifies the maximum timeout in milliseconds for synchronizing ID/access tokens and shared values between Ingress Controller pods. The default is ``200``. | ``int`` | No |
+|``accessTokenEnable`` | Option of whether Bearer token is used to authorize NGINX to access protected backend. | ``boolean`` | No |
```
shawnhankim, line 389: @jputrino, @travisamartin: 
- I would appreciate it if you could either review or fix this description when you get a chance. (FYI: This feature has been merged into the [NGINX Plus OIDC repo](https://github.com/nginxinc/nginx-openid-connect#access-tokens). 
- You could also feel free to raise any PR if you want to fix docs regarding the [access token in the repo](https://github.com/nginxinc/nginx-openid-connect#access-tokens).)

shawnhankim:
@brianehlert : This PR is one of OIDC enhancements which are discussed in the Slack channel.
@shaun-nx, @brianehlert : I would appreciate it if you could review this PR. Please let me know if I miss anything.

brianehlert:
@shawnhankim 
This appears to support passing the access token to the upstream / backend.  Is that correct?
When this project takes updates from the OIDC repository, we try to take all the changes and therefore maintain the OIDC reference implementation as a source of truth, and not that this is some random fork.  Does this PR include all enhancements?  (Whether exposed or not)

shawnhankim:
@brianehlert 
> This appears to support passing the access token to the upstream / backend. Is that correct?
* Yes, it is correct.
* So, we can answer to the question of `ID token` (for user authentication) / `access token` (for API/protected-resource authorization) such as this kind of [article](https://auth0.com/blog/id-token-access-token-what-is-the-difference/) based on this additional feature of `access token` as we already have the `session_jwt` which is the `id_token`.

shawnhankim:
@brianehlert 
> When this project takes updates from the OIDC repository, we try to take all the changes and therefore maintain the OIDC reference implementation as a source of truth, and not that this is some random fork. Does this PR include all enhancements? (Whether exposed or not)

* This PR is to synchronize the latest version of OIDC reference implementation which is [the source of truth](https://github.com/nginxinc/nginx-openid-connect/pull/71).  
* The rest of enhancements are being reviewed by NGINX Plus Team. Those PRs are going to be separately synchronized here once they are merged in the source repo of truth based on N+ team's schedule.

Please let me know if I miss anything about your question.

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -423,6 +423,11 @@ server {
         {{ $proxyOrGRPC }}_set_header username $jwt_claim_sub;
         {{ end }}
 
+        {{ if $s.OIDC.AccessTokenEnable }}
+        {{ $proxyOrGRPC }}_set_header Authorization "Bearer $access_token";
+        {{ $proxyOrGRPC }}_intercept_errors on;
```
lucacome, line 428: do we always want to intercept errors? 
I see this comment in the repo
```
# Intercept and redirect "401 Unauthorized" proxied responses to nginx
# for processing with the error_page directive. Necessary if Access Token
# can expire before ID Token.
```
but I'm not sure if we could have side effects if it's always enabled 🤔

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -423,6 +423,11 @@ server {
         {{ $proxyOrGRPC }}_set_header username $jwt_claim_sub;
         {{ end }}
 
+        {{ if $s.OIDC.AccessTokenEnable }}
```
lucacome, line 426: This could cause a nil pointer exception, we should check if OIDC is set first like we do above
```
{{ with $oidc := $s.OIDC }}
```

lucacome (changes requested):

internal/configs/oidc/openid_connect.js
```diff
@@ -101,9 +101,14 @@ function auth(r, afterSyncCheck) {
                             return;
                         }
 
-                        // ID Token is valid, update keyval
+                        // Update ID token and access token in the key-value store
```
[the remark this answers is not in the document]
shawnhankim: Sure, I have reverted this comment change.

internal/configs/version2/nginx-plus.virtualserver.tmpl
shawnhankim: - Thanks for the good question. 

- Yeah, we could think of several scenarios.
  - **use case 1**. if customers always require to login via IdP to access their proxied apps:
     - The`{{ $proxyOrGRPC }}_intercept_errors on` might help based on the N+ Ref. Implementation.
  - **use case 2**. if customers access a landing page (which isn't required the login for the first time) to show public information without OIDC flow, and if they want to access proxied APIs to show private information after they click `login` button via UI:
     - The `{{ $proxyOrGRPC }}_intercept_errors on` can be removed. Because the UI app can either just show error message or call `/login` endpoint (app level location) to start `@do_oidc_flow` from their browser.
- So, I have added an additional option (`InterceptErrorEnable`) to enable intercepting error for customers to choose their preference.

Let me know if you have other opinion.

internal/configs/version2/nginx-plus.virtualserver.tmpl
shawnhankim: Thanks for your feedback. I have addressed it. Let me know if I miss anything.

internal/configs/version2/nginx-plus.virtualserver.tmpl
jasonwilliams14: @lucacome I agree. I think we should go with errorPages. Since we already use `proxy_intercept_errors` I am concerned that we will generate duplicate directives of `proxy_intercept_errors` which will cause NGINX to fail.
I think safer approach is `error_pages`.

lucacome:
@shawnhankim can you rebase the PR with `main` and run `make update-crds`?

shawnhankim:
>

shawnhankim:
> @shawnhankim can you rebase the PR with `main` and run `make update-crds`?

Thanks @lucacome for this feedback and letting me know. Addressed it.

internal/configs/version2/nginx-plus.virtualserver.tmpl
shawnhankim: - Thanks @lucacome and @jasonwilliams14 for your feedback.
- I have just removed the `InterceptErrorEnable` and `{{ $proxyOrGRPC }}_intercept_errors` for this PR as the following reasons:
  - There is already the same directive.
  - There is already `error_page 401 = @do_oidc_flow` that is created in the config once the OIDC is enabled.
     - If `@do_oidc_flow` is part of the `ErrorPage`, and if we remove the existing directive, then it might be inconvenient for customers.
     - If we just add `error_page 401 = @do_oidc_flow` using the `ErrorPage`, it would be duplicated config as well. 
- Therefore, I removed it for this PR as for now.
- Feel free to let me know if you have better ideas.

lucacome (approved):
Thanks @shawnhankim !

shaun-nx (approved):

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -429,6 +429,12 @@ server {
         {{ $proxyOrGRPC }}_set_header username $jwt_claim_sub;
         {{ end }}
 
+        {{ with $oidc := $s.OIDC }}
+            {{ if $oidc.AccessTokenEnable }}
+        {{ $proxyOrGRPC }}_set_header Authorization "Bearer $access_token";
+            {{ end }}
+        {{ end }}
```
lucacome, line 436:
```suggestion
        {{- if $s.OIDC.AccessTokenEnable }}
        {{ $proxyOrGRPC }}_set_header Authorization "Bearer $access_token";
        {{- end }}
        {{ end }}
```

@shawnhankim after more testing, we need to make sure that OIDC is enabled at all before writing the directive in the location block. The `$l.OIDC` is true only if there's a policy configured, so it's safe to move it inside the block. Sorry for the back and forth, but after this change and a rebase with `main`, we're good to merge 🚀

lucacome (changes requested):

internal/configs/version2/nginx-plus.virtualserver.tmpl
shawnhankim: Thanks @lucacome for your careful review and time. I am happy to change it. 😸 I have changed, rebased and manually tested it.

lucacome (approved):

Issue #3455, opened 2023-01-19
Disabling Leader Election in helm chart


trevor-jack:
**Describe the bug**
There is no way to disable leader election and report-ingress-status in conjunction when deploying nginx to kubernetes using the provided helm templates. Also related to this, currently the "controller-leader-election-configmap.yaml" will be deployed regardless of the controller.reportIngressStatus.enableLeaderElection setting.

**To Reproduce**
1. Set controller.reportIngressStatus.enable: false and controller.reportIngressStatus.enableLeaderElection: false in values.yaml
2. Running a helm template will show that enable-leader-election is not being explicitly set. The default value for this is true, so nginx will still attempt a leader election + attempt to create/update configmap resources.
3. When installed using helm the deployment is successful and nginx seems to run normally, but constantly throws errors
```
error initially creating leader election record: configmaps is forbidden: User "system:serviceaccount:<>" cannot create resource "configmaps" in API group "" in the namespace "<>"
```
This is because the RBAC permissions required are being removed when controller.reportIngressStatus.enableLeaderElection is false. This specific error is also present because leader-election-lock-name is not being explicitly set in this scenario.

**Expected behavior**
The nginx deployment should pass -enable-leader-election=false as an argument to in the container spec when controller.reportIngressStatus.enableLeaderElection=false. This setting is not a dependency of reportIngressStatus.

**Your environment**
* Version of the Ingress Controller - 3.0.0
* Version of Kubernetes 1.21
* Kubernetes platform GCP

shaun-nx:
Hi Trevor. You confirm the version of the Ingress Controller?
We currently don't have a version 1.23.3

brianehlert:
To clarify what Shaun stated.
The project where you posted this is: https://github.com/nginxinc/kubernetes-ingress
And this project has never released a version 1.23, so we can only assume your issue might be related to the community ingress controller project which is here: https://github.com/kubernetes/ingress-nginx/
Thanks.

brianehlert:
Looking at the other project I am not seeing a 1.23 as an Ingress version.  So maybe I am simply a bit confused.

trevor-jack:
Apologies for the confusion. In relation to this project the release is v3.0.0.
The resulting version of nginx deployed in this release is 1.23.3

brianehlert:
Thanks for the issue @trevor-jack .  
Disabling leader election is something we don't see often (so we may have missed something) - can you help us understand why you are disabling leader election?

[...]

PR #3475, opened 2023-01-24
Move logic for `-enable-leader-election` flag in helm templates

shaun-nx:
### Proposed changes
This change ensures the `-enable-leader-election` flag it set to `false` when `controller.reportIngressStatus.enable` it set to false.

Closes [#3455](https://github.com/nginxinc/kubernetes-ingress/issues/3455)

### Checklist


deployments/helm-chart/templates/controller-daemonset.yaml
```diff
@@ -191,10 +191,10 @@ spec:
           - -external-service={{ .Values.controller.reportIngressStatus.externalService }}
 {{- else if and (.Values.controller.service.create) (eq .Values.controller.service.type "LoadBalancer") }}
           - -external-service={{ include "nginx-ingress.serviceName" . }}
+{{- end }}
 {{- end }}
           - -enable-leader-election={{ .Values.controller.reportIngressStatus.enableLeaderElection }}
           - -leader-election-lock-name={{ include "nginx-ingress.leaderElectionName" . }}
```
ciarams87, line 197: This should be under an if statement here as well - this should only be set if `enableLeaderElection` is set (not a blocker, but neater).

ciarams87 (changes requested):
Need to modify `deployments/helm-chart/templates/controller-leader-election-configmap.yaml` as well so that file is only created when the leader election is set to true.

e.g. 
```
{{- if .Values.controller.reportIngressStatus.enableLeaderElection }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ include "nginx-ingress.leaderElectionName" . }}
  labels:
{{- if .Values.controller.reportIngressStatus.annotations }}
  annotations:
{{ toYaml .Values.controller.reportIngressStatus.annotations | indent 4 }}
{{- end }}
{{- end }}
```

vepatel (approved):

PR #3482, opened 2023-01-25
Use new NSM Spiffe and Cert rotation library

avahahn:
Signed-off-by: Ava Hahn <a.hahn@f5.com>

# NOTE
This PR should be merged only after the following:
- NGINX Service Mesh 1.7 is released
- go.mod is updated to point to the release version of the relevant NSM libraries.

### Proposed changes

### Checklist


internal/k8s/controller.go
```diff
@@ -645,10 +649,27 @@ func (lbc *LoadBalancerController) Run() {
...
-		err := lbc.spiffeCertFetcher.Start(lbc.ctx, lbc.addInternalRouteServer)
+		_, _, err := lbc.spiffeCertFetcher.Start(lbc.ctx)
+		lbc.addInternalRouteServer()
 		if err != nil {
 			glog.Fatal(err)
 		}
+
+		// wait for initial bundle
+		cert := <-lbc.spiffeCertFetcher.CertCh
+		lbc.syncSVIDRotation(cert)
```
sjberman, line 660: The previous code in spiffe.go that waited for this had a timeout involved so we didn't hang forever. We should do the same here.

internal/k8s/controller.go
```diff
@@ -246,11 +247,14 @@ func NewLoadBalancerController(input NewLoadBalancerControllerInput) *LoadBalanc
 
 	lbc.syncQueue = newTaskQueue(lbc.sync)
 	if input.SpireAgentAddress != "" {
-		var err error
-		lbc.spiffeCertFetcher, err = NewSpiffeCertFetcher(lbc.syncSVIDRotation, input.SpireAgentAddress)
+		client, err := workloadapi.New(lbc.ctx, workloadapi.WithAddr("unix://"+input.SpireAgentAddress))
```
sjberman, line 250: What if our library just did this initialization in the constructor of the CertFetcher?

sjberman: Then we could likely remove NIC's dependency on the spiffe library altogether.

avahahn: Unfortunately the cert synchronization callback still requires workloadapi for the argument type. Would you like me to typedef that in our library? then we can remove it.

internal/k8s/controller.go
```diff
@@ -245,11 +246,14 @@ func NewLoadBalancerController(input NewLoadBalancerControllerInput) *LoadBalanc
...
 	if input.SpireAgentAddress != "" {
-		var err error
-		lbc.spiffeCertFetcher, err = NewSpiffeCertFetcher(lbc.syncSVIDRotation, input.SpireAgentAddress)
+		lbc.spiffeCertFetcher, err = spiffe.NewX509CertFetcherFromAddr(
+			lbc.ctx,
+			input.SpireAgentAddress,
+		)
 		if err != nil {
-			glog.Fatalf("failed to create Spiffe Controller: %v", err)
+			glog.Fatalf("failed to initialize certfetcher: %v", err)
```
sjberman, line 256:
```suggestion
			glog.Fatalf("failed to initialize spiffe certfetcher: %v", err)
```

internal/k8s/controller.go
```diff
@@ -645,10 +649,33 @@ func (lbc *LoadBalancerController) Run() {
...
 		}
+
+		// wait for initial bundle
+		timeoutch := make(chan bool, 1)
+		go func() { time.Sleep(time.Second * 30); timeoutch <- true }()
+		select {
+		case cert := <-lbc.spiffeCertFetcher.CertCh:
+			lbc.syncSVIDRotation(cert)
+		case <-timeoutch:
+			glog.Fatal("Failed to download initial trust bundle")
```
sjberman, line 665:
```suggestion
			glog.Fatal("Failed to download initial spiffe trust bundle")
```

internal/k8s/controller.go
sjberman: No that's okay, I didn't realize it was used elsewhere.

sjberman (approved):

jasonwilliams14 (approved):

lucacome (approved):

PR #3491, opened 2023-01-26
Egress via Ingress VirtualServer Resource

chase-kiefer:
### Proposed changes
NSM would like the ability to egress traffic through a KIC VirtualServer. This PR adds the functionality and templating necessary to configure a VS resource for NSM egress. It diverges from the pattern implemented for ingress by using a CRD field instead of an annotation.

- added internalRoute field to the virtualserver CRD
- updated templates for internal routes in virtualserver for n+ and oss
- added unit test to validate virtual server internal routes
- updated virtualServerConfigurator type to have an enableInternalRoutes boolean 
- updated virtualserver configuration items to include internRoute docs

### Checklist


chase-kiefer:
@nginxinc/service-mesh Looking for a review from someone on our team also. I can't add you as reviewers so just adding a comment tagging our team.

internal/configs/annotations.go
```diff
@@ -388,6 +388,20 @@ func parseAnnotations(ingEx *IngressEx, baseCfgParams *ConfigParams, isPlus bool
 	return cfgParams
 }
 
+func parseVirtualServerAnnotations(virtualServerEx *VirtualServerEx, enableInternalRoutes bool) bool {
```
sjberman, line 391: Usually when a function returns a boolean, the name of the function should be indicative of the boolean value that is being returned. Something like "isInternalRouteServer" or along those lines.

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -68,8 +68,18 @@ match {{ $m.Name }} {
...
 server {
+
+    {{ if $.InternalRouteServer }}
+    listen 443 ssl;
+    {{if not $s.DisableIPV6}}listen [::]:443 ssl;{{end}}
+    ssl_certificate /etc/nginx/secrets/spiffe_cert.pem;
+    ssl_certificate_key /etc/nginx/secrets/spiffe_key.pem;
+    {{end}}
+
+    {{ if not $.InternalRouteServer}}
```
sjberman, line 79: could just do an "else" here instead of a new if statement.

internal/configs/version2/nginx.virtualserver.tmpl
```diff
@@ -40,8 +40,18 @@ limit_req_zone {{ $z.Key }} zone={{ $z.ZoneName }}:{{ $z.ZoneSize }} rate={{ $z.
...
 server {
+
+    {{ if $.InternalRouteServer }}
+    listen 443 ssl;
+    {{if not $s.DisableIPV6}}listen [::]:443 ssl;{{end}}
+    ssl_certificate /etc/nginx/secrets/spiffe_cert.pem;
+    ssl_certificate_key /etc/nginx/secrets/spiffe_key.pem;
+    {{end}}
+
+    {{ if not $.InternalRouteServer}}
```
sjberman, line 51: Ditto

jbyers19 (approved):
LGTM

internal/configs/version2/nginx-plus.virtualserver.tmpl
chase-kiefer: Is there a preferred pattern for templating logic? If not the `if not` block seems more clear.

sjberman: I would just write it based on the usual if/else standards.

if internalRouteServer is set:
...
otherwise (else) it's not set:
...

chase-kiefer: Updated.

sjberman (approved):

internal/configs/version2/nginx.virtualserver.tmpl
```diff
@@ -310,8 +319,7 @@ server {
             {{ if $l.ProxyBufferSize }}
         {{ $proxyOrGRPC }}_buffer_size {{ $l.ProxyBufferSize }};
             {{ end }}
-
-            {{ if not $l.GRPCPass }}
+        {{ if not $l.GRPCPass }}
```
ciarams87, line 322: Minor: Revert indentation change
```suggestion
            {{ if not $l.GRPCPass }}
```

internal/configs/version2/nginx.virtualserver.tmpl
```diff
@@ -40,8 +40,17 @@ limit_req_zone {{ $z.Key }} zone={{ $z.ZoneName }}:{{ $z.ZoneSize }} rate={{ $z.
...
+    {{ if $.InternalRouteServer }}
+    listen 443 ssl;
+    {{if not $s.DisableIPV6}}listen [::]:443 ssl;{{end}}
+    ssl_certificate /etc/nginx/secrets/spiffe_cert.pem;
+    ssl_certificate_key /etc/nginx/secrets/spiffe_key.pem;
+
+    {{ else }}
     listen 80{{ if $s.ProxyProtocol }} proxy_protocol{{ end }};
     {{ if not $s.DisableIPV6 }}listen [::]:80{{ if $s.ProxyProtocol }} proxy_protocol{{ end }};{{ end }}
+	{{ end }}
```
ciarams87, line 53: Nit: Indentation should be inline with the rest of the block
```suggestion
    {{ end }}
```

internal/configs/virtualserver_test.go
```diff
@@ -1275,6 +1275,119 @@ func TestGenerateVirtualServerConfigWithSpiffeCerts(t *testing.T) {
...
 }
 
+func TestGenerateVirtualServerConfigWithInternalRoutes(t *testing.T) {
+	t.Parallel()
+	virtualServerEx := VirtualServerEx{
+		VirtualServer: &conf_v1.VirtualServer{
+			ObjectMeta: meta_v1.ObjectMeta{
+				Name:        "cafe",
+				Namespace:   "default",
+				Annotations: map[string]string{"nsm.nginx.com/internal-route": "true"},
```
brianehlert, line 1285: @ciarams87 Do we want an annotation on a CRD resource?
Or do we want a property?
I realize an annotation is probably faster to implement, since Ingress uses an annotation.
But annotation is an anti-pattern to VS, VSR, TS, etc.  Isn't it?

ciarams87: Sorry, I didn't notice this comment until now! 

Using an annotation to enable a feature is generally an anti-pattern for our CRDs, however it is for an internal only use case with nothing customer facing (I am making an assumption here, so please correct me if I am wrong!) so in this case I think an annotation is OK. We may not want to modify the CRD for an internal only flag - but it's not something we have encountered before.

ciarams87: Oh, ok I didn't realise this - In that case, yes this is an anti-pattern, and we should extend the VS CRD instead to add this as a field/ property to the definition. I'm happy to help if you need anything!

deployments/common/crds/k8s.nginx.org_virtualservers.yaml
```diff
@@ -93,6 +93,8 @@ spec:
                   type: string
                 ingressClassName:
                   type: string
+                internalRoute:
+                  type: boolean
```
sjberman, line 97: May be worth adding a description to this field for more context.

deployments/helm-chart/crds/k8s.nginx.org_virtualservers.yaml
```diff
@@ -93,6 +93,8 @@ spec:
                   type: string
                 ingressClassName:
                   type: string
+                internalRoute:
+                  type: boolean
```
sjberman, line 97: Ditto

sjberman (commented):
LGTM, be sure to update commit messages and PR description with the new approach.

internal/configs/virtualserver_test.go
chase-kiefer: I pushed an update that uses a CRD field instead of an annotation. Let me know if any other changes are necessary.

internal/configs/virtualserver_test.go
```diff
@@ -8102,7 +8216,7 @@ func TestIsTLSEnabled(t *testing.T) {
 	}
```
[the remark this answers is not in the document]
chase-kiefer: I added a test case to cover nsmEgress being set to true. You are right in that it should return false. nsmEgress and TLS Enable are conflicting configuration items so nsmEgress will need to take precedence.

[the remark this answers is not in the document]
chase-kiefer: Having a user set both shouldn't really matter. They are still either enabling the internal route for NSM or not. The underlying configuration would be invisible to the user. That being said, I can add a validation check if you believe it would be valuable to users.

ciarams87 (approved):
Thanks so much @chase-kiefer ! Great work 😄

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -68,8 +68,17 @@ match {{ $m.Name }} {
 {{ end }}
 
 server {
+
+    {{ if $.InternalRouteServer }}
+    listen 443 ssl;
+    {{if not $s.DisableIPV6}}listen [::]:443 ssl;{{end}}
+    ssl_certificate /etc/nginx/secrets/spiffe_cert.pem;
+    ssl_certificate_key /etc/nginx/secrets/spiffe_key.pem;
```
lucacome, line 76: This is a separate server block in `Ingress`...don't we need to do the same here? And anyway I don't think it should "remove" the listen on port 80?

internal/configs/virtualserver.go
```diff
@@ -2452,7 +2455,11 @@ func generateProxySSLName(svcName, ns string) string {
 	return fmt.Sprintf("%s.%s.svc", svcName, ns)
 }
 
-func isTLSEnabled(u conf_v1.Upstream, spiffeCerts bool) bool {
+func isTLSEnabled(u conf_v1.Upstream, spiffeCerts, spiffeEgress bool) bool {
```
lucacome, line 2458: Could you add some comments around this function? It's not immediately clear to me what we're checking and why if `spiffeEgress` is set TLS is disabled and I think a description will help in the future 🙂

internal/configs/version2/nginx-plus.virtualserver.tmpl
chase-kiefer: In ingress it also switches off different features. https://github.com/nginxinc/kubernetes-ingress/blob/da4566c4e3a09a6387aa339066b7ba2a70bd1d9a/internal/configs/version1/nginx-plus.ingress.tmpl#L23

Does port 80 provide functionality to the internal route that I'm switching off? If so I can adjust the logic.

internal/configs/virtualserver.go
chase-kiefer: Good catch, yeah that seems really confusing. Updated for clarity.

internal/configs/version2/nginx-plus.virtualserver.tmpl
lucacome: I guess I got confused because in ingress the same logic is for `SpiffeCerts` instead of `InternalRouteServer`...right?

chase-kiefer: Yeah, I agree that was a bit confusing. But yes it's the same logic that `SpiffeCerts` provides.

lucacome: shouldn't we be using `SpiffeCerts` then?

chase-kiefer: It's been a while so I had to do some digging. Doesn't this have to do with the way vs and ingress diverge? We follow the similar logic for ingress here: https://github.com/nginxinc/kubernetes-ingress/blob/da4566c4e3a09a6387aa339066b7ba2a70bd1d9a/internal/configs/version1/nginx.tmpl#L213

chase-kiefer: Spoke with Luca offline and provided clarification.

lucacome:
@chase-kiefer I think we need a couple of changes:
- we're missing the check for the CLI argument `enable-internal-routes`, if the flag is not set and we add `internalRoute` in the VirtualServer we should throw a warning and not write the config for it.
- we should align the logic and variables with the Ingress side to make it easier to maintain the code. Right now in the template we have `InternalRouteServer` for the Spiffe server certs and `SpiffeCerts` for the client certs, but in Ingress we use `SpiffeCerts` for the server certs and `SpiffeClientCerts` for the client certs. I'm fine with whatever names you think make the most sense, I'd just like to have them consistent. Maybe for the sake of merging this PR, you can just use the same logic of Ingress and then refactor for both later. Up to you.

Also #3602 might remove the CLI arguments, so we might need to change the logic and look at the annotation instead...

ciarams87 (approved):

internal/configs/virtualserver.go
```diff
@@ -676,7 +685,8 @@ func (vsc *virtualServerConfigurator) GenerateVirtualServerConfig(
 			VSName:                    vsEx.VirtualServer.Name,
 			DisableIPV6:               vsc.isIPV6Disabled,
 		},
-		SpiffeCerts: vsc.spiffeCerts,
+		SpiffeCerts:       enabledInternalRoutes,
+		SpiffeClientCerts: vsc.spiffeCerts,
```
lucacome, line 689: does this need to be similar to what we have for Ingress?
```
SpiffeClientCerts: staticParams.NginxServiceMesh && !cfgParams.SpiffeServerCerts
```

chase-kiefer: Yes, that is necessary to prevent customer misconfiguration. I just pushed a fix.

chase-kiefer: Thanks for catching that!

Issue #3498, opened 2023-01-28
Non standard Ingress path requirement for pathType: ImplementationSpecific breaks k8s compatibility (and 3rd party services)


thecrazyrussian:
**Describe the bug**
Unable to get nginx ingress controller (3.0.1) to validate a k8s compatible Rancher (any version to 2.7.1) ingress that utilizes "pathType: ImplementationSpecific" and does not specify a path. This prevents easy usage of Rancher with later (>2.3.0) nginx controller.

**To Reproduce**
1. On K8s (tested on k3s) k8s v1.24
2. Deploy Nginx controller (3.0.1) via helmchart
3. Deploy Cert Manager (1.8.2)
4. Deploy Rancher (any, tested 2.6.9, 2.6.10, 2.7.1)
5. Nginx fails to validate ingress.

```
I0128 20:30:52.803167       1 event.go:285] Event(v1.ObjectReference{Kind:"Ingress", Namespace:"cattle-system", Name:"rancher", UID:"0d2ea447-9f6d-4ef4-ae0f-a207aa8e8c43", APIVersion:"networking.k8s.io/v1", ResourceVersion:"28689", FieldPath:""}): type: 'Warning' reason: 'Rejected' cattle-system/rancher was rejected: with error: spec.rules[0].http.path[0].path: Required value
```

**Expected behavior**
Nginx properly handles `ImplementationSpecific` with no path specified and defaults to '/' or provides ability to input default implementation specific path.

**Your environment**
* Version of the Ingress Controller -[v3.0.1](https://github.com/nginxinc/kubernetes-ingress/releases/tag/v3.0.1)
* Version of Kubernetes 1.24
* Kubernetes platform k3s
* Using NGINX

**Additional context**
Looks like this is related to the change in 2.3.0 https://github.com/nginxinc/kubernetes-ingress/pull/2775
The error was later qualified: https://github.com/nginxinc/kubernetes-ingress/pull/2971

brianehlert:
If you are using Rancher and the 'built-in' nginx ingress controller you are deploying the Kubernetes Community ingress controller.
Using the Helm chart of this project you will be deploying the one maintained by NGINX.
These two are unique.

This project does not support an empty path.
see the Restrictions section here: https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/basic-configuration/

PR #3507, opened 2023-01-31
Support empty path for ImplementationSpecific pathType

ciarams87:
### Proposed changes
Path is required for Exact and Prefix PathTypes, but should not be required for ImplementationSpecific PathTypes according to the [Kubernetes specs](https://pkg.go.dev/k8s.io/api@v0.26.1/networking/v1#HTTPIngressPath)

Our validation rules require a path be set for all PathTypes. This commit closes this gap - If an Ingress is configured with an ImplementationSpecific PathType and an empty Path, this will be allowed, and the Path will default to `/`. Otherwise ImplementationSpecific PathType is treated identically to the Prefix PathType.

Closes https://github.com/nginxinc/kubernetes-ingress/issues/3498

### Checklist

### DIFF
--- a/internal/k8s/validation_test.go
+++ b/internal/k8s/validation_test.go
@@ -3420,8 +3420,10 @@ func TestValidatePath(t *testing.T) {
 		"/abc}{abc",
 	}
 
+	pathType := networking.PathTypeExact
+
 	for _, path := range validPaths {
-		allErrs := validatePath(path, field.NewPath("path"))
+		allErrs := validatePath(path, &pathType, field.NewPath("path"))
 		if len(allErrs) > 0 {
 			t.Errorf("validatePath(%q) returned errors %v for valid input", path, allErrs)
 		}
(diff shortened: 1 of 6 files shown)


internal/k8s/validation_test.go
```diff
@@ -3420,8 +3420,10 @@ func TestValidatePath(t *testing.T) {
 		"/abc}{abc",
 	}
 
+	pathType := networking.PathTypeExact
+
 	for _, path := range validPaths {
-		allErrs := validatePath(path, field.NewPath("path"))
+		allErrs := validatePath(path, &pathType, field.NewPath("path"))
```
lucacome, line 3426: we don't have any tests with `Prefix`?

lucacome (approved):

jasonwilliams14 (approved):

internal/k8s/validation_test.go
ciarams87: `validatePath` only has different behaviour if the `pathType` is set `ImplementationSpecific`, running the tests with `Prefix` is identical to running with `Exact` (we don't check which it is for this validation test), so there wouldn't be anything new checked.

PR #3540, opened 2023-02-10
Decouple UBI images for Plus and NAP

lucacome:
This will decouple the UBI images so that NAP images don't depend on the Plus image. By doing this we can update the Plus image to UBI 9 and benefit from the newer image with less CVEs. It's also now possible to use the minimal image.

This also fixes CVE-2022-47629 in NAP images.

Closes https://github.com/nginxinc/kubernetes-ingress/security/code-scanning/10037


vepatel (approved):
👍🏼

jjngx (approved):

PR #3548, opened 2023-02-13
Configure IC root filesystem as read-only

### LINKED ISSUE #1677: Run Ngninx Kubernetes Ingress controller with `readOnlyRootFilesystem: true`

Users should be able to run nginx Kubernetes ingress controller with minimum permissions in the security context. Current configuration runs with `AllowPrivilegeEscalation: true` and `readOnlyRootFilesystem: false` but there should be a way to restrict the permissions to a bare minimum.

I was able to remove `AllowPrivilegeEscalation: true` by changing the hard code privileged ports (80, 443) in code & templates and building a custom docker image out of it. But for the `readOnlyRootFilesystem: false` workaround I'm trying to change `/etc/nginx` directory where all configurations are stored/created to one emptyDir volume.  

But after these changes, I'm facing sock connection failure and other issues in IC pod. 
Eg: 
`F0616 14:26:09.191702       1 manager.go:284] Could not get newest config version: could not get expected version: 0 after 4s`

Is it possible to run IC with `readOnlyRootFilesystem: true` ? or some specific steps are required to do so ?

Aha! Link: https://nginx.aha.io/features/IC-97

sigv:
### Proposed changes

The Nginx Ingress Controller has various protections against attacks, such as running the service as non-root to avoid changes to files. An additional industry best practice is having root filesystems set as read-only so that the attack surface is further reduced by limiting changes to binaries and libraries. This commit ensures such a defense in depth is enabled for the Nginx Ingress Controller.

To accomplish read-only rootfs, we are creating emptyDir Volumes for
- `/etc/nginx`
- `/var/cache/nginx`
- `/var/lib/nginx`
- `/var/log/nginx`

We populate `/etc` mount with an Init Container, as there are generic configuration files here. If the base image is updated via Helm chart, then Init Container will pull files from the new base image.

The other three empty volumes are populated during runtime, as they are not expected to have valuable files as part of base image:
- `/var/cache` mount will be used in case caching is enabled.
- `/var/lib` is used for some sockets and temporary files.
- `/var/log` is used for logging in the default configuration.

By default, the emptyDir will be persisted as long as the Pod is not disposed of. Simple Pod crashes will retain the data.

There is no maximum size limit implemented currently, but may be added at some point in time as a further protection. For improved performance, cache could be moved to ram tmpfs as well (currently backed by disk).

Resolves #1677.

### Checklist


sigv:
_I have added tests that prove my fix is effective or that my feature works_
Should I add tests that check whether the applied K8s pods have read-only filesystems?
I presume that's a no - in which case the bullet-point is not applicable.

_I have checked that all unit tests pass after adding my changes_
I have no errors, and all tests as passed/skipped/warning, when running from my Windows machine. Had to tweak the `rsplit` call [here](https://github.com/nginxinc/kubernetes-ingress/blob/e439f67a5f08f2ec7f84467c5cd92a9a19de8d6b/tests/suite/utils/resources_utils.py#L1626) which expects `"/"` as path separator.
Will try to re-test later on from a Linux machine to see how it looks there.

sigv:
Pushed a new commit which configures Nginx Plus with its own different base image in respective static YAML files.
@vepatel, sorry for wasted runner CPU time!

sigv:
Spun up an Ubuntu 22.04 (Jammy) VM in Azure, installed Docker and Minikube, and ran tests.
The test suite fails at Prometheus for some reason - will try to figure that out:

`suite/test_prometheus_metrics.py::TestPrometheusExporter::test_https_metrics[ingress_controller0-expected_metrics0] - Failed: Connection failed after several attempts`

sigv:
Has to be multiple Nginx Pods spinning up in one Deployment as a sidecar, and sharing `/var/lib/nginx/`, resulting in:
```text
[emerg] 14#14: bind() to unix:/var/lib/nginx/nginx-status.sock failed (98: Address already in use)
[emerg] 14#14: bind() to unix:/var/lib/nginx/nginx-config-version.sock failed (98: Address already in use)
[emerg] 14#14: bind() to unix:/var/lib/nginx/nginx-502-server.sock failed (98: Address already in use)
[emerg] 14#14: bind() to unix:/var/lib/nginx/nginx-418-server.sock failed (98: Address already in use)
```

sigv:
I am thinking if this is somewhat similar to [trac.nginx.org#753 (Nginx leaves UNIX domain sockets after SIGQUIT)](https://trac.nginx.org/nginx/ticket/753).
Not the easiest to troubleshoot as part of the test suite.

My main workstation is Windows-based - running Docker in WSL2, with K8s v1.25.4 on top.
It _annoyingly_ succeeds when running `suite/test_prometheus_metrics.py`, while I am seeing the failure on Ubuntu.
So I'm still trying to narrow down the root cause of "Address already in use".

sigv:
Cause appears to be left-over sockets after an unclean shutdown for `nginx/1.23.3`. These are not properly handled when `nginx` restarts using the same `/var/lib/nginx` volume. A very basic example is deletion of the active Pod.

sigv:
Not seeing a nice way how to get a clean `/var/lib/nginx` for Pod restarts.
So most likely course of action is to change the [Nginx image entrypoint](https://github.com/nginxinc/docker-nginx/blob/cc7ca22aee900542c05594d7a2b18623d84c7d12/entrypoint/docker-entrypoint.sh#L12), to check there are no leftover sockets:

```bash
find /var/lib/nginx -type s -print0 | while IFS= read -r -d '' s; do
    grep -qF " $s" /proc/net/unix || (entrypoint_log "$0: removing leftover socket: $s"; rm -v "$s")
done
```

What are the thoughts on this? Any better approaches that come to mind before I open a PR on that project?

sigv:
@shaun-nx, how should the self-assign be interpreted in this case? Will you handle the socket cleanup?

Noticed that there are new commits on `main` - rebased the WIP branch.

shaun-nx:
Hi @sigv
We're discussing internally in the team where this feature will line up with our current priorities.
I've assigned myself to it as I am driving this discussion in our team now. I'll update you as soon as I can.

brianehlert:
Sorry about any perceived delays.  The team is heads down on some existing work and we will take a more serious look as soon as we can guarantee some focus time.
Some pretty interesting things in here.  I am really grateful for your effort, and we look forward to working with you to finish this off and taking it.

sigv:
All good - thank you for the update! Hopefully all the active tasks go smoothly 🙏🏻

I will circle back around to this PR in two weeks, just to check on the availability of involved team members.
Not pursuing a change in the base image until I hear back from you as to avoid additional workload 👍🏻

sigv:
Interesting CI failure on Nginx Plus:
```text
WARNING: Ignoring https://pkgs.nginx.com/plus/R28/alpine/v3.17/main: Protocol error
ERROR: unable to select packages:
  nginx-plus (no such package):
    required by: world[nginx-plus]
  nginx-plus-module-njs (no such package):
    required by: world[nginx-plus-module-njs]
  nginx-plus-module-opentracing (no such package):
    required by: world[nginx-plus-module-opentracing]
```

Is something up with `pkgs.nginx.com` connectivity, or is there really something breaking in the PR?  👀

sigv:
@shaun-nx, I see you were experimenting inside of the repository ([`readOnlyRootFilesystem`](https://github.com/nginxinc/kubernetes-ingress/commits/readOnlyRootFilesystem)).
Any luck? Anything I can help with to get this moving again?

shaun-nx:
Hi @sigv thanks for checking in, you pinged at the right time! We have been making progress with testing this internally and we have made a decision on how we want to approach the initial implementation of this feature.

Full feedback comment coming soon today, just need time to write it all up!

deployments/daemon-set/nginx-ingress.yaml
```diff
@@ -48,12 +48,22 @@ spec:
          #  memory: "1Gi"
         securityContext:
           allowPrivilegeEscalation: true
+          readOnlyRootFilesystem: true
```
shaun-nx, line 51:
```suggestion
          readOnlyRootFilesystem: false
```
Users will set this back to `true` when they uncomment the volumes below.
This same update can be made to `deployments/deployment/nginx-ingress.yaml`, `deployments/deployment/nginx-plus-ingress.yaml` and `deployments/deployment/nginx-plus-ingress.yaml`

deployments/daemon-set/nginx-ingress.yaml
```diff
@@ -48,12 +48,22 @@ spec:
...
             - NET_BIND_SERVICE
+        volumeMounts:
+        - mountPath: /etc/nginx
+          name: nginx-etc
+        - mountPath: /var/cache/nginx
+          name: nginx-cache
+        - mountPath: /var/lib/nginx
+          name: nginx-lib
+        - mountPath: /var/log/nginx
+          name: nginx-log
```
shaun-nx, line 66:
```suggestion
#        volumeMounts:
#        - mountPath: /etc/nginx
#          name: nginx-etc
#        - mountPath: /var/cache/nginx
#          name: nginx-cache
#        - mountPath: /var/lib/nginx
#          name: nginx-lib
#        - mountPath: /var/log/nginx
#          name: nginx-log
```
Volumes will be in the manifest, but commented out to allow users to add them when using `readOnlyRootFilesystem: true`

deployments/daemon-set/nginx-ingress.yaml
```diff
@@ -72,3 +82,30 @@ spec:
...
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          runAsUser: 101 #nginx
+          runAsNonRoot: true
+          capabilities:
+            drop:
+            - ALL
+        volumeMounts:
+        - mountPath: /mnt/etc
+          name: nginx-etc
```
shaun-nx, line 111:
```suggestion
#      securityContext:
#        fsGroup: 101 #nginx
#      volumes:
#      - name: nginx-etc
#        emptyDir: {}
#      - name: nginx-cache
#        emptyDir: {}
#      - name: nginx-lib
#        emptyDir: {}
#      - name: nginx-log
#        emptyDir: {}
#      initContainers:
#      - image: nginx/nginx-ingress:3.0.2
#        imagePullPolicy: IfNotPresent
#        name: init-nginx-ingress
#        command: ['cp', '-vdR', '/etc/nginx/.', '/mnt/etc']
#        securityContext:
#          allowPrivilegeEscalation: false
#          readOnlyRootFilesystem: true
#          runAsUser: 101 #nginx
#          runAsNonRoot: true
#          capabilities:
#            drop:
#            - ALL
#        volumeMounts:
#        - mountPath: /mnt/etc
#          name: nginx-etc
```
The initContainer will be in the manifest, but commented out to allow users to add them when using `readOnlyRootFilesystem: true`

deployments/helm-chart/templates/controller-daemonset.yaml
```diff
@@ -111,16 +111,23 @@ spec:
 {{- end }}
         securityContext:
           allowPrivilegeEscalation: true
+          readOnlyRootFilesystem: true
```
shaun-nx, line 114:
```suggestion
          readOnlyRootFilesystem: {{ .Values.controller.readOnlyRootFilesystem }}
```
This will require adding `readOnlyRootFilesystem: false` to the `controller` section in `deployments/helm-char/values.yaml`

These same changes can be applied to `controller-deployment.yaml`

deployments/helm-chart/templates/controller-daemonset.yaml
```diff
@@ -111,16 +111,23 @@ spec:
...
         volumeMounts:
-{{- end }}
+        - mountPath: /etc/nginx
+          name: nginx-etc
+        - mountPath: /var/cache/nginx
+          name: nginx-cache
+        - mountPath: /var/lib/nginx
+          name: nginx-lib
+        - mountPath: /var/log/nginx
+          name: nginx-log
```
shaun-nx, line 130:
```suggestion
{{- if .Values.controller.readOnlyRootFilesystem }}
        - mountPath: /etc/nginx
          name: nginx-etc
        - mountPath: /var/cache/nginx
          name: nginx-cache
        - mountPath: /var/lib/nginx
          name: nginx-lib
        - mountPath: /var/log/nginx
          name: nginx-log
{{- end }}
```

deployments/helm-chart/templates/controller-daemonset.yaml
```diff
@@ -236,8 +243,35 @@ spec:
...
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          runAsUser: 101 #nginx
+          runAsNonRoot: true
+          capabilities:
+            drop:
+            - ALL
+        volumeMounts:
+        - mountPath: /mnt/etc
+          name: nginx-etc
```
shaun-nx, line 272:
```suggestion
{{- if .Values.controller.readOnlyRootFilesystem }}
      securityContext:
        fsGroup: 101 #nginx
      volumes:
      - name: nginx-etc
        emptyDir: {}
      - name: nginx-cache
        emptyDir: {}
      - name: nginx-lib
        emptyDir: {}
      - name: nginx-log
        emptyDir: {}
      initContainers:
      - name: init-{{ include "nginx-ingress.name" . }}
        image: {{ include "nginx-ingress.image" . }}
        imagePullPolicy: "{{ .Values.controller.image.pullPolicy }}"
        command: ['cp', '-vdR', '/etc/nginx/.', '/mnt/etc']
        securityContext:
          allowPrivilegeEscalation: false
          readOnlyRootFilesystem: true
          runAsUser: 101 #nginx
          runAsNonRoot: true
          capabilities:
            drop:
            - ALL
        volumeMounts:
        - mountPath: /mnt/etc
          name: nginx-etc
{{- end }}
```

sigv:
Applied the comments, and they should be also applied to all related variants. Hope I didn't mess it up during the rebase/conflict resolution.

@shaun-nx, did you have a chance to investigate the socket cleanup topic that I previously experienced in testing?
Does the upstream project's nginx binary or IC process clean up sockets during startup?

sigv:
The default to not enabling this makes sense, and I fully understand the reasoning behind such a workflow!

Thank you for the suggestions! I additionally tweaked the Helm chart to have modified `if or` calls, so that `volumeMounts` and `initContainers` are not stray.

Should I flag the PR as open for review?

sigv:
Oh my. I did not intend to remove reviewers here. Just wanted to flag this as in need of re-review after latest review was taken into consideration as per my comment.

deployments/deployment/nginx-plus-ingress.yaml
```diff
@@ -82,3 +92,30 @@ spec:
          #- -enable-prometheus-metrics
          #- -enable-service-insight
          #- -global-configuration=$(POD_NAMESPACE)/nginx-configuration
+#      securityContext:
+#        fsGroup: 101 #nginx
```
shaun-nx, line 96: I realise the new `securityContext` block that was added in https://github.com/nginxinc/kubernetes-ingress/pull/3573 is conflicting with this one.  We can remove this `securityContext` and move the `fsGroup: 101 #nginx` line into the new on Line 23 of this file.

https://github.com/nginxinc/kubernetes-ingress/blob/09601f457eda4f5126f1c63baabe314b3e633b98/deployments/deployment/nginx-plus-ingress.yaml#L23-L26

sigv: Indeed - applied to the manifests as well as Helm chart!

sigv:
> > did you have a chance to investigate the socket cleanup topic that I previously experienced in testing?
> 
> Yes, thanks for reminding me! We found that the test `suite/test_prometheus_metrics.py::TestPrometheusExporter::test_https_metrics` was writing TLS keys to a temp directory. This was causing the Ingress Controller to error as it couldn't write to that directory. We have a PR up to update the directory that is written to. Should have that merged today: #3615

Looking forward to the PR being merged! 🤞🏻

> Outside of this change, this was causing an un-graceful shutdown of the NGINX which left those sockets open. This is a general issue and one that is on our roadmap to address.

Is the plan to deliver this `readOnly` opt-in change, while acknowledging the potential impact (until the backlog item is taken on)? Should I update some documentation to reflect this risk?

Or is this being targeted for solution in the upcoming release already?

shaun-nx:
> Is the plan to deliver this readOnly opt-in change, while acknowledging the potential impact (until the backlog item is taken on)? In such case, should I update some documentation to reflect this risk?

The plan right now is to highlight this as part of the release documentation for `v3.1.0` we normally do this as part of the release processes so we don't need to add it as part of this PR.

deployments/helm-chart/templates/controller-daemonset.yaml
```diff
@@ -238,8 +252,39 @@ spec:
...
+      volumes:
+      - name: nginx-etc
+        emptyDir: {}
+      - name: nginx-cache
+        emptyDir: {}
+      - name: nginx-lib
+        emptyDir: {}
+      - name: nginx-log
+        emptyDir: {}
+{{- end }}
```
shaun-nx, line 265: Helm is giving errors when both `controller.readOnlyRootFilesystem: true` and `nginxServiceMesh.enable: true` are set.
On L66 we add a `volumes` block if we set either `controller.volumes` or `nginxServiceMesh.enable`

https://github.com/nginxinc/kubernetes-ingress/pull/3548/files#diff-2eccfbc1886b3803b5035aad15dc3f9f38ecfd0b5c5a540408c828f32d6ab3d3R66-R74

This ends up conflicting with the `volumes` block here on L256
We can fix this by removing the `volumes` block on L256 and then moving the rest of this block below L74 in the mentioned link.

We can also add `.Values.controller.readOnlyRootFilesystem` to the condition block that adds the `volumes` block.

So all together it should look like this:

```
{{- if or .Values.controller.volumes .Values.nginxServiceMesh.enable .Values.controller.readOnlyRootFilesystem}}
      volumes:
{{- end }}
{{- if .Values.nginxServiceMesh.enable }}
      - hostPath:
          path: /run/spire/sockets
          type: DirectoryOrCreate
        name: spire-agent-socket
{{- end }}
{{- if .Values.controller.readOnlyRootFilesystem }}
      volumes:
      - name: nginx-etc
        emptyDir: {}
      - name: nginx-cache
        emptyDir: {}
      - name: nginx-lib
        emptyDir: {}
      - name: nginx-log
        emptyDir: {}
{{- end }}
```

shaun-nx: Same change can be applied to `controller-deployment.yaml`

sigv: Yeah. I applied this to the `volumeMounts` but missed it for the `volumes` themselves.

sigv: Latest commit on the PR resolves this. Thank you!

sigv:
Comments are applied, and I also moved the `volumes` block up in the non-Helm YAML files to align better with the layout that Helm has.

shaun-nx:
Thanks @sigv for the updates! I'll get back to you once I've re-tested the changes. So far everything is looking good and we should close to merging this PR. I'll get more eyes on the PR to make sure nothing has been overlooked. Thanks again for all this work!

sigv:
Heads up, [GitHub Code Scanning / Trivy](https://github.com/nginxinc/kubernetes-ingress/pull/3548/checks?check_run_id=11876981529) reports _gnutls: timing side-channel in the TLS RSA key exchange code_ ([CVE-2023-0361](https://avd.aquasec.com/nvd/cve-2023-0361)).

Prior to nginxinc/kubernetes-ingress@d3e7725469bba179f88014c6cae4205416f75b67, this was being mitigated.
@shaun-nx, will you coordinate the change?
```bash
	# temp fix for CVE-2023-0361
	&& apt-get install -y libgnutls30
```

shaun-nx:
Hi @sigv just wanted to keep you posted. Sorry that it has been taking a bit of time to do re-testing on this PR. Some additional internal work has required my focus. I should have an update for you tomorrow. Thanks for your patience 🙏

sigv:
Thanks for the update!

I'm just keeping the PR rebased and without merge conflicts in the meantime 🙂

deployments/daemon-set/nginx-plus-ingress.yaml
```diff
@@ -54,10 +64,20 @@ spec:
          #  memory: "1Gi"
         securityContext:
           allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: false
```
vepatel, line 67: same here, set to `true` and comment

sigv: As above, applied in [latest change](https://github.com/nginxinc/kubernetes-ingress/compare/bb17a8efaf8b2ed61daf2f9b44bd7f6839c74f38..a10301e4641bc272ec9d7af7978457340967b7eb). 🙏🏻

sigv:
And I feel stupid about my 're-review' button dropping others again. Sorry! 😐

shaun-nx (approved):
Looks all good to me. Thanks again for this work @sigv 🚀

vepatel (approved):
thanks for the patience during review process @sigv, we tested upgrades from previous version with and without readOnlyFileSystem and all looks okay to me 👍🏼

ciarams87 (approved):

sigv:
@vepatel, it's a concern that crept up during my local testing. It should be all safe, but as there are more eyes on it now, wanted to highlight it.

Upgrade and redeployment process should be good. The potential issue I observed is when Pod dies abruptly leaving a socket behind. That seemed similar to [trac.nginx.org#753 (Nginx leaves UNIX domain sockets after SIGQUIT)](https://trac.nginx.org/nginx/ticket/753).

[KIC-813](https://nginxsoftware.atlassian.net/browse/KIC-813) issue title sounds promising.

Looking back at it, I did not test this on older versions.. And as this PR does not impact relevant behavior, it should be reproducible on versions before this PR, if the Pod is abused there. The socket file would simply be on root filesystem, which I believe is also persisted across restarts.

So the concern is most likely moot.

vepatel:
@sigv we're targeting `3.1.0` release by the end of this month!

brianehlert:
We are planning on getting better with the Milestones.   (I must have not pushed an update to that milestone, whoops)

Issue #3544, opened 2023-02-11
NIC should be able to run with the "restricted" POD security level


hafe:
TBD

See discussion in slack

sigv:
Please leave default HTTP/S ports as defaults.

Most users of the Ingress Controller will want ports 80 and 443 used. This would particularly impact those who expose NIC to the Internet, as port change would result in users having to type `https://nginx-ingress.com:8443/`

----

[Docker v20.10.0](https://docs.docker.com/engine/release-notes/20.10/#20100) (released 2020-12-08) can be seen as supporting binding privileged ports with no capabilities, as it automatically sets `sysctl net.ipv4.ip_unprivileged_port_start=0` via https://github.com/moby/moby/pull/41030 merged as https://github.com/moby/moby/commit/888da28d42f7d0f9fa250dd8a75d51c2a6cf3098.

Similarly, [K8s docs](https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#enabling-unsafe-sysctls) state that `net.ipv4.ip_unprivileged_port_start` is considered a _safe_ sysctl since Kubernetes v1.22.
PodSpec could contain `securityContext.sysctls` with `{ "name": "net.ipv4.ip_unprivileged_port_start", "value": "0" }`

I have not yet experimented with this.. but a general solution, setting the `sysctl`, sounds more elegant:
- remove `NET_BIND_SERVICE` capability for Pod,
- change `allowPrivilegeEscalation` to `false` for Pod,
- remove the [`setcap`](https://github.com/nginxinc/kubernetes-ingress/blob/876a0f657eb781b7c3c9263d5b3da746d0214650/build/Dockerfile#L209-L210) from binary.

----

This would however leave Pods running in host network in a peculiar situation.

The mentioned Docker runtime change avoids setting the `sysctl` for host networking (_We do not set network sysctls if network namespace is host_) as that ends up potentially changing the actual host `sysctl` lowering the security expectations there.

I am currently not sure of what the most appropriate solution there would be - should we evaluate changing the `sysctl` during runtime, with network mode detection?

sigv:
@hafe, is there any aspect in which NIC does not comply with Restricted security level, other than `allowPrivilegeEscalation`?

hafe:
> @hafe, is there any aspect in which NIC does not comply with Restricted security level, other than `allowPrivilegeEscalation`?

[...]

PR #3573, opened 2023-02-21
Rework port binding logic without privileges

sigv:
### Proposed changes

This is a breaking change! To start Nginx Ingress Controller with this commit applied, you must use an updated image.

Historically, the Ingress Controller entrypoint launched with restricted capabilities. Then Nginx process escalated privileges (NET_BIND_SERVICE) to bind ports 80 and 443 as non-root user. Allowing privilege escalation is generally frowned upon in various policies.

The Nginx binary in old images was adding NET_BIND_SERVICE to its Permitted capability set and also setting the Effective bit, to enforce the Permitted capability during launch. (That's the escalation there.)

With this change, privilege escalation is no longer allowed and the NET_BIND_SERVICE capability is removed. To allow the binary to start, the capabilities are no longer being adjusted on the binary file.

This works because Kubernetes v1.22+ allows Pods to independently lower unprivileged port range to start with zero without affecting other ports (namespaced/"safe" sysctls).

OBS! An old image may be used if the binary's Effective bit is removed:
```dockerfile
FROM nginx/nginx-ingress:3.0.2
USER root
RUN setcap 'cap_net_bind_service=-e' /usr/sbin/nginx 'cap_net_bind_service=-e' /usr/sbin/nginx-debug \
    && setcap -v 'cap_net_bind_service=-e' /usr/sbin/nginx 'cap_net_bind_service=-e' /usr/sbin/nginx-debug
# 101 is nginx
USER 101
```

Improves #3544. Privilege Escalation is the initial scope of the linked issue. However, as indicated by the issue creator, there may be follow-up tasks to fully close it.

### Checklist

### DIFF
--- a/docs/content/technical-specifications.md
+++ b/docs/content/technical-specifications.md
@@ -12,7 +12,8 @@ docs: "DOCS-617"
 
 We advise users to run the most recent release of the NGINX Ingress Controller, and we issue software updates to the most recent release. We provide technical support for F5 customers who are using the most recent version of the NGINX Ingress Controller, and any version released within two years of the current release.
 
-The current release version 3.x supports `discovery.k8s.io/v1` API version of EndpointSlice, available from Kubernetes 1.21 onwards.
+The current 3.x release does not use `NET_BIND_SERVICE` capability, and instead relies on `net.ipv4.ip_unprivileged_port_start` sysctl to allow port binding. Kubernetes 1.22 or later is required for this sysctl to be [classified as safe](https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#safe-and-unsafe-sysctls).
+The 3.0.0 release supports `discovery.k8s.io/v1` API version of EndpointSlice, available from Kubernetes 1.21 onwards.
 The 2.4.2 release is compatible with the Kubernetes Ingress v1 API. Therefore Kubernetes 1.19 and later.
 The 1.12 release supports the Ingress v1beta1 API and continues to receive security fixes to support those unable to upgrade to Kubernetes 1.19 or later. The v1beta1 Ingress API was deprecated with Kubernetes release 1.19 and removed with the Kubernetes 1.22 release.
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -296,7 +296,7 @@ jobs:
                                                 {\"image\": \"debian-plus-nap\", \"marker\": \"appprotect\"}], \
                                               \"k8s\": [\"${{ needs.checks.outputs.k8s_latest }}\"]}" >> $GITHUB_OUTPUT
           else
-            echo "matrix={\"k8s\": [\"1.21.14\", \"1.22.15\", \"1.23.13\", \"1.24.7\", \"1.25.3\", \"${{ needs.checks.outputs.k8s_latest }}\"], \
+            echo "matrix={\"k8s\": [\"1.22.15\", \"1.23.13\", \"1.24.7\", \"1.25.3\", \"${{ needs.checks.outputs.k8s_latest }}\"], \
                                              \"images\": [{\"image\": \"debian\"}, {\"image\": \"debian-plus\"}]}" >> $GITHUB_OUTPUT
           fi
 
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -186,7 +186,7 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
 # RUN update-ca-trust extract
 
 
-############################################# Create common files, permissions and setcap #############################################
+############################################# Create common files and permissions #############################################
 FROM ${BUILD_OS} as common
 
 ARG BUILD_OS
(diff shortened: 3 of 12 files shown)


sigv:
I have left the NET_BIND_SERVICE capability in the Permitted set on the binary file. But I'm thinking now that there is no scenario where Kubernetes ends up using it.

Is there some strange Docker native scenario where it's helpful? Should probably remove otherwise.

sigv:
As mentioned in description, to test the chart, you would need to [build the image](https://github.com/nginxinc/kubernetes-ingress/blob/main/docs/content/installation/building-ingress-controller-image.md) similar to:
```bash
make debian-image PREFIX=myregistry.example.com/nginx-ingress TARGET=container
```

When installing the chart, `--set controller.image.repository=myregistry.example.com/nginx-ingress --set controller.image.tag=untagged-EXAMPLE`

sigv:
[Smoke Tests (debian-plus-nap, dos, 1.26.0)](https://github.com/nginxinc/kubernetes-ingress/actions/runs/4240724856/jobs/7370620669#logs) failure leads to [this `assert`](https://github.com/nginxinc/kubernetes-ingress/blob/294640d20a93f5a6b04b5251cbec1d5d126efc50/tests/suite/test_dos.py#L366-L376).

The 7th argument is checking `signature_detected`. Being an OSS customer, I am not familiar with the AppProtect/DOS mechanisms, so it's not clear to me currently why `mitigated_by_signatures` would not be detected.

```text
=================================== FAILURES ===================================
_ TestDos.test_dos_under_attack_with_learning[crd_ingress_controller_with_dos0] _
Traceback (most recent call last):
  File "/workspace/tests/suite/test_dos.py", line 366, in test_dos_under_attack_with_learning
    assert (
AssertionError: assert (True and True and True and True and True and 84.0 < 150 and False)
 +  where 84.0 = <built-in method total_seconds of datetime.timedelta object at 0x7fb841a13bd0>()
 +    where <built-in method total_seconds of datetime.timedelta object at 0x7fb841a13bd0> = (datetime.datetime(2023, 2, 22, 9, 5, 5) - datetime.datetime(2023, 2, 22, 9, 3, 41)).total_seconds
```

lucacome:
@sigv thanks for the PR! I've re-run the DoS test and it passed 🎉 

I've been doing some testing and I think we can remove `setcap` altogether, but we need to bump the minimum required versions of k8s to `1.22` in the docs, helm chart, etc. for this to work. I think it's fine, `1.21` it's pretty old at this point. @brianehlert thoughts?

sigv:
- Kubernetes v1.22 configured as lowest supported version.
- Kubernetes v1.21 removed from pipelines.
- Invocation of `setcap` removed outright.
- Added note about K8s support change in `docs/content/technical-specifications.md`.
- Branch rebased (`protobuf` update).

docs/content/technical-specifications.md
```diff
@@ -12,7 +12,8 @@ docs: "DOCS-617"
 
 We advise users to run the most recent release of the NGINX Ingress Controller, and we issue software updates to the most recent release. We provide technical support for F5 customers who are using the most recent version of the NGINX Ingress Controller, and any version released within two years of the current release.
 
-The current release version 3.x supports `discovery.k8s.io/v1` API version of EndpointSlice, available from Kubernetes 1.21 onwards.
+The 3.1.0 release removes `NET_BIND_SERVICE` from binary files' capability sets, and the manifests set `net.ipv4.ip_unprivileged_port_start` sysctl to allow port binding. This requires the sysctl to be [classified as safe](https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#safe-and-unsafe-sysctls), which is done in Kubernetes 1.22 and later.
```
sigv, line 15: Not sure here - this probably should be a major bump of `4.0.0`. Let me know if this should be tweaked.

Also, the table versions will need to be further tweaked to properly represent the release.

sigv:
Alternative approach I mentioned in the linked issue could be that IC process (entrypoint) is modified, so that it gets NET_BIND_SERVICE initially and executes Nginx (which could inherit the capability) and then IC drops that capability. In theory we don't need lowered unprivileged port range then - only change in IC to remove privilege escalation. If that's preferred I can propose the change instead.

docs/content/technical-specifications.md
lucacome: I don't think this should be a major bump (`4.x`), we're just removing a version of k8s that is EOL anyway, as you said, and using new functionalities that are in the new supported versions. I don't see it as breaking change tbh.

lucacome:
> It's kind of a risk assessment question of how to get rid of escalation. Either the current scope which enables the whole Pod to easily bind, or simply better handling of the NET_BIND_SERVICE capability.

I like the approach of enabling the pod to easily bind. Do you see any drawbacks with this that I'm not seeing?

docs/content/technical-specifications.md
sigv: With the change, it would be breaking for users who are staying on an old image tag. Not sure why they would be specifying an old image tag, but who knows. If that's an okay change, then no need for `4.x`.

docs/content/technical-specifications.md
```diff
@@ -23,6 +24,7 @@ We explicitly test the NGINX Ingress Controller (NIC) on a range of Kubernetes p
 {{% table %}}
 | NIC Version | Supported Kubernetes Version | NIC Helm Chart Version | NIC Operator Version | NGINX / NGINX Plus version |
 | --- | --- | --- | --- | --- |
+| 3.1.0 | 1.26 - 1.22 | 0.16.3 | 1.3.0 | 1.23.3 / R28 |
```
sigv, line 27: I am anticipating NIC Version change from `3.0.2` to `3.1.0` here.

I committed NIC Helm Chart Version increase from `0.16.2` to `0.16.3` as it's on the `0.x` series -- is this okay here? Or should I bump it up to `0.17.0` in this commit as well?

brianehlert: The versioning related things will all be handled by the release processes.
you don't need to update these as part of your PR.

sigv: Latest commit removes the manual addition.

lucacome (approved):

lucacome:
> In current setup, the same risk already exists. At least when taking a quick look, it seemed the nginx process does not drop NET_BIND_SERVICE (or I am just missing where it happens?).

yeah we don't drop it

sigv:
[Build Docker Plus (ubi-plus, linux/arm64, linux/amd64, linux/s390x, goreleaser)](https://github.com/nginxinc/kubernetes-ingress/actions/runs/4293983547/jobs/7485866865) failed to resolve the Nginx Plus package repository:

```text
error: cannot update repo 'nginx-plus': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried; Last error: Curl error (58): Problem with the local SSL certificate for https://pkgs.nginx.com/plus/R28/centos/9/x86_64/repodata/repomd.xml [could not load PEM client certificate, OpenSSL error error:80000002:system library::No such file or directory, (no key found, wrong pass phrase, or wrong file format?)]
```

I presume this is an intermittent issue, as nothing should have changed for this step.

Jcahilltorre (approved):

shaun-nx (approved):

PR #3586, opened 2023-02-23
chore: Add new label and metadata to pod

ciarams87:
### Proposed changes
- Add version metadata to a pod in the form of annotations
- Add new static label to pods for filtering all NIC pods in a cluster

### Checklist

### DIFF
--- a/cmd/nginx-ingress/main.go
+++ b/cmd/nginx-ingress/main.go
@@ -42,6 +42,11 @@ import (
 // Injected during build
 var version string
 
+const (
+	nginxVersionAnnotation = "app.nginx.org/version"
+	versionAnnotation      = "app.kubernetes.io/version"
+)
+
 func main() {
 	commitHash, commitTime, dirtyBuild := getBuildInfo()
 	fmt.Printf("NGINX Ingress Controller Version=%v Commit=%v Date=%v DirtyState=%v Arch=%v/%v Go=%v\n", version, commitHash, commitTime, dirtyBuild, runtime.GOOS, runtime.GOARCH, runtime.Version())
@@ -64,7 +69,9 @@ func main() {
 
 	nginxManager, useFakeNginxManager := createNginxManager(managerCollector)
 
-	getNginxVersionInfo(nginxManager)
+	nginxVersion := getNginxVersionInfo(nginxManager)
+
+	updateSelfWithVersionInfo(kubeClient, version, nginxVersion)
 
 	templateExecutor, templateExecutorV2 := createTemplateExecutors()
 
@@ -367,7 +374,7 @@ func createNginxManager(managerCollector collectors.ManagerCollector) (nginx.Man
 	return nginxManager, useFakeNginxManager
 }
 
-func getNginxVersionInfo(nginxManager nginx.Manager) {
+func getNginxVersionInfo(nginxManager nginx.Manager) string {
 	nginxVersion := nginxManager.Version()
 	isPlus := strings.Contains(nginxVersion, "plus")
 	glog.Infof("Using %s", nginxVersion)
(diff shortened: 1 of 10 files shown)


deployments/daemon-set/nginx-ingress.yaml
```diff
@@ -11,6 +11,7 @@ spec:
     metadata:
       labels:
         app: nginx-ingress
+        app.kubernetes.io/core-name: nginx-ingress
```
lucacome, line 14: I've never seen `core-name`, can't we use `name`?

cmd/nginx-ingress/main.go
```diff
@@ -42,6 +42,11 @@ import (
 // Injected during build
 var version string
 
+const (
+	nginxVersionAnnotation = "nginx-ingress-controller/nginx-version"
+	versionAnnotation      = "nginx-ingress-controller/version"
```
lucacome, line 47: I think we should use `app.kubernetes.io/version` for the version of KIC

cmd/nginx-ingress/main.go
```diff
@@ -42,6 +42,11 @@ import (
 // Injected during build
 var version string
 
+const (
+	nginxVersionAnnotation = "nginx-ingress-controller/nginx-version"
```
lucacome, line 46: Does it make sense to have something like `app.nginx.org/version`?

cmd/nginx-ingress/main.go
```diff
@@ -746,3 +754,29 @@ func processConfigMaps(kubeClient *kubernetes.Clientset, cfgParams *configs.Conf
...
+		return
+	}
+
+	// Copy pod and update the annotations.
+	newPod := pod.DeepCopy()
+	ann := newPod.ObjectMeta.Annotations
+	if ann == nil {
+		ann = make(map[string]string)
+	}
+	ann[nginxVersionAnnotation] = strings.Split(nginxVersion, ": ")[1]
```
lucacome, line 771: I think we might want to split on `/` to just get the version for example `1.23.2` instead of `nginx/1.23.2`?

lucacome (commented):
I left some comments. also we need to update the `rbac.yaml` for helm too.

deployments/daemon-set/nginx-ingress.yaml
ciarams87: We're already using `name` in the helm chart, but it's not static because we prepend the release name to it. Thinking about it though, that's not correct - the point is to identify an application, not an instance of it. I'll change that and use name instead.

cmd/nginx-ingress/main.go
ciarams87: I don't see the harm in including whether or not the NGINX installed is plus or not in the annotation - it's very clear then with a glance which one you're dealing with without having to check deployment flags, or cross-referencing which version corresponds with Plus.

jjngx (approved):

cmd/nginx-ingress/main.go
lucacome: oh yeah I wasn't saying to remove plus, the version string looks like:
`nginx version: nginx/1.23.2 (nginx-plus-r28)`
and right we're just getting everything after `:`
`nginx/1.23.2 (nginx-plus-r28)`
I thought it would look cleaner to get everything after `/`
`1.23.2 (nginx-plus-r28)`
but it's just a nitpick 🙂

lucacome (approved):

cmd/nginx-ingress/main.go
ciarams87: Ahhh, my apologies, I misinterpreted your comment completely, I should have read it properly, I thought you were saying to just keep the numbers 🤦🏼‍♀️ That makes more sense!

PR #3602, opened 2023-02-27
Updated NGINX Service Mesh references in Helm templates

jbyers19:
### Proposed changes
This change updates the `deployments/helm-chart/templates/controller-daemonset.yaml` and `deployments/helm-chart/templates/controller-deployment.yaml` templates. In the latest NSM release, the `nsm.nginx.com/enable-ingress` and `nsm.nginx.com/enable-egress` _**annotations**_ were moved to _**labels**_. This change updates the templates accordingly. It also removes most `if .Values.nginxServiceMesh.enable/enableEgress` references as these values are now handled by a mutating webhook run by NSM.

**Update:** reverted this change. ~One thing to note is that I removed the `if` for the `POD_SERVICEACCOUNT` env variable so it is now added regardless of whether NSM is enabled or not. Since this is an environment variable [used by the KIC controller](https://github.com/nginxinc/kubernetes-ingress/blob/main/internal/configs/configurator.go#L1619), I thought it would be better to always include it in the pod spec.~

### Testing: 
* Verified `helm lint` passed on the updated templates:
    * `helm lint . -f values-nsm.yaml --set controller.kind=daemonset`
    *  `helm lint . -f values-nsm.yaml --set controller.kind=deployment`
* Manually verified rendered manifest matched what was expected:
    * `helm install kic -f values-nsm.yaml --set controller.kind=daemonset . --dry-run`
    * `helm install kic -f values-nsm.yaml --set controller.kind=deployment . --dry-run`
* Verified linting and unit tests passed.

### Checklist

### DIFF
--- a/deployments/helm-chart/templates/controller-daemonset.yaml
+++ b/deployments/helm-chart/templates/controller-daemonset.yaml
@@ -18,22 +18,19 @@ spec:
       labels:
         {{- include "nginx-ingress.selectorLabels" . | nindent 8 }}
 {{- if .Values.nginxServiceMesh.enable }}
+        nsm.nginx.com/enable-ingress: "true"
+        nsm.nginx.com/enable-egress: "{{ .Values.nginxServiceMesh.enableEgress }}"
         nsm.nginx.com/daemonset: {{ include "nginx-ingress.controller.fullname" . }}
-        spiffe.io/spiffeid: "true"
 {{- end }}
 {{- if .Values.controller.pod.extraLabels }}
 {{ toYaml .Values.controller.pod.extraLabels | indent 8 }}
 {{- end }}
-{{- if or .Values.prometheus.create (or .Values.controller.pod.annotations .Values.nginxServiceMesh.enable) }}
+{{- if or .Values.prometheus.create .Values.controller.pod.annotations }}
       annotations:
 {{- if .Values.prometheus.create }}
         prometheus.io/scrape: "true"
         prometheus.io/port: "{{ .Values.prometheus.port }}"
         prometheus.io/scheme: "{{ .Values.prometheus.scheme }}"
-{{- end }}
-{{- if .Values.nginxServiceMesh.enable }}
-        nsm.nginx.com/enable-ingress: "true"
-        nsm.nginx.com/enable-egress: "{{ .Values.nginxServiceMesh.enableEgress }}"
 {{- end }}
 {{- if .Values.controller.pod.annotations }}
 {{ toYaml .Values.controller.pod.annotations | indent 8 }}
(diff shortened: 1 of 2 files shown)


jbyers19:
@lucacome, no. The internal functionality is still the same. The webhook just adds all these values to the pod spec when it is deployed. So if a user is deploying the ingress from a plain manifest instead of helm they do not have to manually add all these fields. They just need to add the `nsm.nginx.com/enable-ingress/egress` labels and the webhook will deal with the rest.

lucacome:
so we're still giving users the option to use those flags as well?

lucacome:
I was hoping that we could remove them for good and have fewer moving pieces 😅

lucacome:
@jbyers19 I was actually thinking that IMO it would make more sense for KIC to enable the internal routes, etc. based on the labels, rather than having NSM add CLI arguments to KIC based on KIC's labels. What do you think?

jbyers19:
@lucacome, maybe. We already did the work and added the webhook to do this in the last NSM release though. I think it would only work for args though. We would still need the labels, volumes, etc. in the pod spec.

deployments/helm-chart/templates/controller-daemonset.yaml
```diff
@@ -136,14 +123,12 @@ spec:
...
             fieldRef:
               fieldPath: metadata.name
-{{- if .Values.controller.env }}
-{{ toYaml .Values.controller.env | indent 8 }}
-{{- end }}
-{{- if .Values.nginxServiceMesh.enable }}
         - name: POD_SERVICEACCOUNT
           valueFrom:
             fieldRef:
               fieldPath: spec.serviceAccountName
```
sjberman, line 129: This is actually something that we missed in our injection logic. This value is used by us for the SPIFFE certs. I don't think KIC uses it anywhere else.

sjberman: Probably doesn't hurt to have it as part of the official manifest (like you've done here), since it's just an env var. But we likely need to update our controller to inject this if it's not present.

jbyers19: @sjberman, I created internal issue NSM-3677 to track this issue.

sjberman: I wonder if we just keep this var gated behind the NSM condition, since only we use it.

jbyers19: Reverted this change as suggested.

lucacome (approved):

vepatel:
@jbyers19 looks like pr has fair bit of merge commits, can you please rebase with main and only keep the relevant bits

ciarams87 (approved):

PR #3606, opened 2023-03-01
Fix Helm Chart labels and templates. Move version update to labels

lucacome:
`app.kubernetes.io/version` and `app.nginx.org/version` should be labels and not annotations.

This also fixes name generation for the Helm Chart and adds selector labels.

Removed unnecessary names.


jasonwilliams14 (approved):
Tested changes successfully.

shaun-nx (approved):

jasonwilliams14 (approved):
Tested and looks good.

haywoodsh (approved):

ciarams87 (changes requested):
When I try this locally, the version labels don't update, and I get the following in the logs:

> Error updating pod with labels: Pod "my-chart-nginx-ingress-controller-75c4fc6d5d-fxl7l" is invalid: [metadata.labels: Invalid value: "\"3.0.0-SNAPSHOT-dda91a4\"": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'), metadata.labels: Invalid value: "\"1.23.2 (nginx-plus-r28)\"": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')]

lucacome:
Thanks for catching this @ciarams87 ! I was doing some tests with quoting the versions and apparently, it was double quoting the string now. I've fixed it.

ciarams87 (approved):
LGTM now, thanks!

PR #3615, opened 2023-03-02
Prometheus tls path

shaun-nx:
### Proposed changes
Update the path that prometheus TLS secrets are written to.

### Checklist


internal/metrics/listener.go
```diff
@@ -73,8 +74,13 @@ func runServer(port string, registry prometheus.Gatherer, prometheusSecret *api_
 	}
 }
 
-func writeTempFile(data []byte, name string) (*os.File, error) {
-	f, err := os.CreateTemp("", name)
+func writeToSecretsPath(data []byte, name string) (*os.File, error) {
+	_, err := os.Stat(config.DefaultSecretPath)
```
jjngx, line 78: What if the `err` is caused by sth else than "not existing path"? How could we check a specific err type?

jjngx (commented):
left comment

internal/metrics/listener.go
```diff
@@ -76,7 +76,7 @@ func runServer(port string, registry prometheus.Gatherer, prometheusSecret *api_
 
 func writeToSecretsPath(data []byte, name string) (*os.File, error) {
 	_, err := os.Stat(config.DefaultSecretPath)
-	if err != nil {
+	if os.IsNotExist(err) {
```
jjngx, line 79: https://pkg.go.dev/os#IsNotExist

```
This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).
```

ciarams87 (approved):

PR #3616, opened 2023-03-02
Remove libcap

lucacome:
`libcap` is no longer needed so we don't need to install it anymore.

This also removes temporary fixes for CVEs.


sigv:
Thanks for noticing this as an outcome of my PR! 🙏🏻

lucacome:
Thank you @sigv for making this possible 😃

haywoodsh (approved):

shaun-nx (approved):

PR #3632, opened 2023-03-10
Update nginx.org/ca secret type & crl field to IngressMTLS to support CRL

shaun-nx:
### Proposed changes
This change updates both the `nginx.org/ca` secret type and the `ingressMTLS` policy type to provide users a means to configure a Certificate Revocation List.

The CRL can be configured in one of two ways.

1. Using a Base64 encoded CRL with the `nginx.org/ca` secret type using the `ca.crl`
```yaml
kind: Secret
metadata:
  name: ingress-mtls-secret
apiVersion: v1
type: nginx.org/ca
data:
  ca.crt: <base64encoded-certificate>
  ca.crl: <base64encoded-crl>
```

2. Using the `crl` field in the `ingressMTLS` policy spec providing the name of the CRL. This option assumes that the crl `webapp.crl` is provided by the users to `/etc/nginx/secrets`
```yaml
apiVersion: k8s.nginx.org/v1
kind: Policy
metadata:
  name: ingress-mtls-policy
spec:
  ingressMTLS:
    clientCertSecret: ingress-mtls-secret
    crlFileName: webapp.crl
    verifyClient: "on"
    verifyDepth: 1
```

### Checklist

### DIFF
--- a/docs/content/configuration/policy-resource.md
+++ b/docs/content/configuration/policy-resource.md
@@ -302,7 +302,7 @@ This can be done in one of two ways.
 > Note: Only one of these configurations options can be used at a time.
 
 1. Adding the `ca.crl` field to the `nginx.org/ca` secret type, which accepts a base64 encoded certificate revocation list (crl).
-   Example Yaml:
+   Example YAML:
 ```yaml
 kind: Secret
 metadata:
@@ -314,12 +314,12 @@ data:
   ca.crl: <base64encoded-crl>
 ```
 
-2. Adding the `crlFileName` field to your IngressMTLS policy spec with the name of the CRL.
+2. Adding the `crlFileName` field to your IngressMTLS policy spec with the name of the CRL file.
 
 > Note: This configuration option should only be used when using a CRL that is larger than 1MiB
 > Otherwise we recommend using the `nginx.org/ca` secret type for managing your CRL.
 
-Example Yaml:
+Example YAML:
 ```yaml
 apiVersion: k8s.nginx.org/v1
 kind: Policy
@@ -334,7 +334,7 @@ ingressMTLS:
 ```
 
 **IMPORTANT NOTE**
-When configuring a CRL with the `ingressMTLS.crlFileName` field, there are additional steps to consider and limitations to understand.
+When configuring a CRL with the `ingressMTLS.crlFileName` field, there is additional context to keep in mind:
 1. The Ingress Controller will expect the CRL, in this case `webapp.crl`, will be in `/etc/nginx/secrets`. A volume mount will need to be added to the Ingress Controller deployment add your CRL to `/etc/nginx/secrets`
 2. When updating the content of your CRL (e.g a new certificate has been revoked), NGINX will need to be reloaded to pick up the latest changes. Depending on your environment this may require updating the name of your CRL and applying this update to your `ingress-mtls.yaml` policy to ensure NGINX picks up the latest CRL.
 


internal/configs/virtualserver.go
```diff
@@ -908,10 +907,32 @@ func (p *policiesCfg) addIngressMTLSConfig(
...
 	}
 
-	p.IngressMTLS = &version2.IngressMTLS{
-		ClientCert:   secretRef.Path,
-		VerifyClient: verifyClient,
-		VerifyDepth:  verifyDepth,
+	caFields := strings.Fields(secretRef.Path)
+
+	if _, hasCrlKey := secretRef.Secret.Data[CACrlKey]; hasCrlKey && ingressMTLS.Crl != "" {
+		res.addWarningf("Both ca.crl and ingressMTLS.crl fields cannot be used. ca.crl will be ignored and %s will be applied", polKey)
```
ciarams87, line 913: I would be more explicit here that the ca.crl is coming from the secret, something like
```suggestion
		res.addWarningf("Both ca.crl in the Secret and ingressMTLS.crl fields cannot be used. ca.crl in %s will be ignored and %s will be applied", secretKey, polKey)
```

deployments/common/crds/k8s.nginx.org_policies.yaml
```diff
@@ -92,6 +92,8 @@ spec:
                   properties:
                     clientCertSecret:
                       type: string
+                    crl:
```
lucacome, line 95: I think we should use a more descriptive name than just `crl`.

docs/content/configuration/policy-resource.md
```diff
@@ -262,6 +262,52 @@ ingressMTLS:
...
+Example Yaml:
+```yaml
+apiVersion: k8s.nginx.org/v1
+kind: Policy
+metadata:
+  name: ingress-mtls-policy
+spec:
+ingressMTLS:
+    clientCertSecret: ingress-mtls-secret
+    crl: webapp.crl
```
tstraley, line 304: Can you help me understand the expected this?

I'm assuming it is expected that the user would possibly bake this file into their container image at `/etc/nginx/secrets/webapp.crl`, or possibly mount a volume to this path. Is there some other way they may achieve this?

Is the use-case for CRLs that are larger than the 1MiB limit for Secrets?

Does this have a limitation of not automatically reloading NGINX when the file changes (whereas the `ca.crl` Secret data does trigger reload when changed)?

Adding some clarity to the documentation around these points could be helpful.

internal/configs/configurator.go
```diff
@@ -59,8 +59,11 @@ const JWTKeyKey = "jwk"
 // HtpasswdFileKey is the key of the data field of a Secret where the HTTP basic authorization list must be stored
 const HtpasswdFileKey = "htpasswd"
 
-// CAKey is the key of the data field of a Secret where the cert must be stored.
-const CAKey = "ca.crt"
+// CACrtKey is the key of the data field of a Secret where the cert must be stored.
+const CACrtKey = "ca.crt"
+
+// CACrlKey is the key of the data field of a Secret where the cert revocation list much be stored.
```
tstraley, line 65:
```suggestion
// CACrlKey is the key of the data field of a Secret where the cert revocation list must be stored.
```

May also benefit from clarifying this field as optional?

docs/content/configuration/policy-resource.md
shaun-nx: When using the crl field in the policy we are expecting that the user is providing this file to the Ingress Controller in some way, either through a volume or their container image. There is a Note on Line 309 about this.

> Is the use-case for CRLs that are larger than the 1MiB limit for Secrets?
Yes this is the exact use case for this. I can expand the documentation to highlight this, it's not very obvious.

> Does this have a limitation of not automatically reloading NGINX when the file changes (whereas the ca.crl Secret data does trigger reload when changed)?
Yes this is a very good point. I'll make sure this is highlighted in the document. NGINX will not reload automatically when the file contents are changed. You will need to trigger a reload to pick up the latest CRL.

tests/suite/test_ingress_mtls.py
```diff
@@ -53,7 +62,7 @@ def teardown_policy(kube_apis, test_namespace, tls_secret, pol_name, mtls_secret
     delete_secret(kube_apis.v1, mtls_secret, test_namespace)
 
 
-@pytest.mark.policies
+@pytest.mark.policy
```
vepatel, line 65: @shaun-nx this marker should be `policies`

deployments/common/crds/k8s.nginx.org_policies.yaml
ciarams87: I think the crl part is fine, but maybe we could be more descriptive about it being a file name as opposed to the actual list? `crlFile` or `crlFileName` or something?

docs/content/configuration/policy-resource.md
```diff
@@ -284,12 +295,82 @@ We use the `requestHeaders` of the [Action.Proxy](/nginx-ingress-controller/conf
...
+    verifyClient: "on"
+    verifyDepth: 1
+```
+
+**IMPORTANT NOTE**
+When configuring a CRL with the `ingressMTLS.crl` field, there are additional steps to consider and limitations to understand.
+1. The Ingress Controller will expect the CRL, in this case `webapp.crl`, will be in `/etc/nginx/secrets`. We recommend updating your Ingress Controller deployment to add a Volume to mount your CRL to `/etc/nginx/secrets`
+2. When updating the content of your CRL (e.g a new certificate has been revoked), NGINX will need to be reloaded to pick up the latest changes. We recommend updating the name of your CRL and applying this update to your `ingress-mtls.yaml` policy to ensure NGINX picks up the latest CRL.
```
ciarams87, line 339: This will require a new `volumeMount` entry as well, right, which would require a redeploy of the IC?

deployments/common/crds/k8s.nginx.org_policies.yaml
shaun-nx: I think this makes sense yeah. I'll go with `crlFileName` to be as explicit as possible.

vepatel (approved):

docs/content/configuration/policy-resource.md
brianehlert: > es this is a very good point. I'll make sure this is highlighted in the document. NGINX will not reload automatically when the file contents are changed. You will need to trigger a reload to pick up the latest CRL.

I believe that if you are using NGINX Plus you should not have to worry about reloading.  (I might be incorrect on that, but I think so)
https://www.nginx.com/blog/nginx-plus-r18-released/#dynamic-certificate-loading

tstraley: Thanks - those doc updates look great 👍

docs/content/configuration/policy-resource.md
```diff
@@ -284,12 +295,58 @@ We use the `requestHeaders` of the [Action.Proxy](/nginx-ingress-controller/conf
...
+metadata:
+  name: ingress-mtls-secret
+apiVersion: v1
+type: nginx.org/ca
+data:
+  ca.crt: <base64encoded-certificate>
+  ca.crl: <base64encoded-crl>
+```
+
+2. Adding the `crl` field to your IngressMTLS policy spec with the name of the CRL.
```
vepatel, line 317:
```suggestion
2. Adding the `crlFileName` field to your IngressMTLS policy spec with the name of the CRL.
```

shaun-nx: Nice catch!

ADubhlaoich (approved):
Looks good to me.

PR #3646, opened 2023-03-14
Remove app protect agent

jjngx:
### Proposed changes

NIC NAP (app protect) does not need `app agent` anymore. The only one process required to process policies and bundles is `app protect plugin`.

This PR:
- removes references to app protect agent
- update a list of app protect packages installed in docker containers

### Checklist


build/Dockerfile
```diff
@@ -84,7 +84,7 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
 	&& printf "%s\n" "deb https://pkgs.nginx.com/app-protect/${NGINX_PLUS_VERSION}/debian ${DEBIAN_VERSION} nginx-plus" \
 	"deb https://pkgs.nginx.com/app-protect-security-updates/debian ${DEBIAN_VERSION} nginx-plus" > /etc/apt/sources.list.d/nginx-app-protect.list \
 	&& apt-get update \
-	&& apt-get install --no-install-recommends --no-install-suggests -y app-protect app-protect-attack-signatures app-protect-threat-campaigns \
```
brianehlert, line 87: Does the compiler include a default set of attack-signatures and threat-campaigns?
Since those would need to be available on the data plane for JSON APPolicy processing.
This is the main reason customers that deploy NIC + NAP WAF update their images daily today.
AFAIK - only WAF bundles includes signatures and campaigns.

ciarams87: We should still be installing signatures and campaigns for JSON APPolicy customers. I think these might come included with the app-protect package, but in order to get the latest up-to-date versions on our builds (and as Brian alludes to, currently rebuilding the image is the only way to fetch the latest signatures and campaigns), we need to keep these here in our Dockerfile.

build/Dockerfile
```diff
@@ -166,7 +166,7 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/ssl/nginx/nginx-repo.crt,mode
 	&& if [ -z "${NAP_MODULES##*waf*}" ]; then \
 	curl -fsSL https://cs.nginx.com/static/files/app-protect-8.repo > /etc/yum.repos.d/app-protect-8.repo; \
 	sed -i "0,/centos/s;;${NGINX_PLUS_VERSION}/centos;" /etc/yum.repos.d/app-protect-8.repo; \
-	dnf --nodocs install -y app-protect app-protect-attack-signatures app-protect-threat-campaigns; \
+	dnf --nodocs install -y app-protect app-protect-compiler app-protect-common app-protect-engine; \
```
ciarams87, line 169: Ditto to above. Also do we definitely need theses additional packages? I don't see them mentioned in the docs https://docs.nginx.com/nginx-app-protect-waf/admin-guide/install/#rhel-ubi8-docker-deployment-example

ciarams87 (changes requested):

build/Dockerfile
lucacome: `app-protect-compiler`, `app-protect-common` and `app-protect-engine` are already installed by `app-protect` so we don't need to add them here. 
`app-protect-attack-signatures` and `app-protect-threat-campaigns` are dependents of `app-protect` as well and in theory we could remove those too since this layer would be rebuilt anyway, but I think it's better to keep them here for what @ciarams87 said just in case.

build/Dockerfile
jjngx: applied

build/Dockerfile
jjngx: thank you for providing feedback; updates applied

internal/nginx/manager.go
```diff
@@ -462,35 +458,6 @@ func (lm *LocalManager) SetOpenTracing(openTracing bool) {
...
 }
 
-// AppProtectAgentStart starts the AppProtect agent
-func (lm *LocalManager) AppProtectAgentStart(apaDone chan error, logLevel string) {
-	glog.V(3).Info("Setting log level for App Protect - ", logLevel)
-	appProtectLogLevelCmdfull := fmt.Sprintf("%v %v", appProtectLogLevelCmd, logLevel)
-	logLevelCmd := exec.Command("sh", "-c", appProtectLogLevelCmdfull) // #nosec G204
-	if err := logLevelCmd.Run(); err != nil {
-		glog.Fatalf("Failed to set log level for AppProtect: %v", err)
-	}
```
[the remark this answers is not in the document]
jjngx: done, thanks for reviewing

ciarams87 (approved):

PR #3648, opened 2023-03-14
Remove strict permissions for Helm in CI

lucacome:
### Proposed changes
Looks like `packages: write` is not enough for the pipeline to overwrite packages, removing it for now to fix the pipeline.


jasonwilliams14 (approved):
looks good.

PR #3665, opened 2023-03-19
Apply UID/GID defaults from image

### LINKED ISSUE #5422: Support OpenShift's built-in `restricted-v2` Security Context Constraint

Security teams prefer referencing default (built-in) security restrictions. In OpenShift (v4.11+) the `restricted-v2` Security Context Constraint is default, and previously (up to v4.10) the `restricted` SCC was default. Both of these SCCs require that a pod is run as a user in a pre-allocated range of UIDs. This conflicts current Nginx Ingress Controller set-up which uses UID 101.

Nginx Ingress Controller should stop specifying explicit UID in securityContext. Deployments in vanilla Kubernetes will inherit container image default UID, retaining existing behavior. Deployments in OpenShift will be allowed to choose any UID. Users with OpenShift, with existing SCC for NIC would also retain existing RunAsUser behavior.

This is an explicit security requirement. Only alternative is WONTFIX - to not comply with OpenShift requirements.

**Additional context**

- _Table 1. Default security context constraints_ in [Managing security context constraints - Authentication and authorization - OpenShift Container Platform 4.15](https://docs.openshift.com/container-platform/4.15/authentication/managing-security-context-constraints.html#default-sccs_configuring-internal-oauth)

sigv:
### Proposed changes

[`build/Dockerfile` specifies `USER 101` for `common` target](https://github.com/nginxinc/kubernetes-ingress/blob/v3.2.1/build/Dockerfile#L247-L248), which is re-applied into the final images. Helm Chart/Manifests do not need to specify UID explicitly, and can instead use the image's UID. (PodSecurityContext v1 core specifies `runAsUser` defaults to user specified in image metadata if unspecified.)

The existing `runAsNonRoot: true` flag (already in place) will ensure during runtime that the image is configured with a custom user ID.

This is notably helpful for users running OpenShift, because OpenShift attempts to enforce custom UID/GID ranges for individual namespaces as part of `restricted-v2` Security Context Constraint. When removing hard-coded values from manifests, OpenShift will be able to assign its own UID/GID.

In practice, this means a different model of configuring file system permissions. OpenShift assigns the container process GID 0 as supplemental to assist with that. Locations that are expected to be written to must be owned by GID 0, with group write permissions. Previous changes to `main` have ensured that is the case.

Init container copying files is not a concern, as we will have the same UID as owner there as the main NIC container.

Reference: [A Guide to OpenShift and UIDs](https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids)

Closes #5422.

### Checklist

### DIFF
--- a/deployments/daemon-set/nginx-ingress.yaml
+++ b/deployments/daemon-set/nginx-ingress.yaml
@@ -61,7 +61,6 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
 #          readOnlyRootFilesystem: true
-          runAsUser: 101 #nginx
           runAsNonRoot: true
           capabilities:
             drop:
@@ -103,7 +102,6 @@ spec:
 #        securityContext:
 #          allowPrivilegeEscalation: false
 #          readOnlyRootFilesystem: true
-#          runAsUser: 101 #nginx
 #          runAsNonRoot: true
 #          capabilities:
 #            drop:
--- a/charts/nginx-ingress/templates/controller-daemonset.yaml
+++ b/charts/nginx-ingress/templates/controller-daemonset.yaml
@@ -100,7 +100,6 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: {{ .Values.controller.readOnlyRootFilesystem }}
-          runAsUser: 101 #nginx
           runAsNonRoot: true
           capabilities:
             drop:
@@ -153,7 +152,6 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
-          runAsUser: 101 #nginx
           runAsNonRoot: true
           capabilities:
             drop:
--- a/charts/nginx-ingress/templates/controller-deployment.yaml
+++ b/charts/nginx-ingress/templates/controller-deployment.yaml
@@ -109,7 +109,6 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: {{ .Values.controller.readOnlyRootFilesystem }}
-          runAsUser: 101 #nginx
           runAsNonRoot: true
           capabilities:
             drop:
@@ -160,7 +159,6 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
-          runAsUser: 101 #nginx
           runAsNonRoot: true
           capabilities:
             drop:
--- a/charts/nginx-ingress/values.yaml
+++ b/charts/nginx-ingress/values.yaml
@@ -176,7 +176,6 @@ controller:
   securityContext: {} # Remove curly brackets before adding values
     # allowPrivilegeEscalation: true
     # readOnlyRootFilesystem: true
-    # runAsUser: 101 #nginx
     # runAsNonRoot: true
     # capabilities:
     #   drop:
--- a/deployments/daemon-set/nginx-plus-ingress.yaml
+++ b/deployments/daemon-set/nginx-plus-ingress.yaml
@@ -61,7 +61,6 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
 #          readOnlyRootFilesystem: true
-          runAsUser: 101 #nginx
           runAsNonRoot: true
           capabilities:
             drop:
@@ -106,7 +105,6 @@ spec:
 #        securityContext:
 #          allowPrivilegeEscalation: false
 #          readOnlyRootFilesystem: true
-#          runAsUser: 101 #nginx
 #          runAsNonRoot: true
 #          capabilities:
 #            drop:
--- a/deployments/deployment/nginx-ingress.yaml
+++ b/deployments/deployment/nginx-ingress.yaml
@@ -60,7 +60,6 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
 #          readOnlyRootFilesystem: true
-          runAsUser: 101 #nginx
           runAsNonRoot: true
           capabilities:
             drop:
@@ -104,7 +103,6 @@ spec:
 #        securityContext:
 #          allowPrivilegeEscalation: false
 #          readOnlyRootFilesystem: true
-#          runAsUser: 101 #nginx
 #          runAsNonRoot: true
 #          capabilities:
 #            drop:
--- a/deployments/deployment/nginx-plus-ingress.yaml
+++ b/deployments/deployment/nginx-plus-ingress.yaml
@@ -62,7 +62,6 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
 #          readOnlyRootFilesystem: true
-          runAsUser: 101 #nginx
           runAsNonRoot: true
           capabilities:
             drop:
@@ -110,7 +109,6 @@ spec:
 #        securityContext:
 #          allowPrivilegeEscalation: false
 #          readOnlyRootFilesystem: true
-#          runAsUser: 101 #nginx
 #          runAsNonRoot: true
 #          capabilities:
 #            drop:


sigv:
Not sure if [CI / Smoke Tests (alpine, policies, 1.26.2)](https://github.com/nginxinc/kubernetes-ingress/actions/runs/4483509692/jobs/7885077160?pr=3665) failure of internally observed 502 Bad Gateway response is relevant to this PR. 🤔

deployments/daemon-set/nginx-ingress.yaml
```diff
@@ -65,7 +65,8 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
 #          readOnlyRootFilesystem: true
-          runAsUser: 101 #nginx
+          runAsUser: null
```
lucacome, line 68: could this just be removed? instead of `null`?

lucacome (commented):
Hi @sigv thanks for the PR and sorry for the late reply. 

Did you test this in OpenShift? If it runs as a random UID and we set all the permissions for the files to the `nginx` user, will it cause problems?

Shouldn't this be enforced in the SCC? Can't we just set `101` there? See https://github.com/nginxinc/nginx-ingress-helm-operator/blob/main/resources/scc.yaml#L7-L9

I haven't had time to properly look into this and I was wondering if you know the answers already to my questions 🙂

sigv:
Hey hey. Sorry, been busy with a few initiatives having tight delivery deadlines.

I did not test this out on OpenShift. Opened the PR based on a community member feedback, and myself checked that the K8s perspective is functional.

[OpenShift's blog/guidance](https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids) covers the ownership/access concern:
> The Container user is always a member of the root group, so it can read or write files accessible by GID=0. Any command invoked by the Entrypoint will be executed with this unprivileged UID and GID pair. That means, it is an unprivileged user executing the commands and the UID that will be used during execution is not known in advance. From the technical design perspective, that means, directories and files that may be written to by processes in the Container should be owned by the root group and be read/writable by GID=0. Files to be executed should also have group execute permissions.

Will mark this PR as a draft - need to double-check on the concern raised.

deployments/daemon-set/nginx-ingress.yaml
sigv: I was thinking it might be preferred to explicitly say "this is undefined".
But yes - should be safe to drop the line outright, allowing the default behavior to kick in automatically.

sigv:
Sorry for the lack of reply on this. The open initiatives getting wrapped up so I have some time to look into this again.

> Did you test this in OpenShift? If it runs as a random UID and we set all the permissions for the files to the nginx user, will it cause problems?

I personally operate Kubernetes clusters instead of OpenShift clusters. I don't feel like reading the licensing model by RedHat or registering for a 2-month developer trial, so I will spin up a basic Pay-as-You-Go OpenShift cluster in Azure instead. Microsoft explicitly requires [Dsv3 (3rd generation D-series) VMs](https://learn.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series) for OpenShift master nodes, so I just requested a quota change on my Azure account. Should be able to deploy soon and report how the file operations behave.

> Shouldn't this be enforced in the SCC? Can't we just set 101 there? See https://github.com/nginxinc/nginx-ingress-helm-operator/blob/main/resources/scc.yaml#L7-L9

The linked issue requests a `restricted-v2` SCC to be supported out-of-the-box. In particular, the issue is that the SCC "Requires that a pod is run as a user in a pre-allocated range of UIDs". (Reference: RedHat: [Managing security context constraints](https://docs.openshift.com/container-platform/4.12/authentication/managing-security-context-constraints.html))

I think that means a `runAsUser.type: MustRunAsNonRoot` preferably, so that no default UID is specified in the SCC.
Will double-check this during testing - a good time for me to learn a bit more about the OpenShift policies here.

sigv:
Fun fact. When deploying an OpenShift cluster in Azure, there are actually [three kinds of VMs that get deployed for you](https://docs.openshift.com/container-platform/4.12/installing/installing_azure/installing-azure-user-infra.html#installation-azure-limits_installing-azure-user-infra):
- One bootstrap machine, which is removed after installation
- Three control plane machines
- Three compute machines
The three control plane machines can be D8s_v3, resulting in 24 cores used. But it appears the bootstrap machine also uses the D8s_v3 size, so the actual minimum quota requirement there is 32 cores (excluding worker/compute nodes). So I am waiting for a further quota increase.

sigv:
So, a quick sanity check that it works on local K8s without fancy add-ons:

```bash
# install microk8s locally
sudo snap install microk8s --classic
microk8s status --wait-ready

# or reset an existing microk8s installation
sudo microk8s reset

# enable basic services on microk8s
microk8s enable dashboard dns registry istio

microk8s helm install nginx-ingress deployments/helm-chart \
  --namespace nginx-ingress --create-namespace \
  --set controller.image.repository=ghcr.io/nginxinc/kubernetes-ingress \
  --set controller.image.tag=edge \
  --set controller.service.type=NodePort

microk8s kubectl -n nginx-ingress get pods
# NAME                                        READY   STATUS    RESTARTS   AGE
# nginx-ingress-controller-6bd8954f55-rxsc9   1/1     Running   0          60s

microk8s kubectl -n nginx-ingress exec -it nginx-ingress-controller-6bd8954f55-rxsc9 -- /usr/bin/id -u
# 101
```

---

Currently dropping a few findings and notes. OpenShift has interesting security features built in. To summarize, it wants to run things more restricted than "native Kubernetes".

To be honest, I am a little annoyed by RedHat and/or Microsoft, as I was working based off expectations from latest documentation. When deploying installer-provided infrastructure in Microsoft Azure you get Openshift v4.10 (current is v4.12). Why this matters is that v4.11 introduces the `restricted-v2` security context constraints. When running v4.10, you only have the older `restricted` security context constraints, which does not seem to allow `NET_BIND_SERVICE` capability to be added explicitly -- that's CAPS column below, listing `<no value>` instead of `["NET_BIND_SERVICE"]`.

```bash
oc get scc privileged restricted restricted-v2
# NAME         PRIV    CAPS         SELINUX     RUNASUSER        FSGROUP     SUPGROUP   PRIORITY     READONLYROOTFS   VOLUMES
# privileged   true    ["*"]        RunAsAny    RunAsAny         RunAsAny    RunAsAny   <no value>   false            ["*"]
# restricted   false   <no value>   MustRunAs   MustRunAsRange   MustRunAs   RunAsAny   <no value>   false            ["configMap","downwardAPI","emptyDir","persistentVolumeClaim","projected","secret"]
# Error from server (NotFound): securitycontextconstraints.security.openshift.io "restricted-v2" not found
```

The error event observed for ReplicaSet (not able to deploy Pod) was initially not clear to me, when trying to deploy the Helm chart. But really the "providers" are just "security context constraints" it tries to evaluate.

```text
pods "nginx-ingress-controller-deadbeef00-" is forbidden: unable to validate against any security context constraint: [
provider "anyuid": Forbidden: not usable by user or serviceaccount, spec.containers[0].securityContext.capabilities.add: Invalid value: "NET_BIND_SERVICE": capability may not be added,
provider "nonroot": Forbidden: not usable by user or serviceaccount,
provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount,
provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount,
provider "hostnetwork": Forbidden: not usable by user or serviceaccount,
provider "hostaccess": Forbidden: not usable by user or serviceaccount,
provider "node-exporter": Forbidden: not usable by user or serviceaccount,
provider "privileged": Forbidden: not usable by user or serviceaccount,
provider "privileged-genevalogging": Forbidden: not usable by user or serviceaccount]
```

And `Forbidden: not usable by user or serviceaccount` can be resolved by adding RoleBinding for the security context constraint to the `nginx-ingress` ServiceAccount. I am not sure why `restricted` SCC is not being automatically applied/why there is no failure information visible to me for `restricted` SCC in logs.

```bash
oc new-project nginx-ingress --display-name='Nginx Ingress Controller'

oc -n nginx-ingress adm policy add-scc-to-user restricted -z nginx-ingress
oc -n nginx-ingress adm policy add-scc-to-user privileged -z nginx-ingress

helm install nginx-ingress deployments/helm-chart \
  --namespace nginx-ingress \
  --set controller.image.repository=ghcr.io/nginxinc/kubernetes-ingress \
  --set controller.image.tag=edge

kubectl -n nginx-ingress exec -it nginx-ingress-controller-689d96577b-cj2kv -- /usr/bin/id -u
# 101
```

I want to poke around a bit more -- need to manually configure a `restricted` SCC variant that allows `NET_BIND_SERVICE`, to see how the additional security safeguards change the UID behavior.

sigv:
I need to check on the difference that `edge-ubi` brings with itself, as the [Building the Ingress Controller Image](https://docs.nginx.com/nginx-ingress-controller/installation/building-ingress-controller-image/#makefile-targets) page says its for OpenShift clusters. That's `nginxcontrib/nginx:1.23.4-ubi` as base image, instead of default `nginx:1.23.4`.

sigv:
Red Hat OpenShift Service on AWS (ROSA) needs some manual configuration to authorize access for the RedHat Cloud Console to your AWS account. But after the linking is done, you have a smooth deployment process from RedHat Console. This allows picking specific versions from dropdown, so the deployed cluster is already upgraded for you. Instead of having to rely on the pre-selected version as in Microsoft Azure, and then selecting an upgrade. Small things, but they can matter when you are messing around with cluster deployments.

---

About OpenShift versioning: v4.10 (released March 2022) is in Maintenance Support until September 2023. This is the version that does not have `restricted-v2` security context constraints available. (Docs: [Red Hat OpenShift Container Platform Life Cycle Policy](https://access.redhat.com/support/policy/updates/openshift#dates))

I was not able to configure an SCC based on `restricted` that gets picked up in the `nginx-ingress` namespace. Something is causing the built-in `restricted` SCC to be ignored, so if I base my new template on it then it gets ignored as well. (Or maybe I am missing something else.)

In any case, running NIC as `privileged` SCC seems to be functional, just as in plain K8s.

```bash
oc new-project nginx-ingress --display-name='Nginx Ingress Controller'
oc adm policy add-scc-to-user privileged -z nginx-ingress -n nginx-ingress
helm install nginx-ingress deployments/helm-chart \
  --namespace nginx-ingress \
  --set controller.image.repository=ghcr.io/nginxinc/kubernetes-ingress \
  --set controller.image.tag=edge-ubi

kubectl -n nginx-ingress get pods
# NAME                                        READY   STATUS    RESTARTS   AGE
# nginx-ingress-controller-6988bc87d9-fckm5   1/1     Running   0          30s
kubectl -n nginx-ingress exec -it nginx-ingress-controller-6988bc87d9-fckm5 -- /bin/id -u
# 101
```

I then deleted the `nginx-ingress` namespace, and asked RedHat Console to upgrade my cluster to the 4.11 channel.
When starting in 4.11+, you get `runAsUser: 1001040000` and accordingly in the Pod:
```text
Failed to write main config: failed to open /etc/nginx/nginx.conf: open /etc/nginx/nginx.conf: permission denied
```

If Helm chart is applied with `controller.readOnlyRootFilesystem=true` value, then pod admission doesn't like `fsGroup` being set to `101` explicitly.

---

So I am currently trying to think how I can patch up the project for GID=0 as well as resolve fsGroup being flexible. Will follow up with further changes.

sigv:
OpenShift SCCs are actually fairly straight forward. They are a programmatic way to ensure a policy, and they hook into the pod admission process. Nothing crazy. Except sometimes event messages do not include reason why `restricted` constraint cannot be applied. Go figure.

Overall, ended up being a bit larger of a change than I initially thought. Here is the run-down.

---

The change removes `runAsUser` and `fsGroup` so the UID and GID assignments are not being enforced by Helm chart template directly.

UID selection on Kubernetes is inherited from the image metadata (so it retains UID 101 as now), while the GID logic can be left up to defaults (we don't depend on it).

---

For OpenShift 4.11+, this change means that the default locked down `restricted-v2` security context constraints are functional. They can get randomized UID/GID.

Even though UID/GID are randomized, a supplemental GID 0 is applied. That means the NIC process can work with files and directories where group ownership is set to root.

Dockerfile already configures ownership to GID 0, so we adjust permissions to allow group the same rights as owner user.

---

I know `/var/log/nginx` directory was added as a potential write location when implementing read-only root filesystem. The Dockerfile now also changes ownership of this directory, in the same manner as for `/etc/nginx` and the other `/var` locations.

The directory was previously mode `755` owned by `root:root`.
Two files existed in it -- `access.log` and `error.log` -- with modes `640` owned by `nginx:adm`.
The ownerships are changes to `nginx:root` now, with mode `775` for directory and `660` for files.

---

Built an UBI image with `make ubi-image TARGET=container` and validated that it deploys to OpenShift 4.11 with:
```bash
helm install nginx-ingress deployments/helm-chart \
  --namespace nginx-ingress --create-namespace \
  --set controller.image.repository=docker.io/sigv/nginx-ingress \
  --set controller.image.tag=cbe5594-ubi
```

I did a quick sanity check that the controller responds with an HTTP 404.

I then applied r/o rootfs with:
```bash
helm upgrade nginx-ingress deployments/helm-chart -n nginx-ingress \
  --set controller.image.repository=docker.io/sigv/nginx-ingress \
  --set controller.image.tag=cbe5594-ubi \
  --set controller.readOnlyRootFilesystem=true
```
..and observed the same HTTP 404.

---

On local machine,
```bash
microk8s helm install nginx-ingress deployments/helm-chart \
  --namespace nginx-ingress --create-namespace \
  --set controller.image.repository=docker.io/sigv/nginx-ingress \
  --set controller.image.tag=cbe5594-debian \
  --set controller.service.type=NodePort
```
..starts up successfully.

---

I am redeploying OpenShift to version 4.10. (Should not have done an in-place upgrade previously probably. I would avoid waiting on the deployment now. Oh well.) When it's up, I will verify NIC still launches on OpenShift 4.10 with `privileged` SCC assigned like it did before.

Will run the test suite as well to verify there is no strangeness going on.

OBS: If the workflow gets run on Github Actions, then it needs to be run as protected inside of the official repository -- there is a Dockerfile change, which means the cache cannot be reused, and the Plus secrets are needed for the new build.

sigv:
Executive summary: This change allows development to OpenShift without having to manually configuring custom security constraints.

sigv:
@lucacome, should I maybe split this apart into smaller PRs based on the technical changes to be reviewed independently? Or is this okay as-is, one functional change for OpenShift users?

sigv:
I split off #3925 and #3926 to deal with clean-up changes, that were initially included in this change but can be reviewed individually. Hopefully this makes review easier on those two tasks.

I will rebase this PR, when they are approved, so that change scope here is halved, and update the description accordingly. 🤞🏻

sigv:
@lucacome, should I split the Dockerfile changes into separate PRs - for `chown 101:0 /var/log/nginx` and for `chmod g=u`?

That way this PR only consists of the `runAsUser` removal? That part is the tricky part that can break some custom images, where people add `USER 0` in the Dockerfile but do not add `USER 101` later.

andrescolodrero:
HI @sigv 
From a bare metall openshift, im having this error in the replica set:
But i dont see where to override runAsUser

Error creating: pods "nginxingress-sample-nginx-ingress-controller-b44ccd556-" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, spec.containers[0].securityContext.runAsUser: Invalid value: 101: must be in the ranges: [1000880000, 1000889999], provider "restricted": Forbidden: not usable by user or serviceaccount, provider "nonroot-v2": Forbidden: not usable by user or serviceaccount, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount, provider "hostnetwork-v2": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or

sigv:
The proposed diff is visible in this Pull Request's [Files changed tab](https://github.com/nginxinc/kubernetes-ingress/pull/3665/files).

If you are using the Helm chart, the relevant references are in [`deployments/helm-chart/templates/`](https://github.com/nginxinc/kubernetes-ingress/tree/83d8b7d97914221fcb0682a542290868e51b2138/deployments/helm-chart/templates) -- depending on whether a DaemonSet or Deployment is used, it's either [`controller-daemonset.yaml`](https://github.com/nginxinc/kubernetes-ingress/blob/83d8b7d97914221fcb0682a542290868e51b2138/deployments/helm-chart/templates/controller-daemonset.yaml) or [`controller-deployment.yaml`](https://github.com/nginxinc/kubernetes-ingress/blob/83d8b7d97914221fcb0682a542290868e51b2138/deployments/helm-chart/templates/controller-deployment.yaml) respectively.

If you are instead using the static manifests, the relevant references are in [`deployments/daemon-set`](https://github.com/nginxinc/kubernetes-ingress/tree/83d8b7d97914221fcb0682a542290868e51b2138/deployments/daemon-set) or [`deployments/deployment`](https://github.com/nginxinc/kubernetes-ingress/tree/83d8b7d97914221fcb0682a542290868e51b2138/deployments/deployment) depending on installation type, as above. Either `nginx-ingress.yaml` or `nginx-plus-ingress.yaml` file will be relevant, depending on the service type.

However, do keep in mind that there is a proposed change to [`build/Dockerfile`](https://github.com/nginxinc/kubernetes-ingress/blob/83d8b7d97914221fcb0682a542290868e51b2138/build/Dockerfile) pending as well, so you would have to manually build an image that you load into your Openshift cluster. The latest release build does not include these prerequisite changes.

A notable part of the Dockerfile change is pending review as #3962. There is still a concern that if you are currently writing anything to `/var/log/nginx` then this scope would not be enough. However for a vanilla configuration, when that scope is merged, the latest edge build should satisfy your requirements.

sigv:
@andrescolodrero, if you still experience issues, let me know what Openshift version you are running and what exact commands you are using to carry out the installation.

andrescolodrero:
@sigv 
IM using Openshift 4.12

Installation is done via Operator, basic setup from the UI.
Operator version: nginx-ingress-operator.v1.5.0

sigv:
Yes. The Operator is not yet scoped in for removal of anyuid. This is a slow work-in-progress.

andrescolodrero:
ok thanks @sigv 
I tried also Helm install with same results, i will check building my own image of ingress.

brianehlert:
> Yes. The Operator is not yet scoped in for removal of anyuid. This is a slow work-in-progress.

Considering that our Operator is derived from the Helm chart...  We need to make sure it works.

build/Dockerfile
```diff
@@ -234,8 +234,8 @@ RUN --mount=type=bind,target=/tmp mkdir -p /var/lib/nginx /etc/nginx/secrets /et
 	&& setcap -v 'cap_net_bind_service=+eip' /usr/sbin/nginx 'cap_net_bind_service=+eip' /usr/sbin/nginx-debug \
 	&& [ -z "${BUILD_OS##*plus*}" ] && PLUS=-plus; cp -a /tmp/internal/configs/version1/nginx$PLUS.ingress.tmpl /tmp/internal/configs/version1/nginx$PLUS.tmpl \
 	/tmp/internal/configs/version2/nginx$PLUS.virtualserver.tmpl /tmp/internal/configs/version2/nginx$PLUS.transportserver.tmpl / \
-	&& chown -R 101:0 /etc/nginx /var/cache/nginx /var/lib/nginx /*.tmpl \
-	&& chmod -R g=u /etc/nginx /var/cache/nginx /var/lib/nginx /*.tmpl \
+	&& chown -R 101:0 /etc/nginx /var/cache/nginx /var/lib/nginx /var/log/nginx /*.tmpl \
+	&& chmod -R g=u /etc/nginx /var/cache/nginx /var/lib/nginx /var/log/nginx /*.tmpl \
```
sigv, line 238: @lucacome, back in a24f8372cc1f7d46e01e117d65e68c5ec8026d97 you changed the default logging targets to `stderr` and `stdout`. Is it preferred to make `/var/log/nginx` writeable for compatibility?

If so, I am also considering splitting this into a separate small PR, as it would be a no-op for current `runAsUser: 101` deployments. Then it would be more straight forward to test the OpenShift change scope by using `edge-ubi` image, instead having to build manually.

brianehlert: The problem with writing logs to file is that there is no logrotate within a container, so we run into a disk filling problem.
Also, stderr and stdout means that K8s intercepts the messages and they are associated with the pod through kubectl.

We do have customers that write log to file for very specific 3rd party tool purposes, but it is rare and the logrotate solution is complicated.

Therefore as a rule stderr and stdout is the common assumption.

sigv: I am just wondering whether it is considered a supported solution. Sounds like it is for those specific customers. Will open a PR in that case.

brianehlert: NGINX supports any path for any files.  There just happen to be defaults.
Are you thinking of how this interacts with readOnlyRootFileSystem or just the default log path permissions in general?
I mean, there is no harm with setting to correct group membership for the rare case.

sigv: r/o root filesystem has a mountpoint to support `/var/log/nginx` scenario. No issues expected there.

This PR is concerned additional constraints that OpenShift wants to impose by default. Notably, it wants to run with a randomized UID/GID, where GID 0 is assigned as supplemental for filesystem actions.

Will try and phrase it nicely in the separate PR. 🤞🏻

sigv: Opened #4269 for that scope. Removing the log directory change from this PR.

sigv:
With #4269 merged, I will dust this off (merge conflicts) and look into relevant Operator changes.

There are currently other initiatives that I am working on, but later this month I should be able to work on this change too.

sigv:
Not seeing any issues with Operator from my end. Opened nginxinc/nginx-ingress-helm-operator#254 to modify the SCC included in the repository there.

Let me know what testing would be preferable to be done from my side.

ADubhlaoich (commented):
LGTM docs-wise: most of the PR is not docs-related, so I am not giving an approval.

vepatel:
Hi @sigv can you create a new issue for this as the one linked in description seems incomplete and WIP, we can then prioritise and review the PR, thanks!

sigv:
@vepatel, opened a fresh #5422 for this scope. 🤞🏻

j1m-ryan:
Hi @sigv, now that securityContext is configurably using helm, is this still required?

PR #3690, opened 2023-03-27
Move build docker steps after tests

lucacome:
This re-arranges the steps and moves the build of the docker images after the smoke tests. This way the docker images can only be built after all the tests pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,65 +129,10 @@ jobs:
           path: ${{ github.workspace }}/dist
           key: nginx-ingress-${{ github.run_id }}-${{ github.run_number }}
 
-  build-docker:
-    name: Build Docker OSS
-    needs: binaries
-    strategy:
-        fail-fast: false
-        matrix:
-          image: [debian, alpine]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -333,6 +278,61 @@ jobs:
...
+    uses: ./.github/workflows/build-plus.yml
+    with:
+      platforms: ${{ matrix.platforms }}
+      image: ${{ matrix.image }}
+      target: ${{ matrix.target }}
+    secrets: inherit
+
+  build-docker-nap:
+    name: Build Docker NAP
+    needs: build-docker-plus
```
ciarams87, line 320: Why do we need to build the 3 types of images consecutively instead of concurrently?

lucacome: It was to limit the number of cache access and get fewer `429`s from GitHub. But it seems like it's not doing much anyway, we can try to remove it again...

ciarams87 (approved):

shaun-nx (approved):

vepatel (approved):

PR #3726, opened 2023-04-05
Update VirtualServer template to allow internal JWKS locations to use a common proxy_cache

shaun-nx:
### Proposed changes
This change:
- Updates virtual server template to allow all internal JWKS locations to use a common proxy_cache_path.
- Add additional python tests for subpaths, VirtualServerRoute and multiple VS deployments.

### Checklist


tests/suite/test_jwt_policies_jwksuri.py
```diff
@@ -41,7 +56,7 @@ def get_token(request):
 
 @pytest.mark.skip_for_nginx_oss
 @pytest.mark.policies
-@pytest.mark.skip(reason="issues with ingressClass")
+# @pytest.mark.skip(reason="issues with ingressClass")
```
vepatel, line 59: @shaun-nx can we remove this comment, probably not needed after fixing the count issue

shaun-nx: I'll remove it from the new VSR test as well

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -38,6 +38,8 @@ http {
     include       /etc/nginx/mime.types;
     default_type  application/octet-stream;
 
+    proxy_cache_path /var/cache/nginx/jwks_uri levels=1 keys_zone=jwks_uri:1m max_size=10m;
```
lucacome, line 41: I don't think we want to set this at the `http` level. This will create a cache unconditionally and it's not needed when somebody's not using `jwks`.

internal/configs/version2/nginx-plus.virtualserver.tmpl
lucacome, line 1: It won't let me comment on line 374, but `{{ if .KeyCache }}` there is not needed, it's independent of `proxy_cache_path`.

internal/configs/version1/nginx-plus.tmpl
shaun-nx: [proxy_cache_path](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_path) can only be in the http context. This was originally in the http context, but was set in the virtual server config file. So in cases where we deployed two Virtual Servers with this policy we could get a conflict.

e.g.
```
stderr: "2023/04/04 09:25:30 [emerg] 62#62: the same path name \"/var/cache/nginx/jwks_uri\" used in /etc/nginx/conf.d/vs_default_cafe.conf:53 and in /etc/nginx/conf.d/vs_default_webapp.conf:33\n"
finished with error: exit status 1
```

shaun-nx: You are right about it being there unconditionally. With OIDC this directive was only set when using the `-enable-oidc` flag in the deployment. Since proxy_cache_path can only exist in the http context we would need to add a similar flag for JWKS.

lucacome: @shaun-nx Sorry poor choice of words. I meant that we should keep it inside the VirtualServer template. We can avoid conflicts by naming the cache uniquely for the VirtualServer, maybe something like adding the policy name.

It might also be a security concern to store all in a single cache for all the VSs, we might need to look into it.

brianehlert: This needs to be in the server context.  At the VS level.
Was this resolved?

shaun-nx: @brianehlert it can be in a VS config, but it can't exist in a server context in NGINX. If the proxy_cache_path is going to be in the VS config then the cache path's will need to be uniquely named per VirtualServer as @lucacome mentioned.

brianehlert: Sorry, I phrased my question that way since a VirtualServer is analogous to a server block in the nginx.conf
Which sets a context for the cache path and for the jwks_uri request path itself.

Issue #3732, opened 2023-04-06
proxy_ssl_trusted_certificate getting cert and then crl on the same line


brianehlert:
### Discussed in https://github.com/nginxinc/kubernetes-ingress/discussions/3731

<div type='discussions-op-text'>

<sup>Originally posted by **coolbry95** April  6, 2023</sup>
I believe this is breaking from this line here https://github.com/nginxinc/kubernetes-ingress/pull/3632/files#diff-0a308ab187fcddbbc7a73510aacc512d8a04022d75c6f95044b1e3fd4c483770R738.

Before it would just return the full path for the cert but now its returning the cert and the crl.

```
proxy_ssl_trusted_certificate /etc/nginx/secrets/asdf-trusted-ca-ca.crt /etc/nginx/secrets/asdf-trusted-ca-ca.crl;

apiVersion: k8s.nginx.org/v1
kind: Policy
metadata:
  name: egress-mtls-oauth2
spec:
  egressMTLS:
    trustedCertSecret: egress-trusted-ca
    verifyServer: true
    verifyDepth: 25
    serverName: true
    sessionReuse: true
    sslName: oauth2.googleapis.com
apiVersion: k8s.nginx.org/v1
kind: Policy
metadata:
  name: egress-mtls-pubsub
spec:
  egressMTLS:
    trustedCertSecret: egress-trusted-ca
    verifyServer: true
    verifyDepth: 25
    serverName: true
    sessionReuse: true
    sslName: pubsub.googleapis.com
apiVersion: k8s.nginx.org/v1
kind: VirtualServer
metadata:
  name: oauth
spec:
  host: oauth2.googleapis.com
  tls:
    secret: egress-proxy-tls
  policies:
    - name: ingress-mtls
    - name: egress-mtls-oauth2
  upstreams:
  - name: oauth
    service: oauth
    port: 443
    tls:
      enable: true
  routes:
  - path: /
    action:
      pass: oauth
apiVersion: k8s.nginx.org/v1
kind: VirtualServer
metadata:
  name: pubsub
spec:
  host: pubsub.googleapis.com
  tls:
    secret: egress-proxy-tls
  policies:
    - name: ingress-mtls
    - name: egress-mtls-pubsub
  upstreams:
  - name: pubsub
    service: pubsub
    port: 443
    type: grpc
    tls:
      enable: true
  routes:
  - path: /
    action:
      pass: pubsub
```</div>

brianehlert:
[...]

PR #3737, opened 2023-04-07
Update VirtualServer to ignore CRL for EgressMTLS

shaun-nx:
### Proposed changes
This change updates the logic which applies policies to VirtualServer to ignore CRL for EgressMTLS.
Resolves https://github.com/nginxinc/kubernetes-ingress/issues/3732

### Checklist


lucacome (changes requested):
I don't think we want to ignore the CRL. We should use the right directive for it `proxy_ssl_crl`.

brianehlert:
> I don't think we want to ignore the CRL. We should use the right directive for it `proxy_ssl_crl`.

That would be the right thing to do.

shaun-nx:
@brianehlert @lucacome 
The aim of this PR isn't to support using CRLs with EgressMTLs. This only addresses the issue posted in https://github.com/nginxinc/kubernetes-ingress/issues/3732

We only added support for CRLs with IngressMTLs but not EgressMTLs.
The PR for IngressMTLs update is here for reference: https://github.com/nginxinc/kubernetes-ingress/pull/3632

brianehlert:
To close this conversation, we will be adding proper EgressMTLS support for CRL to match IngressMTLS behavior.
That is not work contained in this PR.

vepatel (approved):

jjngx (approved):

ciarams87 (approved):

Issue #3769, opened 2023-04-13
Need backward support for autoscaling/v2beta2


johnzeng2010:
Per https://github.com/argoproj/argo-cd/issues/9145, the argocd does not support autoscaling/v2.  The current support to auto scale of nginx is using v2.  This blocks us to use nginx auto scale feature since we use argocd to do the deployment.

The solution is probably best served by a conditional in the Helm template.  This allows to support both v2 and v2beta2.

A new knob under autoscaling should work for us. 

Couple of more issues related:

https://github.com/argoproj/argo-cd/issues/7846
https://github.com/kubernetes/kubernetes/issues/74099

PR #3773, opened 2023-04-17
Swap cpu and memory in HPA template

coolbry95:
### Proposed changes
There is a kubernetes issue where the cpu and memory get swapped at some point. This make Argocd think the manifest is always out of sync. Swap them here so that Argocd is happy.

#3769

### Checklist

### DIFF
--- a/deployments/helm-chart/templates/controller-deployment.yaml
+++ b/deployments/helm-chart/templates/controller-deployment.yaml
@@ -10,7 +10,9 @@ metadata:
   annotations: {{ toYaml .Values.controller.annotations | nindent 4 }}
 {{- end }}
 spec:
+  {{- if not .Values.controller.autoscaling.enabled }}
   replicas: {{ .Values.controller.replicaCount }}
+  {{- end }}
   selector:
     matchLabels:
       {{- include "nginx-ingress.selectorLabels" . | nindent 6 }}
--- a/deployments/helm-chart/templates/controller-hpa.yaml
+++ b/deployments/helm-chart/templates/controller-hpa.yaml
@@ -18,14 +18,6 @@ spec:
   minReplicas: {{ .Values.controller.autoscaling.minReplicas }}
   maxReplicas: {{ .Values.controller.autoscaling.maxReplicas }}
   metrics:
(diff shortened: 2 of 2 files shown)


deployments/helm-chart/templates/controller-deployment.yaml
```diff
@@ -10,7 +10,9 @@ metadata:
   annotations: {{ toYaml .Values.controller.annotations | nindent 4 }}
 {{- end }}
 spec:
+  {{- if not .Values.controller.autoscaling.enabled }}
```
lucacome, line 13: This is `1` by default, also not sure it should be connected to autoscaling

deployments/helm-chart/templates/controller-hpa.yaml
```diff
@@ -18,20 +18,20 @@ spec:
   minReplicas: {{ .Values.controller.autoscaling.minReplicas }}
   maxReplicas: {{ .Values.controller.autoscaling.maxReplicas }}
   metrics:
-    {{- if .Values.controller.autoscaling.targetCPUUtilizationPercentage }}
+    {{- if .Values.controller.autoscaling.targetMemoryUtilizationPercentage }}
```
lucacome, line 21: I'm not really sure what difference it makes to swap the two. Do you have a link to any docs about this?

coolbry95: You can see the k8s issue here. https://github.com/kubernetes/kubernetes/issues/74099

The problem is somewhere after the manifest is applied in k8s the order swaps. This causes Argocd to think the manifest is always out of sync. By swapping these Argocd will not think the manifest is out of sync.

lucacome: Yeah I saw that, still seems pretty weird to me 🤷

lucacome (approved):

brianehlert:
This fix follows prior art from here: https://github.com/kubernetes/kubernetes/issues/74099
Further discussed here: https://github.com/argoproj/argo-cd/issues/9145#issuecomment-1507820227
This fix is probably the lowest impact and has the effect of making ArgoCD content.

shaun-nx (approved):

PR #3790, opened 2023-04-18
Gunzip for VS

jjngx:
### Proposed changes

This PR introduces `gunzip` functionality setup on the server (VS).

### Checklist


docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -53,6 +54,7 @@ spec:
 | ---| ---| ---| --- |
 |``host`` | The host (domain name) of the server. Must be a valid subdomain as defined in RFC 1123, such as ``my-app`` or ``hello.example.com``. When using a wildcard domain like ``*.example.com`` the domain must be contained in double quotes.  The ``host`` value needs to be unique among all Ingress and VirtualServer resources. See also [Handling Host and Listener Collisions](/nginx-ingress-controller/configuration/handling-host-and-listener-collisions). | ``string`` | Yes |
 |``tls`` | The TLS termination configuration. | [tls](#virtualservertls) | No |
+|``gunzip`` | Enables or disables decompression of gzipped responses for clients. | ``string`` | No |
```
haywoodsh, line 57: Wouldn't it be better to mention the default and allow values, like other `string` fields in the spec, or maybe also add a link to the nginx documention on the `gunzip` directive?
Something like `next-upstream-timeout` above:
>The time during which a request can be passed to the next upstream server. See the proxy_next_upstream_timeout directive. The 0 value turns off the time limit. The default is 0.

shaun-nx (approved):

haywoodsh (approved):

Issue #3752, opened 2023-04-12
AKS LoadBalancer IP address changed when upgrading from v2 to v3


jannekem:
**Describe the bug**
After upgrading NGINX Ingress Controller my services became unavailable as the LoadBalancer IP address changed on an Azure AKS cluster.

I couldn't find any mention about such a possible side effect from the release notes.

**To Reproduce**
I have reproduced this on two AKS clusters. My services are exposed with `VirtualServers`. After I run `helm upgrade` the VirtualServer `status.externalEndpoints` still shows the old IP address but the service is not reachable anymore. The new IP address can be found from Azure Portal in the "Services and ingresses" page of the AKS cluster instead.

Here are the versions I'm upgrading from/to.

```
$ helm history ingress-controller
REVISION        UPDATED                         STATUS          CHART                   APP VERSION     DESCRIPTION     
5               Fri Nov 18 09:10:41 2022        superseded      nginx-ingress-0.15.1    2.4.1           Upgrade complete
6               Wed Apr 12 09:38:06 2023        deployed        nginx-ingress-0.17.0    3.1.0           Upgrade complete
```

**Expected behavior**
The IP addresses should not change during an upgrade.

**Your environment**
* Version of the Ingress Controller 3.1.0
* Version of Kubernetes 1.25.5
* Kubernetes platform AKS
* Using NGINX

**Additional context**
Helm settings:

```yaml
controller:
  enableSnippets: true
  enableCustomResources: true
  enableCertManager: true
  resources:
    requests:
      cpu: 100m
      memory: 128Mi
    limits:
      cpu: 1000m
      memory: 1Gi

```

haywoodsh:
Hi @jannekem, thank you for reaching out to us. I am not sure that there is a guarantee that the IP address won't change during a helm upgrade. If you have upgraded to version 2.4.1 from an earlier version before, could you please let me know if you have also experienced such a problem?

jannekem:
With the production cluster we've gone through the following versions: 1.8.1 -> 2.0.0 -> 2.0.3 -> 2.3.0 -> 2.3.1. The IP address has stayed the same through all these revisions.

I upgraded the development cluster from 2.3.1 and the test cluster from 2.4.1, and they both had the same behavior. I haven't upgraded the production cluster yet because that would require coordination to update the DNS records if/when the IP changes.

De1taE1even:
[...]

PR #3802, opened 2023-04-20
added serviceNameOverride

timnee:
### Proposed changes

Add a name override to the helm controller-server template.  This can be used to prevent cloud load balancers from being replaced due to service name change during helm upgrades.  This addresses https://github.com/nginxinc/kubernetes-ingress/issues/3752

### Checklist

- [x]  I have checked that all unit tests pass after adding my changes


coolbry95:
I think if a change like this is made it should be for all resources not just one by one. Maybe fullNameOverride should be for all resources and it should not append "-nginx-ingress".

lucacome:
@coolbry95 we already support `fullnameOverride` and `nameOverride`, but there's no way to use it for this use-case

lucacome (approved):
@timnee thanks for the fix, and sorry again for the mishap!

PR #3814, opened 2023-04-25
Remove semverCompare for allocateLoadBalancerNodePorts

centromere:
Kubernetes 1.22.0 is now a minimum requirement for the entire chart.

### Proposed changes

### Checklist


ciarams87 (approved):

jjngx (approved):

deployments/helm-chart/templates/controller-service.yaml
```diff
@@ -20,7 +20,7 @@ spec:
   {{- end }}
 {{- end }}
 {{- if eq .Values.controller.service.type "LoadBalancer" }}
-  {{- if and (semverCompare ">=1.22.0-0" .Capabilities.KubeVersion.Version) (.Values.controller.service.allocateLoadBalancerNodePorts) }}
+  {{- if hasKey .Values.controller.service "allocateLoadBalancerNodePorts" }}
```
lucacome, line 23: what is the difference between using `hasKey` and
```
{{- if .Values.controller.service.allocateLoadBalancerNodePorts }}
```

if it's the same, does it make sense to keep it consistent with all the other `if`s in the file?

centromere: If I specify `allocateLoadBalancerNodePorts: false`, then your code snippet will cause the setting to be elided from the output entirely.

lucacome: Thanks for the explanation @centromere ! This actually fixes a bug then as `allocateLoadBalancerNodePorts` could never be set to false before (and the default value is `true`)

lucacome (approved):

PR #3841, opened 2023-05-03
Added Alpine 3.16 NGINX Plus + App protect WAF image option

fabriziofiorucci:
### Proposed changes
Added Alpine 3.16 NGINX Plus + App protect WAF image option

### Checklist

- [ X] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc
- [ X] I have added tests that prove my fix is effective or that my feature works
- [ X] I have checked that all unit tests pass after adding my changes
- [ X] I have updated necessary documentation
- [ X] I have rebased my branch onto main
- [ X] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork


brianehlert:
I am guessing the @lucacome should be also working through this at the moment to produce images for the 3.1.1 patch release.
It is not unusual for either of the NAP products to have special image requirements at the time of a NIC release.

brianehlert:
We have been discussing this and are not entirely keen on adding yet another build to the existing process.
Especially one that is hard coded and essentially bespoke.

Can you help us understand a bit more of the reasoning and the value to pushing this into upstream?
Since there is no issue tied to this we don't fully understand the goal of the change.

brianehlert:
Is the intention to replace an existing image option?  Or simply add more?
We are trying to avoid the case of 'simply adding more'.  
As the maintenance each release cycle is getting to be pretty time consuming.
It would be far more useful for this project to choose an option that there is commitment to remain current with.

"requests" is plural.  How many?

brianehlert:
I believe this has been resolved a different way.

PR #3844, opened 2023-05-03
Fix gunzip support for VS and add python tests

jjngx:
### Proposed changes

This PR:
- fixes gunzip support for VS (gunzip is of type `bool` instead of `string`)
- updates NIC documentation
- updates examples (`custom-resources/jwt`)
- adds Python integration tests

### Checklist


pkg/apis/configuration/validation/virtualserver_test.go
```diff
@@ -69,215 +69,6 @@ func TestValidateVirtualServer(t *testing.T) {
 	}
 }
```
vepatel, line 71: any specific reason to remove these unit tests other than datatype change?

jjngx: The tests are removed because of datatype change (they are not relevant now). Now we have boolean and it can be either `true` or `false`. Validation of this type is enforced  in the OpenAPI spec (updated CRD for VirtualServer).

examples/custom-resources/jwt/README.md
```diff
@@ -67,3 +67,21 @@ Date: 10/Sep/2020:18:20:03 +0000
...
+
+Example:
+```yaml
+apiVersion: k8s.nginx.org/v1
+kind: VirtualServer
+metadata:
+  name: webapp
+spec:
+  host: webapp.example.com
+  gunzip: on
```
haywoodsh, line 85: Should this be 'true'/'false' now? Same for the *Note* above.
> When the ``gunzip`` value is set to ``on``,

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
```diff
@@ -54,7 +54,7 @@ spec:
 | ---| ---| ---| --- |
 |``host`` | The host (domain name) of the server. Must be a valid subdomain as defined in RFC 1123, such as ``my-app`` or ``hello.example.com``. When using a wildcard domain like ``*.example.com`` the domain must be contained in double quotes.  The ``host`` value needs to be unique among all Ingress and VirtualServer resources. See also [Handling Host and Listener Collisions](/nginx-ingress-controller/configuration/handling-host-and-listener-collisions). | ``string`` | Yes |
 |``tls`` | The TLS termination configuration. | [tls](#virtualservertls) | No |
-|``gunzip`` | Enables or disables [decompression](https://docs.nginx.com/nginx/admin-guide/web-server/compression/) of gzipped responses for clients. Allowed values are: "on" or "off". If the ``gunzip`` value is not set, it defaults to ``off``.   | ``string`` | No |
+|``gunzip`` | Enables or disables [decompression](https://docs.nginx.com/nginx/admin-guide/web-server/compression/) of gzipped responses for clients. Allowed values are: "on" or "off". If the ``gunzip`` value is not set, it defaults to ``off``.   | ``boolean`` | No |
```
haywoodsh, line 57: The allowed and default values here should probably be changed too.

ciarams87, line 57: If this is a bool, allowed values are `“true”/“false”, “yes”/“no” or “on”/“off”` AFAIK

examples/custom-resources/jwt/README.md
jjngx: `on` has the same meaning as `true` when parsing yaml. [docs](https://nginx.org/en/docs/http/ngx_http_gunzip_module.html#gunzip) refers to values `on` / `off` for the gunzip directive. So, kept the same for NIC naming convention.

docs/content/configuration/virtualserver-and-virtualserverroute-resources.md
jjngx: The question is if we keep only `on`/`off` to be consistent with [N docs](https://nginx.org/en/docs/http/ngx_http_gunzip_module.html), and [this one](https://docs.nginx.com/nginx/admin-guide/web-server/compression/). From user's perspective it will be either `gunzip: on` in VS or no gunzip field.

ciarams87: I understand that, but the openAPI validation will still allow any of the allowed bool values. Sticking to "on"/"off" in the examples is definitely the right thing, but stating `Allowed values are: "on" or "off".` is technically not true here.

jjngx: Updated docs and pushed the change

ciarams87 (approved):

Issue #3893, opened 2023-05-12
Ingress enable health check but no use.


chengxus:
**Describe the bug**
Ingress enables health checks, and the profile URL is sometimes updated when the pod probe URL is updated, but never reloaded.

**To Reproduce**
The file used in the current repository 1.12.4 version.
1. apply  examples/complete-example/cafe.yaml
2. apply examples/complete-example/cafe-secret.yaml
3. edit examples/complete-example/cafe-ingress.yaml add annotations 'nginx.com/health-checks: "true"' and apply
4. edit examples/complete-example/cafe.yaml add probe '
       readinessProbe:
          httpGet:
            port: 8080
            path: /healthz/tea
          initialDelaySeconds: 1
          periodSeconds: 5
          timeoutSeconds: 4
          successThreshold: 2
          failureThreshold: 3 '
5. get /api/6/http/upstreams not found health check.

**Expected behavior**
I expect to update the nginx config file and reload when I modify the probe。

* Version of the Ingress Controller：1.12.4
* Version of Kubernetes ：v1.17.4
* Kubernetes platform Mini-kube 
* Using NGINX Plus

I've fixed this。
cache pod probe config，and reload nginx when probe update。

brianehlert:
Thanks for bringing this to our attention @chengxus we are looking into this.

PR #3894, opened 2023-05-12
fix Ingress enable health check but no use

chengxus:
Ingress enables health checks, and the profile URL is sometimes updated when the pod probe URL is updated, but never reloaded.

### Proposed changes
fix issues https://github.com/nginxinc/kubernetes-ingress/issues/3893

cache pods probe config ，and compare with old when endpoints update,if updated reload nginx

### DIFF
--- a/internal/configs/configurator.go
+++ b/internal/configs/configurator.go
@@ -8,6 +8,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"os"
+	"reflect"
 	"strings"
 
 	"github.com/nginxinc/kubernetes-ingress/internal/k8s/secrets"
@@ -100,6 +101,7 @@ type Configurator struct {
 	cfgParams               *ConfigParams
 	templateExecutor        *version1.TemplateExecutor
 	templateExecutorV2      *version2.TemplateExecutor
+	HealthChecks            map[string]*api_v1.Probe
 	ingresses               map[string]*IngressEx
 	minions                 map[string]map[string]bool
 	virtualServers          map[string]*VirtualServerEx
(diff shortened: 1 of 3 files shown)


brianehlert:
Can I ask why you are providing a fix for a release that is over 1 year old?

chengxus:
> Can I ask why you are providing a fix for a release that is over 1 year old?
> 
> Does your fix still apply to the current release? https://github.com/nginxinc/kubernetes-ingress/tree/v3.1.1/examples/ingress-resources/health-checks
> 
> Does this behavior also manifest with the VirtualServer and TransportServer objects? (I would hate to see a fix for one resource that does not include the others when they all support the same capabilities).

Because I'm using version 1.12.4 .
current release the bug  still exists,but current release can’t automatically merge. 
VirtualServer and TransportServer objects used another way to not have the problem.

internal/configs/ingress.go
```diff
@@ -230,7 +232,9 @@ func generateNginxCfg(ingEx *IngressEx, apResources AppProtectResources, isMinio
 
 			if cfgParams.HealthCheckEnabled {
 				if hc, exists := ingEx.HealthChecks[ingEx.Ingress.Spec.Backend.ServiceName+ingEx.Ingress.Spec.Backend.ServicePort.String()]; exists {
```
jjngx, line 234: How glanceability could be improved here? How about using a shorter key?

internal/configs/configurator.go
```diff
@@ -785,9 +788,17 @@ func (cnf *Configurator) deleteTransportServer(key string) error {
 // UpdateEndpoints updates endpoints in NGINX configuration for the Ingress resources.
 func (cnf *Configurator) UpdateEndpoints(ingExes []*IngressEx) error {
 	reloadPlus := false
+	forceReload := false
 
 	for _, ingEx := range ingExes {
 		// It is safe to ignore warnings here as no new warnings should appear when updating Endpoints for Ingresses
+		for k, v := range ingEx.HealthChecks {
+			data, exist := cnf.HealthChecks[ingEx.Ingress.Namespace+"-"+k]
```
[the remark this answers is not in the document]
chengxus: "data" is deployment config upstream health check. you can change it to a better name.

internal/configs/ingress.go
chengxus: service and port can identify a upstream

PR #3926, opened 2023-05-19
Remove explicit `fsGroup` definition

sigv:
### Proposed changes

There is no use-case that is relevant for enforcing the `fsGroup` as the default user's group. Our `nginx` user will already have access to the volumes being created, as they are made with `g+rwx`.

### Checklist


lucacome (approved):
Makes sense, we also set the group to `0` (root) for our files, so we don't need to enforce `101`.

Thanks @sigv !

jjngx (approved):

PR #4004, opened 2023-06-14
Update labeler to official action

lucacome:
### Proposed changes
Changes:
- Updates the version of the labeler to the official actions/labeler
- Updates the config to the new format
- Adds missing `change` label to the config
- Adds PRs from pre-commit to the`dependencies` label
- Adds new logic (now possible with the new format) that prevents from adding the `test` label in dependency updates

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -2,7 +2,7 @@ name: "Pull Request Labeler"
 on:
   - pull_request_target
 
-permissions:  # added using https://github.com/step-security/secure-workflows
+permissions:
   contents: read
 
 jobs:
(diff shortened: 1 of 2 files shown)


jjngx (approved):

.github/workflows/labeler.yml
```diff
@@ -2,16 +2,16 @@ name: "Pull Request Labeler"
...
+permissions:
   contents: read
 
 jobs:
   triage:
     permissions:
       contents: read
       pull-requests: write
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
shaun-nx, line 13: Why are we using a specific version here and not just keeping at at latest? Do we risk forgetting to update this or will it be automatically updated with dependabot?

.github/workflows/labeler.yml
```diff
@@ -2,16 +2,16 @@ name: "Pull Request Labeler"
...
     steps:
-    - uses: joshdales/labeler@7b1327b4c44a8794dfc7573d60637cd60ce4b697 # if https://github.com/actions/labeler/pull/203 is merged, use the official action actions/labeler
-      with:
-        repo-token: "${{ secrets.GITHUB_TOKEN }}"
+      - uses: actions/labeler@4f052778de9a9b80cb16cfb9079b02287285a4cb # v5.0.0-alpha.1
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
```
shaun-nx, line 17: These three lines are indented more than they were. Is this okay?

.github/workflows/labeler.yml
lucacome: We prefer to use specific versions to avoid ambiguity. We have `ubuntu-22.04` in all the other workflows and this change is to align all of them.

Dependabot won't update this, we'll have to do it manually but (historically) they only use Ubuntu LTS versions so the update will be every couple of years

.github/workflows/labeler.yml
lucacome: yeah the file is still valid

PR #4031, opened 2023-06-20
Add Alpine image with FIPS inside

lucacome:
### Proposed changes
Adds a new image with FIPS module and necessary configuration. The nginx-plus-module-fips-check was added to the base images. This lightweight module adds a line to the nginx output log printing whether FIPS mode is enabled or not.

### DIFF
--- a/internal/configs/version1/nginx-plus.tmpl
+++ b/internal/configs/version1/nginx-plus.tmpl
@@ -21,6 +21,7 @@ load_module modules/ngx_http_app_protect_module.so;
 {{- if .AppProtectDosLoadModule}}
 load_module modules/ngx_http_app_protect_dos_module.so;
 {{- end}}
+load_module modules/ngx_fips_check_module.so;
 {{- if .MainSnippets}}
 {{range $value := .MainSnippets}}
 {{$value}}{{end}}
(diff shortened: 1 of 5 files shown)


internal/configs/version1/nginx-plus.tmpl
```diff
@@ -21,6 +21,7 @@ load_module modules/ngx_http_app_protect_module.so;
 {{- if .AppProtectDosLoadModule}}
 load_module modules/ngx_http_app_protect_dos_module.so;
 {{- end}}
+load_module modules/ngx_fips_check_module.so;
```
vepatel, line 24: does this adds fips to all the images?

vepatel:
I think we should add a test-run job for alpine-plus-fips image as well
`{\"image\": \"alpine-plus-fips\", \"marker\":\"policies\"}, \`

internal/configs/version1/nginx-plus.tmpl
lucacome: This module only prints the line in the logs saying if FIPS is enabled or not. I've added it to all the plus images (it's only a few KBs)

shaun-nx: That lines up with what @vepatel showed us in the logs from his testing. Thanks for confirming.

shaun-nx (approved):

PR #3977, opened 2023-06-08
Add support for `controller.selectorLabels`

hanyouqing:
### Proposed changes
Fix issue https://github.com/nginxinc/kubernetes-ingress/issues/3976

### Checklist
Before creating a PR, run through this checklist and mark each as complete.

- [x] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that all unit tests pass after adding my changes
- [ ] I have updated necessary documentation
- [x] I have rebased my branch onto main
- [x] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

Issue #3976, opened 2023-06-08
Introduced changes to the immutable MatchLabels to the daemonset in #3606


hanyouqing:
**Describe the bug**
The immutable MatchLabels for DaemonSet was changed in #3606 with no override values option defined in the templates which will cause unexpected downtime for production, cause we have to upgrade by deleting and reinstalling.

```bash
cannot patch "nginx-ingress-nginx-ingress" with kind DaemonSet: DaemonSet.apps "nginx-ingress-nginx-ingress" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"nginx-ingress", "app.kubernetes.io/name":"nginx-ingress"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

**To Reproduce**
Upgrade ingress-nginx 3.0.2 to 3.1.1 

**Expected behavior**
Upgrade ingress-nginx (or make some changes in vaules.yaml in advance) to the target version successfully. 

**Your environment**
* Version of the Ingress Controller: `v3.0.2`
* Version of Kubernetes: `v1.24.10`
* Kubernetes platform (e.g. Mini-kube or GCP): 
* Using NGINX or NGINX Plus: `NGINX Plus`

**Additional context**
Revert this [changes](https://github.com/nginxinc/kubernetes-ingress/pull/3606/files#diff-2eccfbc1886b3803b5035aad15dc3f9f38ecfd0b5c5a540408c828f32d6ab3d3R12-R15) or keep it be compatible with the previous version would be better. FYI @lucacome

vepatel:
Hi @hanyouqing thanks for creating this issue, can you please give us the steps to reproduce the error in the description. We did an upgrade from 3.0.2 to 3.1.1 using `helm upgrade...` and didn't run into it. Please see describe outputs from 3.0.2 and 3.1.1 below.

```
~/nginx/kubernetes-ingress/deployments/helm-chart on tags/v3.0.2 λ k describe daemonsets.apps 
Name:           nginx-ingress-nginx-ingress
Selector:       app=nginx-ingress-nginx-ingress
Node-Selector:  <none>
Labels:         app.kubernetes.io/instance=nginx-ingress
                app.kubernetes.io/managed-by=Helm
                app.kubernetes.io/name=nginx-ingress-nginx-ingress
                helm.sh/chart=nginx-ingress-0.16.2


[...]

PR #3933, opened 2023-05-23
fix --enternal-service flag when using serviceNameOverride

timnee:
### Proposed changes

Add a template helper for nginx-ingress.controller.service.name to fix the --external-service flag on deployement/daemonset.  This fixes a bug introduced in PR https://github.com/nginxinc/kubernetes-ingress/pull/3802 which resulted in ingress status to be empty

### Checklist
Before creating a PR, run through this checklist and mark each as complete.

- [x] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that all unit tests pass after adding my changes
- [x] I have updated necessary documentation
- [x] I have rebased my branch onto main
- [x] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

PR #4032, opened 2023-06-20
Make fullnameOverride override

coolbry95:
Before 3.0.0 all resources were named `nginx-ingress`.  We need to keep that functionality. 

The fullnameOverride currently will use the fullNameOverride plus nginx-ingress or whatever .Values.controller.name is set to. The final result looks like `nginx-ingress-something`. The fullNameOverride should be the final name set and should not be modified after.

Since the deployment or daemonset matchLabels cannot be updated there needs to be a way to revert to the old labels. Adding selectorLabelsOverride allows a user to use the old labels and not force a recreation of the deployment/daemonset.

### Proposed changes

### Checklist

### DIFF
--- a/deployments/helm-chart/templates/_helpers.tpl
+++ b/deployments/helm-chart/templates/_helpers.tpl
@@ -68,8 +68,12 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 Selector labels
 */}}
 {{- define "nginx-ingress.selectorLabels" -}}
+{{- if .Values.selectorLabelsOverride }}
+{{- .Values.selectorLabelsOverride | toYaml }}
+{{- else }}
 app.kubernetes.io/name: {{ include "nginx-ingress.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end}}
 {{- end }}
 
 {{/*
--- a/deployments/helm-chart/templates/_helpers.tpl
+++ b/deployments/helm-chart/templates/_helpers.tpl
@@ -30,7 +30,11 @@ Create a default fully qualified controller name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
 {{- define "nginx-ingress.controller.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride -}}
+{{- else }}
 {{- printf "%s-%s" (include "nginx-ingress.fullname" .) .Values.controller.name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
 {{- end -}}
 
 {{/*


brianehlert:
@coolbry95 
Can you please describe the problem this is solving?  Just to make it easy for the reviewers to really understand the problem this is fixing.

brianehlert:
Is this an improvement on: https://github.com/nginxinc/kubernetes-ingress/pull/3802
or
https://github.com/nginxinc/kubernetes-ingress/pull/3933

coolbry95:
Updated the PR message. It is in similar vein to #3802.

This is a blocker for us to upgrade to 3.x.x because of the resource name changes.

I also believe the current behavior is a bug when considering the name `fullnameOverride` because its not actually overriding when its being modified still.

lucacome:
`fullnameOverride ` is overriding `nginx-ingress.fullname` and the `fullname` is used in the controller name. It's not a bug.

Why is it preventing you from upgrading?

coolbry95:
> Why is it preventing you from upgrading?

Because it is changing the names of resources which is unacceptable for my team and application. 

We are unable to preserve previous behavior. We would like all resources to be nginx-ingress.

There is also the option of introducing another override variable. Is that more acceptable than this? We could introduce an override variable that trumps all others.

coolbry95:
> > fullnameOverride  is overriding nginx-ingress.fullname and the fullname is used in the controller name.
> 
> @coolbry95 is it that you want to be able to have `fullname `be an empty string? And just have `"nginx-ingress"`

Yes but that is still wrong. No matter what fullname or controller.name are set to you will always have '-' somewhere in the mix. 

We would like to just be able to control the names ourselves.

vepatel:
hi @coolbry95 we're looking at the issue and noticed doesn't covers daemon-set so we need to check if there's something else in the need of similar change.

vepatel:
thanks for clarification @coolbry95 , we're also looking at https://github.com/nginxinc/kubernetes-ingress/issues/3976 as the issue seems related and just need to make sure changes don't conflict with each other. Thanks for your patience!

coolbry95:
I don't see an issue.

Selector labels uses the `nginx-ingress.name`
```
{{- define "nginx-ingress.name" -}}
{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
{{- end }}
```
Selector label
`app.kubernetes.io/name: {{ include "nginx-ingress.name" . }}`

This change affects the `nginx-ingress.controller.fullname` only. It doesn't look like this is used for any labels just names of manifests.

coolbry95:
I think this could be solved by doing instead of the if statement. This would match what `nginx-ingress.controller.service.name` is doing.

```
{{- default (printf "%s-%s" (include "nginx-ingress.fullname" .) .Values.controller.name) .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
```

vepatel:
I tried changes from this PR and one below by editing `controller-daemonset.yaml` separately:
`  name: {{ default (include "nginx-ingress.controller.fullname" .) .Values.fullnameOverride }}` 
but both resulted in failed upgrade:
```
Error: UPGRADE FAILED: cannot patch "test-release-nginx-ingress" with kind DaemonSet: DaemonSet.apps "test-release-nginx-ingress" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"test-release", "app.kubernetes.io/name":"nginx-ingress"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

coolbry95:
> I tried changes from this PR and one below by editing `controller-daemonset.yaml` separately: ` name: {{ default (include "nginx-ingress.controller.fullname" .) .Values.fullnameOverride }}` but both resulted in failed upgrade:
> 
> ```
> Error: UPGRADE FAILED: cannot patch "test-release-nginx-ingress" with kind DaemonSet: DaemonSet.apps "test-release-nginx-ingress" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"test-release", "app.kubernetes.io/name":"nginx-ingress"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
> ```

Your error is still unrelated to this PR. It looks like you are still hitting the issue in #3976.

Not sure what the exact steps you are taking. If you change the name of the daemonset it will create a new one. The workflow here for us is to keep all the names of all resources the same. In #3976 they are trying to keep the selector labels the same so the daemonset doesn't need to be recreated. The only way to change the daemonset labels is to change the `.Release.Name` or the `nginx-ingress.name`. 

Before the labels change the daemonset would have the label `app: {{ include "nginx-ingress.appName" . }}` and after would be
```
app.kubernetes.io/name: {{ include "nginx-ingress.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
```
This is a breaking change. There is no way to not recreate the daemonset without an override to the selector labels used.

vepatel:
Pointed that out because the changes in PR weren't working because of the other issue and it seems appropriate to fix these together, now that PR is updated with selector changes we can try both both overrides.

vepatel:
@coolbry95 is there anything else I need to add, as for me even after these changes deployment is being re-created:
```
helm upgrade test-release --set controller.kind=deployment --set controller.nginxplus=false --set controller.image.pullPolicy=Always --set controller.fullnameOverride=true .

Release "test-release" has been upgraded. Happy Helming!
NAME: test-release
LAST DEPLOYED: Mon Jul 24 13:53:18 2023
NAMESPACE: default
STATUS: deployed
REVISION: 2
TEST SUITE: None
NOTES:
The NGINX Ingress Controller has been installed.

 λ k get po
NAME                                                     READY   STATUS    RESTARTS   AGE
test-release-nginx-ingress-controller-6b6dc55cd4-wbvmv   1/1     Running   0          5s

 λ k describe deployments.apps 
Name:                   test-release-nginx-ingress-controller
Namespace:              default
CreationTimestamp:      Mon, 24 Jul 2023 13:53:19 +0100
Labels:                 app=test-release-nginx-ingress

```

vepatel:
Hi @coolbry95, I think we should be able to fix the issue without the changes in this PR, we've accepted https://github.com/nginxinc/kubernetes-ingress/pull/3977 to fix the issue with selectorLabels. Can you please try:

1. taking the changes from above pr,
2. perform a helm upgrade with `--set fullnameOverride=nginx-ingress-nginx-ingress --set controller.name=""` if your release name is `nginx-ingress`
3. perform a helm upgrade with `--set controller.name=""` if your release name is anything other than `nginx-ingress`

coolbry95:
> Hi @coolbry95, I think we should be able to fix the issue without the changes in this PR, we've accepted #3977 to fix the issue with selectorLabels. Can you please try:
> 
> 1. taking the changes from [Add support for `controller.selectorLabels` #3977](https://github.com/nginxinc/kubernetes-ingress/pull/3977),
> 2. perform a helm upgrade with `--set controller.serviceNameOverride="<old_service_name>" --set fullnameOverride="<old_deployment_name>" --set controller.name=""` if your helm release name is `nginx-ingress`
> 3. perform a helm upgrade with `--set controller.serviceNameOverride="<old_service_name>" --set controller.name=""` if your helm release name is anything other than `nginx-ingress`

This flow works. I would still prefer fullNameOverride to work how it does in this PR. I think it is confusing that overriding the name doesn't actually override it.

PR #4053, opened 2023-06-23
Bump alpine-fips to 0.1.1-alpine3.18

lucacome:
### Proposed changes
Bumps alpine-fips to the latest version. Also moves the image to FROM so dependabot can automatically update it.


jasonwilliams14 (approved):
very nice!

PR #4054, opened 2023-06-23
Make NAP independent from Plus

lucacome:
### Proposed changes
NAP doesn't support `debian 12`, this creates a separate image using `debian 11` instead of using `debian-plus` as base


jasonwilliams14 (approved):
Looks good

PR #4093, opened 2023-07-12
Add explicit note to that docs apply to latest release

ADubhlaoich:
### Proposed changes
These commits add a small note near the prerequisites part of a few installation-based sections of the documentation to draw the reader's attention towards the fact that documentation applies only to the latest release for NGINX Ingress Controller.

### Checklist


vepatel (approved):

brianehlert:
Is this to add a disclaimer to main as opposed to making folks aware that the URL is release sensitive?

I ask because it reads like a disclaimer, rather than noting that the URL is release version sensitive.  <- which I think is a more instructive way to address the problem.

I assume that we have some feedback that main is being used to reference documents while using some older release.  You can DM me any details of the reasoning.

lucacome:
we also don't have a way to point people to a different version of the docs if they're not using the latest release, so this might be confusing.

PR #4137, opened 2023-07-24
Added if to hostport on hostnetwork config value

EutiziStefano:
### Proposed changes
The chart inserts "HostPort" in the daemonset also when "HostNetwork" is setted to "false", do not set hostPort in this case.

Setting the controller.hostNetwork: false should not set hostPort in the deployment or daemonset.
This can be fixed in helm chart, inserting in templates/controller-daemonset.yaml a if statement around the two host port.

Thanks
Stefano


lucacome:
@EutiziStefano `hostNetwork` and `hostPort` are independent of each other. I've opened #4252 to make `hostPort` customizable

PR #4149, opened 2023-07-25
Upgrade nginx-hugo-theme version to support newer Hugo releases

jputrino:
### Proposed changes

This PR updates the nginx-hugo-theme module to v0.34.0, which adds support for Hugo v0.115.3 (and subsequent newer releases). This upgrade is necessary to address security vulnerabilities in the earlier pinned version of Hugo.

- The docs/README has been updated to match that used across the NGINX projects where we develop docs.
- The docs/makefile and netlify configs have been updated to include the deploy-preview make command, which is needed to build docs for the deploy-preview and branch contexts in Netlify. These settings are also now in parity with the configurations used across the NGINX documentation projects.

Related PR: https://github.com/nginxinc/nginx-hugo-theme/pull/16

Once this PR is merged, the environment variables that are stored in Netlify for the nginx-hugo-theme version and hugo version will be updated accordingly.

### Checklist


docs/README.md
```diff
@@ -1,24 +1,136 @@
...
+make hugo-mod       -   cleans the Hugo module cache and fetches the latest version of the theme module
+make docs-drafts    -   runs the local hugo server and includes all docs marked with `draft: true`
+```
+
+## Linting
+
+- To run the markdownlint check, run the following command from the docs directory:
+
+    ```bash
+    markdownlint -c docs/mdlint_conf.json content
```
lucacome, line 35: do we need to add this file `docs/mdlint_conf.json` ?

lucacome: we just recently added `.markdownlint-cli2.yaml` to the root of the project for all other markdown files, but I'm not sure if we want to apply the same rules here for the docs

jputrino: I'll take a look, thanks for catching that!

jputrino: @lucacome I'm taking this out of the README and we will address it in a separate PR. It's a much larger can of worms than I care to bring into the context of this PR.

jputrino: We don't want to use the main markdownlint config file for the docs, but we need to standardize on the rules across all of the repos before going further.

lucacome (approved):

PR #4184, opened 2023-08-07
add helm upgrade document

vepatel:
### Proposed changes

- Add documentation for no-downtime helm upgrade

Preview: https://deploy-preview-4184--nginx-kubernetes-ingress.netlify.app/nginx-ingress-controller/installation/upgrade-with-helm/

### Checklist


ADubhlaoich (approved):
Approved, though I pushed some changes directly that should be reviewed.

- Ensured that product and command nouns were consistent
- Added the note callout and tab components to the upgrade steps section
- Reduced and rephrased text for conciseness and clarity
- Changed inline comments to single backticks instead of triples

docs/content/installation/upgrade-with-helm.md
```diff
@@ -0,0 +1,102 @@
...
+doctypes: [""]
+toc: true
+docs: "DOCS-603"
+---
+
+## Background
+
+In NGINX Ingress Controller version 3.1.0, [changes were introduced](https://github.com/nginxinc/kubernetes-ingress/pull/3606) to Helm resource names, labels and annotations to fit with Helm best practices.
+When using Helm to upgrade from a version prior to to 3.1.0, certain resources like Deployment, DaemonSet and Service will be recreated due to the aforementioned changes, which will result in downtime.
+Although the advisory is to update all resources in accordance with new naming convention, to avoid the downtime please follow the steps listed in this page.
```
ADubhlaoich, line 14:
```suggestion
When using Helm to upgrade from a version prior to to 3.1.0, certain resources like Deployment, DaemonSet and Service will be recreated due to the aforementioned changes, which will result in downtime.

Although the advisory is to update all resources in accordance with new naming convention, to avoid the downtime please follow the steps listed in this page.
```

Really small change, but improves the readability by breaking up what would otherwise be a chunky paragraph of text into two separate texts - also draws more attention to the last line.

docs/content/installation/upgrade-with-helm.md
```diff
@@ -0,0 +1,107 @@
...
+description: "This document describes how to use Helm to upgrade NGINX Ingress Controller from 2.x and 3.0.x to the latest available release."
+weight: 1800
+doctypes: [""]
+toc: true
+docs: "DOCS-603"
+---
+
+## Background
+
+In NGINX Ingress Controller version 3.1.0, [changes were introduced](https://github.com/nginxinc/kubernetes-ingress/pull/3606) to Helm resource names, labels and annotations to fit with Helm best practices.
```
jasonwilliams14, line 12: Should we make this document part of our `helm upgrade` section in our existing doc?

https://docs.nginx.com/nginx-ingress-controller/installation/installation-with-helm/#upgrading-the-chart-1

docs/content/installation/upgrade-with-helm.md
```diff
@@ -0,0 +1,107 @@
...
+    ```shell
+    selectorLabels: {app: nginx-ingress-nginx-ingress}
+    ```
+
+1. Run `helm upgrade` with following arguments set:
+    ```shell
+    --set controller.serviceNameOverride="nginx-ingress-nginx-ingress"
+    --set controller.name=""
+    --set fullnameOverride="nginx-ingress-nginx-ingress"
+    ```
```
jasonwilliams14, line 49: Should we also provide an example of how to do a upgrade to point to a `values.yaml` file as well?
They can make the changes in their `values.yaml` file and upgrade with OCI method:
Something like:

```
helm upgrade <nic_release_name> oci://ghcr.io/nginxinc/charts/nginx-ingress--version 0.18 -n nginx-ingress  -f values.yaml
```
Thoughts? 🤔

jasonwilliams14 (commented):
Leave some comments

docs/content/installation/upgrade-with-helm.md
vepatel: good shout, I can link this page from there. Having it separate makes it more visible imo.

docs/content/installation/upgrade-with-helm.md
vepatel: for OCI method, all the user has to do is `repository: "oci://ghcr.io/nginxinc/charts/nginx-ingress"` in values.yaml or `--set controller.repository="repository: "oci://ghcr.io/nginxinc/charts/nginx-ingress" ` which isn't upgrade specific and applies to both installation and upgrade.

Pointing to values is separate from OCI so can defiantly add it as modifying values.yaml is integral part for this.

haywoodsh (approved):

docs/content/installation/upgrade-with-helm.md
lucacome: I agree that this document should be just a note in the upgrade section we already have. Or this should be a more general document on how to upgrade if we really want to have a separate one.

docs/content/installation/upgrade-with-helm.md
```diff
@@ -0,0 +1,107 @@
...
+    It could look as follows:
+
+    ```shell
+    helm upgrade nginx-ingress --set controller.kind=deployment --set controller.nginxplus=false --set controller.image.pullPolicy=Always --set controller.serviceNameOverride="nginx-ingress-nginx-ingress" --set controller.name="" --set fullnameOverride="nginx-ingress-nginx-ingress"
```
lucacome, line 53:
```suggestion
    helm upgrade nginx-ingress oci://ghcr.io/nginxinc/charts/nginx-ingress--version 0.18 --set controller.kind=deployment --set controller.nginxplus=false --set controller.image.pullPolicy=Always --set controller.serviceNameOverride="nginx-ingress-nginx-ingress" --set controller.name="" --set fullnameOverride="nginx-ingress-nginx-ingress"
```

This command uses the old repository instead of OC registry. The other document uses OCI, so we should be consistent.
And we should probably keep in general without `--set controller.kind=deployment`?

docs/content/installation/upgrade-with-helm.md
lucacome: It's not NIC repository, it's the repository for the Helm Chart. It's a separate argument.

lucacome (changes requested):

docs/content/installation/upgrade-with-helm.md
vepatel: gotcha I guess we don't need it then, user can use whatever desired to keep it more generalized

PR #4214, opened 2023-08-11
Add docs for path-regex annotation for mergeable ingresses

jjngx:
### Proposed changes

This PR updates docs and add examples for `path-regex` annotation feature for Mergeable Ingress.

### Checklist


examples/ingress-resources/path-regex/README.md
```diff
@@ -41,7 +41,9 @@ as a case-insensitive regular expression match
 - **=** : Equal sign modifier (annotation value ``exact``) - the location is considered a match if the request
  URI exactly matches the location provided.
 
-## Example 1: Case Sensitive RegEx
+## Path-Regex Annotation in Ingress Types
```
vepatel, line 44:
```suggestion
## Path-Regex Annotation in Simple Ingress Type
```

docs/content/configuration/ingress-resources/advanced-configuration-with-annotations.md
```diff
@@ -112,7 +112,7 @@ The table below summarizes the available annotations.
...
-|``nginx.org/path-regex`` | N/A | Enables regular expression modifiers for [location](https://nginx.org/en/docs/http/ngx_http_core_module.html#location) directive. You can specify one of these values: "case_sensitive", "case_insensitive" or "exact". The annotation is applied to the entire Ingress resource and its paths. |  |  [Path Regex](https://github.com/nginxinc/kubernetes-ingress/tree/examples/ingress-resources/path-regex). |
+|``nginx.org/path-regex`` | N/A | Enables regular expression modifiers for [location](https://nginx.org/en/docs/http/ngx_http_core_module.html#location) directive. You can specify one of these values: "case_sensitive", "case_insensitive", or "exact". The annotation is applied to the entire Ingress resource and its paths. If you use a Master and Minion Ingresses (Mergeable Ingresses) you can specify different regex annotations values on all minions or a subset of minions and/or a master. |  |  [Path Regex](https://github.com/nginxinc/kubernetes-ingress/tree/examples/ingress-resources/path-regex). |
```
vepatel, line 115:
```suggestion
|``nginx.org/path-regex`` | N/A | Enables regular expression modifiers for [location](https://nginx.org/en/docs/http/ngx_http_core_module.html#location) directive. You can specify one of these values: "case_sensitive", "case_insensitive", or "exact". The annotation is applied to the entire Ingress resource and its paths. While using Master and Minion Ingresses i.e. Mergeable Ingresses, this annotation can be specified on both types at once but value from Minion takes precedence. |  |  [Path Regex](https://github.com/nginxinc/kubernetes-ingress/tree/examples/ingress-resources/path-regex). |
```

examples/ingress-resources/path-regex/README.md
```diff
@@ -189,3 +191,305 @@ Corresponding NGINX config file snippet:
...
+
+Content of `coffee-minion.yaml`:
+
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: cafe-ingress-coffee-minion
+  annotations:
+    nginx.org/mergeable-ingress-type: "minion"
```
brianehlert, line 232: I see that path-regex is set at the master but not at the minion.
Therefore I would not expect the minion paths to support regex, only paths defined in the master.
I would expect this path to cause an error.

brianehlert: Each ingress is a configuration to itself.  They just get flattened into a single larger server block configuration.

vepatel: @brianehlert we don't support that particular use-case by design, where one have have a master with its own paths on top of minions so imo this behaviour of inheriting master annotation makes sense.
If one tries to add paths to master:
`spec.rules[0].http.paths: Too many: 1: must have at most 0 items`
`Ingress master is invalid or doesn't exist`

examples/ingress-resources/path-regex/README.md
```diff
@@ -189,3 +191,305 @@ Corresponding NGINX config file snippet:
...
+spec:
+  ingressClassName: nginx
+  tls:
+  - hosts:
+    - cafe.example.com
+    secretName: cafe-secret
+  rules:
+  - host: cafe.example.com
+```
+
```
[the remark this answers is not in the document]
vepatel: same as above

vepatel (approved):

shaun-nx (approved):

examples/ingress-resources/path-regex/README.md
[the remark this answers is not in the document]
jjngx: updated

examples/ingress-resources/path-regex/README.md
jjngx: fixed

PR #4269, opened 2023-08-18
Ensure `/var/log/nginx` is writeable by `101:0`

sigv:
### Proposed changes

In a standard deployment, error log is written to `/dev/stderr` and access log is written to `/dev/stdout`. Furthermore, `error.log` and `access.log` in `/var/log/nginx` are mapped to the respective stdio. However, a deployment may override configuration, and remove the symbolic links, to write to the container storage directly.

OpenShift tries to impose various restrictions by default. One of these is for UID/GID used by the container process. If these restrictions are supported in future, adjustments to file system permissions need to be done so that /var/log/nginx remains writeable. Specifically, OpenShift adds GID 0 as supplemental to container process for file system operations.

This PR ensures the nginx user (UID `101`) and root group (GID `0`) owns the log directory, and that owner group permissions match the owner user permissions (`g=u`). This ensures that OpenShift deployments retain write permissions in future.

In particular, this change is a pre-requisite of PR #3665 (enabling randomized UID/GID for issue #3544), in order to not break the edge-case of writing to disk for OpenShift users.

### Checklist


vepatel:
Hi @sigv can you please open an issue for this PR: https://github.com/nginxinc/kubernetes-ingress/blob/main/CONTRIBUTING.md#open-a-pull-request thanks!

sigv:
@vepatel, opened and referenced #4279.

sigv:
@vepatel, would it be possible to re-surface this fairly small patch that was separated out of #3665 scope?

lucacome:
@sigv since we don't write to `/var/log/nginx` and we don't provide a mechanism to easily change this right now, I'm not sure we should make this change.
It would be very hard to cover all the use cases of users making custom changes to their deployments...

sigv:
@lucacome

One part of why I opened this is for backwards compatibility. As I remember, there were some customers of NIC that had their own logrotate solutions set up in containers. If #3665 is implemented without this PR, such scenarios would be broken in OpenShift.

Other part is as precaution. I am an OSS user, and I am not sure if DOS variant also wants to write to the logs directory (similar to how WAF variant is previously ensured to have rights).

Feel free to close this as `WONTDO` from your side if these are not concerns for the project.

brianehlert:
> One part of why I opened this is for backwards compatibility. As I remember, there were some customers of NIC that had their own logrotate solutions set up in containers. If #3665 is implemented without this PR, such scenarios would be broken in OpenShift.
> 
There are customers who insist on writing logs.  Regardless of the logrotate situation.  This is generally done for some 3rd party agent that is added to the image or as a sidecar.  Think classic Splunk agent for NGINX type of situation.

lucacome (approved):
Sorry for the long wait again @sigv 

I guess it doesn't cost us anything to set the permissions for `/var/log/nginx` and be on the safe side.

sigv:
> Sorry for the long wait again

No worries! I fully understand things sometimes take time 🙂 

0c8462d8dd244cba00ac872a34888eb751f869dd shuffled lines around and there's now a merge conflict to resolve.
I will squeeze this in around end of day (Europe), so the blocker doesn't stay there over the weekend. 🤞🏻

sigv:
Tests are failing as they want to re-build the Nginx Plus components for which my repo doesn't have the certificate. 😞

shaun-nx (approved):
Thanks again for being patient with us @sigv !

sigv:
I'm not sure how the Alpine error could be triggered by this change (`Error getting Endpoints for Upstream backend1: error retrieving endpoints for the service backend1-svc: no endpointslices for target port 8080 in service backend1-svc`)

shaun-nx:
@sigv That particular log message happens when we're enumerating pods before they are ready. From the logs its looks like the pods for that test took too long to become ready.

shaun-nx:
@sigv I've re-run the failed test and everything is green now. Just waiting on images to be built and we should be good.

shaun-nx:
🚀

PR #4276, opened 2023-08-21
Use Lease for leader election

lucacome:
### Proposed changes

This PR removes the need for a `configmap` for leader election by using a `lease`. It also creates a `Role` file to grant the necessary permissions.


shaun-nx (approved):
Did a local test of this and it worked perfectly.

Not only that, the IC acquired the lease for leader election right at the start. This normally took about 15-20 seconds after the IC started.

Great stuff Luca!

shaun-nx:
@lucacome can you create and attach an issue for this PR as well? Thanks!

jjngx (approved):

PR #4278, opened 2023-08-22
Generate mainfests automatically from Helm

lucacome:
### Proposed changes

Uses Helm Charts in examples/helm-chart to template single file manifests in `deploy/`.


ADubhlaoich (approved):
Docs-related changes LGTM!

deploy/README.md
shaun-nx, line 1: I feel like this README file should have a one liner on what each `deploy.yaml` file will deploy.
I see you already have something like this in the README in `examples/helm-chart` with a link to this folder too.

We can probably just re-use the same descriptions in this file.

deployments/rbac/rbac.yaml
shaun-nx, line 1: We will need to re-add this file for the python tests as it's references by `tests/suite/utils/resources_utils.py`

https://github.com/nginxinc/kubernetes-ingress/blob/main/tests/suite/utils/resources_utils.py#L40

shaun-nx: Same needs to happen for `deployments/rbac/apdos-rbac.yaml` and `deployments/rbac/ap-rbac.yaml`.
Both are references in `tests/suite/utils/resources_utils.py`

https://github.com/nginxinc/kubernetes-ingress/blob/main/tests/suite/utils/resources_utils.py#L64
https://github.com/nginxinc/kubernetes-ingress/blob/main/tests/suite/utils/resources_utils.py#L88

tests/suite/test_default_server.py
```diff
@@ -2,8 +2,7 @@
 
 import pytest
 import requests
-from requests.exceptions import ConnectionError
-from settings import BASEDIR, DEPLOYMENTS, TEST_DATA
+from settings import BASEDIR, TEST_DATA
```
vepatel, line 5:
```suggestion
from requests.exceptions import ConnectionError
from settings import BASEDIR, TEST_DATA
```

Makefile
```diff
@@ -69,6 +69,10 @@ update-crds: ## Update CRDs
 	kustomize build config/crd/app-protect-dos --load-restrictor='LoadRestrictionsNone' >deploy/crds-nap-dos.yaml
 	kustomize build config/crd/app-protect-waf --load-restrictor='LoadRestrictionsNone' >deploy/crds-nap-waf.yaml
 
+.PHONY: generate-manifests
+generate-manifests: ## Generate manifests
+	./hack/generate-manifests.sh
+
```
shaun-nx, line 75: All looks great on my end

I wanted to add one suggestions, but it's non-blocking
It might be useful to add another make command here for generating just a single deployment manifest.

For example
```
TYPE ?= default
.PHONY: generate-manifest
generate-manifest: ## Generate a single manifest
	helm template nginx-ingress --namespace nginx-ingress --values examples/helm-chart/$(TYPE)/values.yaml charts/nginx-ingress > deploy/$(TYPE)/deploy.yaml
```

shaun-nx (approved):

charts/nginx-ingress/values.yaml
```diff
@@ -380,7 +380,7 @@ controller:
 
     ## The name of the service account of the Ingress Controller pods. Used for RBAC.
     ## Autogenerated if not set or set to "".
-    # name: nginx-ingress
+    name: nginx-ingress
```
shaun-nx, line 383: By setting this label we are forcing the name of the Service Account to be `nginx-ingress` right? This would mean that Service Accounts that were created using the autogenerated name would be deleted and a new one made with the name `nginx-ingress` right?

config/base/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: nginx-ingress
```
shaun-nx, line 4: Do we want to also force users into using the `nginx-ingress` namespace?

charts/nginx-ingress/README.md
```diff
@@ -100,15 +100,15 @@ CRDs](#upgrading-the-crds).
 To upgrade the release `my-release`:
 
 ```console
-helm upgrade my-release oci://ghcr.io/nginxinc/charts/nginx-ingress --version 1.0.1
+helm upgrade my-release -n nginx-ingress oci://ghcr.io/nginxinc/charts/nginx-ingress --version 1.0.1
```
vepatel, line 103: according to old command the installation was in default ns and now user is running upgrade in a different ns? I understand that these command as a guide but most users usually straight-up copy-paste and upgrade will fail due to ownership of clusterroles as well.

charts/nginx-ingress/values.yaml
lucacome: the autogenerated name is `nginx-ingress`

config/base/namespace.yaml
lucacome: it's what we do already

charts/nginx-ingress/README.md
lucacome: How do you propose we change it?

Makefile
lucacome: What's the use case?

jasonwilliams14 (approved):
looks good.
This will be great to provide a singe .yaml file for customers to use for helm upgrdes.

charts/nginx-ingress/values.yaml
shaun-nx: If I do a Helm deploy today the auto generated name is <release-name>-nginx-ingress

So for example if I do `helm install test-release ...` I get `test-release-nginx-ingress` as the name of the ServiceAccount.

Issue #4284, opened 2023-08-23
Update example spec comment for GlobalConfiguration in helm values.yaml


shaun-nx:
**Is your feature request related to a problem? Please describe.**
This request updates the example listener block for GlobalConfiguration in the helm/values.yaml file to make it more clear on how to add listeners to the GlobalConfiguration resource. the `controller.globalConfiguration.spec` block requires the `listeners` to be added in JSON format. The current example comment does not make that as clear to the user.

**Describe the solution you'd like**
Snippet of current yaml with example listener block:
```yaml
controller:
  globalConfiguration:
    ## Creates the GlobalConfiguration custom resource. Requires controller.enableCustomResources.
    create: false

    ## The spec of the GlobalConfiguration for defining the global configuration parameters of the Ingress Controller.
    spec: {}
      # listeners:
      # - name: dns-udp
      #   port: 5353
      #   protocol: UDP
      # - name: dns-tcp
      #   port: 5353
      #   protocol: TCP
```

Snippet of suggested yaml with update listener block comment:
```yaml
controller:
  globalConfiguration:
    ## Creates the GlobalConfiguration custom resource. Requires controller.enableCustomResources.
    create: false

    ## The spec of the GlobalConfiguration for defining the global configuration parameters of the Ingress Controller.
    spec: {
    # "listeners": [
    #   {
    #     "name": "dns-udp",
    #     "port": 5353,
    #     "protocol": "UDP"
    #   },
    #   {
    #     "name": "dns-tcp",
    #     "port": 5353,
    #     "protocol": "TCP"
    #   }
    # ]
    }
```

UPDATE:
Given that the spec field for controller.globalConfiguration was appended with curly brackets `{}` I had assumed that this field only worked with JSON notation. However if the curly brackets `{}` are removed we can use YAML notation instead. The only catch is that the curly braces `{}` need to be there if the spec is being omitted.

```yaml
controller:
  globalConfiguration:
    ## Creates the GlobalConfiguration custom resource. Requires controller.enableCustomResources.
    create: false

    ## The spec of the GlobalConfiguration for defining the global configuration parameters of the Ingress Controller.
    spec:
      listeners:
      - name: dns-udp
        port: 5353
        protocol: UDP
      - name: dns-tcp
        port: 5353
        protocol: TCP
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

shaun-nx:
UPDATE:
[...]

PR #4285, opened 2023-08-23
Update example spec comment for GlobalConfiguration in helm values.yaml

shaun-nx:
### Proposed changes

This change updates the example comment for the `controller.globalConfiguration.spec` field in `deployments/helm-chart/values.yaml` to provide more clarity on the exact format for adding listeners.

Resolves https://github.com/nginxinc/kubernetes-ingress/issues/4284

### Checklist


deployments/helm-chart/values.yaml
```diff
@@ -271,14 +271,20 @@ controller:
...
+    #     "port": 5353,
+    #     "protocol": "UDP"
+    #   },
+    #   {
+    #     "name": "dns-tcp",
+    #     "port": 5353,
+    #     "protocol": "TCP"
+    #   }
+    # ]
+    }
```
brianehlert, line 287: This looks like a JSON array in a YAML document.
why are we doing it JSON style and not YAML style in a yaml doc.

shaun-nx: You are right. Since we use `spec: {}` as the default I assumed that we only used JSON notation here. However I seems like the empty curly brace `{}` is only required when leaving the field empty. I've updated the attached issue around this.

jjngx (approved):

haywoodsh (approved):

Issue #4209, opened 2023-08-09
Enhance Helm with an option to not generate a new IngressClass object


brianehlert:
### Discussed in https://github.com/nginxinc/kubernetes-ingress/discussions/4179

<div type='discussions-op-text'>

<sup>Originally posted by **brianehlert** August  3, 2023</sup>
One upgrade pattern for services in Kubernetes is a blue/green.  We tend to focus on this pattern for backend services behind the ingress controller.  However, this is a valid upgrade pattern for the ingress controller itself.

This is currently not possible using the Helm chart provided with this project.  
What the Helm chart enforces is that each ingress controller deployment has a unique IngressClass.  
While this is true to the IngressClass intention of being a unique management construct, it blocks the possibility of using the blue/green pattern to upgrade the ingress controller service itself.
The additional impact of this behavior is that the customer is not able to generate an IngressClass out of band with the Helm chart, such as part of some automation workflow prior to deploying the ingress controller using the Helm chart.

This is believed to be possible with manifest (not validated) however that does not meet the automation needs when a customer has expressly instrumented the Helm chart.

What I am proposing is some option of the Helm chart to:

- not generate a new ingress class when using the Helm chart to deploy ingress controller
- support adding an additional deployment of ingress controller to an existing IngressClass

</div>

PR #4333, opened 2023-09-07
Allow installing IC without creating a new ingress class

haywoodsh:
### Proposed changes

### Checklist


ADubhlaoich (approved):
LGTM, docs-wise!

vepatel:
@haywoodsh upgrading existing installation with a diff ingressClass, eg: `name: nginx2` removes the old ingressClass, is this expected behaviour?

vepatel:
If ingressClass named `nginx3` doesn't exists, then:
install/Upgrade fails `create: false` as IC looks for IngressClass specified in `name: nginx3`

vepatel:
qq: If two IC deployments are sharing same ingress class then is it possible to switch between the two for serving the requests? 
cc@brianehlert

haywoodsh:
> @haywoodsh upgrading existing installation using:
> 
> * diff ingressClass and setting `create: True`, eg: `name: nginx2` removes the old ingressClass
> * same ingressClass and setting `create: False` removes the old ingressClass
> 
> ```
> I0914 11:43:45.163449       1 main.go:236] Kubernetes version: 1.25.13
> F0914 11:43:45.167742       1 main.go:251] Error when getting IngressClass nginx: ingressclasses.networking.k8s.io "nginx" not found
> ```
> 
> Is this expected?

Appreciate your testing @vepatel! `Create: True` is equivalent to deploying the IC without this parameter in current versions. Did you initially install an older IC version without the parameter or `create: True`, and then use `helm upgrade` to upgrade with `create: False`? I didn't expect this PR would handle such a scenario. If the prior IngressClass was installed and managed by helm, I can see Helm deleting it during upgrade when you don't include it in the new release with `create: False`.
I wonder if there is a way to use schema to enforce a constant value, but at the very least, I should make it clear in the documentation this is not recommended for users to switch the value if they're using helm upgrade.

vepatel:
Thanks @haywoodsh there's a doc task linked in https://github.com/nginxinc/kubernetes-ingress/issues/4209 now 👍🏼

deployments/helm-chart/values.yaml
```diff
@@ -221,17 +221,22 @@ controller:
   ## The number of replicas of the Ingress Controller deployment.
   replicaCount: 1
 
-  ## A class of the Ingress Controller.
+  # Configures the ingress class the Ingress Controller uses.
+  ingressClass:
```
lucacome, line 225: This might be a breaking change? Since we're changing `ingressClass` from a string to an object?

haywoodsh: Yes, we will require the users to use the correct version of the helm chart and update their helm value yaml file before installing a new release, otherwise the helm schema will reject it. Is it what you meant?

lucacome: what I mean is that we probably have to bump the helm chart to `1.0.0`

vepatel (approved):

deployments/helm-chart/values.yaml
danielnginx: We decided to bump the helm chart to 1.0.0

shaun-nx (approved):

deployments/helm-chart/values.yaml
```diff
@@ -221,17 +221,22 @@ controller:
...
 
-  ## The Ingress Controller processes all the resources that do not have the "ingressClassName" field for all versions of kubernetes.
-  ingressClass: nginx
+    ## The Ingress Controller processes all the resources that do not have the "ingressClassName" field for all versions of kubernetes.
+    name: nginx
 
-  ## New Ingresses without an ingressClassName field specified will be assigned the class specified in `controller.ingressClass`.
-  setAsDefaultIngress: false
+    ## Creates a new IngressClass object with the name "controller.ingressClass.name". Set to "False" to use an exising IngressClass with the same name. If you use helm upgrade, do not change the value from the previous release. If you are upgrading from a release earlier than 3.3.0, do not set the value to false.
+    create: True
```
vepatel, line 236: @haywoodsh can we change this to lowercase here and in docs to keep it inline with other booleans?

vepatel (changes requested):
requesting some changes

docs/content/installation/installation-with-helm.md
```diff
@@ -327,8 +327,9 @@ The following tables lists the configurable parameters of the NGINX Ingress Cont
...
-|`controller.setAsDefaultIngress` | New Ingresses without an `"ingressClassName"` field specified will be assigned the class specified in `controller.ingressClass`. | false |
+|`controller.ingressClass.name` | A class of the Ingress Controller. An IngressClass resource with the name equal to the class must be deployed. Otherwise, the Ingress Controller will fail to start. The Ingress Controller only processes resources that belong to its class - i.e. have the "ingressClassName" field resource equal to the class. The Ingress Controller processes all the VirtualServer/VirtualServerRoute/TransportServer resources that do not have the "ingressClassName" field for all versions of Kubernetes. | nginx |
+|`controller.ingressClass.create` | Creates a new ingressClass object with the name `controller.ingressClass.name`. Set to `False` to use an exising ingressClass with the same name. | true |
```
vepatel, line 331: Should this section be same as above?

deployments/helm-chart/values.yaml
haywoodsh: Changed in the latest commit

vepatel (approved):

PR #4342, opened 2023-09-12
prevent pip to look for dependency during install

vepatel:
### Proposed changes

- prevent pip from looking for hashes of dependencies at the installation time

### Checklist


lucacome:
>prevent pip from looking for hashes of dependencies at the installation time

Why do we want this? We don't want to use hashes?

vepatel:
@lucacome not for the dependencies no, see the linked issue for more details.

jjngx (approved):

lucacome:
I feel like we should figure out why it only happens _sometimes_ before disabling it. It kind of defeats the purpose of the hashes if we don't check them...

PR #4351, opened 2023-09-13
Update helm values file to move controller.serviceMonitor to prometheus.serviceMonitor

shaun-nx:
### Proposed changes

This PR moves the ServiceMonitor fields from `controller.serviceMonitor` to `prometheus.serviceMonitor` in the helm values files.
This is being done to bring the creation of ServiceMonitor closer to where the Prometheus Service itself is created.

Helm command used to deploy:
```bash
~/deployments/helm-chart » helm install prometheus \
--set controller.kind=deployment \
--set controller.nginxplus=true \
--set controller.image.repository=nginx-plus-ingress \
--set controller.image.tag="moveServiceMonitor" \
--set prometheus.service.create=true \
--set prometheus.serviceMonitor.create=true \
--set prometheus.serviceMonitor.labels.product="nic" \
--set controller.replicaCount=5 .
```

### Checklist


ADubhlaoich (approved):
LGTM!

vepatel (approved):

deployments/helm-chart/templates/controller-prometheus-service.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
 spec:
   clusterIP: None
   ports:
-  - name: prometheus
+  - name: {{ .Values.prometheus.service.endpoint }}
```
lucacome, line 15: why do we need the name of the port to be configurable?

shaun-nx: I made it configurable mostly to make it more clear how this field links to the value of `prometheus.serviceMonitor.endpoints`

When you configure the `endpoints` for a ServiceMonitor to scrape, it uses the name of the port in the service to know which ports to scrape.
So this might look like this:

```yaml
prometheus:
  serviceMonitor:
    endpoints:
      - port: prometheus
```

I wanted to represent the value of the name here as `endpoint` so that it was more obvious how it linked together with the ServiceMonitor `endpoints` field
So for example, in the values yaml it would look like this

lucacome: IMHO we don't need to make it configurable, and the name `prometheus.service.endpoint` is not very clear anyway

deployments/helm-chart/values.yaml
```diff
@@ -474,8 +461,26 @@ prometheus:
...
+    ## Kubernetes object labels to attach to the serviceMonitor object.
+    labels: {}
+
+    ## A set of labels to allow the selection of endpoints for the ServiceMonitor.
+    selectorMatchLabels: {}
+
+    ## A list of endpoints allowed as part of this ServiceMonitor.
+    endpoints: []
+    # - port: prometheus
+
```
jasonwilliams14, line 483: I was doing some testing with this. As is, with the `- port: prometheus` commented out, metric scrapping will not work. (prometheus dashboard will not pick up the servicemonitor)

The service will get generated as the following:

```
spec:
  endpoints: []
  selector:
    matchLabels:
      app: nic-servicemonitor
      app.kubernetes.io/instance: test01
      app.kubernetes.io/name: nginx-ingress
```

Would it make more sense to just hard code these values in and we can generate them upon install for the users?

brianehlert: > I was doing some testing with this. As is, with the `- port: prometheus` commented out, metric scrapping will not work. (prometheus dashboard will not pick up the servicemonitor)

This tells me that:
1 - this needs to be uncommented with a note to not change it ( values.prometheus.serviceMonitor.endpoints )
2 - also needs a note not to change it "prometheus" line ( values,prometheus.endpoint )

This will let us get this out the door with minimal change

In a future release we need to hide these two options.  I really don't see any value in these fields being configurable as it really just sets someone up to break their monitor.

lucacome: since we're moving all the fields in this PR (and this hasn't been released yet), I think we can remove the `endpoints` as part of the changes here.

deployments/helm-chart/templates/controller-prometheus-service.yaml
lucacome: so from the other thread I think we can revert this change

deployments/helm-chart/values.yaml
```diff
@@ -474,8 +461,26 @@ prometheus:
...
 
+  serviceMonitor:
+    ## Creates a serviceMonitor to expose statistics on the kubernetes pods.
+    create: false
+
+    ## Kubernetes object labels to attach to the serviceMonitor object.
+    labels: {}
+
+    ## A set of labels to allow the selection of endpoints for the ServiceMonitor.
+    selectorMatchLabels: {}
```
lucacome, line 478: I don't know if we even need to have these customizable. And the `labels` in the `service`.
I think we can remove the extra fields and set the labels in the yaml files directly. I can't see why somebody would want to customize these lables...

deployments/helm-chart/templates/controller-prometheus-service.yaml
shaun-nx: Sounds good.

deployments/helm-chart/values.yaml
shaun-nx: I feel it is still useful to keep this customizable.
I think it would be good if we have a default label selection there so if users just do a helm install without setting any arguments it will still work out of the box.

haywoodsh (approved):

PR #4508, opened 2023-10-12
add nginx-plus-module-otel to base image

justdan96:
### Proposed changes

This PR is just to add nginx-plus-module-otel to the base images, so that OpenTelemetry can be configured easily.

### Checklist

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -43,7 +43,7 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/apk/cert.pem,mode=0644 \
 	--mount=type=bind,from=alpine-opentracing-lib,target=/tmp/ot/ \
 	wget -nv -O /etc/apk/keys/nginx_signing.rsa.pub https://cs.nginx.com/static/keys/nginx_signing.rsa.pub \
 	&& printf "%s\n" "https://pkgs.nginx.com/plus/${NGINX_PLUS_VERSION}/alpine/v$(grep -E -o '^[0-9]+\.[0-9]+' /etc/alpine-release)/main" >> /etc/apk/repositories \
-	&& apk add --no-cache nginx-plus nginx-plus-module-njs nginx-plus-module-opentracing nginx-plus-module-fips-check libcap libcurl \
+	&& apk add --no-cache nginx-plus nginx-plus-module-njs nginx-plus-module-opentracing nginx-plus-module-otel nginx-plus-module-fips-check libcap libcurl \
 	&& cp -av /tmp/ot/usr/local/lib/libjaegertracing*so* /tmp/ot/usr/local/lib/libzipkin*so* /tmp/ot/usr/local/lib/libdd*so* /tmp/ot/usr/local/lib/libyaml*so* /usr/local/lib/ \
 	&& ldconfig /usr/local/lib/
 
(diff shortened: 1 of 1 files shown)


jjngx:
Hi @justdan96, thank you for opening the PR. Please follow the project contributing guideline described [here](https://github.com/nginxinc/kubernetes-ingress/blob/main/CONTRIBUTING.md#contributing).

justdan96:
> an Issue that describes the enhancement (this informs release notes and watchers)

There isn't a lot to describe, this just avoids users having to compile their own images to add the otel module (or use a custom init container). 

> at least an example showing other users how to use the module

It currently isn't possible the use the module (even with this merged) as it needs to be enabled in the template. It would need code similar to that for OpenTracing - i.e.: https://github.com/nginxinc/kubernetes-ingress/blob/2641fbb237352112912e2698994015286863d21b/internal/configs/config_params.go#L32
I don't have time to do that at the moment so this PR is just the first step. Users can otherwise refer to this page for module configuration parameters: https://docs.nginx.com/nginx/admin-guide/dynamic-modules/opentelemetry/.

build/Dockerfile
```diff
@@ -43,7 +43,7 @@ RUN --mount=type=secret,id=nginx-repo.crt,dst=/etc/apk/cert.pem,mode=0644 \
 	--mount=type=bind,from=alpine-opentracing-lib,target=/tmp/ot/ \
 	wget -nv -O /etc/apk/keys/nginx_signing.rsa.pub https://cs.nginx.com/static/keys/nginx_signing.rsa.pub \
 	&& printf "%s\n" "https://pkgs.nginx.com/plus/${NGINX_PLUS_VERSION}/alpine/v$(grep -E -o '^[0-9]+\.[0-9]+' /etc/alpine-release)/main" >> /etc/apk/repositories \
-	&& apk add --no-cache nginx-plus nginx-plus-module-njs nginx-plus-module-opentracing nginx-plus-module-fips-check libcap libcurl \
+	&& apk add --no-cache nginx-plus nginx-plus-module-njs nginx-plus-module-opentracing nginx-plus-module-otel nginx-plus-module-fips-check libcap libcurl \
```
lucacome, line 46: do we want to install both OpenTracing and OpenTelemetry?

brianehlert: Is there any problem with doing so?

the project has supported OpenTracing for so long that this would essentially be a breaking change for tracing, and there is not any enablement through YAML - OpenTel tracing would be all snippets based configuration until we can develop the YAML patterns.

brianehlert:
The key question that we need an answer for before accepting this PR is:
- can both the OpenTracing and OpenTelemetry modules be installed at the same time and not interact in a negative way with each other?

This is so we can give a path for customers to migrate from OpenTracing to OpenTelemetry in-place and not introduce an immediately breaking change.
We totally embrace moving to the new native NGINX OpenTelemetry module but want to be sure to also do the right thing for anyone using OpenTracing and not introduce a release that immediately breaks them if we can avoid it.

Can anyone / has anyone investigated this?

brianehlert:
Also, the NGINX OpenTelemetry module should be an option for both NGINX Plus and NGINX open source.

justdan96:
1. We have both modules installed at the same time with no ill effects. We have only one loaded though. 
2. I can amend the Dockerfile to add it to further images.

brianehlert:
Since this will be configured exclusively using snippets at this time, it would be great if there was also an example included.
Probably under /shared-examples (not entirely sure though)
That could help us cover a documentation gap.

justdan96:
Actually for point 2 there doesn't seem to be Docker images I could copy the .so files from, so I am assuming we would want to compile from source?

brianehlert:
I think the key here is that we need a solution for this for both the free and N+ releases.

defanator:
> * can both the OpenTracing and OpenTelemetry modules be installed at the same time and not interact in a negative way with each other?

I believe they can, there's nothing that should prevent having both and using just the one. I'm tagging @dplotnikov-f5 and @thresheek as they could share first-hand feedback.

brianehlert:
If we can finish this with the N+ OTEL module that is fine.
While any N OSS module sorts itself out.

justdan96:
I need to look at this again at some point.

PR #4539, opened 2023-10-18
Update k8s versions in nightly

lucacome:
### Proposed changes

Update the list of k8s versions with 1.27


.github/workflows/ci.yml
```diff
@@ -268,7 +268,7 @@ jobs:
                                                 {\"image\": \"debian-plus-nap\", \"marker\": \"dos\"}], \
                                               \"k8s\": [\"${{ needs.checks.outputs.k8s_latest }}\"]}" >> $GITHUB_OUTPUT
           else
-            echo "matrix={\"k8s\": [\"1.22.17\", \"1.23.17\", \"1.24.15\", \"1.25.11\", \"1.26.6\", \"${{ needs.checks.outputs.k8s_latest }}\"], \
+            echo "matrix={\"k8s\": [\"1.22.17\", \"1.23.17\", \"1.24.15\", \"1.25.11\", \"1.26.6\", \"1.27.3\", \"${{ needs.checks.outputs.k8s_latest }}\"], \
```
vepatel, line 271: we remove a few right I guess now, 5 should be sufficient

lucacome: we usually had all the supported versions I think

jjngx (approved):

vepatel (approved):

.github/workflows/ci.yml
vepatel: we usually had last 5 but I think we just kept on adding new versions without removing.

Issue #4503, opened 2023-10-11
whoami: cannot find name for user ID 101 after ssh-ing in pod container


vepatel:
**Describe the bug**
- when running command whoami or 'cat /etc/passwd' after exec into NIC (v3.2.0 >= plus only) pod:
```
    $ whoami
          whoami: cannot find name for user ID 101

    $ cat /etc/passwd
          root:x:0:0:root:/root:/bin/bash
          [...]
          nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin
          nginx:x:999:999:nginx user:/nonexistent:/usr/sbin/nologin
```
- when running command whoami or 'cat /etc/passwd' after exec into NIC (v3.2.0 < plus only) pod:
```
     $ whoami
            nginx
            
      $ cat /etc/passwd
            root:x:0:0:root:/root:/bin/bash
            [...]
            _apt:x:100:65534::/nonexistent:/usr/sbin/nologin
            nginx:x:101:101:nginx user,,,:/nonexistent:/bin/false  
```

**To Reproduce**
- Install an IC instance with plus images for v3.2.0+ and ssh into pod container and run above commands.

vepatel:
hey @sigv, any opinion or thoughts on this one?

sigv:
This happens on Plus variants only? Not reproducible on the OSS variant?

sigv:
@vepatel, this happens on Plus variants only? Not reproducible on the OSS variant?

sigv:
Nothing in this repo's [`v3.1.1...v3.2.0`](https://github.com/nginxinc/kubernetes-ingress/compare/v3.1.1...v3.2.0) jumps right out.

```
$ docker run -it --rm --entrypoint=/bin/sh ghcr.io/nginxinc/kubernetes-ingress:3.2.0 -c 'whoami; grep nginx /etc/passwd'
nginx
nginx:x:101:101:nginx user:/nonexistent:/bin/false
```

---

What I am initially thinking is the OSS variant and Plus variants are built from different bases.
Looking into who adds the user.

sigv:
@vepatel, additional question: is it some specific base distro that is affected -- Debian/Alpine/UBI -- or across the board?

sigv:
Alpine has BusyBox's `adduser` and the behavior there appears unchanged.
My gut feel tells me the Alpine Plus images are fine as well.

```
[...]

PR #4540, opened 2023-10-18
Ensure Debian Plus starts assigning users from 101

sigv:
### Proposed changes

When running `adduser` or `addgroup`, adduser.conf controls various configuration flags. Notably for us, there is `FIRST_SYSTEM_UID` and `FIRST_SYSTEM_GID` which determine what UIDs and GIDs are assigned for system users.

The Nginx images are historically built with UID 101 for `nginx` user. However, Debian 12 cleaned up its image and UID 100 became available. By default, the range `100-999` is used for system users. That is bad for the Nginx image as UID 101 is expected.

This patch simply ensures the first ID used for package install is 101.

Improves #4503.

### Checklist


sigv:
NB! Currently this change will be a no-op, as the Nginx Plus package appears to run `useradd`. See referenced issue for details.

sigv:
`sed` expression `/^FIRST_SYSTEM_GID\s*=/d` was missing the `d`elete at end. Sorry about that - pushed a fix for that.

sigv:
Decision was made to create the user explicitly before installation of the package. This PR is redundant as result.

PR #4551, opened 2023-10-20
Move static files to scratch image

lucacome:
### Proposed changes

Moves nginx static files to a `scratch` image.
With `ADD --link` we can cache the files instead of downloading them every time.
It also removes the need for `curl`/`wget` and switches from `gpg` to https://sequoia-pgp.org/ that has fewer dependencies.


jjngx (approved):

jjngx (approved):

jjngx (approved):

shaun-nx (approved):
👏

Issue #4558, opened 2023-10-24
Add the possibility for the controller to be in a shared namespace


panzouh:
**Is your feature request related to a problem? Please describe.**
I would like to use nginx plugged with a sidecar, the purpose of the sidecar is to pull a configuration from an API and then pushing it to Nginx via a shared volume. My problem is that when the configuration is reloaded i would like to send a SIGHUP signal to nginx process to gracefully shutdown and reload configuration.

**Describe the solution you'd like**
The solution would be simple it would be to add this key like in the `values.yaml` :

```yaml
controller:
  ## The name of the Ingress Controller daemonset or deployment.
  name: controller

  ## The kind of the Ingress Controller installation - deployment or daemonset.
  kind: deployment

  ## Shared process namespace between containers in the Ingress Controller pod.
  sharedProcessNamespace: false
```

And this in `./templates/deployment.yaml` & `./templates/daemonset.yaml` :

```yaml
apiVersion: apps/v1
kind: Deployment # Same on daemonset
metadata:
  name: {{ include "nginx-ingress.controller.fullname" . }}
  namespace: {{ .Release.Namespace }}
  labels:
    {{- include "nginx-ingress.labels" . | nindent 4 }}
{{- if .Values.controller.annotations }}
  annotations: {{ toYaml .Values.controller.annotations | nindent 4 }}
{{- end }}
spec:
  {{- if not .Values.controller.autoscaling.enabled }}
  replicas: {{ .Values.controller.replicaCount }}
  {{- end }}
  selector:
    matchLabels:
      {{- include "nginx-ingress.selectorLabels" . | nindent 6 }}
  template:
    metadata:
      labels:
        {{- include "nginx-ingress.selectorLabels" . | nindent 8 }}
{{- if .Values.nginxServiceMesh.enable }}
        nsm.nginx.com/enable-ingress: "true"
        nsm.nginx.com/enable-egress: "{{ .Values.nginxServiceMesh.enableEgress }}"
        nsm.nginx.com/deployment: {{ include "nginx-ingress.controller.fullname" . }}
{{- end }}
{{- if .Values.controller.pod.extraLabels }}
{{ toYaml .Values.controller.pod.extraLabels | indent 8 }}
{{- end }}
{{- if or .Values.prometheus.create .Values.controller.pod.annotations }}
      annotations:
{{- if .Values.prometheus.create }}
        prometheus.io/scrape: "true"
[...]

PR #4559, opened 2023-10-24
Add process namespace sharing for ingress controller

panzouh:
### Proposed changes

The changes allow to configure process namespace sharing for a pod.
 
Solves : #4558

### Checklist

### DIFF
--- a/charts/nginx-ingress/values.yaml
+++ b/charts/nginx-ingress/values.yaml
@@ -64,6 +64,9 @@ controller:
   ## Enables debugging for NGINX. Uses the nginx-debug binary. Requires error-log-level: debug in the ConfigMap via `controller.config.entries`.
   nginxDebug: false
 
+  ## Shared process namespace between containers in the Ingress Controller pod.
+  sharedProcessNamespace: false
+
   ## The log level of the Ingress Controller.
   logLevel: 1
 
--- a/charts/nginx-ingress/values.yaml
+++ b/charts/nginx-ingress/values.yaml
@@ -64,8 +64,8 @@ controller:
   ## Enables debugging for NGINX. Uses the nginx-debug binary. Requires error-log-level: debug in the ConfigMap via `controller.config.entries`.
   nginxDebug: false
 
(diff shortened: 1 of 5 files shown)


charts/nginx-ingress/values.yaml
```diff
@@ -64,6 +64,9 @@ controller:
   ## Enables debugging for NGINX. Uses the nginx-debug binary. Requires error-log-level: debug in the ConfigMap via `controller.config.entries`.
   nginxDebug: false
 
+  ## Shared process namespace between containers in the Ingress Controller pod.
+  sharedProcessNamespace: false
```
lucacome, line 68: should we name it `share` instead of `shared`?

panzouh: The attribute in the pod template is indeed `shareProcessNamespace` i thought it was more logic but i don't mind changing it

panzouh: > should we name it `share` instead of `shared`?

Updated on 8961b32f8f4a1a62883ba28fa92d8c88f657bf61

panzouh:
> @panzouh there are a few things missing from this PR.
> 
> 1. Docs will need to be updated for this new value. You can add a new entry for this in `charts/nginx-ingress/README.md`. This is to ensure we provide a breakdown of what the setting does.
> 2. The `values.json.schema` file in `charts/nginx-ingress/` will need to be updated as well.

Of course, given the fact that these modifications are not described in the contribute guide nor in the precommit config how it should be done ?

lucacome (approved):
@panzouh LGTM, but seems like you need to rebase your branch.

shaun-nx (approved):

shaun-nx:
HI @panzouh was having some problems with the pipeline. I'm re-running the failed jobs now. I will check back in on this PR in my morning (bit late as I'm checking this 😅)

shaun-nx:
Thanks for adding this @panzouh ! Keep us informed on https://github.com/nginxinc/kubernetes-ingress/issues/4558 is this setting helps you!

PR #4560, opened 2023-10-24
Add string comparison functions to Go template executors

### LINKED ISSUE #4555: Add string comparison Go template functions

I would like to intercept proxied responses with codes greater than or equal to 300 and return an error HTML page, but only for non-API routes. To achieve this, I would like to be able to make use of the following ingress template:
```yaml
{{range $location := $server.Locations}}
location {{$location.Path}} {
  ...
  {{- if not contains $location.Path "/api/"}}
  proxy_intercept_errors on;
  {{- end}}
  ...
}
```

However, the `contains` function is not defined in the Go template, nor are any other string comparison functions.

I would like to add the following helper string comparison functions to the Go templates so as to enable the possibility of performing string comparisons in the templates:
* `contains` - [`strings.Contains`](https://pkg.go.dev/strings#Contains)
* `hasPrefix` - [`strings.HasPrefix`](https://pkg.go.dev/strings#HasPrefix)
* `hasSuffix` - [`strings.HasSuffix`](https://pkg.go.dev/strings#HasSuffix)
* `toLower` - [`strings.ToLower`](https://pkg.go.dev/strings#ToLower)
* `toUpper` - [`strings.ToUpper`](https://pkg.go.dev/strings#ToUpper)

The latter 2 functions are not exactly string comparison functions, but they do help in comparing strings with different casing.

I have considered making use of the `if` directive but I'm told that [it is evil when used in a location context](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/).

**Additional context**

I am willing to work on the PR to implement this change. This issue is being opened so as to get approval from the core development team first.

ianyong:
### Proposed changes

Add the following helper functions to the Go template executors to enable string comparisons:
* `contains` - [`strings.Contains`](https://pkg.go.dev/strings#Contains)
* `hasPrefix` - [`strings.HasPrefix`](https://pkg.go.dev/strings#HasPrefix)
* `hasSuffix` - [`strings.HasSuffix`](https://pkg.go.dev/strings#HasSuffix)
* `toLower` - [`strings.ToLower`](https://pkg.go.dev/strings#ToLower)
* `toUpper` - [`strings.ToUpper`](https://pkg.go.dev/strings#ToUpper)

Not very sure why there are two versions of the template executors. I presume `version1` will eventually be phased out in favour of `version2` given the naming? Either way, I've made the same changes to both versions.

Resolves #4555. 

### Checklist


internal/configs/version2/template_helper.go
```diff
@@ -20,17 +20,12 @@ func hasCIKey(key string, d map[string]string) bool {
...
 }
 
-// toLower takes a string and make it lowercase.
-//
-// Example:
-//
-//	{{ if .SameSite}} samesite={{.SameSite | toLower }}{{ end }}
-func toLower(s string) string {
-	return strings.ToLower(s)
-}
```
ianyong, line 30: I removed this function because all it does it call `strings.ToLower` with the exact same parameters, which is unnecessary.

jjngx (changes requested):
Thank you for contributing! Could you please keep tests in the same style as tests other packages in the project?

- call `t.Parallel()` as a first line in a test
- use anonymous struct literals for table tests (as in other test files)
- keep test struct fields not exported

ianyong:
@jjngx Thanks for reviewing! I've made the changes for the first 2 points. For the last point, the test struct fields need to be exported for them to be accessed within the templates, else the tests will fail with `<struct field name> is an unexported field of struct type struct <struct type>`. It's also why the `expected` struct field is not exported across all the test cases - because it does not need to be accessed within the templates.

jjngx (approved):
@ianyong thank you again for contributing! 🚀

jjngx:
@ianyong could you please update and push your branch?

jjngx (approved):

ianyong:
Is there anything I can do to get this PR merged? The failing CI jobs don't seem to be caused by my changes, or at least I see other PRs with the same failed jobs being merged.

vepatel:
@hanyouqing we'll prioritise merging this in coming week, thanks! You can monitor the state from parent issue.

jjngx:
@ianyong could you please update and push your branch?

jjngx:
@ianyong  could you please update and push your branch?

shaun-nx (approved):
👍

haywoodsh (approved):

Issue #4550, opened 2023-10-20
NIC throws warnings re deprecated annotations when `ingressClassName` isn't used


vepatel:
**Describe the bug**
NIC throws warnings re deprecated annotation `kubernetes.io/ingress.class` when `ingressClassName` isn't used

**To Reproduce**
Steps to reproduce the behaviour:

1. Deploy IC on a k8s cluster with default IngressClass
2. Deploy an application with Ingress resource (omit `ingressClassName` field under `spec:`)
3. Check IC logs

**Expected behaviour**
There shouldn't be any mentions of the said Ingress in the logs

**Current behaviour**
```
W1019 10:17:52.179344       1 controller.go:3985] Using the DEPRECATED annotation 'kubernetes.io/ingress.class'. The 'ingressClassName' field will be ignored.
W1019 10:17:52.727485       1 controller.go:3985] Using the DEPRECATED annotation 'kubernetes.io/ingress.class'. The 'ingressClassName' field will be ignored.
W1019 10:20:49.168243       1 controller.go:3985] Using the DEPRECATED annotation 'kubernetes.io/ingress.class'. The 'ingressClassName' field will be ignored.
```

PR #4564, opened 2023-10-25
Don't throw annotation deprication warning when `ingressClassName` is not used in an Ingress resource

shaun-nx:
### Proposed changes

The below warning message was being thrown when an `Ingress` resource did not have an `ingressClassName` AND the annotation `kubernetes.io/ingress.class` was not used.

```
W1025 09:30:31.434146       1 controller.go:3985] Using the DEPRECATED annotation 'kubernetes.io/ingress.class'. The 'ingressClassName' field will be ignored.
```

This warning message should only be displayed when `kubernetes.io/ingress.class` is used AND `ingressClassName` is not used.

**Testing**: Confirmed warning message was absent when deploying an Ingress resource with both `ingressClassName` AND the annotation `kubernetes.io/ingress.class` missing. This same test was performed for the mergable ingress type (master -minion) to ensure configurations are applied correctly.

Resolves https://github.com/nginxinc/kubernetes-ingress/issues/4550

### Checklist


internal/k8s/controller.go
```diff
@@ -3980,7 +3980,7 @@ func (lbc *LoadBalancerController) HasCorrectIngressClass(obj interface{}) bool
 		class = obj.Annotations[ingressClassKey]
 		if class == "" && obj.Spec.IngressClassName != nil {
 			class = *obj.Spec.IngressClassName
-		} else {
+		} else if class != "" {
```
vepatel, line 3983: @jjngx should there still be an ending`else {}` block according to go patterns?

shaun-nx: Not that I'm aware of.
This is the same as doing this:
```
class = obj.Annotations[ingressClassKey]
if class == "" && obj.Spec.IngressClassName != nil {
    class = *obj.Spec.IngressClassName
}

// True only if there is a value for obj.Annotations[ingressClassKey]
if class != "" {
    glog.Warningln("Using the DEPRECATED annotation 'kubernetes.io/ingress.class'. The 'ingressClassName' field will be ignored.")
}
```

vepatel (approved):

haywoodsh (approved):

Issue #4251, opened 2023-08-16
Graduate TransportServer and GlobalConfiguration to v1


brianehlert:
During the process of raising the NGINX Ingress Controller specific CRDs to v1, the TransportServer object was forgotten.
We need to remedy this.

Acceptance Criteria:

- advance TransportServer versioning to v1
- provide in-box backward compatibility for the current v1alpha1
- provide in-box deprecation message about TransportServer v1alpha1 (log message, Event on object)
- advance GlobalConfiguration versioning to v1
- provide in-box backward compatibility for the current v1alpha1
- provide in-box deprecation message about GlobalConfiguration v1alpha1 (log message, Event on object)
- Remove previewpolicy flag (long deprecated and no longer in use)
```[tasklist]
### Tasks
- [x] Update CRDs in both `chats/nginx-ingress/crds/` and `deploy/crds.yaml`
- [x] Graduate TransportServer to V1
- [x] Graduate GlobalConfiguration to V1
- [x] Update documentation around graduation of TS and GC
- [x] Test deployments of TS and GC using Helm and deploy manually with `deploy/crds.yaml`
```

vepatel:
@brianehlert should we create one for GlobalConfiguration as well?

brianehlert:
> @brianehlert should we create one for GlobalConfiguration as well?

We can extend this to cover both and have sub tasks for implementation if that is needed.
Or write a second one.  It is simply about how the work is tracked.

danielnginx:
As reminder: Run make crds before check in the code.

PR #4574, opened 2023-10-26
Graduate TransportServer and GlobalConfiguration to v1

shaun-nx:
### Proposed changes

Resolves: https://github.com/nginxinc/kubernetes-ingress/issues/4251

This change:
- Updates `TransportServer` resource from `v1alpha1` to `v1`
- Updates `GlobalConfiguration` resource from `v1alpha1` to `v1`

**Testing**

- Confirmed `v1alpha1` of `TransportServer` and `GlobalConfiguration` are not deleted when the new CRD is applied
- Confirmed that `v1alpha1` of `TransportServer` will operates correctly when new CRD is applied
- Confirmed that v1 and v1alpha1 of `TransportServer` and `GlobalConfiguration` can be deployed side-by-side
- Confirmed that all relevant configurations for `TransportServer` will operate correctly. These properties were tested in isolation as they were renamed when being moved from `v1alphav1/types.go` to `v1/types.go` :
  - HealthCheck
  - Match (property of HealthCheck)
  - UpstreamParameters
  - Action
  - TLS
- Confirmed all python tests for `TransportServer` and `GlobalConfiguration` pass when deployed as `v1alpha1` with using the updated CRD
- Confirmed all python tests for `TransportServer` and `GlobalConfiguration` pass when deployed as `v1` AND using the updated CRD

### Checklist


config/crd/bases/k8s.nginx.org_virtualservers.yaml
```diff
@@ -105,8 +105,8 @@ spec:
                   routing.
                 type: boolean
               listener:
-                description: Listener references a custom http and/or https listener
-                  defined in GlobalConfiguration.
+                description: VirtualServerListener references a custom http and/or
```
vepatel, line 108: did we change a property name for VS?

tests/suite/utils/custom_resources_utils.py
```diff
@@ -125,7 +125,7 @@ def read_ts(custom_objects: CustomObjectsApi, namespace, name) -> object:
     """
     Read TransportService resource.
     """
-    return read_custom_resource_v1alpha1(custom_objects, namespace, "transportservers", name)
+    return read_custom_resource_v1(custom_objects, namespace, "transportservers", name)
```
vepatel, line 128: think we can now replace the usage of this func with https://github.com/nginxinc/kubernetes-ingress/blob/main/tests/suite/utils/custom_resources_utils.py#L63, basically all tests that used `read_custom_resource_v1alpha1` can use the one in link above

config/crd/bases/k8s.nginx.org_virtualservers.yaml
shaun-nx: This description shouldn't have been changed. In `v1.types.go` we changed the struct `Listener` to `VirtualServerListener` since `GlobalConfiguration` also had a `Listener` struct that it used, so they were conflicting with each other.

I'll change this description back though. It should be referring to the name used in the YAML definition.

tests/suite/utils/custom_resources_utils.py
shaun-nx: That's a good point. I'll change the name of this one back and do the usage update.

config/crd/bases/k8s.nginx.org_virtualservers.yaml
shaun-nx: Actually nevermind I was wrong about that. The description that is put here is taken from the comment that is put above the struct in `v1/types.go` so they have to stay in line

```
// VirtualServerListener references a custom http and/or https listener defined in GlobalConfiguration.
type VirtualServerListener struct {
	HTTP  string `json:"http"`
	HTTPS string `json:"https"`
}
```

tests/suite/utils/custom_resources_utils.py
```diff
@@ -163,7 +163,7 @@ def patch_gc_from_yaml(custom_objects: CustomObjectsApi, name, yaml_manifest, na
     :return: a dictionary representing the resource
     """
     print(f"Load {yaml_manifest}")
-    return patch_custom_resource_v1alpha1(custom_objects, name, yaml_manifest, namespace, "globalconfigurations")
+    return patch_custom_resource_v1(custom_objects, name, yaml_manifest, namespace, "globalconfigurations")
```
shaun-nx, line 166: Opening a discussion on this one.
From what I see we only ever had `patch_custom_resource_v1alpha1` and never had a generic ``patch_custom_resource` in this file.

So we could keep this change as is or change it to `patch_custom_resource` and also add a `patch_custom_resource_v1alpha1` to have both

config/crd/bases/k8s.nginx.org_virtualservers.yaml
shaun-nx: To answer the original question, no we didn't change the property. You will still use the keyword `listener` in the Yaml definition when using custom listeners.

tests/suite/utils/custom_resources_utils.py
vepatel: we can have one function now `def patch_custom_resource()` then below listed ones can use that:
`def patch_ts()`
`def patch_gc_from_yaml()`
`def patch_ts_from_yaml()`

```
def patch_custom_resource_v1alpha1(custom_objects: CustomObjectsApi, name, yaml_manifest, namespace, plural) -> None:
    """
    Patch a custom resource based on yaml manifest
    """
    print(f"Update a Resource: {name}")
    with open(yaml_manifest) as f:
        dep = yaml.safe_load(f)

    try:
        custom_objects.patch_namespaced_custom_object("k8s.nginx.org", "v1", namespace, plural, name, dep)
    except ApiException:
        logging.exception(f"Failed with exception while patching custom resource: {name}")
        raise
```

internal/configs/config_params.go
```diff
@@ -196,8 +196,8 @@ func NewDefaultGlobalConfigParams() *GlobalConfigParams {
 func NewGlobalConfigParamsWithTLSPassthrough() *GlobalConfigParams {
 	return &GlobalConfigParams{
 		Listeners: map[string]Listener{
-			conf_v1alpha1.TLSPassthroughListenerName: {
-				Protocol: conf_v1alpha1.TLSPassthroughListenerProtocol,
+			conf_v1.TLSPassthroughListenerName: {
```
vepatel, line 199: were these lines covered by a unit test earlier?

shaun-nx: No, these never had unit tests.

jjngx (approved):
🚀

examples/custom-resources/basic-tcp-udp/transport-server-udp.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: k8s.nginx.org/v1alpha1
+apiVersion: k8s.nginx.org/v
```
haywoodsh, line 1: seems to be a typo here?

shaun-nx: Very nice catch!

internal/configs/config_params.go
danielnginx: Should we address the Codecov warnings? Is that viable?

haywoodsh (approved):

internal/configs/config_params.go
shaun-nx: @danielnginx a lot of that code can't be coverd by unit tests, or it doesn't make sense to unit test it

PR #4655, opened 2023-11-15
Support weights 0 and 100 in traffic splitting

haywoodsh:
### Proposed changes

Allow split traffic's weight value in virtual server to be 0% or 100%. 

This change is proposed because "0%" is not currently a valid syntax in `split_clients`. When the value is set to 0%, it will not be included in the generated config, but the effect will be the same as if it were. A value of 100% enables full traffic routing to a particular upstream.

### Checklist


jjngx (approved):
🚀 
@haywoodsh could you add description to the Proposed Changes section?

pkg/apis/configuration/validation/virtualserver_test.go
```diff
@@ -1583,32 +1583,89 @@ func TestValidatePath(t *testing.T) {
 
 func TestValidateSplits(t *testing.T) {
```
shaun-nx, line 1584: I would add a test to `TestValidateSplitsFails` as well to test that a negative number can't be used.

This might look like this:
```
{
	splits: []v1.Split{
		{
			Weight: 100,
			Action: &v1.Action{
				Pass: "test-1",
			},
		},
		{
			Weight: -2,
			Action: &v1.Action{
				Pass: "test-2",
			},
		},
	},
	upstreamNames: map[string]sets.Empty{
		"test-1": {},
		"test-2": {},
	},
	msg: "negative weight",
},
```

haywoodsh: test case added

shaun-nx (approved):

Issue #4603, opened 2023-11-02
Support rate-limiting for ingress-resources


dbaumgarten:
**Is your feature request related to a problem? Please describe.**
To properly secure our applications running behind nginx-ingress we need to implement rate-limiting on source-IP basis. Nginx does already support this. However the ingress-controller has no direct support for this. You need to use snippets (in the configmap AND in the ingress-resources).
Additionally rate-limits implemented with snippets do not work well with pod-autoscaling. The rate-limits are per-pod, and if autoscaling adds more pods, it will accidentally also increase the total rate-limit.

**Describe the solution you'd like**
It would be great to have the rate-limiting feature available via the usual annotations. Something like:
```
nginx.org/request-limit: "100/s"
```
Additionally it would be great if the ingress controller would (optionally) scale this value by the amount of currently running ingress-pods.

**Describe alternatives you've considered**
Implement rate-limiting via snippets and implement a custom controller that dynamically modifies the nginx-configmap according to the current amount of running nginx-pods.

**Additional context**
I would like to implement this feature myself and open a PR.
But I though it would be good to first ask if you would accept such a PR at all.

dbaumgarten:
Re-using the existing policies is an interesting idea. Being able to use policies with ingress-objects would bring a lot of interesting features. (For example currently you also need to use snippets fo IP-based white/black-listing.
It does however sound like something that requires quite a bit more effort, as the mixing of CRDs and Ingress-Objects is something completely new.

Currently we simply have fixed rate-limits in place via snippets. A http-snippet in the configmap for creating the zone and location-snippets in the ingresses to apply the zone to them.
We are also using the horizontal pod autoscaler to automatically scale the nginx-ingress according to the current load.

However in a recent incident we noticed that the rate-limiting becomes less effective the more ingress-replicas are running.
Therefore we started looking for possible solutions.

The "simplest" way would be to create a seperate controller that automatically modifies the configmap accordingly.
However as this would add more moving parts that would need to be maintained, I though it couldn't hurt to ask if that is something that could be added to the nginx-ingress-controller.

Rate-limiting is quite essential for reliably running a service. So supporting this properly would probably help quite a few people.

shaun-nx:
Hi @dbaumgarten this would be a great addition, and thank you for offering to do the PR as well! I'm going to pass this issue by the team before we suggest opening an PR.

brianehlert:
[...]

PR #4660, opened 2023-11-16
Add annotations for controlling request rate limiting

dbaumgarten:
### Proposed changes

This implements #4603.
Ingress resources can now have annotations to control request-rate-limiting.

Key points:
- Rate-limits apply per ingress object (all routes/paths of an ingress share a common limit_req_zone and have common settings)
- Supports different limits per path using minions. (Each mergeable minion can have separate rate-limiting configuration. Each minion uses it's own zone.
- By default, no rate-limit is applied. limit_req_status defaults to 429. Zone size defaults to 10m.

### Checklist

### DIFF
--- a/internal/configs/annotations_test.go
+++ b/internal/configs/annotations_test.go
@@ -182,7 +182,7 @@ func TestParseRateLimitAnnotations(t *testing.T) {
 		"nginx.org/limit-req-dry-run":     "true",
 		"nginx.org/limit-req-log-level":   "info",
 	}, NewDefaultConfigParams(false), context); len(errors) > 0 {
-		t.Errorf("Errors when parsing valid limit-req annotations")
+		t.Error("Errors when parsing valid limit-req annotations")
 	}
 
 	if errors := parseRateLimitAnnotations(map[string]string{
@@ -214,5 +214,4 @@ func TestParseRateLimitAnnotations(t *testing.T) {
 	}, NewDefaultConfigParams(false), context); len(errors) == 0 {
 		t.Errorf("No Errors when parsing invalid log level")
 	}
-
 }
--- a/internal/configs/annotations_test.go
+++ b/internal/configs/annotations_test.go
@@ -188,30 +188,30 @@ func TestParseRateLimitAnnotations(t *testing.T) {
 	if errors := parseRateLimitAnnotations(map[string]string{
 		"nginx.org/limit-req-rate": "200",
 	}, NewDefaultConfigParams(false), context); len(errors) == 0 {
-		t.Errorf("No Errors when parsing invalid request rate")
+		t.Error("No Errors when parsing invalid request rate")
 	}
 
 	if errors := parseRateLimitAnnotations(map[string]string{
 		"nginx.org/limit-req-rate": "200r/h",
 	}, NewDefaultConfigParams(false), context); len(errors) == 0 {
-		t.Errorf("No Errors when parsing invalid request rate")
+		t.Error("No Errors when parsing invalid request rate")
 	}
 
 	if errors := parseRateLimitAnnotations(map[string]string{
 		"nginx.org/limit-req-rate": "0r/s",
 	}, NewDefaultConfigParams(false), context); len(errors) == 0 {
-		t.Errorf("No Errors when parsing invalid request rate")
+		t.Error("No Errors when parsing invalid request rate")
 	}
 
 	if errors := parseRateLimitAnnotations(map[string]string{
 		"nginx.org/limit-req-zone-size": "10abc",
 	}, NewDefaultConfigParams(false), context); len(errors) == 0 {
-		t.Errorf("No Errors when parsing invalid zone size")
+		t.Error("No Errors when parsing invalid zone size")
 	}
 
 	if errors := parseRateLimitAnnotations(map[string]string{
 		"nginx.org/limit-req-log-level": "foobar",
 	}, NewDefaultConfigParams(false), context); len(errors) == 0 {
-		t.Errorf("No Errors when parsing invalid log level")
+		t.Error("No Errors when parsing invalid log level")
 	}
 }
--- a/internal/configs/annotations_test.go
+++ b/internal/configs/annotations_test.go
@@ -4,6 +4,9 @@ import (
 	"reflect"
 	"sort"
 	"testing"
+
+	networking "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestParseRewrites(t *testing.T) {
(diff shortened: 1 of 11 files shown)


jjngx:
@dbaumgarten could you please fix conflicts and push updates?

jjngx:
@dbaumgarten could you please resolve conflicts and push updates?

haywoodsh:
Hi @dbaumgarten I pulled down your PR and it worked well! 

One suggestion I have is that you can align the annotation names to the corresponding fields of the existing VirtuaServer [rateLimit policy](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#ratelimit) instead of the underlying NGINX directives, e.g. `nginx.org/limit-req` can be `nginx.org/limit-req-rate`, and `nginx.org/limit-req-status` can be `nginx.org/limit-req-reject-code`. What do you think?

Another thing I noticed is that, the `key` of `limit_req_zone` is hardcoded to `$binary_remote_addr`, unlike in the virtualserver policy where it is configurable, as well as other fields like `dryRun` and `logLevel` are available as configurable annotations. Is there any plan to include them in this PR?

internal/configs/ingress_test.go
```diff
@@ -568,6 +568,126 @@ func TestGenerateNginxCfgForMergeableIngressesForBasicAuth(t *testing.T) {
 	}
 }
 
+func TestGenerateNginxCfgForLimitReq(t *testing.T) {
+	t.Parallel()
+	cafeIngressEx := createCafeIngressEx()
+	cafeIngressEx.Ingress.Annotations["nginx.org/limit-req"] = "200r/s"
```
jjngx, line 574: What happens if the suffix 'r/s' or 'r/m' is omitted?
ref: [NGINX docs](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req)

```shell
The rate is specified in requests per second (r/s). If a rate of less than one request per second is desired, it is specified in request per minute (r/m). For example, half-request per second is 30r/m.
```

Is validation of the string representing number of requests required?

cc @haywoodsh @vepatel

jjngx (commented):
Left one question regarding validation.

dbaumgarten:
@haywoodsh aligning the annotations with the keys of VirtualServer sounds like a good idea. Will do.
Making the key for limit_req_zone configurable sounds like a good idea.
For my first draft I wanted to keep the number of added annotations to a minimum.
But if you think dryRun and LogLevel would be useful I can add these.

@jjngx Validating if the string ends in a proper unit should be pretty simple. Necessary or not it might save some people some headaches

jjngx:
@dbaumgarten thank you for contributing!

Could you please add validators and positive and negative tests for requests limiting and zone sizes annotations? The rules (suffixes) are listed [here](https://nginx.org/en/docs/syntax.html).
Please reach out if you need assistance.

cc @haywoodsh @j1m-ryan @vepatel

j1m-ryan:
Hi @dbaumgarten, heres from info from some manual testing I did which should help you in the validation for the `nginx.org/limit-req-rate` annotation
The `rate` argugment can accept two types of values `r/s` (requests per second) and `r/m` (requests per minute). The other time intervals will be rejected by nginx (for example `r/ms` requests per millisecond or `r/h` requests per hour would **not** be valid).
If no requests per time-interval is specificed, it assumes it is `r/s`, however, for the validation in NIC, I'd be happy if we were explicit in requiring either `r/s` or `r/m` for this annotation.
As a last piece of validation ensure the number before `r/s` or `r/m` is a whole number greater than 0. Thanks again.

dbaumgarten:
Hi,
PR has been rebased, annotations are aligned with the virtual-server resource and validations have been added

j1m-ryan:
Hi @dbaumgarten, looks good to me so far. Just failing the lint. Can you install `pre-commit` (via pip or brew), the run `pre-commit install` in your repo. It picked up a couple small formatting things. We might need to document this in the readme also.

```Running [/home/runner/golangci-lint-1.55.2-linux-amd64/golangci-lint run --out-format=github-actions --new-from-patch=/tmp/tmp-1857-Do5KUniBBV9u/pull.patch --new=false --new-from-rev=] in [] ...
  Error: cyclomatic complexity 18 of func `parseRateLimitAnnotations` is high (> 15) (gocyclo)
  Error: File is not `gofumpt`-ed (gofumpt)
  Error: `Occuring` is a misspelling of `Occurring` (misspell)
  Error: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)
  Error: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)```

File: internal/configs/annotations_test.go [hunk 2]
```diff
@@ -159,3 +162,57 @@ func TestMergeMasterAnnotationsIntoMinion(t *testing.T) {
...
+		"nginx.org/limit-req-key":         "${request_uri}",
+		"nginx.org/limit-req-burst":       "100",
+		"nginx.org/limit-req-delay":       "80",
+		"nginx.org/limit-req-no-delay":    "true",
+		"nginx.org/limit-req-reject-code": "429",
+		"nginx.org/limit-req-zone-size":   "11m",
+		"nginx.org/limit-req-dry-run":     "true",
+		"nginx.org/limit-req-log-level":   "info",
+	}, NewDefaultConfigParams(false), context); len(errors) > 0 {
+		t.Errorf("Errors when parsing valid limit-req annotations")
```
jjngx, line 185: May be worth to consider using `t.Error`

dbaumgarten: Done

dbaumgarten:
I fixed the linter-findings.
However I "fixed" the too-high complexity of parseRateLimitAnnotations by simply excluding it from gocyclo, as I don't really see any value in splitting it up.
Is that okay?

j1m-ryan:
I'm ok with this @dbaumgarten as the cyclometric complexity is coming from linear if's, rather than nested ifs, and this feature had a lot of annotations. So thats fine IMO.

As @jjngx pointed out, can you change the `t.Errorf`'s to `t.Errors`, where you are not using format strings. Theres still [a few examples of this in annotations_test.go](https://github.com/nginxinc/kubernetes-ingress/pull/4660/files#diff-b5221bb3303c813f81f9f0737dca24fcb4baa311776757ec09638a957c87ac94R188-R215)

Other than that its good to go imo

j1m-ryan (approved):

jjngx (approved):
👍🏻 thanks for contributing @dbaumgarten

ADubhlaoich (approved):
LGTM on the documentation side.

dbaumgarten:
There is so much going on on main, the PR is effectively always outdated...

Issue #4727, opened 2023-11-30
Virtual Server validation does not reject regex subroute paths that do not match virtual server path 


haywoodsh:
**Describe the bug**

In virtual server validation, if a virtual server route does not have the virtual server path as prefix, it will be rejected. In the specific where the subroute is a regex path, however, the validation fails to reject an invalid virtual server route definition, where a subroute path does not have the virtual server path as prefix.

**To Reproduce**
deploy the following virtual server and virtual server route:
Virtual server path:
```
- path: /team1/map/
  route: team1/map-ingress
```

virtual server route:
```
spec:
  host: xyz-x.fqdn
  upstreams:
    - name: map-svc
      service map-svc
      port: 80
  subroutes:
  - path: ~*^/map(/|$)(.*)
    action:
      proxy:
        upstream: map-svc
        rewritePath: /$2
```

**Expected behavior**
The virtual server route with incorrect subroute path should be rejected during validation

**Additional context**
This could be the possible code snippet that should return an error message but does not. 
https://github.com/nginxinc/kubernetes-ingress/blob/main/pkg/apis/configuration/validation/virtualserver.go#L1482-L1485

PR #4744, opened 2023-12-04
Fixed validation for VSR exact & regex subroutes

jo-carter:
Since a VSR may only contain a single subroute when using exact or regex matching skipping route validation here is incorrect.

Correctly configured VSRs for regex or exact will be validated (and terminate) earlier in function, ie. where VS path is regex + only 1 subroute + subroute syntax check passes.

Resolves ticket [4727](https://github.com/nginxinc/kubernetes-ingress/issues/4727)

### Checklist


haywoodsh:
Thank you @jo-carter for your PR!  Could you please add some unit test cases for the validation failures? 
https://github.com/nginxinc/kubernetes-ingress/blob/main/pkg/apis/configuration/validation/virtualserver_test.go#L821

jo-carter:
@haywoodsh Sure, I've added a couple of unit tests.

vepatel:
@jo-carter this change removes the capability of using regex or exact matches altogether

pkg/apis/configuration/validation/virtualserver_test.go
```diff
@@ -2394,6 +2394,46 @@ func TestValidateVirtualServerRouteSubroutesFails(t *testing.T) {
...
+				{	Path: "/abc",
+					Action: &v1.Action{
+						Pass: "test-1",
+					},
+				},
+			},
+			upstreamNames: map[string]sets.Empty{
+				"test-1": {},
+			},
+			pathPrefix: "/abc",
```
haywoodsh, line 2434: I think `pathPrefix` might not accurately describe what we're testing anymore since we could be using prefix/exact/regex path for route in virtual server here.

jo-carter:
@haywoodsh Do we need to squash / amend any of these commits ? They undo each other in places (the unit tests).

ADubhlaoich (commented):
LGTM for docs changes: not approving as most of the PR content is code-based.

jjngx (approved):

PR #4747, opened 2023-12-05
docs: add docker fallback for hugo and markdownlint

nginx-jack:
### Proposed changes

Adds a docker fallback for hugo and markdownlint. The goal is to make it possible to contribute to documentation without needing to install additional tooling (expect docker!).

### Checklist

- [x] I have checked that all unit tests pass after adding my changes NA

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,6 +1,46 @@
+HUGO?=hugo
+HUGO_IMG?=hugomods/hugo:0.115.3
+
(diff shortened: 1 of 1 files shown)


Jcahilltorre (approved):

ADubhlaoich (approved):
LGTM but you should be aware that NGINX Ingress Controller does not use conventional commits.

PR #4764, opened 2023-12-07
Dynamic reload of SSL certificates for NGINX Plus

oseoin:
### Proposed changes

Enable [lazy loading of SSL Certificates](https://www.nginx.com/blog/nginx-plus-r18-released/#dynamic-certificate-loading_disk) for NGINX Plus. This works by replacing the `/etc/nginx/secrets` with a `map` which returns that path by default for the variable `$secret_dir_path`.

E.g.
```
ssl_certificate /etc/nginx/secrets/default-my-secret.crt
```

Becomes:
```
map $nginx_version $secret_dir_path {
    default "/etc/nginx/secrets";
}

...

ssl_certificate $secret_dir_path/default-my-secret.crt
```

When this is enabled NIC checks secret updates and skips a reload of NGINX if there are no changes to the configuration files which reference that secret.

### Checklist


charts/nginx-ingress/values.yaml
```diff
@@ -463,6 +463,9 @@ controller:
   ## Configure root filesystem as read-only and add volumes for temporary data.
   readOnlyRootFilesystem: false
 
+  ## Enable dynamic reloading of certificates for NGINX Plus
+  enableSSLDynamicReload: true
```
vepatel, line 467: new flag is required to be added in chart README.md

vepatel:
new cli arg needs to be added in https://github.com/nginxinc/kubernetes-ingress/blob/dynamic-reload-ssl/docs/content/configuration/global-configuration/command-line-arguments.md

tests/suite/test_tls.py
```diff
@@ -131,3 +144,41 @@ def test_tls_termination(self, kube_apis, ingress_controller_endpoint, test_name
         replace_secret(kube_apis.v1, tls_setup.secret_name, test_namespace, tls_setup.new_secret_path)
         wait_before_test(1)
         assert_gb_subject(ingress_controller_endpoint, tls_setup.ingress_host)
+
+        count_after = get_reload_count(tls_setup.metrics_url)
+        reloads = count_after - count_before
+        expected_reloads = 8
```
vepatel, line 150: this count will always be deterministic?

cmd/nginx-ingress/flags.go
```diff
@@ -269,6 +275,11 @@ func parseFlags() {
 	if *ingressLink != "" && *externalService != "" {
 		glog.Fatal("ingresslink and external-service cannot both be set")
 	}
+
+	if *enableDynamicSSLReload && !*nginxPlus {
+		glog.V(3).Infof("%s flag requires -nginx-plus and will not be enabled", dynamicSSLReloadParam)
```
vepatel, line 280: should this be fatal? similar to maybe:
`*enableAppProtect && !*nginxPlus`

oseoin: Since it is opt-out rather than opt-in we would fail for OSS with fatal, so I think it should stay info?

vepatel:
https://github.com/nginxinc/kubernetes-ingress/actions/runs/7170099267/job/19524954747?pr=4764#step:3:1741 seems to have some assertions based on ssl certs resulting in failure

pdabelf5 (approved):

tests/suite/test_tls.py
oseoin: This has changed in the latest updates, it now only expects a 1 or 0 depending on is dynamic loads is enabled.

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -617,5 +623,88 @@ def test_secure_tcp_request_load_balanced(
...
+        print(f"replacing: {sec_name} in {transport_server_setup.namespace}")
+        replace_secret(kube_apis.v1, sec_name, transport_server_setup.namespace, src_new_sec_yaml)
+        wait_before_test()
+        print(f"waited to {datetime.now().strftime('%m/%d/%Y, %H:%M:%S')}")
+        count_after = get_reload_count(transport_server_setup.metrics_url)
+        reloads = count_after - count_before_replace
+        expected_reloads = 0
+        assert reloads == expected_reloads, f"expected {expected_reloads} reloads, got {reloads}"
+
+        # self.restore_ts(kube_apis, transport_server_setup)
```
vepatel, line 709: might need to uncomment this line to reset TS back to its original state

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -617,5 +623,88 @@ def test_secure_tcp_request_load_balanced(
...
+                    "-v=3",
+                ],
+            },
+            {"example": "transport-server-tcp-load-balance"},
+        )
+    ],
+    indirect=True,
+)
+class TestTransportServerTcpLoadBalanceDynamicReload:
+    def test_secure_tcp_request_load_balanced(
```
vepatel, line 664: I think we need to change the test name here to better reflect what's being tested, something like:
`test_dynamic_reload()` should be sufficient

tests/suite/test_transport_server_tcp_load_balance.py
```diff
@@ -617,5 +623,88 @@ def test_secure_tcp_request_load_balanced(
...
+        self, kube_apis, crd_ingress_controller, transport_server_setup, ingress_controller_prerequisites
+    ):
+        """
+        Sends requests to a TLS enabled load balanced TCP service.
```
vepatel, line 668: need to update the description inline with test name.

oseoin:
> ```
>   File "/workspace/tests/suite/test_virtual_server_tls.py", line 201, in test_tls_termination
>     assert reloads == expected_reloads, f"expected {expected_reloads} reloads, got {reloads}"
> AssertionError: expected 0 reloads, got 1
> assert 1 == 0
> ```

Should hopefully be resolved in the latest commit, always passed locally but upstream ordering for the virtualserver resource was not stable before now.

shaun-nx (approved):
👏

vepatel (approved):

PR #4831, opened 2023-12-21
Add the ability to have Nginx version checks in templates

oseoin:
### Proposed changes

Add the ability to add version dependent template elements via a new `nginx.Version` struct which is available in the main

### Checklist


jjngx (approved):
👍🏻

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -346,3 +346,9 @@ stream {
 
     include /etc/nginx/stream-conf.d/*.conf;
 }
+
+{{- if and (.NginxVersion.IsPlus) (.NginxVersion.PlusGreaterThanOrEqualTo "nginx-plus-r31") }}
```
vepatel, line 350: do we need `(.NginxVersion.IsPlus)` here as string in next condition is only possible on plus versions?

cmd/nginx-ingress/main.go
```diff
@@ -79,7 +79,7 @@ func main() {
 		appProtectVersion = getAppProtectVersionInfo()
 	}
 
-	updateSelfWithVersionInfo(kubeClient, version, nginxVersion, appProtectVersion)
+	updateSelfWithVersionInfo(kubeClient, version, nginxVersion.String(), appProtectVersion)
```
vepatel, line 82: is its possible to check if pod labels are being assigned correctly after this change?

oseoin: Not sure if we have a test for it, but locally seems fine: `app.nginx.org/version: 1.25.3-nginx-plus-r31`

j1m-ryan (approved):

vepatel (approved):

Issue #4665, opened 2023-11-16
Extend use-cluster-ip to Ingress Resource


brianehlert:
use-cluster-ip is currently a key capability to target the service endpoint instead of routing traffic directly to the backend service pods.
This is important for any customer running a service mesh behind the ingress controller.  
It is also important is some edge use cases with dynamically scaling backend applications and their interaction with queue.

To expand the appeal of the project use-cluster-ip needs to be extended to support the Ingress resource using an annotation.

AC:
- introduce the annotation `nginx.org/use-cluster-ip`
- mirror the behavior of use-cluster-ip from the VS and VR resources https://docs.nginx.com/nginx-ingress-controller/configuration/virtualserver-and-virtualserverroute-resources/#common-parts-of-the-virtualserver-and-virtualserverroute
- This cannot be applied to a master Ingress
- This can be applied to an Ingress that is not a master or minion
- If this is applied to a minion Ingress then it applies to only that minion (like it would for VirtualServerRoute/VirtualServer)

PR #4862, opened 2024-01-03
Add use-cluster-ip annotation for ingress resources

j1m-ryan:
### Proposed changes

#4665 

This adds the `nginx.org/use-cluster-ip` boolean annotation to the ingress resource. It defaults to `false`.

### Manual Tests
- [x] Works for an ingress that is neither minion nor master 
image
- [x] Is ingored when a master ingress is annotated with it 
```bash
E0103 12:08:06.928220       1 ingress.go:619] Ingress Resource default/cafe-ingress-master with the annotation 'nginx.org/mergeable-ingress-type' set to 'master' cannot contain the 'nginx.org/use-cluster-ip' annotation(s). They will be ignored
```
- [x] Does not follow minion inheritence
- [x] Works for minions on the individual level (coffee minion is annotated with `use-cluster-ip: true`, but the tea minion is not below)
image

### Checklist


ADubhlaoich (approved):

oseoin (approved):

jjngx (approved):
👍🏻

pdabelf5:
@j1m-ryan has this been tested outside the default namespace?

j1m-ryan:
> @j1m-ryan has this been tested outside the default namespace?

I haven't, good catch @pdabelf5 , I'll do this now

shaun-nx (approved):

btsuhako:
this is a welcomed feature, we're planning to use this once it's released.

for context in the issue below, i'm using the `edge` release of the ingress controller with no Helm values overrides, everything is set to the default

we're doing some testing, and possibly found a bug. that or some documentation needs to be added

the following manifests work fine:
```yaml
---
kind: Ingress
metadata:
  name: podinfo
  namespace: podinfo
  annotations:
    nginx.org/use-cluster-ip: "True"
spec:
  ingressClassName: nginx
  rules:
  - host: localhost
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: podinfo
            port:
              number: 9898
---
apiVersion: v1
kind: Service
metadata:
  name: podinfo
  labels:
    helm.sh/chart: podinfo-6.5.3
    app.kubernetes.io/name: podinfo
    app.kubernetes.io/version: "6.5.3"
    app.kubernetes.io/managed-by: Helm
spec:
  type: ClusterIP
  ports:
    - port: 9898
      targetPort: http
      protocol: TCP
      name: http
    - port: 9999
      targetPort: grpc
      protocol: TCP
      name: grpc
  selector:
    app.kubernetes.io/name: podinfo
```

the following manifest does not work. note the comment in the `backend.service.port` part of the Ingress

```yaml
---
kind: Ingress
metadata:
  name: podinfo
  namespace: podinfo
  annotations:
    nginx.org/use-cluster-ip: "True"
spec:
  ingressClassName: nginx
  rules:
  - host: localhost
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: podinfo
            port:
              name: http # cannot use named port
```

it seems like the ingress controller won't generate a valid config with the above

```shell
$ k exec -it deploy/nginxinc-ingress-nginx-ingress-controller -c nginx-ingress -- nginx -T
2024/01/22 21:34:39 [emerg] 47#47: invalid port in upstream "podinfo.podinfo.svc.cluster.local:0" in /etc/nginx/conf.d/podinfo-podinfo.conf:3
nginx: [emerg] invalid port in upstream "podinfo.podinfo.svc.cluster.local:0" in /etc/nginx/conf.d/podinfo-podinfo.conf:3
nginx: configuration file /etc/nginx/nginx.conf test failed
command terminated with exit code 1
```

if it helps, the following does work:
```yaml
---
kind: Ingress
metadata:
  name: podinfo
  namespace: podinfo
  annotations:
    nginx.org/use-cluster-ip: "False" # use Pod IP
spec:
  ingressClassName: nginx
  rules:
  - host: localhost
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: podinfo
            port:
              name: http # named port now works
```

is this intentional behavior of the `nginx.org/use-cluster-ip` annotation or a bug? if it's a bug, let me know if I need to create an issue. It seems like something is lost when mapping the named port to container the port number when using the `nginx.org/use-cluster-ip: "True"` annotation

j1m-ryan:
Hi @btsuhako, thank you for reporting this. I can see in the implementation that the `name`'d ports are not handled, and it defaults to `number`. I have replicated the error you experienced myself. I made an issue for it here #4970 
Thanks again.

brianehlert:
Does the trailing comment on the port line cause any error by itself?

btsuhako:
> Does the trailing comment on the port line cause any error by itself?

The error happens with and without the comment on that line. I added the comment to call out the difference from the working manifest

Issue #4877, opened 2024-01-09
Epic - Adoption Telemetry


shaun-nx:
As a maintainer of NIC, I would like to provide the product with a means to collect Product Telemetry so that I can better understand how our users configure our product

**Data Collection**
Some data points that NIC is interested in collecting has been implemented by the [nginx-cluster-connector](https://github.com/nginxinc/nginx-cluster-connector)

Details can be found in the [data-design](https://github.com/nginxinc/nginx-cluster-connector/blob/main/data-design.md) for the cluster connector

```[tasklist]
### Tasks
- [x] Update all `int` data types to `int64`
- [ ] Confirm data is sent to XCDF devendpoint
- [ ] Document process for changes/updates to schema
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/4863
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/4902
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/4903
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5123
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/4883
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5136
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5143
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5151
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/4882
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5144
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5135
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5137
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5138
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5139
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5140
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5141
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/5142
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/4884
- [ ] https://github.com/nginxinc/kubernetes-ingress/issues/4898
- [x] Define Schema for NIC
- [ ] Enable Telemetry Job to collect version of App protect
- [x] Add Telemetry opt-out option in Helm
[...]

PR #4896, opened 2024-01-10
Telemetry Job

shaun-nx:
### Proposed changes

This PR adds the base telemetry job that will be responsible for collecting telemetry data for each deployment of the Ingress Controller.

The current implementation of the Job writes collected, fake data to a dummy exporter.
 
Tasks to gather and report specific telemetry data points are outlined here: https://github.com/nginxinc/kubernetes-ingress/issues/4877

**Tests**
- [x] Ensure only the leader pod reports data:
```
I0124 15:43:17.788228       1 leaderelection.go:260] successfully acquired lease nginx-ingress/nginx-ingress-leader-election
I0124 15:43:17.789698       1 leader.go:60] started leading
I0124 15:43:17.789842       1 leader.go:77] updating VirtualServer and VirtualServerRoutes status
I0124 15:43:17.791000       1 telemetry.go:104] Collecting telemetry data
I0124 15:43:17.791729       1 telemetry.go:113] Exported telemetry data

```
- [x] Ensure new leader pod start reporting if leader is lost
- [x] Ensure telemetry data is not reported when `enable-telemetry-reporting` flag is set to `false`
- [x] Ensure telemetry is reported at the period that it is configured at
- [x] Update Helm values to add new CLI arguments
- [x] Confirm telemetry is not reported when `enable-telemetry-reporting` it set to `false` in Helm values
- [x] Update Helm install documentation

### Checklist


ADubhlaoich:
Should this be mentioned somewhere in user-facing documentation?

shaun-nx:
> Should this be mentioned somewhere in user-facing documentation?

There will be documentation added as part of the overall implementation.
This specific piece of work is just to create the base "dummy" job. It currently doesn't actually export any data.
It still needs to be fleshed out, but the issue is tracked here: https://github.com/nginxinc/kubernetes-ingress/issues/4898

internal/telemetry/exporter_test.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func TestExportData(t *testing.T) {
+	t.Parallel()
+
+	exporter := NewLogExporter()
+
+	err := exporter.Export(context.Background(), Data{})
+	if err != nil {
+		t.Fatalf("Expeceted no error, but got %s", err.Error())
```
pdabelf5, line 15:
```suggestion
		t.Fatalf("Expected no error, but got %s", err.Error())
```

vepatel:
@jjngx maybe we can implement a something like https://github.com/nginxinc/nginx-gateway-fabric/blob/9f4248e0d629f7953fde17f4f03bd95034306f7f/internal/mode/static/telemetry/telemetryfakes/fake_exporter.go for completeness in unit testing

cmd/nginx-ingress/main.go
```diff
@@ -790,9 +794,12 @@ func processConfigMaps(kubeClient *kubernetes.Clientset, cfgParams *configs.Conf
...
 func updateSelfWithVersionInfo(kubeClient *kubernetes.Clientset, version, nginxVersion, appProtectVersion string, maxRetries int, waitTime time.Duration) {
-	nginxVer := strings.TrimSuffix(strings.Split(nginxVersion, "/")[1], "\n")
-	replacer := strings.NewReplacer(" ", "-", "(", "", ")", "")
-	nginxVer = replacer.Replace(nginxVer)
+	var nginxVer string
+	if nginxVersion != "" {
+		nginxVer = strings.TrimSuffix(strings.Split(nginxVersion, "/")[1], "\n")
+		replacer := strings.NewReplacer(" ", "-", "(", "", ")", "")
+		nginxVer = replacer.Replace(nginxVer)
+	}
```
pdabelf5, line 802: How is this change related to the telemetry work?

jjngx: This was addressed in a separate PR.

cmd/nginx-ingress/flags.go
```diff
@@ -201,6 +203,10 @@ var (
 
 	enableDynamicSSLReload = flag.Bool(dynamicSSLReloadParam, true, "Enable reloading of SSL Certificates without restarting the NGINX process.")
 
+	enableTelemetryReporting = flag.Bool("enable-telemetry-reporting", true, "Enable gathering and reporting of product related telemetry.")
```
oseoin, line 206: Do we have a task for controlling this via helm or should it be added as part of this PR?

jjngx: We don't have helm-related tasks atm. We would need to add this to Helm before releasing v3.5.

shaun-nx: Thanks @oseoin nice catch. I'll make this change as part of this PR.

j1m-ryan (approved):

cmd/nginx-ingress/flags.go
jjngx: Task is also added to the list.

jjngx (approved):

internal/nginx/fake_manager.go
```diff
@@ -103,7 +103,7 @@ func (fm *FakeManager) CreateDHParam(_ string) (string, error) {
 // Version provides a fake implementation of Version.
 func (*FakeManager) Version() Version {
 	glog.V(3).Info("Printing nginx version")
-	return Version{}
+	return NewVersion("nginx version: nginx/1.25.3 (nginx-plus-r31)")
```
vepatel, line 106: is this change related?

jjngx: change related

jjngx (changes requested):
We need consensus on how telemetry job time period is handled - as string (that comes at the moment from flags) or as `time.Duration`. If the latter, then unit tests would need to be updated.

vepatel (approved):

PR #4946, opened 2024-01-19
Build test image and re-use in Smoke tests

pdabelf5:
### Proposed changes

- Build test image based in hash of `requirements.txt` file and save to GCR for re-use in further pipelines
- Mount tests into test-runner image so latest tests are always used.

### Checklist


.github/workflows/build-test-image.yml
```diff
@@ -0,0 +1,61 @@
+name: Build Test Image
+
+on:
+  schedule:
+    - cron: "0 03 * * *" # run every day at 03:00 UTC
```
pdabelf5, line 5: Do you think we need to have a scheduled rebuild of the image if it also will be built in a PR?

vepatel: no I think if we're building on hash updates then it should be okay 👍🏼

jjngx (approved):

PR #5009, opened 2024-01-30
Add version formatting for NGINX pod label

jjngx:
### Proposed changes

This PR adds a method for formatting NGINX version.

### Checklist


pdabelf5 (commented):
Would it make sense to have the version functionality in its own file?

shaun-nx (approved):

internal/nginx/version.go
```diff
@@ -0,0 +1,100 @@
+package nginx
```
vepatel, line 1: seems like few lines here aren't covered by version_test.go, Is it possible to add few negatives cases?

internal/nginx/version.go
```diff
@@ -0,0 +1,100 @@
...
+	return nv
+}
+
+// String returns the raw Nginx version string from `nginx -v`.
+func (v Version) String() string {
+	return v.Raw
+}
+
+// Format returns a string representing Nginx version.
+func (v Version) Format() string {
```
pdabelf5, line 56: This is to be called in `main.go` ?

jjngx: Yes, will be dressed in a separate PR

> This is to be called in `main.go` ?

Yes, will be dressed in a separate PR

pdabelf5 (approved):

PR #5057, opened 2024-02-09
fix: ensure subsequent calls of "make clean" succeed

defanator:
### Proposed changes

Simple fix to Makefile ensuring that any subsequent invocations of `make clean` will succeed.

### Checklist


j1m-ryan (approved):

Makefile
```diff
@@ -171,8 +171,8 @@ push: ## Docker push to PREFIX and TAG
 
 .PHONY: clean
 clean:  ## Remove nginx-ingress binary
-	-rm nginx-ingress
-	-rm -r dist
+	-rm -f nginx-ingress
+	-rm -rf dist
```
pdabelf5, line 175: @defanator thank you for contributing.  The leading `-` in the Makefile means a failing command here does not fail the target (see `ignored` below).  
```
$ make clean
rm nginx-ingress
rm: nginx-ingress: No such file or directory
make: [clean] Error 1 (ignored)
rm -r dist
rm: dist: No such file or directory
make: [clean] Error 1 (ignored)
$ echo $?
0
```
However, using `-f` is more explicit and will prevent `No such file or directory` messages.  Maybe prefixing with `@` would be a good choice here with your suggestion.

pdabelf5 (approved):

jjngx (approved):

Issue #4902, opened 2024-01-10
Count of VirtualServers in watched namespaces


shaun-nx:
As part of the Adoption Telemetry, we want to gather and report a count of deployed VirtualServers that are observed by a deployment of Ingress Controller

Acceptance Criteria:
- Gather a count of VirtualServer that are observed by the Ingress Controller
- Ensure the count of VirtualServers are logged 
```[tasklist]
### Tasks
- [ ] Gather a count of VirtualServer that are observed by the Ingress Controller
- [ ] Ensure the count of VirtualServers are logged
```

PR #5095, opened 2024-02-14
Telemety/vs and ts counts

shaun-nx:
### Proposed changes

This change updates the base telemetry job to collect a count of VirtualServer and TransportServer resources that are observed by the Ingress Controller.

Closes https://github.com/nginxinc/kubernetes-ingress/issues/4902 and https://github.com/nginxinc/kubernetes-ingress/issues/4903

### Testing

Test cases in `collector.go` cover scenarios where:
1. Resources in a watched namespace is counted
2. Resources not in a watched namespace is not counted
3. Resources deployed across multiple namespaces are counted if those namespaces are watched

#### Additional changes
Renamed `telemetry.go` to `collector.go`
Moved code related to the `DiscardExporter` out of `collector.go` and into `exporter.go`

Note: Tests in `collector_test.go` cover calls to the Export function in `exporter.go`

### Checklist


internal/k8s/controller.go
```diff
@@ -277,20 +277,27 @@ func NewLoadBalancerController(input NewLoadBalancerControllerInput) *LoadBalanc
...
+		Namespaces:            lbc.namespaceList,
+	}
+
 	// NIC Telemetry Reporting
 	if input.EnableTelemetryReporting {
 		lbc.telemetryChan = make(chan struct{})
 		collector, err := telemetry.NewCollector(
-			telemetry.WithTimePeriod(input.TelemetryReportingPeriod),
+			collectorConfig,
+			telemetry.WithTimePeriod("24h"),
```
jjngx, line 293: since the new collector takes a config struct, it may be worth to consider adding the time period param as a field. this way `NewCollector` would take only one param - the config struct. less paperwork.

internal/k8s/controller.go
```diff
@@ -277,20 +277,27 @@ func NewLoadBalancerController(input NewLoadBalancerControllerInput) *LoadBalanc
 		lbc.externalDNSController = ed_controller.NewController(ed_controller.BuildOpts(context.TODO(), lbc.namespaceList, lbc.recorder, lbc.confClient, input.ResyncPeriod, isDynamicNs))
 	}
 
+	glog.V(3).Infof("Nginx Ingress Controller has class: %v", input.IngressClass)
+
+	collectorConfig := telemetry.CollectorConfig{
```
jjngx, line 282: collector config needs to be created only when user opts in, correct? if yes, maybe to move it after L289 - `if input.EnableTelemetryReporting` and take care about config only when user wants or not telemetry?

internal/telemetry/collector.go
```diff
@@ -110,26 +97,47 @@ func (c *Collector) Collect(ctx context.Context) {
...
+func (c *Collector) BuildReport(ctx context.Context) (TraceData, error) {
+	dt := TraceData{}
+	var err error
+
+	for _, namespace := range c.Config.Namespaces {
+		dt.VirtualServers += c.GetVirtualServerCount(ctx, namespace)
+		dt.TransportServers += c.GetTransportServerCount(ctx, namespace)
+	}
+
+	return dt, err
```
jjngx, line 113: are we planning to handle errors here? atm the err value it's always `nil`. may be worth to simplify the func signature, unless the err would be used in the future.

internal/telemetry/collector.go
```diff
@@ -110,26 +97,47 @@ func (c *Collector) Collect(ctx context.Context) {
...
-	// Placeholder function
-	return 0
+// GetVirtualServerCount returns the number of VirtualServers in a namespace.
+func (c *Collector) GetVirtualServerCount(ctx context.Context, namespace string) int {
+	count, err := c.Config.CustomK8sClientReader.
+		K8sV1().
+		VirtualServers(namespace).
+		List(ctx, meta_v1.ListOptions{})
+	if err != nil {
+		glog.Errorf("unable to list VirtualServers in namespace %s Error: %x", namespace, err)
```
jjngx, line 128: if there is an error here it may be worth to consider returning 0

internal/telemetry/collector.go
```diff
@@ -110,26 +97,47 @@ func (c *Collector) Collect(ctx context.Context) {
...
 }
 
-// GetVSCount returns number of VirtualServers in watched namespaces.
-//
-// Note: this is a placeholder function.
-func (c *Collector) GetVSCount() int {
-	// Placeholder function
-	return 0
+// SetConfig is used to overwrite the existing config for the Collector.
+func (c *Collector) SetConfig(cfg CollectorConfig) {
```
jjngx, line 117: is this method used only to support a table tests, or we do have a case when we want to update the collector config at runtime?

shaun-nx: This was added to support the tests. It makes me think the tests could be structured better. Any thoughts on how to go about that?

jjngx: One test per one config sounds reasonable

internal/telemetry/collector.go
shaun-nx: That should work. If we're not able to list a resource in a certain namespace, then it should just add `0`

jjngx (approved):

PR #5110, opened 2024-02-16
Draft: issue 4837 - chore: main.go refactor

mrajagopal:
### Proposed changes

### Checklist


cmd/nginx-ingress/main.go
```diff
@@ -241,70 +256,72 @@ func createConfigAndKubeClient() (*rest.Config, *kubernetes.Clientset) {
...
+}
 
-	return config, kubeClient
+func getKubeClient(config *rest.Config) (kubeClient *kubernetes.Clientset, err error) {
+	kubeClient, err = kubernetes.NewForConfig(config)
+	return kubeClient, err
 }
 
-func kubernetesVersionInfo(kubeClient kubernetes.Interface) {
+func confirmMinimumkubernetesVersionCriteria(kubeClient kubernetes.Interface) (err error) {
```
j1m-ryan, line 271: capitalise the K in the function name

jjngx:
Hi @mrajagopal, I want to ask a more general question about what is the intention behind this PR?

mrajagopal:
> Hi @mrajagopal, I want to ask a more general question about what is the intention behind this PR?

@jjngx , the objective of this PR is to give test coverage for `main.go`.
This particular commit attempts to improve functions by:
* returning the error instead of throwing a fatal error
* Split functions to improve readability and testability

cmd/nginx-ingress/main.go
mrajagopal: Sorted with a further rename to `confirmMinimumK8sVersionCriteria`.

cmd/nginx-ingress/main.go
```diff
@@ -58,43 +58,90 @@ func main() {
...
+	kubeClient, err := getKubeClient(config)
+	if err != nil {
+		glog.Fatalf("Failed to create client: %v.", err)
+	}
 
-	validateIngressClass(kubeClient)
+	err = confirmMinimumK8sVersionCriteria(kubeClient)
+	if err != nil {
+		glog.Fatal(err)
+	}
```
shaun-nx, line 74: Since the error that is used here is never re-used outside of the scope of the return, you can also format the check like this. This is more of a `go-lang` way of doing things. It's useful when you want reduce the scope of the return value (in this case the error)
```suggestion
	if err := confirmMinimumK8sVersionCriteria(kubeClient); err != nil {
		glog.Fatal(err)
	}
```

jjngx:
An idea that may be worth considering:

We have many functions in the `main` that call `log.Fatal` in case of an error. It maybe worth considering to the `must` prefix (eg. `mustParse()`, `mustXYZ()`), check the error and call `fatal` inside the functions. This way we would communicate to the reader that the func either works or NIC won't start, we would have less error checking in the `main`.

mrajagopal:
> An idea that may be worth considering:
> 
> We have many functions in the `main` that call `log.Fatal` in case of an error. It maybe worth considering to the `must` prefix (eg. `mustParse()`, `mustXYZ()`), check the error and call `fatal` inside the functions. This way we would communicate to the reader that the func either works or NIC won't start, we would have less error checking in the `main`.

This is a good idea, is shall adopt this.

mrajagopal:
> > An idea that may be worth considering:
> > We have many functions in the `main` that call `log.Fatal` in case of an error. It maybe worth considering to the `must` prefix (eg. `mustParse()`, `mustXYZ()`), check the error and call `fatal` inside the functions. This way we would communicate to the reader that the func either works or NIC won't start, we would have less error checking in the `main`.
> 
> @jjngx , this is a good idea, I shall adopt this where possible.

Having reviewed all these functions, changing them to not return an error but log the log fatal within the function would make the unit-tests for these function redundant. Therefore, for the time being we can continue with returning an error.

jjngx:
> > > An idea that may be worth considering:
> > > We have many functions in the `main` that call `log.Fatal` in case of an error. It maybe worth considering to the `must` prefix (eg. `mustParse()`, `mustXYZ()`), check the error and call `fatal` inside the functions. This way we would communicate to the reader that the func either works or NIC won't start, we would have less error checking in the `main`.
> > 
> > 
> > @jjngx , this is a good idea, I shall adopt this where possible.
> 
> Having reviewed all these functions, changing them to not return an error but log the log fatal within the function would make the unit-tests for these function redundant. Therefore, for the time being we can continue with returning an error.

Excellent, what's better than reducing complexity by removing or not adding more code 👍🏻

danielnginx:
We will get back to this work in the future, thank you for the contribution.

PR #5176, opened 2024-02-27
Add NIC metadata (name and version) to telemetry data

jjngx:
### Proposed changes

This PR adds NIC metadata (Name, Version) to telemetry.

NIC log example:
```shell
I0227 16:58:42.719659       1 collector.go:85] Collecting telemetry data
I0227 16:58:42.746579       1 collector.go:94] Exported telemetry data: {ProjectMeta:{Name:NIC Version:3.5.0-SNAPSHOT} NICResourceCounts:{VirtualServers:0 VirtualServerRoutes:0 TransportServers:0} NodeCount:1 ClusterID:c232007d-76bf-48a5-8dc2-60e017467f0f K8sVersion:v1.29.2 Arch:amd64}
```

### Checklist

### DIFF
--- a/internal/telemetry/collector.go
+++ b/internal/telemetry/collector.go
@@ -53,6 +53,9 @@ type CollectorConfig struct {
 	Period time.Duration
 
 	Configurator *configs.Configurator
+
+	// Version represents NIC version.
+	Version string
 }
 
 // NewCollector takes 0 or more options and creates a new TraceReporter.
@@ -93,9 +96,15 @@ func (c *Collector) Collect(ctx context.Context) {
 
 // BuildReport takes context and builds report from gathered telemetry data.
 func (c *Collector) BuildReport(ctx context.Context) (Data, error) {
-	d := Data{
-		Arch: runtime.GOARCH,
+	pm := ProjectMeta{
+		Name:    "NIC",
+		Version: c.Config.Version,
 	}
+	d := Data{
+		ProjectMeta: pm,
+		Arch:        runtime.GOARCH,
+	}
+
 	var err error
 
 	if c.Config.Configurator != nil {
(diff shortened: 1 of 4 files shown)


oseoin (approved):

internal/telemetry/collector.go
```diff
@@ -93,9 +96,15 @@ func (c *Collector) Collect(ctx context.Context) {
 
 // BuildReport takes context and builds report from gathered telemetry data.
 func (c *Collector) BuildReport(ctx context.Context) (Data, error) {
+	pm := ProjectMeta{
+		Name:    "NIC",
```
pdabelf5, line 100: Can we use acronyms/abbreviations on docs/logs etc?

pdabelf5 (approved):

internal/telemetry/collector.go
jjngx: Using this name ("NIC") mirrors the NGF approach. We can use full name, but my understanding is that the "NIC" will be a "code name" in the telemetry payload.

vepatel: Can we make `NGINX Ingress Controller` from https://github.com/nginxinc/kubernetes-ingress/blob/main/cmd/nginx-ingress/main.go#L57 a var and then pass it to collector?

jjngx: "NIC" is a codename for using in telemetry. It's in line with NGF approach: [link](https://github.com/nginxinc/nginx-gateway-fabric/blob/57d6d6a2527264c9ce8e75eab7e7a6499cf29556/internal/mode/static/telemetry/collector.go#L125L137)

vepatel (approved):

PR #5205, opened 2024-03-06
Preserve valid listeners when invalid listeners are present in GlobalConfiguration

haywoodsh:
### Proposed changes

Improve validation process for the GlobalConfiguration. The updated logic will allow the application of a GlobalConfiguration resource even if it includes invalid listeners. Invalid listeners will be disregarded while the valid ones will remain, which ensures valid VS and TS configs are operational and continue to accept traffics. NIC will emit a `Warning` Kubernetes event for the GlobalConfiguration with the reason `AddedOrUpdatedWithError` for any invalid listeners.

### Checklist


jjngx (approved):

docs/content/configuration/global-configuration/globalconfiguration-resource.md
```diff
@@ -124,12 +124,7 @@ If a resource is not rejected (it doesn't violate the structural schema), the In
 
 #### Comprehensive Validation
 
-The Ingress Controller validates the fields of a GlobalConfiguration resource. If a resource is invalid, the Ingress Controller will not use it. Consider the following two cases:
-
-1. When the Ingress Controller pod starts, if the GlobalConfiguration resource is invalid, the Ingress Controller will fail to start and exit with an error.
-1. When the Ingress Controller is running, if the GlobalConfiguration resource becomes invalid, the Ingress Controller will ignore the new version. It will report an error and continue to use the previous version. When the resource becomes valid again, the Ingress Controller will start using it.
-
-**Note**: If a GlobalConfiguration is deleted while the Ingress Controller is running, the controller will keep using the previous version of the resource.
+The Ingress Controller validates the fields of a GlobalConfiguration resource. If a GlobalConfiguration resource is partially invalid, the Ingress Controller use the valid listeners and emit events about invalid listeners.
```
shaun-nx, line 127: > If a GlobalConfiguration resource is partially invalid, the Ingress Controller use the valid listeners and emit events about invalid listeners.

How do we behave if all of the listeners are invalid?

haywoodsh: The behaviour should be the same as when only some listeners are invalid, which it would emit a `Warning` status but not `Rejected`.

shaun-nx (approved):

Issue #5082, opened 2024-02-13
Warning in NGINX logs which says "using uninitialized `default_connection_header` variable" when using `rewritePath`


j1m-ryan:
**Describe the bug**
There is a warning in the NGINX logs which looks like this
```
2024/02/13 12:30:28 [warn] 89#89: *23 using uninitialized "default_connection_header" variable, client: 10.244.0.1, server: cafe.example.com, request: "GET /coffee HTTP/1.1", host: "cafe.example.com"
```

**To Reproduce**

1. Modify the `examples/custom-resources/traffic-splitting/cafe-virtual-server.yaml` file to look like this
```yaml
apiVersion: k8s.nginx.org/v1
kind: VirtualServer
metadata:
  name: cafe
spec:
  host: cafe.example.com
  upstreams:
  - name: coffee-v1
    service: coffee-v1-svc
    port: 80
  - name: coffee-v2
    service: coffee-v2-svc
    port: 80
  routes:
  - path: /coffee
    splits:
    - weight: 50
      action:
        proxy:
          upstream: coffee-v1 
          rewritePath: /coffee
    - weight: 50
      action:
        pass: coffee-v2

```
2. `kaf examples/custom-resources/traffic-splitting`
3.  `curl cafe.example.com/coffee` a few times
4.  See error in the logs about 50% of the time (due to it being a 50 50 split).

**Expected behavior**
Expectation is that I should be able to `rewrite` inside these location blocks without a warning in the logs about an uninitialized variable

**Your environment**

- Version of the Ingress Controller - `3.5.0-snapshot`
- Version of Kubernetes -  Client Version: v1.28.4 Server Version: v1.28.3
- Kubernetes platform - minikube
- Using NGINX Plus

**Additional context**
@defanator experienced this and provided a few potential solutions here https://github.com/defanator/rewrite-using-uninitialized-variable
Solution 1: move the `set` before the `rewrite` in the go tmpl files
Solution 2: use a `map` instead of a `set`

relevant part of the generated nginx config
```nginx
location /internal_location_splits_0_split_0 {
        set $service "coffee-v1-svc";
        status_zone "coffee-v1-svc";
        internal;


        rewrite ^ $request_uri_no_args;
        rewrite "^/coffee(.*)$" "/coffee$1" break;
        proxy_connect_timeout 60s;
        proxy_read_timeout 60s;
        proxy_send_timeout 60s;
        client_max_body_size 1m;

        proxy_buffering on;
[...]

PR #5211, opened 2024-03-07
Move set above rewrite to fix uninitialized variable

AlexFenlon:
### Proposed changes

Moved set default_connection_header above rewrites to fix a warning in the logs about that variable being uninitialized.

I did extensive testing before making this change to make sure this won't affect anyone. Using my own web socket app and  the ` examples/custom-resources/traffic-splitting/cafe-virtual-server.yaml` in the issue to test what changes are made if I moved this set before the rewrites.

Doing a websocket connecton via `npx wscat -c ws://cafe.example.com:80` does not give a warning which means it's only curl giving the warning

Curl hasn't recognised websockets due to curl being designed for standard http only and therefore gives the warning as http_upgrade is blank doing a curl and not using websockets as websockets automatically sets the upgrade and connection headers. That is why there is a warning.

Doing the websocket connection defines the http_upgrade which then maps the default_connection_header so therefore it is defined and initialised so it doesn't give the warning.

This curl command doesn't give a warning before and after moving the set so therefore everything will work as normal, this change will just mask the warning
  `curl --include \
     --header "Connection: Upgrade" \
     --header "Upgrade: websocket" \
     http://cafe.example.com:80/`  
     
Resolves: https://github.com/nginxinc/kubernetes-ingress/issues/5082

### Checklist


pdabelf5 (approved):
👍

jjngx (approved):
🚀

j1m-ryan (approved):

PR #5212, opened 2024-03-07
Weight changes Dynamic Reload

j1m-ryan:
### Proposed changes

### Checklist
- [x] Weight changes behind flag 
- [x] Go tests for VS
- [x] Python Tests (Weight updates vs/vsr)
- [x] Python Tests (No reload vs/vsr)
- [x] Manual tests ([weight_changes_without_reload_manual_tests.zip](https://github.com/nginxinc/kubernetes-ingress/files/14558719/weight_changes_without_reload_manual_tests.zip))


ADubhlaoich (commented):
Singular doc change LGTM, ping if there's more when the PR gets close to completion.

build/Dockerfile
```diff
@@ -355,6 +355,9 @@ ARG NAP_MODULES=none
 # copy oidc files on plus build
 RUN --mount=type=bind,target=/tmp [ -n "${BUILD_OS##*plus*}" ] && exit 0; mkdir -p /etc/nginx/oidc/ && cp -a /tmp/internal/configs/oidc/* /etc/nginx/oidc/
 
+# create state_files directory on plus build
+RUN --mount=type=bind,target=/tmp [ -n "${BUILD_OS##*plus*}" ] && exit 0; mkdir -p /etc/nginx/state_files/
```
oseoin, line 359: I think this will work without`--mount=type=bind,target=/tmp`, but it might be better to include it in an existing `RUN` to avoid extra layers, e.g. if you add it after line 383 it could be:

```Dockerfile
    && rm -f /etc/nginx/conf.d/* \
    if [ -n "${BUILD_OS##*plus*}" ]; then mkdir -p /etc/nginx/state_files/; fi
```

internal/configs/version1/nginx.tmpl
```diff
@@ -49,6 +51,7 @@ http {
         '' $sent_http_grpc_status;
     }
 
+
```
oseoin, line 54: nit, extra line (unless it's necessary, in which case we might be better off removing the next `-` :)

build/Dockerfile
j1m-ryan: That makes sense thanks. I've change it to this. Needed an `&&` on the last line also.

jjngx (approved):
🚀

charts/nginx-ingress/README.md
```diff
@@ -471,6 +471,7 @@ The following tables lists the configurable parameters of the NGINX Ingress Cont
 |`controller.readOnlyRootFilesystem` | Configure root filesystem as read-only and add volumes for temporary data. Three major releases after 3.5.x this argument will be moved permanently to the `controller.securityContext` section. | false |
 |`controller.enableSSLDynamicReload` | Enable lazy loading for SSL Certificates. | true |
 |`controller.telemetryReporting.enable` | Enable telemetry reporting. | true |
+|`controller.enableWeightChangesWithoutReload` | Enable weight changes without reloading the NGINX configuration. | false |
```
oseoin, line 474: Do we also need to call out `May require increasing variable_hash_bucket_size, and variable_hash_max_size in the ConfigMap based on the number of two-way splits` here?

cmd/nginx-ingress/flags.go
```diff
@@ -16,7 +16,8 @@ import (
 )
 
 const (
-	dynamicSSLReloadParam = "ssl-dynamic-reload"
+	dynamicSSLReloadParam           = "ssl-dynamic-reload"
+	weightChangesWithoutReloadParam = "weight-changes-without-reload"
```
oseoin, line 20: I'm only just realising that these are possibly at odds in terms of naming... Maybe it should be `weight-changes-dynamic-reload`? It's probably not a big deal to leave it 🤔

charts/nginx-ingress/README.md
j1m-ryan: yes! thanks

cmd/nginx-ingress/flags.go
j1m-ryan: yeah we should probably keep them the same, I will change this to `weight-changes-dynamic-reload`

cmd/nginx-ingress/flags.go
```diff
@@ -281,6 +284,10 @@ func parseFlags() {
 		glog.Fatal("ingresslink and external-service cannot both be set")
 	}
 
+	if *enableDynamicWeightChangesReload && !*nginxPlus {
+		*enableDynamicWeightChangesReload = false
```
oseoin, line 288: Might be good to add a warning log, see `enableServiceInsight` above

shaun-nx (approved):

PR #5217, opened 2024-03-07
Feat/telemetry platform

jjngx:
### Proposed changes

This PR adds Platform to telemetry data. The k8s platform or a cloud provider is determined by checking the `platformID` field.

If the ProviderID value is present, it is added to the telemetry payload in the original form (in lower case). If the `ProviderID` is not present, the value `other` is used and added to the telemetry payload.

This approach allows reporting actual values used by cloud/platform vendors. It also decouples vendor's changes from NIC codebase.

`kind` example:

```shell
I0307 19:31:31.619500       1 collector.go:85] Collecting telemetry data
I0307 19:31:31.650602       1 collector.go:94] Exported telemetry data: {ProjectMeta:{Name:NIC Version:3.5.0-SNAPSHOT} NICResourceCounts:{VirtualServers:0 VirtualServerRoutes:0 TransportServers:0} NodeCount:1 ClusterID:c232007d-76bf-48a5-8dc2-60e017467f0f K8sVersion:v1.29.2 Arch:amd64 Platform:kind}
```

### Checklist

### DIFF
--- a/internal/telemetry/cluster.go
+++ b/internal/telemetry/cluster.go
@@ -102,6 +102,9 @@ func lookupPlatform(providerID string) string {
 	if strings.HasPrefix(provider, "equinixmetal") {
 		return "equinixmetal"
 	}
+	if strings.HasPrefix(provider, "alicloud") {
+		return "alicloud"
+	}
 
 	p := strings.Split(provider, ":")
 	if len(p) == 0 {
--- a/internal/telemetry/cluster.go
+++ b/internal/telemetry/cluster.go
@@ -2,6 +2,8 @@ package telemetry
 
 import (
 	"context"
+	"errors"
+	"strings"
 
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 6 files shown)


internal/telemetry/cluster.go
```diff
@@ -35,3 +37,81 @@ func (c *Collector) K8sVersion() (string, error) {
...
+//
+//gocyclo:ignore
+func lookupPlatform(providerID string) string {
+	provider := strings.TrimSpace(providerID)
+	// The case when the ProviderID field not used by the cloud provider.
+	if provider == "" {
+		return "other"
+	}
+
+	provider = strings.ToLower(providerID)
```
shaun-nx, line 68: I feel like we can simplify this function by putting the platforms in a map and iterating that map.

For example:
```go
// key = providerId; value = platform
var mapProviderIDToPlatform = map[string]string{
	"aws":          "aws",
	"azure":        "azure",
	"gce":          "gke",
	"kind":         "kind",
	"vsphere":      "vsphere",
	"k3s":          "k3s",
	"ibmcloud":     "ibmcloud",
	"ibmpowervs":   "ibmpowervs",
	"cloudstack":   "cloudstack",
	"openstack":    "openstack",
	"digitalocean": "digitalocean",
	"equinixmetal": "equinixmetal",
	"alicloud":     "alicloud",
}
```

```go
func lookupPlatform(providerID string) string {
	provider := strings.TrimSpace(providerID)
	// The case when the ProviderID field not used by the cloud provider.
	if provider == "" {
		return "other"
	}

	provider = strings.ToLower(providerID)

	for id, platform := range mapProviderIDToPlatform {
		if strings.HasPrefix(provider, id) {
			return platform
		}
	}
}
```

shaun-nx: There was a comment @pleshakov made on [here](https://github.com/nginxinc/nginx-gateway-fabric/pull/1651#discussion_r1516245954) about defining a type for constants, and then mapping to those constants [here](https://github.com/nginxinc/nginx-gateway-fabric/pull/1651#discussion_r1516250232)

What are your thoughts on this? I'm in the same boat as @bjee19. I don't fully see the benefit of setting `type CusterPlatform string`

internal/telemetry/cluster_test.go
```diff
@@ -80,6 +80,276 @@ func TestK8sVersionRetrievesClusterVersion(t *testing.T) {
...
+func TestGCPPlatformDeterminesOwnName(t *testing.T) {
+	t.Parallel()
+
+	c := newTestCollectorForClusterWithNodes(t, nodeGCP)
+	got, err := c.Platform(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	want := "gce"
```
vepatel, line 122:
```suggestion
	want := "gcp"
```

jjngx: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/660b340964b01bbf4367d0e3c00e2874fd54d8a1/cloud/providerid/providerid.go#L29

jjngx: value of the provider is taken from the ProviderID field set by the Cloud/Platform Provider. If it's `gcp` it will be reported in the Telemetry payload as `gcp`. We do not hard-code any values apart from `other` (used when ProviderID field is not present - populated by third party cloud/platform provider.

jjngx:
```shell

I0311 17:22:40.840321       1 collector.go:87] Collecting telemetry data
I0311 17:22:40.861822       1 collector.go:114] Exported telemetry data: {Data:{ProjectName:NIC ProjectVersion:3.5.0-SNAPSHOT ProjectArchitecture:amd64 ClusterID:c232007d-76bf-48a5-8dc2-60e017467f0f ClusterVersion:v1.29.2 ClusterPlatform:kind InstallationID: ClusterNodeCount:1} NICResourceCounts:{VirtualServers:0 VirtualServerRoutes:0 TransportServers:0}}
```

pdabelf5 (approved):

shaun-nx (approved):

PR #5229, opened 2024-03-10
helm: add knob to control cluster level rbac rendering

### LINKED ISSUE #5228: Allow helm chart to support external cluster level RBAC
The helm chart comes with a ClusterRole and a ClusterRoleBinding. The ClusterRoleBinding gives very wide access to the whole cluster for example read access to every secret in the cluster. This does not align well with function provided by the -watch-secret-namespace option.

A simple knob to disable the cluster level RBAC resource generation from the helm chart. This gives users a way to provide more suited cluster level RBAC externally yet use the upstream helm chart.

**Additional context**

hafe:
Closes #5228

### Proposed changes

### Checklist


charts/nginx-ingress/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nginx-ingress
-version: 1.3.0
+version: 1.4.0
```
vepatel, line 3:
```suggestion
version: 1.3.0
```
These are automatically updated during major release

vepatel (approved):

ADubhlaoich (approved):
LGTM docs-wise

charts/nginx-ingress/values.yaml
```diff
@@ -501,6 +501,14 @@ rbac:
...
   create: true
 
+  clusterrole:
+    ## Create ClusterRole
+    create: true
+
+  clusterrolebinding:
+    ## Create ClusterRoleBinding
+    create: true
+
```
shaun-nx, line 511: Hi @hafe 
Would there ever be a scenario where someone would set:
1. `clusterrole.create=true` and `clusterrolebinding.create=false`
2. `clusterrole.create=false` and `clusterrolebinding.create=true`

Do you think it makes sense to just have `clusterrole.create` that will create both `ClusterRole` and `ClusterRoleBinding` resources?

hafe: I think clusterrbac.create makes more sense if you want to combine the two.

vepatel: agree with `clusterrbac.create`, in this case combining these args into one is better because they're pretty much hard-coded to each other

charts/nginx-ingress/values.yaml
```diff
@@ -501,6 +501,10 @@ rbac:
   ## Configures RBAC.
   create: true
 
+  cluster:
+    ## Create ClusterRole and ClusterRoleBinding
+    create: true
+
 prometheus:
```
shaun-nx, line 508: Hi @hafe
Can we have this be `clusterrole.create` instead?
I feel like `cluster.create` might be confusing.

vepatel: or `clusterrbac.create`

hafe: Sure. 

Was thinking, maybe the crb can be left as is. It really needed. It is just the content of the cr that would need modification

shaun-nx (approved):

PR #5237, opened 2024-03-12
Update Telemetry Job with GRPC endpoint

shaun-nx:
### Proposed changes

This change sets the GRPC endpoint that telemetry data will be sent to.

Additional changes:
The layout of the Helm arguments for telemetry have been changed.
We originally had `enableTelemetryReporting` as the only option.

This was done to future proof these settings.
```yaml
controller:
  ## Configure telemetry reporting options
  telemetryReporting:
    ## Enable telemetry reporting
    enable: true
```

### Checklist


cmd/nginx-ingress/flags.go
```diff
@@ -203,8 +201,9 @@ var (
 
 	enableDynamicSSLReload = flag.Bool(dynamicSSLReloadParam, true, "Enable reloading of SSL Certificates without restarting the NGINX process.")
 
-	enableTelemetryReporting = flag.Bool("enable-telemetry-reporting", true, "Enable gathering and reporting of product related telemetry.")
-	telemetryReportingPeriod = flag.String("telemetry-reporting-period", "24h", "Sets a telemetry reporting period.")
+	enableTelemetryReporting   = flag.Bool("enable-telemetry-reporting", true, "Enable gathering and reporting of product related telemetry.")
+	telemetryReportingEndpoint = flag.String("telemetry-reporting-endpoint", "dev-oss.edge.df.f5.com:443", "Set the GRPC endpoint to send data to.")
```
jjngx, line 205: @shaun-nx, general question: Could you help me to understand: Why does NIC allow users to change telemetry endpoint? The same question about having `insecure` endpoint option in NIC.

shaun-nx: @jjngx Having this be configurable is useful for both development and in testing.
For testing, if we want to send data to a local otel collector, we may need to set. the `secure` option to false. I've had to do this when I was testing locally and switching between my local collector and the dev XCDF endpoint.

That being said, it may not make sense to present this as a configuration option here, and instead set it as a build option. I know that NFG did it this way too.

jjngx: Let's follow the same route. For local testing we do not need even build params. We can change it directly in the code and then remove (before sending a PR). This would be the same approach as with the reporting period. Users should not be given an option to alter these params.

charts/nginx-ingress/README.md
```diff
@@ -466,11 +466,13 @@ The following tables lists the configurable parameters of the NGINX Ingress Cont
...
 |`controller.disableIPV6` | Disable IPV6 listeners explicitly for nodes that do not support the IPV6 stack. | false |
-|`controller.defaultHTTPListenerPort`  | Sets the port for the HTTP `default_server` listener. | 80 |
+|`controller.defaultHTTPListenerPort` | Sets the port for the HTTP `default_server` listener. | 80 |
 |`controller.defaultHTTPSListenerPort` | Sets the port for the HTTPS `default_server` listener. | 443 |
 |`controller.readOnlyRootFilesystem` | Configure root filesystem as read-only and add volumes for temporary data. Three major releases after 3.5.x this argument will be moved permanently to the `controller.securityContext` section. | false |
 |`controller.enableSSLDynamicReload` | Enable lazy loading for SSL Certificates. | true |
-|`controller.enableTelemetryReporting` | Enable telemetry reporting. | true |
+|`controller.telemetryReporting.enable` | Enable telemetry reporting. | true |
```
vepatel, line 473: need to add these in https://github.com/nginxinc/kubernetes-ingress/blob/main/docs/content/installation/installing-nic/installation-with-helm.md as well

cmd/nginx-ingress/flags.go
```diff
@@ -203,8 +201,9 @@ var (
 
 	enableDynamicSSLReload = flag.Bool(dynamicSSLReloadParam, true, "Enable reloading of SSL Certificates without restarting the NGINX process.")
 
-	enableTelemetryReporting = flag.Bool("enable-telemetry-reporting", true, "Enable gathering and reporting of product related telemetry.")
-	telemetryReportingPeriod = flag.String("telemetry-reporting-period", "24h", "Sets a telemetry reporting period.")
+	enableTelemetryReporting   = flag.Bool("enable-telemetry-reporting", true, "Enable gathering and reporting of product related telemetry.")
```
vepatel, line 204: docs for cli https://github.com/nginxinc/kubernetes-ingress/blob/main/docs/content/configuration/global-configuration/command-line-arguments.md

cmd/nginx-ingress/flags.go
jjngx: What do python tests actually test and prove in this particular case? Allowing users to change the telemetry endpoint does not make any sense from business perspective. What other product on the market, that gathers usage stats, allows end-users to mess with the endpoint?

My2c

charts/nginx-ingress/README.md
shaun-nx: Good catch. Thanks!

internal/k8s/controller_test.go
```diff
@@ -3758,6 +3758,20 @@ func TestPreSyncSecrets(t *testing.T) {
 	}
 }
 
+//func TestNewTelemetryExporter(t *testing.T) {
```
vepatel, line 3761: can this be uncommented?

jjngx (approved):
🚀 @shaun-nx can we merge this PR after NIC Replicas?

charts/nginx-ingress/values.yaml
```diff
@@ -487,8 +487,10 @@ controller:
   ## Enable dynamic reloading of certificates
   enableSSLDynamicReload: true
 
-  ## Enable telemetry reporting
-  enableTelemetryReporting: true
+  ## Configure telemetry reporting options
+  telemetryReporting:
```
oseoin, line 491: [the schema](charts/nginx-ingress/values.schema.json) needs to be updated along with this change

oseoin (approved):

docs/content/installation/installing-nic/installation-with-helm.md
```diff
@@ -22,10 +22,10 @@ h2 {
...
-  - Alternatively, you can build an Ingress Controller image with NGINX Plus and push it to your private registry by following the instructions from [here]({{< relref "installation/building-nginx-ingress-controller.md" >}}).
-  - Update the `controller.image.repository` field of the `values-plus.yaml` accordingly.
+    - To pull from the F5 Container registry, configure a docker registry secret using your JWT token from the MyF5 portal by following the instructions from [here](https://docs.nginx.com/nginx-ingress-controller/installation/nic-images/using-the-jwt-token-docker-secret). Make sure to specify the secret using `controller.serviceAccount.imagePullSecretName` or `controller.serviceAccount.imagePullSecretsNames` parameter.
+    - Alternatively, pull an Ingress Controller image with NGINX Plus and push it to your private registry by following the instructions from [here]({{< relref "installation/nic-images/pulling-ingress-controller-image" >}}).
+    - Alternatively, you can build an Ingress Controller image with NGINX Plus and push it to your private registry by following the instructions from [here]({{< relref "installation/building-nginx-ingress-controller.md" >}}).
```
ADubhlaoich, line 27:
```suggestion
    - Alternatively, pull an NGINX Ingress Controller image with NGINX Plus and push it to your private registry by following the instructions from [here]({{< relref "installation/nic-images/pulling-ingress-controller-image" >}}).
    - Alternatively, you can build an NGINX Ingress Controller image with NGINX Plus and push it to your private registry by following the instructions from [here]({{< relref "installation/building-nginx-ingress-controller.md" >}}).
```

Noun fix - I know you just changed the indentation of these list items, but we'll never be rid of this if we don't address it everywhere see it!

docs/content/installation/installing-nic/installation-with-helm.md
```diff
@@ -22,10 +22,10 @@ h2 {
...
+    - Update the `controller.image.repository` field of the `values-plus.yaml` accordingly.
 - If you’d like to use App Protect DoS, please install App Protect DoS Arbitrator [helm chart](https://github.com/nginxinc/nap-dos-arbitrator-helm-chart). Make sure to install in the same namespace as the NGINX Ingress Controller. Note that if you install multiple NGINX Ingress Controllers in the same namespace, they will need to share the same Arbitrator because it is not possible to install more than one Arbitrator in a single namespace.
```
ADubhlaoich, line 29:
```suggestion
- To use App Protect DoS, install the App Protect DoS Arbitrator [Helm Chart](https://github.com/nginxinc/nap-dos-arbitrator-helm-chart) in the same namespace as the NGINX Ingress Controller. If you install multiple NGINX Ingress Controllers in the same namespace, they will need to share the same Arbitrator because there can only be one Arbitrator in a single namespace.
```

Removing unnecessary words.

docs/content/configuration/global-configuration/command-line-arguments.md
```diff
@@ -536,6 +536,14 @@ The default value is `true`.
 
 <a name="cmdoption-ssl-dynamic-reload"></a>
 
+### -enable-telemetry-reporting
+
+Enable gathering and reporting of product related telemetry.
```
ADubhlaoich, line 541:
```suggestion
Enable gathering and reporting of software telemetry.
```

Small thing, but the word product has marketing ick attached to it. No need to mention "related" since there's an assumption it is related to the software you're configuring.

ADubhlaoich (approved):
LGTM, some minor changes.

shaun-nx:
@ADubhlaoich Thanks for the input.
I'm going to add those changes into another PR that I'm submitting today just in the interest of not re-running the pipeline.
I'll link you that same PR when it's up.

Issue #4772, opened 2023-12-08
Add ability to pass X-Forwarded-* headers unchanged to upstream


dzhlobo:
**Problem description**

In our app we generate full urls based on `Host` or `X-Forwarded-Host` and `X-Forwarded-Proto` headers. This is kind of default way in Django apps and there are no much flexibility present to control it, especially when a url is generated not in our code but in 3rd party library code.

On the ingress we have a "private" domain name because the app is behind a CloudFront CDN so the "public" domain is configured on CDN. This is what happens with clients requests:
1. Original request comes to CDN with `Host: example.com`. It sets `X-Forwarded-Host` and passes request to nginx ingress with following headers: 
```
Host: api.services.example.com
X-Forwarded-Host: example.com
```
2. nginx ingress in its turn rewrites `X-Forwarded-Host` and passes request to upstream with following headers:
```
Host: api.services.example.com
X-Forwarded-Host: api.services.example.com
X-Forwarded-Proto: https
```
You can notice that `X-Forwarded-Host` from CDN gets lost and therefore urls generated by the app looks like `https://api.services.example.com/*`.

**Solution I'd like to have**

I'm thinking about a couple of options.

Option 1.
Add an annotation config key for ingress resource to tell nginx to pass particular headers to upstream without changes, something like this:
```
nginx.org/proxy-set-headers-from-downstream: "X-Forwarded-Host,ABC"
```
Then in location block in nginx config we would add configuration like this:
```
location / {
  proxy_set_header "X-Forwarded-Host" $http_x_forwarded_host;
  proxy_set_header "ABC" $http_abc;
}
```

I don't 100% like it because it is a custom one and not very intuitive due to different semantics comparing to other annotations that maps 1 to 1 to nginx directives (e.g. `nginx.org/proxy-pass-headers` adds `proxy_pass_headers` in config while there are no `proxy_set_headers_from_downstream` directive in nginx).

Option 2.
Add an annotation that allows to add `proxy_set_header` to location block:
```
nginx.org/proxy-set-header.x-forwarded-host: "${http_x_forwarded_host}"
nginx.org/proxy-set-header.abc: "${http_abc}"
```

I don't 100% like it because of dynamic annotation keys (wouldn't it be hard to parse in controller?). It seems we currently have no dynamic annotation keys.

**Alternatives I've considered**

We ended up using VirtualHost resource instead of Ingress with following configuration:
```
routes:
  - path: "/"
    action:
      proxy:
[...]

PR #5366, opened 2024-04-09
Add proxy-set-headers annotation for ingress

AlexFenlon:
### Proposed changes
 
This adds the nginx.org/proxy-set-headers annotation to ingress.
As shown in option 1 in the issue, It can be used to set a custom header by setting the annotation by doing `nginx.org/proxy-set-headers: "X-Forwarded-ABC"`

It can also allow multiple headers by doing `nginx.org/proxy-set-headers: "X-Forwarded-ABC,ABC2"`

It also allows the user to set a custom value in the annotation by doing `nginx.org/proxy-set-headers: "X-Forwarded-ABC: test"`

It can be used in Minion and Master, as well in a non mergable/standard ingress.
The annotation set in minion, if it has the same header name eg. X-Forwarded-ABC, but different value, minion will override the master

If there are different headers in both master and minions, both master and the minion annotations will show up.

There are pytests and go tests for this functionality.

Resolves: #4772

### Checklist


internal/configs/version1/nginx.ingress.tmpl
```diff
@@ -180,23 +180,23 @@ server {
...
 		{{- with $location.BasicAuth }}
 		auth_basic {{ printf "%q" .Realm }};
 		auth_basic_user_file {{ .Secret }};
 		{{- end }}
-
 		proxy_connect_timeout {{$location.ProxyConnectTimeout}};
 		proxy_read_timeout {{$location.ProxyReadTimeout}};
 		proxy_send_timeout {{$location.ProxySendTimeout}};
 		client_max_body_size {{$location.ClientMaxBodySize}};
+        {{- $proxySetHeaders := generateProxySetHeaders $location $.Ingress.Annotations -}}
```
j1m-ryan, line 198: Can these two lines be indented to the right? Or does that throw off indentation in the generated config?

AlexFenlon: They are indented on my side, weird that it saved like that

internal/configs/version1/template_helper.go
```diff
@@ -64,14 +66,162 @@ func makePathWithRegex(path, regexType string) string {
...
+// and generates proxy_set_header headers for minions based on the nginx.org/proxy-set-headers annotation in a mergable ingress.
+// It returns a string containing the generated proxy headers and a header.
+func minionProxySetHeaders(loc *Location, minionHeaders map[string]bool) (string, map[string]bool, error) {
+	proxySetHeaders, ok := loc.MinionIngress.Annotations["nginx.org/proxy-set-headers"]
+	if !ok {
+		return "", nil, nil
+	}
+	var combinedMinions string
+	headers := strings.Split(proxySetHeaders, ",")
+	for _, header := range headers {
```
pdabelf5, line 111: can this for loop be a function?

AlexFenlon: I have tried and for the moment, that doesn't seem possible.

tests/data/annotations/mergeable/minion-annotations-differ.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     nginx.org/mergeable-ingress-type: "master"
     nginx.org/proxy-send-timeout: "10s"
     nginx.org/max-conns: "108"
+    nginx.org/proxy-set-headers: "X-Forwarded-ABC"
```
vepatel, line 10: @AlexFenlon lets add another test for standard ingress when annotation is applied to all the paths

internal/configs/annotations_test.go
```diff
@@ -163,6 +167,59 @@ func TestMergeMasterAnnotationsIntoMinion(t *testing.T) {
...
+			t.Errorf("want error on input %+v, got nil", header)
+		}
+	}
+}
+
+// ParseProxySetHeader ensures that the string value contains only letters
+func ParseProxySetHeader(headers []string) error {
+	for _, header := range headers {
+		if err := ValidateProxySetHeader(header); err != nil {
+			fmt.Println("errored header = ", header)
```
shaun-nx, line 205: You can remove this print statement since each test that uses this function reports the header that caused the error.

AlexFenlon: Removed.

internal/configs/annotations_test.go
```diff
@@ -163,6 +166,58 @@ func TestMergeMasterAnnotationsIntoMinion(t *testing.T) {
...
+		err := ParseProxySetHeader(header)
+		if err != nil {
+			t.Errorf("want nil, got error on valid input: %+v", header)
+		}
+	}
+}
+
+func TestParseProxySetHeaderInvalidInputString(t *testing.T) {
+	t.Parallel()
+	headers1 := []string{"X-Forwarded-For1"}
```
j1m-ryan, line 186: Why is this header invalid? I believe $http_x_forwarded_for1 would be a valid nginx variable

AlexFenlon: The header cannot have any numbers in it so therefore it breaks. If the user wants the value to have a number, that is okay.

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -260,9 +260,9 @@ server {
 		proxy_set_header Upgrade $http_upgrade;
 		proxy_set_header Connection $connection_upgrade;
 		{{- else}}
-		{{- if $.Keepalive}}proxy_set_header Connection "";{{end}}
+		{{- if $.Keepalive}}
```
j1m-ryan, line 263: why does this change?

AlexFenlon: There was formatting issues such as weird spacing and blank lines, can revert if needed

internal/configs/version1/template_helper.go
```diff
@@ -64,14 +66,131 @@ func makePathWithRegex(path, regexType string) string {
...
+			combinedMinions += output
+		}
+	}
+	return combinedMinions, minionHeaders, nil
+}
+
+// standardIngressOrMasterProxySetHeaders takes two strings, two bools and a bool map
+// and generates proxy_set_header headers based on the nginx.org/proxy-set-headers annotation in a standard ingress.
+// It returns a string containing the generated proxy headers for either standardIngress/NonMergeable or Master.
+func standardIngressOrMasterProxySetHeaders(result string, minionHeaders map[string]bool, proxySetHeaders string, ok bool, isMergeable bool) (string, error) {
```
j1m-ryan, line 129: Can get rid of the `result string` argument, as it is always an empty string at the start.

internal/configs/version1/template_helper.go
```diff
@@ -64,14 +66,131 @@ func makePathWithRegex(path, regexType string) string {
...
+	if !ok {
```
j1m-ryan, line 130: Can this `ok` be handled one layer up so it doesnt have to be passed to the `standardIngressOrMasterProxySetHeaders`  function?

internal/configs/version1/template_helper.go
j1m-ryan: Disregard this, I read further on

j1m-ryan (approved):

jjngx (approved):
👍🏻

PR #5431, opened 2024-04-19
Add examples for ProxySetHeaders Annotation

AlexFenlon:
### Proposed changes

Adds examples for both standard and mergeable ingress for the ProxySetHeaders Annotation

### Checklist


AlexFenlon:
@ADubhlaoich I added link to the examples here but I am not sure if it is correct, would the relative path be better instead of the full link?

docs/content/tutorials/ingress-proxy-set-headers-annotation.md
```diff
@@ -0,0 +1,364 @@
+---
```
j1m-ryan, line 1: I don't think we need a tutorial for this annotation. The inline annotation documentation in the Advanced Configuration with Annotations page should really be enough to communicate what it does IMO.

j1m-ryan (approved):

jjngx (approved):
🚀

PR #5468, opened 2024-04-29
Added replaceAll support for go templates

fabriziofiorucci:
### Proposed changes

Enabled support for `replaceAll` in go templates

### Checklist

### DIFF
--- a/internal/configs/version2/template_helper.go
+++ b/internal/configs/version2/template_helper.go
@@ -129,6 +129,7 @@ var helperFunctions = template.FuncMap{
 	"hasSuffix":         strings.HasSuffix,
 	"toLower":           strings.ToLower,
 	"toUpper":           strings.ToUpper,
+	"replaceAll":        strings.ReplaceAll,
 	"makeHTTPListener":  makeHTTPListener,
 	"makeHTTPSListener": makeHTTPSListener,
 	"makeSecretPath":    commonhelpers.MakeSecretPath,


vepatel (approved):

j1m-ryan (approved):

haywoodsh:
It could be potentially confusing for the users since this PR makes changes to helper functions in `internal/configs/version2` but the docs are for ingress template only. 
> Helper functions can be used in the Ingress template to parse the values of custom annotations.
https://github.com/nginxinc/kubernetes-ingress/pull/5468/files#diff-41a03d28be17f49b16b7644d03d1dafb5073b6aee2ffafe099446b1649dba1cfR103

Until we change the page's wording or move it to a better location, I would suggest following this PR #4560 and add the `replaceAll` function to both `version1` and `version2`

I also think we should add a test case for the function to `template_helper_test.go` in both `version1` and `version2`

haywoodsh (commented):
It could be potentially confusing for the users since this PR makes changes to helper functions in `internal/configs/version2` but the docs are for ingress template only. 
> Helper functions can be used in the Ingress template to parse the values of custom annotations.
https://github.com/nginxinc/kubernetes-ingress/pull/5468/files#diff-41a03d28be17f49b16b7644d03d1dafb5073b6aee2ffafe099446b1649dba1cfR103

Until we change the page's wording or move it to a better location, I would suggest following this PR #4560 and add the `replaceAll` function to both `version1` and `version2`

I also think we should add a test case for the function to `template_helper_test.go` in both `version1` and `version2` like the PR mentioned above

PR #5559, opened 2024-05-16
Remove deprecated Go code

jjngx:
### Proposed changes

This PR removes deprecated code.

References:

- `staticcheck`:
```shell
➜  kubernetes-ingress git:(test/staticcheck) ✗ staticcheck ./internal/certmanager/test_files/*.go
internal/certmanager/test_files/util.go:26:2: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: As of Go 1.20 there is no reason to call Seed with a random value. Programs that call Seed with a known value to get a specific sequence of results should use New(NewSource(seed)) to obtain a local random generator.  (SA1019)
```

- Go docs: [ref](https://pkg.go.dev/math/rand#Seed)

### Checklist


internal/certmanager/test_files/util.go
```diff
@@ -19,13 +19,8 @@ package test
 
 import (
 	"math/rand"
-	"time"
 )
 
-func init() {
-	rand.Seed(time.Now().UnixNano())
```
j1m-ryan, line 26: Was this dead code? (In addition to use a deprecated api)

jjngx: No, the func was called implicitly when the package was imported. Seeding random gen explicitly is no longer needed (since v1.20). It's done automatically under the hood when the program starts.

j1m-ryan: ah ok. Thanks

j1m-ryan (approved):

oseoin (approved):

pdabelf5 (approved):

PR #5580, opened 2024-05-20
Support APIKey authentication

haywoodsh:
### Proposed changes

Create API Key Policy to allow authentication with API key header or query in request in VirtualServer and VirtualServerRoute.
Add telemetry for API Key Policy.

### Checklist


internal/configs/version1/nginx-plus.tmpl
```diff
@@ -27,9 +27,7 @@ load_module modules/ngx_fips_check_module.so;
 {{$value}}{{end}}
 {{- end}}
 
-{{- if .OIDC}}
 load_module modules/ngx_http_js_module.so;
-{{- end}}
```
oseoin, line 32: I wonder if this could be considered a breaking change if a user is building their own image without NJS installed 🤔. Maybe we could give a config option for `enableNJS` which we could default to true and fail this and the OIDC policy if it's not enabled?

internal/k8s/secrets/validation.go
```diff
@@ -109,6 +112,23 @@ func ValidateOIDCSecret(secret *api_v1.Secret) error {
...
+
+	uniqueKeys := make(map[string]bool)
+	for _, key := range secret.Data {
+		if uniqueKeys[string(key)] {
+			return fmt.Errorf("API Keys cannot be repeated")
+		}
+		uniqueKeys[string(key)] = true
+	}
+
+	return nil
```
oseoin, line 129: Might be worth validating that uniqueKeys has at least one entry?

internal/configs/version1/nginx-plus.tmpl
j1m-ryan: You're right that loading the same module twice will cause an NGINX emerg error. Which would be breaking for that user. However I don't know if I agree that I would consider it a breaking change as it requires a user to be using their own image, but nevertheless I like the idea of the flag, and we would then have an easy way to conditionally add the JS directives in the main config files.

```nginx
 {{- if .NJSEnabled}}
 js_import /etc/nginx/njs/apikey_auth.js;
 js_set $apikey_auth_hash apikey_auth.hash;
```

image

internal/k8s/secrets/validation.go
haywoodsh: Enforcing uniqueKeys having at least one entry would mean we will reject an empty kubernetes secrets for API key policy. We should allow users to supply a secret without any client/api key combination. if we reject empty secrets, the policy would become invalid and returns 500 for all requests. 

We think an API policy without any client/API key is a legit use case. If user deploys a policy with empty secret, the policy would not become invalid, and the map in the nginx config would become empty, therefore nginx would return 403 for all requests using the specified headers or queries, the same behavior as having the wrong API key in the request. Users can add clients/api keys to the secret without changing the policy or virtual server.

internal/configs/version1/nginx-plus.tmpl
j1m-ryan: we will do this in another PR as there is a lot in this already 
[#5761](https://github.com/nginxinc/kubernetes-ingress/issues/5761)

oseoin (approved):

pdabelf5 (approved):

PR #5866, opened 2024-06-26
Chore/python clean up, and unused variable check to pre commit

j1m-ryan:
### Proposed changes

- removed an obscured function
- adding pre commit tooling to error on unused imports/ unused variables.
- autopep8 format

### DIFF
--- a/tests/suite/fixtures/ic_fixtures.py
+++ b/tests/suite/fixtures/ic_fixtures.py
@@ -267,7 +267,6 @@ def crd_ingress_controller_with_dos(
 
         print("------------------------- Create syslog svc -----------------------")
         src_syslog_yaml = f"{TEST_DATA}/dos/dos-syslog.yaml"
-        f"/var/log/messages"
         create_items_from_yaml(kube_apis, src_syslog_yaml, namespace)
         before = time.time()
         wait_until_all_pods_are_ready(kube_apis.v1, namespace)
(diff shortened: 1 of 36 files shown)


jjngx (approved):

haywoodsh (approved):

tests/suite/fixtures/ic_fixtures.py
```diff
@@ -267,7 +267,7 @@ def crd_ingress_controller_with_dos(
 
         print("------------------------- Create syslog svc -----------------------")
         src_syslog_yaml = f"{TEST_DATA}/dos/dos-syslog.yaml"
-        log_loc = f"/var/log/messages"
+        f"/var/log/messages"
```
pdabelf5, line 270: is this needed at all?

tests/suite/test_annotations.py
```diff
@@ -566,7 +566,7 @@ def test_standard_ingress(
         unexpected_strings,
     ):
         initial_events = get_events(kube_apis.v1, annotations_setup.namespace)
-        initial_count = get_event_count(annotations_setup.ingress_event_text, initial_events)
+        get_event_count(annotations_setup.ingress_event_text, initial_events)
```
pdabelf5, line 569: This whole line can be removed?

j1m-ryan: good catch.  And the one above as initial events is not used elsewhere. I am assuming these get functions have no side effects, but we will know when the tests run

tests/suite/fixtures/ic_fixtures.py
j1m-ryan: Good catch. no its not, I have removed it now. Thanks

PR #5936, opened 2024-07-03
Update link to doc that compares NGINX Ingress Controller to the Kubernetes Ingress Controller

shaun-nx:
Updates link to doc that compares NGINX Ingress Controller to the Kubernetes Ingress Controller


j1m-ryan (approved):

oseoin:
Does this need the `needs cherry pick` label?

ADubhlaoich (changes requested):
I'm not sure why this link even exists in the repo.

The document is in a draft state: I asked @brianehlert about whether or not to take it out of draft, and he said that since nobody has kept on top of checking the community controller's feature parity, it would be irresponsible to publish the document.

As best I can tell this is still the case, so it makes no sense to make a document in a draft state even more prominent when it may contain false information.

PR #6005, opened 2024-07-11
Workflow to build ppc64le binary for ubi

oseoin:
### Proposed changes

Add workflow to build `ppc64le` binaries in a scratch image which can be used by #6008.

### Checklist

### DIFF
--- a/.github/workflows/build-ubi-dependency.yml
+++ b/.github/workflows/build-ubi-dependency.yml
@@ -1,10 +1,11 @@
 name: Build UBI ppc64le Dependency
 
 on:
-  # TODO: remove push
   push:
-  schedule:
-    - cron: "30 4 * * 1" # run Mon at 04:30 UTC
+    branches:
+      - main
+    paths:
+      - build/dependencies/Dockerfile.ubi-ppc64le
   workflow_dispatch:
     inputs:
       nginx_version:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-ubi-dependency.yml
```diff
@@ -0,0 +1,136 @@
+name: Build UBI ppc64le Dependency
+
+on:
+  # TODO: remove push
+  push:
```
oseoin, line 5: Remove after testing

.github/workflows/build-ubi-dependency.yml
```diff
@@ -0,0 +1,137 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Output Variables
+        id: var
+        run: |
+          if [ -n "${{ inputs.nginx_version }}" ]; then
+            nginx_v=${{ inputs.nginx_version }}
+          else
+            nginx_v=$(grep -m1 'FROM nginx:' <build/dependencies/Dockerfile.ubi-ppc64le | cut -d '@' -f1 | awk -F'[: ]' '{print $3}')
+          fi
```
shaun-nx, line 59: Should we add an `echo` here to if the first condition doesn't happen?
Some thing like "nginx version not found, getting version dockerfile"?

oseoin: The `[ -n "${{ inputs.nginx_version }}" ]` will only evaluate to true when the workflow is run manually, so the `else` is the default. It should also be clear which is used from the echo on line 70.

shaun-nx (approved):

jjngx (approved):

PR #6052, opened 2024-07-24
replace promlog with go-kit

pdabelf5:
### Proposed changes

recent `prometheus-common` version cause issues with go linter & tests, replace the `promlog` logging library with `go-kit`, a similar logger with the same interface.

### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/dlclark/regexp2 v1.11.2
 	github.com/gkampitakis/go-snaps v0.5.4
 	github.com/go-chi/chi/v5 v5.1.0
+	github.com/go-kit/log v0.2.1
 	github.com/golang-jwt/jwt/v4 v4.5.0
 	github.com/golang/glog v1.2.0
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -96,6 +95,7 @@ require (
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/prometheus/client_model v0.6.1 // indirect
+	github.com/prometheus/common v0.47.0 // indirect
```
vepatel, line 98: is this still required?

pdabelf5: it's pulled in as a dependency of a dependency, it's not directly needed anymore so shouldn't affect tests & the linter

shaun-nx (approved):

go.mod
pdabelf5: This will get updated as the prometheus exporter updates

Issue #4837, opened 2023-12-22
Refactor main_test.go and add more coverage for main.go


vepatel:
- `main_test.go` requires refactoring as almost all tests are covering `flags.go`,
- `main.go` requires more test coverage.

mrajagopal:
So, as a start, would it be fair to say that `main_test.go` should be renamed `flags_test.go` with those tests intended for `main_test.go` moved into such a file?

j1m-ryan:
Reopening because the second part is not done yet
> `main.go` requires more test coverage

mrajagopal:
* There are 28 functions in `main.go`.
* taking `getAndValidateSecret()` as an example, there is k8s [fake Clientset](https://pkg.go.dev/k8s.io/client-go/kubernetes/fake) but this does not match the input to `getAndValidateSecret()` which requires
`Cannot use 'kubeClient' (type *"[k8s.io/client-go/kubernetes/fake](http://k8s.io/client-go/kubernetes/fake)".Clientset) as the type *"[k8s.io/client-go/kubernetes](http://k8s.io/client-go/kubernetes)".Clientset`
* I suspect some of this boils down to my unfamiliarity with the codebase

main.go | Test Required Yes/No | Notes
-- | -- | --
func main | No |  
func createConfigAndKubeClient | No | No return parameter
func kubernetesVersionInfo | No | No return parameter
func validateIngressClass | No | No return parameter
func checkNamespaces | No | No return parameter
func checkNamespaceExists | No | No return parameter
func createCustomClients | Yes |  
func createPlusClient | Yes |  
func createTemplateExecutors | Yes |  
func createNginxManager | Yes | There is NewFakeManager but the intent of this is is different, not what I am thinking of
func getNginxVersionInfo | Yes | But NewFakeManager will need to implement non-empty Version()
func getAppProtectVersionInfo | No? | The function reads /opt/app_protect/VERSION file
func startApAgentsAndPlugins | No | This requires starting an external agent unless we mock that as well
func processDefaultServerSecret | Yes |  
func processWildcardSecret | Yes | Will need to be adapted to use fake.Clientset
func createGlobalConfigurationValidator | No | This seems to be tested already in globalconfiguration_test.go
func processNginxConfig | No | No return parameter; best tested at the individual module level
func handleTermination | No |  
func getSocketClient | No |  
func getAndValidateSecret | Yes | Conflicts with kubernetes.Clientset vs fake.Clientset
[...]

PR #6314, opened 2024-08-30
Issue 4837: chore: main_test.go unit-tests

mrajagopal:
This is the first of potentially a few to get testability for `main.go`:
* Added unit-tests for the following functions:
* `mustCreateConfigClient`, `mustValidateIngressClass`, `mustConfirmMinimumK8sVersionCriteria`
* Some of the function names were ported from a PR based off an earlier revision of `main` branch

`NOTE:` At present there isn't any test for `mustCreateDynamicClient` as it doesn't seem testable

### Proposed changes
The objective of this PR is to give test coverage for main.go as outlined in https://github.com/nginxinc/kubernetes-ingress/issues/4837.
This includes improving functions by:

returning the error instead of throwing a fatal error
Split functions to improve readability and testability

### Checklist

### DIFF
--- a/cmd/nginx-ingress/main.go
+++ b/cmd/nginx-ingress/main.go
@@ -68,13 +68,32 @@ func main() {
 
 	buildOS := os.Getenv("BUILD_OS")
 
-	config, kubeClient := mustCreateConfigAndKubeClient()
-	mustValidateKubernetesVersionInfo(kubeClient)
-	mustValidateIngressClass(kubeClient)
+	config, err := mustGetClientConfig()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


jjngx (commented):
The point of having functions with the prefix `must` is to communicate that those functions do not return errors (see Go std lib). They either call panic or os.Exit (`log.Fatal`). They prevent the NIC from starting in a state that would make it impossible to run.

There is no need to add additional error handling (or move call to `os.Exit` from inside the func) as there is no error returned. The function either validate or set some values or terminate the program.

PR #6554, opened 2024-10-01
Update documentation folder, dependencies and deployment action

ADubhlaoich:
### Proposed changes

This commit makes a number of changes related to documentation:

- It changes the folder from /docs to /site, matching other NGINX software, which retain /docs for development documentation
- It updates go.mod with the new theme file
- It updates the Docs section of the .gitignore file
- It updates the deployment action to match the folder change, and adds the ability to change the theme version during deployment

Notably, the latest theme update has a new shortcode which allows for code files to be embedded from GitHub links, creating the possibility of single sources of truth in documentation examples.

### Checklist

Before creating a PR, run through this checklist and mark each as complete.

- [ ] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that all unit tests pass after adding my changes
- [ ] I have updated necessary documentation
- [ ] I have rebased my branch onto main
- [ ] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

PR #6596, opened 2024-10-07
[cherry-pick] docs: enableGitInfo config (#6577)

nginx-jack:
Cherry-pick enableGitInfo config and docs-action bump

This enables enableGitInfo config to fix incorrect last-modified docs values.

### DIFF
--- a/docs/config/_default/config.toml
+++ b/docs/config/_default/config.toml
@@ -1,6 +1,6 @@
 title = "NGINX Ingress Controller"
 baseURL = "/"
-enableGitInfo = false
+enableGitInfo = true
 staticDir = ["static"]
 languageCode = "en-us"
 description = "Enterprise-grade Ingress load balancing on Kubernetes platforms."


vepatel (approved):

AlexFenlon (approved):

vepatel:
@nginx-jack looks like make-docs is failing

nginx-jack:
> @nginx-jack looks like make-docs is failing
By default the checkout action clones at depth `1`, hugo now needs full history so it can add last modified dates as part of it's build. Can I update `ci.yml` to increase the clone depth?

vepatel:
@nginx-jack sure, it the checkout is specific to this job the shouldn't be an issue. Probably need to get it in `main` first

nginx-jack:
> @nginx-jack sure, if the checkout is specific to this job the shouldn't be an issue. Probably need to get it in `main` first

I'll need to make some more tweaks, since the action is using docker to build, it doesn't have the git context at all as it's only mounting the `site` directory, and no the root level folder. Will try get it sorted today!

nginx-jack:
@vepatel PR here for a workaround https://github.com/nginxinc/kubernetes-ingress/pull/6597.
I'd like a better solution, that still provided git info for the hugo build, so I'll add an item in our backlog to address it.

nginx-jack:
I'm going to close this for now. There's an outstanding docs PR that wasn't cherry-picked yet. https://github.com/nginxinc/kubernetes-ingress/pull/6554
I'll get that in first, and then queue up the rest of the changes.
Would you prefer I do them as multiple cherry pick MRs, or a single one?

PR #6520, opened 2024-09-26
Add LogLevel & LogFormat flags

AlexFenlon:
### Proposed changes

- added LogLevel parameter to Ingress Controller
- added LogFormat parameter to Ingress Controller
- refactored logLevel in helm chart
- deprecated glog `-v=` format
- added tests for LogFormat and LogLevel
- initiated the log handler requred for structured logs in preperation for future work
- glog is set as default for logFormat to keep consistancy until glog is fully replaced

### Checklist

Before creating a PR, run through this checklist and mark each as complete.

- [x] I have read the [CONTRIBUTING](https://github.com/nginxinc/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that all unit tests pass after adding my changes
- [x] I have updated necessary documentation
- [x] I have rebased my branch onto main
- [x] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

PR #6620, opened 2024-10-08
refactor: fix nil pointer when glog is removed & update old `-v=` to `-log-level=`

AlexFenlon:
### Proposed changes

- fix nil pointer when glog is removed in flags.go to prepare for full glog removal in the `healthcheck` package soon by removing a check for`logtostderr` and `include_year` flags as `logtostderr` was included with glog and `include_year` has been removed. 
- update old `-v=` to `-log-level=` as it was only used for glog which has now been switched out and `-v=` has no longer been used since #6520  

What logtostderr flag was doing in glog 
```
logtostderr (bool, default=false)
Log messages to stderr instead of logfiles.
```

Helm-chart will need to be updated to v2.0.0 as we have a breaking change changing from `-v` to `-log-level` so this needs to be done

### Checklist


deployments/daemon-set/nginx-ingress.yaml
```diff
@@ -91,7 +91,8 @@ spec:
           - -report-ingress-status
           - -external-service=nginx-ingress
          #- -default-server-tls-secret=$(POD_NAMESPACE)/default-server-secret
-         #- -v=3 # Enables extensive logging. Useful for troubleshooting.
+         #- -log-level=debug # Enables extensive logging. Useful for troubleshooting.
```
jjngx, line 94: How about to add available level options (same as L95 for `-log-format`)?

jjngx (approved):
🚀

ADubhlaoich (approved):
LGTM, though the PR is largely not docs-based.

j1m-ryan:
can you update the docs here https://github.com/nginxinc/kubernetes-ingress/blob/main/site/content/logging-and-monitoring/logging.md?

j1m-ryan (approved):

Issue #6679, opened 2024-10-17
GlobalConfiguration custom name in helm chart


frantsao:
**Is your feature request related to a problem? Please describe.**
This issue is related to a situation that occurs when we want to provision the ingress controller using a helm chart but the GlobalConfiguration resource is previously created using a manifest (i.e. operating with Terraform).

**Describe the solution you'd like**
I would like to provide a custom name to helm chart to configure the ingress controller with the GlobalConfiguration previously created, in order to achieve [the prerrequisites](https://github.com/nginxinc/kubernetes-ingress/tree/v3.6.2/examples/custom-resources/custom-listeners#prerequisites).

**Describe alternatives you've considered**
The easiest alterantive that doesn't break my work flow has been forking the chart and adding the parameter by myself (so, PR is on the way ;-)).

**Additional context**
N/A

j1m-ryan:
Hi @frantsao, thanks for working on this.

We would like this parameter to be in the [globalConfiguration](https://github.com/nginxinc/kubernetes-ingress/blob/582198a0270349c74aed4099ec1e5cb2a01e7eea/charts/nginx-ingress/values.yaml#L361) object in the helm chart.

mohamadaldawamnah:
hi @frantsao ill be taking over this issue , i will implement the changes my collegue @pdabelf5 have suggested and merge it with main

frantsao:
> hi [@frantsao](https://github.com/frantsao) ill be taking over this issue , i will implement the changes my collegue [@pdabelf5](https://github.com/pdabelf5) have suggested and merge it with main

Hello,
Sorry I didn't reply earlier. I'm having a very busy few days. I'd be grateful if you continue this :-)
Thanks!

frantsao:
Thanks, @mohamadaldawamnah :-)

vepatel:
hey @frantsao this is now released and available in 5.2.0 https://github.com/nginx/kubernetes-ingress/releases/tag/v5.2.0

frantsao:
> hey [@frantsao](https://github.com/frantsao) this is now released and available in 5.2.0 https://github.com/nginx/kubernetes-ingress/releases/tag/v5.2.0

Thanks for letting me know! :-)

PR #6680, opened 2024-10-17
#6679 add globalConfigurationCustomName parameter 

frantsao:
### Proposed changes

Add globalConfigurationCustomName parameter  to controller confuguration in helm charts. It is helpful when the GlobalConfiguration is created separately from the helm provisioning, as explained in #6679 

### Checklist


frantsao:

j1m-ryan:
Hi @frantsao,

I just tested this out, and it works well. Thanks again for working on it. There's one other thing that would be ideal: if the customName and create options could work together.

This would require an update to [this file](https://github.com/nginxinc/kubernetes-ingress/blob/main/charts/nginx-ingress/templates/controller-globalconfiguration.yaml) to use the name and namespace parts of the customName, if it exists.

The logic in _helpers.tpl would also need to be adjusted slightly to accommodate this change.

frantsao:
> Hi @frantsao,
> 
> I just tested this out, and it works well. Thanks again for working on it. There's one other thing that would be ideal: if the customName and create options could work together.
> 
> This would require an update to [this file](https://github.com/nginxinc/kubernetes-ingress/blob/main/charts/nginx-ingress/templates/controller-globalconfiguration.yaml) to use the name and namespace parts of the customName, if it exists.
> 
> The logic in _helpers.tpl would also need to be adjusted slightly to accommodate this change.

Ok, I think it's a good idea too, but I didn't know if the change could be interesting for you too. I'll work on it as soon as I have some free time again :-)

j1m-ryan:
Hi @frantsao,
No worries at all, take your time. When you look at this again, you might find that making definitions in _helpers.tpl could be useful for getting the name and namespace defined. You can then use the defintions in the [global configuration template file](https://github.com/nginxinc/kubernetes-ingress/blob/main/charts/nginx-ingress/templates/controller-globalconfiguration.yaml).

frantsao:
Hi @j1m-ryan ,
I added the changes that we talked about previously. I hope it's what you expect :-)

charts/nginx-ingress/templates/_helpers.tpl
```diff
@@ -213,6 +213,20 @@ false
 {{- end -}}
 {{- end -}}
 
+{{/*
```
pdabelf5, line 216: it would be good to have the chart apply fail if `.Values.controller.globalConfiguration.customName` doesn't contain a `/`

charts/nginx-ingress/values.yaml
```diff
@@ -397,6 +397,9 @@ controller:
     ## Creates the GlobalConfiguration custom resource. Requires controller.enableCustomResources.
     create: false
 
+    ## customName: "the-namespace/the-name-of-the-global-configuration-custom-resource"
```
pdabelf5, line 410: Maybe add a comment here about making sure this namespace is watched when `watchNamespace` or `watchNamespaceLabel` parameters are in use

danielnginx:
Fixed on https://github.com/nginx/kubernetes-ingress/pull/8142

PR #6888, opened 2024-12-02
add license secret and mgmt config map for r33

AlexFenlon:
### Proposed changes

add initial r33 commit adds license token secret name and mgmt config map. 

### Checklist

### DIFF
--- a/cmd/nginx-ingress/main.go
+++ b/cmd/nginx-ingress/main.go
@@ -78,6 +78,7 @@ const (
 	appProtectv5BundleFolder = "/etc/app_protect/bundles/"
 	fatalEventFlushTime      = 200 * time.Millisecond
 	secretErrorReason        = "SecretError"
+	configMapErrorReason     = "ConfigMapError"
 )
 
 func main() {
@@ -150,6 +151,14 @@ func main() {
 
 	go updateSelfWithVersionInfo(ctx, eventRecorder, kubeClient, version, appProtectVersion, agentVersion, nginxVersion, 10, time.Second*5)
 
+	var mgmtCfgParams *configs.MGMTConfigParams
+	if *nginxPlus {
+		mgmtCfgParams = processMGMTConfigMap(kubeClient, configs.NewDefaultMGMTConfigParams(ctx), eventRecorder, pod)
+		if err := processLicenseSecret(kubeClient, nginxManager, mgmtCfgParams, controllerNamespace); err != nil {
+			logEventAndExit(ctx, eventRecorder, pod, secretErrorReason, err)
+		}
+	}
+
 	templateExecutor, templateExecutorV2 := createTemplateExecutors(ctx)
 
 	sslRejectHandshake, err := processDefaultServerSecret(kubeClient, nginxManager)
@@ -199,7 +208,7 @@ func main() {
 		AppProtectBundlePath:           appProtectBundlePath,
 	}
 
-	mustProcessNginxConfig(staticCfgParams, cfgParams, templateExecutor, nginxManager)
+	mustProcessNginxConfig(staticCfgParams, cfgParams, mgmtCfgParams, templateExecutor, nginxManager)
 
 	if *enableTLSPassthrough {
 		var emptyFile []byte
@@ -215,6 +224,7 @@ func main() {
 		NginxManager:                        nginxManager,
 		StaticCfgParams:                     staticCfgParams,
 		Config:                              cfgParams,
+		MGMTCfgParams:                       mgmtCfgParams,
 		TemplateExecutor:                    templateExecutor,
 		TemplateExecutorV2:                  templateExecutorV2,
 		LatencyCollector:                    latencyCollector,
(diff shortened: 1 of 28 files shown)


internal/configs/configmaps_test.go
```diff
@@ -285,6 +285,106 @@ func TestParseConfigMapAccessLogDefault(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.msg, func(t *testing.T) {
+			result, _, err := ParseMGMTConfigMap(context.Background(), test.configMap, makeEventLogger())
+			if err == nil {
+				t.Fatal(err)
+			}
+			if result.Secrets.License != test.want.Secrets.License {
```
jjngx: var `want` here is always `nil`, correct? Am I right that the table tests here contain values that should make the `ParseMGMTConfigMap` func return error? If that's true, other returned values are irrelevant.

AlexFenlon: Yes, the only way we actually get an error from `ParseMGMTConfigMap` is from the license

internal/k8s/secrets/validation_test.go
```diff
@@ -457,6 +458,64 @@ func TestValidateOIDCSecretFails(t *testing.T) {
...
+		},
+		Type: SecretTypeLicense,
+		Data: map[string][]byte{
+			"license.jwt": []byte(base64.StdEncoding.EncodeToString([]byte("license-token"))),
+		},
+	}
+
+	err := ValidateLicenseSecret(secret)
+	if err != nil {
+		t.Errorf("TestValidateLicenseSecret() returned error %v", err)
```
jjngx, line 476: typo: the failing func name is `ValidateLicenseSecret`

internal/k8s/secrets/validation_test.go
```diff
@@ -457,6 +458,64 @@ func TestValidateOIDCSecretFails(t *testing.T) {
...
+				Type: SecretTypeLicense,
+			},
+			msg: "Missing license.jwt for license secret",
+		},
+	}
+
+	for _, test := range tests {
+		err := ValidateLicenseSecret(test.secret)
+		if err == nil {
+			t.Errorf("TestValidateLicenseSecretFails() returned no error for the case of %s", test.msg)
```
jjngx, line 514: same typo here

internal/configs/configmaps_test.go
jjngx: So, if the test inspects the `License` then error should be not `nil`. In this example if the `err` == `nil` the test fails.

AlexFenlon: We have refactored the test, can we have another review on that please

cmd/nginx-ingress/main.go
```diff
@@ -909,6 +941,24 @@ func processConfigMaps(kubeClient *kubernetes.Clientset, cfgParams *configs.Conf
...
+
+	ns, name, err := k8s.ParseNamespaceName(*mgmtConfigMap)
+	if err != nil {
+		logEventAndExit(ctx, eventLog, pod, configMapErrorReason, fmt.Errorf("error parsing the mgmt-configmap argument: %w", err))
+	}
+	cfm, err := kubeClient.CoreV1().ConfigMaps(ns).Get(context.TODO(), name, meta_v1.GetOptions{})
+	if err != nil {
+		logEventAndExit(ctx, eventLog, cfm, configMapErrorReason, fmt.Errorf("error when getting mgmt-configmap [%v]: %w", *mgmtConfigMap, err))
+	}
+	if mgmtCfgParams, _, fatalErr = configs.ParseMGMTConfigMap(ctx, cfm, eventLog); fatalErr != nil {
```
jjngx, line 956: is it a difference between `fatalErr` and `err` ?

internal/configs/configmaps_test.go
```diff
@@ -285,6 +285,168 @@ func TestParseConfigMapAccessLogDefault(t *testing.T) {
...
+					"enforce-initial-report":    "",
+				},
+			},
+			msg: "enforce-initial-report set empty",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.msg, func(t *testing.T) {
+			_, configWarnings, _ := ParseMGMTConfigMap(context.Background(), test.configMap, makeEventLogger())
```
jjngx, line 349: is err checking silenced here?

internal/configs/configmaps_test.go
```diff
@@ -285,6 +285,168 @@ func TestParseConfigMapAccessLogDefault(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			if warnings {
+				t.Fatal("Unexpected warnings")
+			}
+
+			if result.EnforceInitialReport == nil {
+				t.Fatalf("EnforceInitialReport: want %v, got nil", *test.want.EnforceInitialReport)
+			} else if *result.EnforceInitialReport != *test.want.EnforceInitialReport {
```
jjngx, line 443: `t.Fatal` will finish execution in case of error. Then `else` is obsolete.

internal/configs/configmaps_test.go
AlexFenlon: We are only testing warnings here, we have other test for errors

internal/configs/configmaps_test.go
AlexFenlon: thanks, I deleted the else and added another if block

internal/configs/configmaps_test.go
AlexFenlon: license secret returns a both a warning and error but i did add this check, thanks.

PR #7191, opened 2025-01-22
Fix mountPath directory for NAP5 policy documentation

ADubhlaoich:
### Proposed changes

This commit updates the path for the volumeMounts shown in the NAP5 compile documentation to use the newest convention, and also adds a missing parameter from the `kubectl cp` example at the end of the page.

Issue reported by Fabrizio Fiorucci, as it can cause confusion and possible error for a reader following along and reading the NAP documentation concurrently.

### Checklist


ADubhlaoich:
Fabrizio has indicated that a user will also need to copy a logging profile bundle as part of this process.

The final section of this page is written in a generic enough way to show the user what command to use: they're expected to know what they need to copy. 

The logging profile bundle isn't actually mentioned anywhere in the existing text, so I'm unsure what modifications would need to be made for it to have the correct context.

pdabelf5 (approved):

jjngx (approved):

AlexFenlon (approved):

Issue #7218, opened 2025-01-27
`zone_sync` conversion from snippet to ConfigMap values


AlexFenlon:
**Is your feature request related to a problem? Please describe.**

As an engineer, I want `zone_sync` to a feature for NIC to allow state sharing. First this needs to be a K8s object.

**Describe the solution you'd like**

- provide config data for the `zone_sync` in the nginx-config ConfigMap

jjngx:
add zone_sync data to nginx-config
- zone-sync true/false
- zone-sync port
- calculate domain for the service using the NIC's namespace

PR #7239, opened 2025-01-29
Add initial `zone_sync` values to the ConfigMap

AlexFenlon:
### Proposed changes

This pr adds the following `zone_sync` values to the default `nginx-config` ConfigMap.

- zone-sync true/false
- zone-sync-port - int

`zone-sync-port` requires `zone-sync` to be enabled, which in turn needs NGINX Plus enabled.

Domain is also added and is written dynamically based on the namespace of the controller eg.
`nginx-ingress-headless.nginx-ingress.svc.cluster.local`

This PR only just adds the values and the validation for it, the user cannot enter anyhing until future implementation.
Resolves: #7218 
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: nginx-config
  namespace: nginx-ingress
data:
  zone-sync: true
  zone-sync-port: 1234
```
### Checklist


jjngx (approved):

shaun-nx:
@AlexFenlon Just a heads up, we will need to update our [ConfigMap keys](https://docs.nginx.com/nginx-ingress-controller/configuration/global-configuration/configmap-resource/#configmap-keys) in confluence with this new setting. This doesn't have to do in this PR, but just to make sure it's captured

internal/configs/configmaps.go
```diff
@@ -399,6 +400,47 @@ func ParseConfigMap(ctx context.Context, cfgm *v1.ConfigMap, nginxPlus bool, has
...
 	}
 
+	if zoneSync, exists, err := GetMapKeyAsBool(cfgm.Data, "zone-sync", cfgm); exists {
+		if err != nil {
+			nl.Error(l, err)
+			eventLog.Event(cfgm, v1.EventTypeWarning, nl.EventReasonInvalidValue, err.Error())
+			configOk = false
+		} else {
+			if nginxPlus {
+				cfgParams.ZoneSync.EnableZoneSync = zoneSync
```
javorszky, line 410: zone sync is always enabled when using nginx plus even if the passed in config has

```
zone-sync-enabled: false
```
?

javorszky: Or rather, the `zoneSync` variable has the entire `ZoneSync` struct from above, right? In which case does this assignment make sense?

Wouldn't it be better to have

```go
cfgParams.ZoneSync = zoneSync
```
?

javorszky: or perhaps
```go
cfgParams.ZoneSync.EnableZoneSync = zoneSync.EnableZoneSync
```

javorszky (approved):
Approved with questions around naming things.

internal/configs/config_params.go
```diff
@@ -175,6 +176,13 @@ type Listener struct {
 	Protocol string
 }
 
+// ZoneSync holds zone sync values for state sharing.
+type ZoneSync struct {
+	EnableZoneSync bool
```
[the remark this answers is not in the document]
AlexFenlon: Great point, changes has been made.

pdabelf5 (approved):

PR #7318, opened 2025-02-06
allow customization of service http and https port names through helm

arussellf5:
### Proposed changes

This change allows users to customize the names of the NIC service's http and https ports through a helm chart. It adds defaults which are the same as the previous hardcoded values in the helm chart, thereby ensuring backwards compatibility. 

This change will enable users of NIC to more smoothly integrate with NGINXaaS's AKS functionality. Our [load balancer for kubernetes](https://gitlab.com/f5/nginx/nginxazurelb/nginxaas-loadbalancer-kubernetes) discovers the upstream name and upstream type that it needs to update by parsing service port names, so we need to allow users of our service to easily customize port names.   

### Checklist


puneetsarna:
As an FYI: This unlocks a product use-case where we want NLK to integrate seemlesly with NIC in that, we don't want users making any changes to their configs outside of what's proposed in this MR (to set port names), and then NIC should just integrate with NLK and NGINXaaS.

haywoodsh:
The helm schema should also be updated to reflect the change: https://github.com/nginx/kubernetes-ingress/blob/817956be61494e09d6c3357df458630306685ee0/charts/nginx-ingress/values.schema.json

arussellf5:
> The helm schema should also be updated to reflect the change: https://github.com/nginx/kubernetes-ingress/blob/817956be61494e09d6c3357df458630306685ee0/charts/nginx-ingress/values.schema.json

Thanks @haywoodsh for catching this! I've updated the schema.

puneetsarna:
Is there any documentation update that we may want to do or does that get done as part of the changelog (for helm or the product in general)?

haywoodsh:
We also have this doc on our website documenting all the customizable Helm values but we are happy to update it ourselves after this PR is merged. https://github.com/nginx/kubernetes-ingress/blob/817956be61494e09d6c3357df458630306685ee0/site/content/installation/installing-nic/installation-with-helm.md

The product changelog should be automatically generated upon release

AlexFenlon (approved):

arussellf5:
@AlexFenlon @haywoodsh thanks for your reviews! Would one of you please merge this for me? I don't have the permission to do so.

PR #7695, opened 2025-04-24
Add ConfigMapKeys & MGMTConfigMapKeys to Telemetry

AlexFenlon:
### Proposed changes

This PR adds telemetry for the ConfigMap Keys. It only adds the ConfigMap keys managed by NIC.

Example below shows that the ConfigMap kety "test" is not being reported as NIC does not use it.
(the filtered list will have to be manually updated every time we add a new ConfigMap or MGMTConfigMap key)

Log output example
```
ConfigMapKeys:[error-log-level zone-sync] MGMTConfigMapKeys:[license-token-secret-name]}}
```

contents of the main ConfigMap:
```
kubectl get cm my-release-nginx-ingress -o yaml

apiVersion: v1
data:
  error-log-level: debug
  zone-sync: true
  test: test
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: my-release
    meta.helm.sh/release-namespace: default
  labels:
    app.kubernetes.io/instance: my-release
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: nginx-ingress
    app.kubernetes.io/version: 5.0.0
    helm.sh/chart: nginx-ingress-2.1.0
  name: my-release-nginx-ingress
  namespace: default
```

contexts of MGMT ConfigMap:
```
kubectl get cm my-release-nginx-ingress-mgmt -o yaml

apiVersion: v1
data:
  license-token-secret-name: license-token
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: my-release
    meta.helm.sh/release-namespace: default
  labels:
    app.kubernetes.io/instance: my-release
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: nginx-ingress
    app.kubernetes.io/version: 5.0.0
    helm.sh/chart: nginx-ingress-2.1.0
  name: my-release-nginx-ingress-mgmt
```

### Checklist


internal/telemetry/cluster.go
```diff
@@ -199,6 +303,72 @@ func (c *Collector) BuildOS() string {
...
+		filteredKeys[key] = struct{}{}
+	}
+
+	var keys []string
+	for k := range configMap.Data {
+		if _, ok := filteredKeys[k]; ok {
+			keys = append(keys, k)
+		}
+	}
+	sort.Strings(keys)
```
javorszky, line 334: What is this sort used for here?

internal/telemetry/cluster.go
```diff
@@ -199,6 +303,72 @@ func (c *Collector) BuildOS() string {
...
+
+	return keys, nil
+}
+
+// MGMTConfigMapKeys gets the MGMT ConfigMap keys from the K8s API and returns keys that are filtered and used by NIC.
+func (c *Collector) MGMTConfigMapKeys(ctx context.Context) ([]string, error) {
```
javorszky, line 340: this one and the `ConfigMapKeys` are 95% the same method.

The only differences are
* `c.Config.MainConfigMapName` vs `c.Config.MGMTConfigMapName`
* `configMapFilteredKeys` vs `mgmtConfigMapFilteredKeys`

I'd move the common functionality into its own private function (doesn't even need to be a method on the struct) and pass in the variations

internal/telemetry/cluster_test.go
```diff
@@ -533,6 +533,128 @@ func TestGetServices(t *testing.T) {
...
+}
+
+func TestMGMTConfigMapKeys(t *testing.T) {
+	t.Parallel()
+	testCases := []struct {
+		name   string
+		config telemetry.CollectorConfig
+		want   []string
+	}{
+		{
```
javorszky, line 604: can you add a test case with a key that's not used by NIC?

internal/telemetry/collector_test.go
```diff
@@ -1129,6 +1129,178 @@ func TestCollectInstallationFlags(t *testing.T) {
...
+		},
+		{
+			name: "two keys",
+			configMap: map[string]string{
+				"error-log-level": "debug",
+				"zone-sync":       "true",
+			},
+			want: []string{"error-log-level", "zone-sync"},
+		},
+		{
```
javorszky, line 1154: I'd love a test case for one with keys used and unused 🙂

internal/telemetry/collector_test.go
```diff
@@ -1129,6 +1129,178 @@ func TestCollectInstallationFlags(t *testing.T) {
...
+}
+
+func TestCollectMGMTConfigMapKeys(t *testing.T) {
+	t.Parallel()
+	testCases := []struct {
+		name      string
+		configMap map[string]string
+		want      []string
+	}{
+		{
```
javorszky, line 1224: and another test case here with a case with keys that were filtered out

javorszky (changes requested):

internal/telemetry/cluster.go
AlexFenlon: the sort is there due to tests failing, the test was looking for a certain order, in which the output would sometimes give the correct order but not all the time, the sort insists the order remains the same so it can be tested correctly.

internal/telemetry/cluster.go
AlexFenlon: great point actually, yeah should be able to refactor this.

internal/telemetry/cluster.go
javorszky: I wouldn't adapt the code for the sake of the test, I'd adapt the test so it can deal with the code. If the test fails because of bad order of elements in a slice, what I would do are either

* use `sort` in the test, or
* use [`assert.ElementsMatch` in the stretchr/testify package](https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch) to compare strings out of order. We already use stretchr/testify here for example: https://github.com/nginx/kubernetes-ingress/blob/main/internal/certmanager/sync_test.go#L702

AlexFenlon: Noted, will try this.

internal/telemetry/collector_test.go
```diff
@@ -3,22 +3,24 @@ package telemetry_test
...
 	"context"
+	"encoding/json"
 	"fmt"
 	"runtime"
+	"sort"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
```
javorszky, line 14: super tiny nit: go import grouping seems to be:

1. std library
2. same package (nginx/kubernetes-ingress)
3. third party (go-cmp/cmp should stay there)
4. test packages

internal/telemetry/cluster_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
```
javorszky, line 7: similarly to the other one, can you tidy up this import block with

1. internal
2. own repo
3. third party
4. test if applicable

There's a linter that can do this for us if we want to pay attention to this: https://github.com/daixiang0/gci

This is not super important, functionality isn't impacted, it would just look more tidy 🙂

javorszky (dismissed):
Good to go with two minor nits you're welcome to ignore 🙂

vepatel (approved):

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/nginx/nginx-prometheus-exporter v1.4.1
 	github.com/nginx/telemetry-exporter v0.1.3
 	github.com/nginxinc/nginx-service-mesh v1.7.0
+	github.com/pkg/errors v0.9.1
```
vepatel, line 19: lets see if we can use an approved and active library insted

javorszky: do we have an approved and active library for wrapping errors? 🤔

AlexFenlon: we actually had already used this indirectly but great question🤔

AlexFenlon: removed.

javorszky (approved):

vepatel (approved):

ADubhlaoich (approved):
LGTM!

PR #7738, opened 2025-05-02
fix package to nginx instead of nginx_test

AlexFenlon:
### Proposed changes

- change from nginx_test to nginx as there was no reason for it being nginx_test

### Checklist

Before creating a PR, run through this checklist and mark each as complete.

- [x] I have read the [CONTRIBUTING](https://github.com/nginx/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that all unit tests pass after adding my changes
- [ ] I have updated necessary documentation
- [ ] I have rebased my branch onto main
- [ ] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

PR #7754, opened 2025-05-07
Rename ngnix_test package to nginx in version_test.go

AlexFenlon:
### Proposed changes

redoing #7738 and #7752 to rebuild images

### Checklist

### DIFF
--- a/internal/nginx/version_test.go
+++ b/internal/nginx/version_test.go
@@ -1,22 +1,21 @@
-package nginx_test
+package nginx
 
 import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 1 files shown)


vepatel (approved):

internal/nginx/version_test.go
```diff
@@ -1,22 +1,21 @@
-package nginx_test
+package nginx
```
javorszky, line 1: What fails without this change?

javorszky (changes requested):

internal/nginx/version_test.go
AlexFenlon: good point, making the function public as there is no reason to change this whole package and keep the function private.

PR #7923, opened 2025-06-19
Pin mvdan/gofumpt in github workflow ci and fix bash variable quoting

javorszky:
### Proposed changes

* pin gofumpt to current latest (v0.8.0) rather than "latest"
* quote variables in a bash script to prevent globbing and word splitting, see https://github.com/koalaman/shellcheck/wiki/SC2053

### Checklist

### DIFF
--- a/hack/docker.sh
+++ b/hack/docker.sh
@@ -4,12 +4,12 @@ git_tag=$1
 docker_tag=edge
 
 git_commit=$(git rev-parse HEAD)
-commit_tag=$(git describe --exact-match ${git_commit} 2>/dev/null)
+commit_tag=$(git describe --exact-match "${git_commit}" 2>/dev/null)
 
-if [[ ${commit_tag} == ${git_tag} ]]; then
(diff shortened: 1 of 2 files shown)


hack/docker.sh
```diff
@@ -4,20 +4,20 @@ git_tag=$1
 docker_tag=edge
 
 git_commit=$(git rev-parse HEAD)
-commit_tag=$(git describe --exact-match ${git_commit} 2>/dev/null)
+commit_tag=$(git describe --exact-match "${git_commit}" 2>/dev/null)
 
-if [[ ${commit_tag} == ${git_tag} ]]; then
+if [[ ${commit_tag} == "${git_tag}" ]]; then
```
pdabelf5, line 9: I don't think this file is used any more, it can probably be deleted

pdabelf5: my bad, it's used in the Makefile

pdabelf5 (approved):

PR #7989, opened 2025-07-04
Pin OSS version in Dockerfile

pdabelf5:
### Proposed changes

Pin OSS version in Dockerfile, current version is `1.27`

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 VER = $(shell grep IC_VERSION .github/data/version.txt | cut -d '=' -f 2)
 GIT_TAG = $(shell git describe --exact-match --tags || echo untagged)
 VERSION = $(VER)-SNAPSHOT
+NGINX_OSS_VERSION             ?= 1.27
 NGINX_PLUS_VERSION            ?= R34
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -2,6 +2,7 @@
 VER = $(shell grep IC_VERSION .github/data/version.txt | cut -d '=' -f 2)
 GIT_TAG = $(shell git describe --exact-match --tags || echo untagged)
 VERSION = $(VER)-SNAPSHOT
+NGINX_OSS_VERSION             ?= 1.27
```
vepatel, line 5: can we get rid of NGINX_VERSION and just use NGINX_OSS_VERSION and NGINX_PLUS_VERSION directly instead in all of Dockerfile?

pdabelf5: We'll need to rethink how the labels work.  I'd rather do that when we have more time

vepatel: fair, for OSS i think it gets populated from base images iirc

pdabelf5: yes, that's what I recall too

PR #7991, opened 2025-07-04
Ensure packages are installed on images

pdabelf5:
### Proposed changes

- Run tests in the pipeline to ensure the correct package & versions are installed on each image type.  
- Provide a report of the results attached to the PR as a comment.

### Checklist


javorszky:
I'm also thinking from a code perspective, we have 3 pairs of cmd/regex based on base os. I wonder if it'd be better to specify the OS in the json, and have a map to grab the cmd / regex. Super tiny nit, just preference, so feel free to ignore it. Reason for including: in case a change only affects only some of the regexes / commands for the same base os.

I do really like the idea! Excellent work! 🙌🏻

vepatel:
**stretch goal:**  generate report as a part of workflow after the images are built, similar to DockerScout action in CI jobs where we get  list of CVEs. Instead we get list of modules installed.

vepatel (approved):

PR #8011, opened 2025-07-10
Allow StartupProbe to be configured via helm 

vepatel:
### Proposed changes

- Add startup probe to helm configuration,
- Add an ingress based example
- Docs: https://github.com/nginx/documentation/pull/805

### Checklist


charts/nginx-ingress/values.yaml
```diff
@@ -566,6 +566,32 @@ controller:
...
+  startupStatus:
+
+    ## Enable the startup probe.
+    enable: false
+
+    # ## Set the port where the startup endpoint is exposed.
+    # port: 9999
+
+    # # # path to the startup endpoint.
+    # path: /
```
pdabelf5: if these are required when `startupStatus` is enabled, should they be uncommented to avoid errors.  Or maybe mark them as required in the comments.

charts/tests/testdata/startupstatus.yaml
```diff
@@ -0,0 +1,10 @@
+controller:
```
pdabelf5: it would be good to have a negative case, where not all required values are set

charts/nginx-ingress/values.yaml
vepatel: enforcement is via values.json schema, so the deployment will fail without path and port, I'll update the comments to reflect the same 👍🏼

pdabelf5 (approved):

AlexFenlon (approved):

PR #8133, opened 2025-08-13
Move proxy values balancing to the beginning of parsing

javorszky:
### Proposed changes

This PR moves the proxy values balancing from the template helpers to the time where the annotations and config maps are being parsed. There are a few notable changes:

* proxy value types (like `4 8k` and `16k`) are now custom structs with iota constants for the unit (`k`, `m`), and `uint64` for the numbers. This guarantees that once the values are parsed and corrected, they won't become invalid later through any changes, and we can offload a lot of the error checking to the compiler. One such is that we know none of the values will be negative, because a `uint` can't be negative
* some tests had to be changed. For example previously an empty `proxy_buffers` value passed in via a configmap would generate a bad config, now it gets auto-corrected to the default `2 4k` value. This is the cost of moving the proxy balancing to be right after parsing the values
* similarly the e2e pytest also had to change because if the passed in values are actually invalid, then the result will have those configuration items missing, rather than present, but with invalid values
* I brought on and reused most of the tests that were in the PR originally, though some of them had to be changed. The logic of what value stays, and what values move around it was fuzzy. Right now if proxy-buffers is not defined, that gets set to the default value of `2 4k`, and everything else gets brought down to fit into that one
* rules of how the values relate to each other are based on the NGINX source code
  * `proxy_buffers`: https://github.com/nginx/nginx/blob/master/src/http/modules/ngx_http_proxy_module.c#L3763-L3767
  * `proxy_busy_buffers_size`: https://github.com/nginx/nginx/blob/master/src/http/modules/ngx_http_proxy_module.c#L3787-L3804
 * those rules also mean that the `proxy_buffers_size` is at most all of `proxy_buffers` minus one buffer size 

### Checklist


pdabelf5:
> pkg/apis/configuration/v1/types.go  

afaik should only be used for CRD generation

AlexFenlon:
Docs PR:  https://github.com/nginx/documentation/pull/917

charts/nginx-ingress/templates/_helpers.tpl
```diff
@@ -304,6 +304,10 @@ Build the args for the service binary.
 - -enable-custom-resources={{ .Values.controller.enableCustomResources }}
 - -enable-snippets={{ .Values.controller.enableSnippets }}
 - -disable-ipv6={{ .Values.controller.disableIPV6 }}
+
```
AlexFenlon:
```suggestion

```

charts/nginx-ingress/values.yaml
```diff
@@ -164,6 +164,9 @@ controller:
   ## Sets the log format of Ingress Controller. Options include: glog, json, text
   logFormat: glog
 
+  ## Enables auto adjusting some of the NGINX directives to help with safe configuration and prevent misconfigurations.
```
pdabelf5:
```suggestion
  ## Enables auto adjusting some of the NGINX directives to help with safe configuration and prevent NGINX misconfigurations.
```

pdabelf5: See "here (doc) link" for more details of which configuration options are affected

cmd/nginx-ingress/flags.go
```diff
@@ -225,6 +225,8 @@ var (
 
 	enableDynamicWeightChangesReload = flag.Bool(dynamicWeightChangesParam, false, "Enable changing weights of split clients without reloading NGINX. Requires -nginx-plus")
 
+	enableDirectiveAutoadjust = flag.Bool("enable-directive-autoadjust", false, "Enable automatic adjustment of NGINX directives to make sure no conflicts will be present. This is an experimental feature and may not work as expected.")
```
pdabelf5: >This is an experimental feature and may not work as expected.

Is experimental features something we do?

cmd/nginx-ingress/main.go
```diff
@@ -274,6 +275,7 @@ func main() {
 		cr_validation.IsDosEnabled(*appProtectDos),
 		cr_validation.IsCertManagerEnabled(*enableCertManager),
 		cr_validation.IsExternalDNSEnabled(*enableExternalDNS),
+		cr_validation.IsDirectiveAutoadjustEnabled(*enableDirectiveAutoadjust),
```
pdabelf5, line 278: nit: `cr_validation` feels like it's mis-named now as it's validating cli flags

internal/configs/annotations.go
```diff
@@ -296,12 +299,48 @@ func parseAnnotations(ingEx *IngressEx, baseCfgParams *ConfigParams, isPlus bool
...
+		nl.Errorf(l, "error reconciling proxy_buffers, proxy_buffer_size, and proxy_busy_buffers_size values: %s", err.Error())
+	}
+	cfgParams.ProxyBuffers = balancedProxyBuffers
+	cfgParams.ProxyBufferSize = balancedProxyBufferSize
+	cfgParams.ProxyBusyBuffersSize = balancedProxyBusyBufferSize
+
+	if len(modifications) > 0 {
+		for _, modification := range modifications {
+			nl.Infof(l, "Changes made to proxy values: %s", modification)
+		}
```
pdabelf5, line 343: Should this only happen if `enableDirectiveAutoadjust` is true?

internal/configs/configmaps.go
```diff
@@ -335,12 +335,53 @@ func ParseConfigMap(ctx context.Context, cfgm *v1.ConfigMap, nginxPlus bool, has
...
 
 	if proxyBufferSize, exists := cfgm.Data["proxy-buffer-size"]; exists {
-		cfgParams.ProxyBufferSize = proxyBufferSize
+		proxyBufferSizeData, err := validation.NewSizeWithUnit(proxyBufferSize)
+		if err != nil {
+			nl.Errorf(l, "error parsing nginx.org/proxy-buffer-size: %s", err)
+		} else {
+			cfgParams.ProxyBufferSize = proxyBufferSizeData
+		}
+		// normalizedProxyBufferSize := validation.NormalizeBufferSize(proxyBufferSize)
```
pdabelf5: Why is this commented code here?

internal/configs/configmaps.go
```diff
@@ -335,12 +335,53 @@ func ParseConfigMap(ctx context.Context, cfgm *v1.ConfigMap, nginxPlus bool, has
...
 	}
+	cfgParams.ProxyBuffers = balancedProxyBuffers
+	cfgParams.ProxyBufferSize = balancedProxyBufferSize
+	cfgParams.ProxyBusyBuffersSize = balancedProxyBusyBufferSize
+
+	if len(modifications) > 0 {
+		for _, modification := range modifications {
+			nl.Infof(l, "Changes made to proxy values: %s", modification)
+		}
+	}
```
pdabelf5: Should this only happen if `enableDirectiveAutoadjust` is true?

internal/configs/configmaps.go
```diff
@@ -335,12 +335,53 @@ func ParseConfigMap(ctx context.Context, cfgm *v1.ConfigMap, nginxPlus bool, has
...
+
+	// Normalise the three proxy buffer values across each other.
```
pdabelf5: is this comment about `proxy-max-temp-file-size` ?

internal/configs/config_params.go
pdabelf5, line 1: Why is `isDirectiveAutoadjustEnabled` added to `Configurator` and not `StaticConfigParams`?  `StaticConfigParams` seems to hold all the static parameters that can only be set at startup.  These values are all still available via the `Configurator`.  Let's be consistent in our approach.

internal/configs/virtualserver.go
```diff
@@ -478,6 +478,13 @@ func (vsc *virtualServerConfigurator) GenerateVirtualServerConfig(
 
 	// generate upstreams for VirtualServer
 	for _, u := range vsEx.VirtualServer.Spec.Upstreams {
+		// // here
+		// err := balanceProxiesForUpstreams(&u)
+		// if err != nil {
+		// 	// if we're here, upstream was unchanged
+		// 	vsc.addWarningf(vsEx.VirtualServer, "Balancing proxy buffer values failed for %s: %s", u.Name, err)
+		// }
+
```
pdabelf5: why is this commented code here?

internal/configs/virtualserver.go
```diff
@@ -496,6 +503,14 @@ func (vsc *virtualServerConfigurator) GenerateVirtualServerConfig(
...
 		upstreamNamer := NewUpstreamNamerForVirtualServerRoute(vsEx.VirtualServer, vsr)
 		for _, u := range vsr.Spec.Upstreams {
+			// here
+			//
+			// err := balanceProxiesForUpstreams(&u)
+			// if err != nil {
+			// 	// if we're here, upstream was unchanged
+			// 	vsc.addWarningf(vsEx.VirtualServer, "Balancing proxy buffer values failed for %s: %s", u.Name, err)
+			// }
+
```
pdabelf5: why is this commented code here?

internal/configs/annotations.go
javorszky: This only happens if `enableDirectiveAutoadjust` is true. The check is inside the `validation.BalanceProxyValues` function. If autoadjust is false, it returns all the incoming values as is with a log line of

> auto adjust is turned off, no changes have been made to the proxy values

internal/configs/configmaps.go
javorszky: Same as above, the check is inside `BalanceProxyValues`

internal/configs/config_params.go
javorszky: The idea was because similar directives are also on the `Configurator`, like `isPrometheusEnabled`, or `isDynamicSSLReloadEnabled`, both of which are also CLI flags set at startup: `--enable-prometheus-metrics` and `--ssl-dynamic-reload`, so I thought that's where they should go. I figured if I put the auto adjust there, it will be consistent with what's already there.

Is there a reason those two aren't on the `StaticConfigParams`?

charts/nginx-ingress/values.schema.json
```diff
@@ -657,6 +657,15 @@
             "json"
           ]
         },
+        "directiveAutoAdjust": {
+          "type": "boolean",
+          "default": true,
```
vepatel:
```suggestion
          "default": false,
```

cmd/nginx-ingress/flags.go
AlexFenlon: Bad wording here, will rephrase to "Results may vary and might not be ideal in all cases" instead of experimental

cmd/nginx-ingress/main.go
AlexFenlon: I agree :/ can be fixed in a separate pr though

internal/configs/configmaps.go
AlexFenlon: Good spot, will clean up

internal/configs/configmaps.go
AlexFenlon: nope, comment was left over, thank you

cmd/nginx-ingress/flags.go
vepatel: I'd also change ` to make sure no conflicts will be present` to `to avoid conflicting Nginx configuration`

cmd/nginx-ingress/flags.go
```diff
@@ -225,6 +225,8 @@ var (
 
 	enableDynamicWeightChangesReload = flag.Bool(dynamicWeightChangesParam, false, "Enable changing weights of split clients without reloading NGINX. Requires -nginx-plus")
 
+	enableDirectiveAutoadjust = flag.Bool("enable-directive-autoadjust", false, "Enable automatic adjustment of NGINX directives to make sure no conflicts will be present. Results may vary and might not be ideal in all cases.")
```
AlexFenlon:
```suggestion
	enableDirectiveAutoadjust = flag.Bool("enable-directive-autoadjust", false, "Enable automatic adjustment of NGINX directives to avoid conflicting NGINX configuration. Results may vary and might not be ideal in all cases.")
```

cmd/nginx-ingress/flags.go
AlexFenlon: Done and done

internal/configs/config_params.go
pdabelf5: I don't know, they should probably move too

pdabelf5: btw, `--ssl-dynamic-reload` is in StaticParams

javorszky: > btw, --ssl-dynamic-reload is in StaticParams

And weirdly, ALSO in the `ConfigParams` 😂

AlexFenlon: It was an easy fix, that is in now, Thanks Paul.

pdabelf5 (approved):

internal/configs/parsing_helpers_test.go
```diff
@@ -617,6 +617,79 @@ func TestParseProxyBuffersSpec(t *testing.T) {
...
+					t.Errorf("ParseProxyBuffersSpecWithAutoAdjust(%q) expected error but got none, result: %q", tc.input, result)
+				}
+			} else {
+				if err != nil {
+					t.Errorf("ParseProxyBuffersSpecWithAutoAdjust(%q) unexpected error: %v", tc.input, err)
+				}
+				if result != tc.expected {
+					t.Errorf("ParseProxyBuffersSpecWithAutoAdjust(%q) = %q, expected %q", tc.input, result, tc.expected)
+				}
+			}
```
pdabelf5: Hopefully this makes understanding the logic a little more readable
```suggestion
		    if tc.hasError && err == nil {
			  t.Errorf("ParseProxyBuffersSpecWithAutoAdjust(%q) expected error but got none, result: %q", tc.input, result)
                    }
		    if err != nil {
			    t.Errorf("ParseProxyBuffersSpecWithAutoAdjust(%q) unexpected error: %v", tc.input, err)
		    }
		    if result != tc.expected {
			    t.Errorf("ParseProxyBuffersSpecWithAutoAdjust(%q) = %q, expected %q", tc.input, result, tc.expected)
		    }
```

internal/configs/parsing_helpers_test.go
```diff
@@ -847,3 +920,82 @@ func TestParseFloat64(t *testing.T) {
...
+				}
+			} else {
+				if err != nil {
+					t.Errorf("ParseSizeWithAutoAdjust(%q) unexpected error: %v", tc.input, err)
+				}
+				if result != tc.expected {
+					t.Errorf("ParseSizeWithAutoAdjust(%q) = %q, expected %q", tc.input, result, tc.expected)
+				}
+			}
+		})
```
pdabelf5: Same issue as above

pdabelf5 (changes requested):

pdabelf5 (approved):

vepatel (approved):

PR #8151, opened 2025-08-15
Setup netrc when forcing build

pdabelf5:
### Proposed changes

- NetRC file was not being setup when forcing the build stage.  This change addresses that.
- Also allow for branch name to contain the work `merge`
- Simplify `inputs.force` logic
- Simplify Docker build logic

### Checklist


vepatel (dismissed):

.github/workflows/ci.yml
```diff
@@ -200,7 +200,7 @@ jobs:
               password ${{ secrets.ARTIFACTORY_TOKEN }}
           EOF
           chmod 600 $HOME/.netrc
-        if: ${{ needs.checks.outputs.binary_cache_hit != 'true' && needs.checks.outputs.forked_workflow != 'true' }}
+        if: ${{ (inputs.force && inputs.force || false) || needs.checks.outputs.binary_cache_hit != 'true' && needs.checks.outputs.forked_workflow != 'true' }}
```
javorszky: what does this do? `inputs.force` is used twice, so given the possible values of true and false

inputs.force | inputs.force | false | result
-------------|--------------|-------|-------
true | true | false | true
false | false | false | false

that evaluates from left to right 🤔

pdabelf5: `inputs.force && inputs.force || false` is a ternary check.  Ideally we would only need `inputs.force`.  This way seems to ensure consistency

javorszky: okay, what I'm confused about is why we're using `inputs.foce` twice with an `&&`.

pdabelf5: https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#operators
> GitHub provides a way to create conditional logic in expressions using binary logical operators (&& and ||). This pattern can be used to achieve similar functionality to the ternary operator (?:) found in many programming languages, while actually using only binary operators.

vepatel (approved):

vepatel (approved):

PR #8283, opened 2025-09-21
Fix populate client max body size on VirtualServer Route match

### LINKED ISSUE #7332: [Bug]: client-max-body-size on (upstream) VirtualServer is never considered
### Version

edge

Other

### Steps to reproduce

# Bug

When applying the `client-max-body-size` option on the upstream, via VirtualServer it ultimately has no effect. This is not say that it's not applied to the conf because it is, this is more to say that it doesn't work - this seems to be that the client max body size is then set on the sub location resource and not the originally matching location.

This may be either intended behaviour or a bug, but since it's nginx that's interpreting that conf, that would be a bug in nginx, however I think there would be ways to workaround this in the operator.

# Reproduction and testing

I've included some examples here but, they're really to show what's going on more than to be things you can just take and run (as the upstream itself is not provided).

```yaml
apiVersion: k8s.nginx.org/v1
kind: VirtualServer
metadata:
  name: vs-test
spec:
  ingressClassName: nginx  
  upstreams:
  - name: max-body-test
    service: max-body-test-service
    port: 3000
    keepalive: 32
    client-max-body-size: 15m
  routes:
  - path: ~* '^/max-body-test$'
    matches:
      - conditions:
        - variable: $request_method
          value: POST
        action:
          proxy:
            upstream: max-body-test
```

Once this is deployed, this becomes the following nginx conf; (this is a snippet as the actual conf is quite a bit longer but the important parts are included);

```
upstream vs-test-max-body-size {
    zone vs-test-max-body-size256k;
    random two least_conn;
    server 10.244.221.35:3000 max_fails=1 fail_timeout=10s max_conns=0;
    keepalive 32;
}

map $request_method $vs_test_matches_0_match_0_cond_0 {
    "POST" 1;
    default 0;
}
map $vs_test_matches_0_match_0_cond_0 $vs_test_matches_0 {
    ...
}
server {
    listen 80;
    ...
    location ~* '^/max-body-size$' {
        rewrite ^ $vs_test_matches_0 last;
    }

    location @return_0 {
        ...
    }

    location /internal_location_matches_0_match_0 {
        set $service "max-body-size-service";
        internal;

        ...
        client_max_body_size 15m;
        ...
    }
    location /internal_location_matches_0_default {
        ...
    }
}
```

In the above example, we can see that our `client_max_body_size 15m;` was added to `location /internal_location_matches_0_match_0 {` as that's the match that proxies to our upstream. However, the default of 1m is still used so we're unable to send anything larger still.

If we modify this slightly to change the originally matching location block;

```
location ~* '^/max-body-size$' {
        rewrite ^ $vs_test_matches_0 last;
    }
```

And add the max body size here;

```
location ~* '^/max-body-size$' {
        client_max_body_size 15m;
        rewrite ^ $vs_test_matches_0 last;
    }
```

And we copy this into nginx;

```bash
kubectl cp ./vs-test.conf nginx-ingress/nginx-ingress-controller-pod:/etc/nginx/conf.d/vs-test.conf
kubectl exec -it -n nginx-ingress nginx-ingress-controller-pod -- sh -c "nginx -s reload"
```

We're now able to send bodies that are larger than 1m and upto our defined 15m.

I have no doubts that sever and http snippets would also fix this but, this would then apply to all defined endpoints and upstreams within that scope which may not be what we want.

# Possible Solutions

1. We can work around this potential bug by adding this configuration into the originally matching location block. It would be a case of adding this configuration option to the `s.InternalRedirectLocations` struct or if the option is within the loop of another struct that could be used, the template for this here;

https://github.com/nginx/kubernetes-ingress/blob/278f84154b694f27f1c61ebb302c43755d6105ff/internal/configs/version2/nginx.virtualserver.tmpl#L190

2. This is reported as a nginx bug and ultimately is fixed there without any changes to this operator.

# Related issues

https://github.com/nginx/kubernetes-ingress/issues/5859
https://github.com/nginx/kubernetes-ingress/issues/5317

andrew-s:
### Proposed changes

This fixes #7332 which when using virtual server/routes with a match condition then the max body size of the upstream is ultimately lost as nginx doesn't inherit from the sub locations, this PR adds it to the internal redirect location (the parent).

I couldn't see another way for this to work other than creating an issue against nginx, but there's no guarantees that would be accepted as I would imagine instant parameter inheritance in a version might break existing systems that are expected to work in the current way.

Some edge cases to this implementation;
- any upstream with a client max body size then applies to all matched routes on that path
- if a path with matched routes has multiple upstreams all with different max client body sizes, only the first would be used (perhaps we log a warning if this happens?)

I've added the initial unit tests, I haven't added the template test (I guess this is still worthwhile to prevent this breaking/removal in the future) - I wanted to gauge if it's also worth while adding to the suite tests? (which the helm template looks to be broke in the main branch currently)

### Checklist

- [ x ] I have read the [CONTRIBUTING](https://github.com/nginx/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc
- [ x ] I have added tests that prove my fix is effective or that my feature works
- [ x ] I have checked that all unit tests pass after adding my changes
- [ x ] I have updated necessary documentation
- [ x ] I have rebased my branch onto main
- [ x ] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

### DIFF
--- a/internal/configs/version2/__snapshots__/templates_test.snap
+++ b/internal/configs/version2/__snapshots__/templates_test.snap
@@ -323,9 +323,11 @@ server {
     app_protect_dos_monitor uri=test.example.com protocol=http timeout=30;
     # server snippet
     location /split {
+        
         rewrite ^ @split_0 last;
     }
     location /coffee {
+        
         rewrite ^ @match last;
     }
     location @hc-coffee {
@@ -750,9 +752,11 @@ server {
     app_protect_dos_monitor uri=test.example.com protocol=http timeout=30;
     # server snippet
     location /split {
+        
         rewrite ^ @split_0 last;
     }
     location /coffee {
+        
         rewrite ^ @match last;
     }
     location @hc-coffee {
@@ -2883,9 +2887,11 @@ server {
     limit_req zone=pol_rl_test_test_test burst=5 delay=10;
     # server snippet
     location /split {
+        
         rewrite ^ @split_0 last;
     }
     location /coffee {
+        
         rewrite ^ @match last;
     }
     location @vs_cafe_cafe_vsr_tea_tea_tea__tea_error_page_0 {
@@ -3267,9 +3273,11 @@ server {
     
     # server snippet
     location /split {
+        
         rewrite ^ @split_0 last;
     }
     location /coffee {
+        
         rewrite ^ @match last;
     }
     location @hc-coffee {
--- a/internal/configs/version2/http.go
+++ b/internal/configs/version2/http.go
@@ -337,8 +337,9 @@ type Distribution struct {
 
 // InternalRedirectLocation defines a location for internally redirecting requests to named locations.
 type InternalRedirectLocation struct {
-	Path        string
-	Destination string
+	Path              string
+	Destination       string
+	ClientMaxBodySize string
 }
 
 // Map defines a map.
--- a/internal/configs/version2/nginx-plus.virtualserver.tmpl
+++ b/internal/configs/version2/nginx-plus.virtualserver.tmpl
@@ -399,6 +399,9 @@ server {
 
     {{- range $l := $s.InternalRedirectLocations }}
     location {{ $l.Path }} {
+        {{ if $l.ClientMaxBodySize }}
+        client_max_body_size {{ $l.ClientMaxBodySize }};
+        {{ end }}
         rewrite ^ {{ $l.Destination }} last;
     }
     {{- end }}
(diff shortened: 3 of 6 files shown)


andrew-s:

AlexFenlon:
Hi @andrew-s, 

Thank you for the pull request!

We will take a look at this as soon as we can!

andrew-s:
No problem @AlexFenlon 

I'll await your initial response but I have been thinking, since this location is a rewrite I would assume (correctly or incorrectly) that all upstream parameters would be lost, I haven't tested this but it would be odd we only lose `client_max_body_size`.

So my thoughts are, would it be best to add Parameters to the `InternalRedirectLocation` of a new type `InternalRedirectLocationParameters` - we still face the issue of having only one of the upstream parameters/attributes take precedent over the others.

PR #8404, opened 2025-10-10
Include `metrics` in the helm agent config

### LINKED ISSUE #8407: [Bug]: Generated nginx-agent config is incorrect for Security Monitoring environments
### Version

edge

Kind

### Steps to reproduce

In the NIC + NAP + NIM Security Monitoring use case (https://docs.nginx.com/nginx-instance-manager/monitoring/security-monitoring/configure/set-up-app-protect-instances/), the agent config that is generated in the Helm chart is wrong. With that config, events are not properly collected and sent because of the family of metrics features we enable:

https://github.com/nginx/kubernetes-ingress/blob/eb6244cc31333646b35607d48ff699a99caa24da/charts/nginx-ingress/templates/_helpers.tpl#L588

Right now, the config file that is generated for the use case above only enables the feature `metrics-sender`, which results in the agent being ready to send events, but not collecting them. To actually collect the events we need to include the `metrics-collection` feature as well.

Alternatively, enabling `metrics` would cover both sending and receiving behaviors.

Repro:
- Deploy the Security Monitoring use case with the latest Helm chart (enable Agent, enable NAPv5).
- Integrate it with NIM.
- Generate security events. 
- Those event won't be seen in NIM.

Fix:
- Replace `metrics-sender` with `metrics` in the agent config.
- Restart the agent.
- Now, security events can be seen in NIM.

dareste:
### Proposed changes

Fixes #8407 

The agent configuration that is included in the Helm chart does not work as expected in scenarios where App Protect is enabled and we are sending security events to NIM (WAF Security Monitoring solution).

Right now, the config file that is generated only enables the feature `metrics-sender`, which results in the agent being ready to send events, but not collecting them. To actually collect the events we need to include the `metrics-collection` feature as well.

The solution proposed in this PR is to replace the `metrics-sender` feature with the `metrics` one, which includes both collect and send features.

### Checklist

- [x ] I have read the [CONTRIBUTING](https://github.com/nginx/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc
- [x ] I have rebased my branch onto main
- [x ] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork


pdabelf5:
@dareste Can you log an issue describing the problem so the team can review? Thanks!

AlexFenlon:
Hi @dareste, 

Can you also run `make test-update-snaps` to fix the unit tests!

PR #8424, opened 2025-10-14
Allow renovate to run postUpgradeTasks

pdabelf5:
### Proposed changes

Allow renovate to run postUpgradeTasks
https://docs.renovatebot.com/self-hosted-configuration/#allowedcommands

### Checklist

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -115,80 +115,80 @@ jobs:
           key: nginx-ingress-${{ inputs.go-md5 }}
         if: ${{ inputs.force }}
 
-  generate-assertion-doc:
-    if: ${{ github.event_name != 'pull_request' }}
-    name: Assertion Doc ${{ matrix.nic.arch }}
-    needs: [binaries]
-    runs-on: ${{ inputs.runner }}
-    permissions:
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -245,6 +245,9 @@
     ],
     "executionMode": "branch"
   },
+  "allowedCommands": [
+    "^make .+"
```
javorszky, line 249: does this allow renovate to run any arbitrary make command? Even if the command is `make delete-everything-and-send-us-the-secrets`?

pdabelf5: it needs to match a command in `postUpgradeTasks`, also, these commands get run on renovate PR's.  We can make it more specific

pdabelf5: here are the docs for a more indepth explanation
https://docs.renovatebot.com/self-hosted-configuration/#allowedcommands

javorszky: What are the make commands that we want renovate to run? Or expect to run at the end of it? As that's an array, I'd feel better if we specified the actual command(s) here with `^make ...$`

pdabelf5: So to answer the question, no renovate cannot run arbitrary make commands

PR #8454, opened 2025-10-22
Add resource reference validation function

pdabelf5:
### Proposed changes

Add a function that checks if a resource reference has a namespace prefix.  This make the code more readable at a glance.

### Checklist

### DIFF
--- a/examples/custom-resources/jwks/keycloak.yaml
+++ b/examples/custom-resources/jwks/keycloak.yaml
@@ -31,7 +31,7 @@ spec:
     spec:
       containers:
       - name: keycloak
-        image: quay.io/keycloak/keycloak:26.5.1
+        image: quay.io/keycloak/keycloak:26.5.2
         args: ["start-dev"]
         env:
         - name: KC_BOOTSTRAP_ADMIN_USERNAME
--- a/examples/custom-resources/oidc/keycloak.yaml
+++ b/examples/custom-resources/oidc/keycloak.yaml
@@ -34,7 +34,7 @@ spec:
     spec:
       containers:
         - name: keycloak
-          image: quay.io/keycloak/keycloak:26.5.1
+          image: quay.io/keycloak/keycloak:26.5.2
           args: ["start"]
           env:
             - name: KC_BOOTSTRAP_ADMIN_USERNAME


haywoodsh:
This function might be better located in a util go file than in a validation.go, as it looks more like a helper function that abstracts logic than one that returns an error for invalid configs.

pdabelf5:
> This function might be better located in a util go file than in a validation.go, as it looks more like a helper function that abstracts logic than one that returns an error for invalid configs.

Moving this to the `k8s` package introduces circular dependencies.  It also seems strange to be using k8s utils outside the k8s package

vepatel:
does this needs more changes due to foreign namespace support in upstreams now?

internal/configs/virtualserver.go
```diff
@@ -147,11 +148,9 @@ func GenerateEndpointsKey(
...
-	if strings.Contains(serviceRef, "/") {
-		parts := strings.Split(serviceRef, "/")
-		if len(parts) == 2 {
-			return parts[0], parts[1]
-		}
+	if nsutils.HasNamespace(serviceRef) {
+		ns, name, _ := nsutils.ParseNamespaceName(serviceRef)
+		return ns, name
 	}
 	return defaultNamespace, serviceRef
```
javorszky, line 157: I feel like `HasNamespace` is redundant here. The following achieves the same cutting down one function call


```suggestion
	ns, name, err := nsutils.ParseNamespaceName(serviceRef)
	if err != nil {
	    return defaultNamespace, serviceRef
	}
	
	return ns, name
```

internal/configs/virtualserver.go
```diff
@@ -1605,11 +1604,12 @@ func (p *policiesCfg) addAPIKeyConfig(
 
 	p.APIKey.Clients = generateAPIKeyClients(secretRef.Secret.Data)
 
+	_, polName, _ := nsutils.ParseNamespaceName(rfc1123ToSnake(polKey))
```
javorszky: are we guaranteed to have a valid namespace/name format for `polKey` here?

internal/k8s/controller.go
```diff
@@ -329,13 +330,13 @@ func NewLoadBalancerController(input NewLoadBalancerControllerInput) *LoadBalanc
 	if lbc.areCustomResourcesEnabled {
 		if input.GlobalConfiguration != "" {
 			lbc.watchGlobalConfiguration = true
-			ns, name, _ := ParseNamespaceName(input.GlobalConfiguration)
+			ns, name, _ := nsutils.ParseNamespaceName(input.GlobalConfiguration)
```
javorszky: are we guaranteed to have a valid namespace/name here?

internal/k8s/policy.go
```diff
@@ -96,7 +97,7 @@ func (lbc *LoadBalancerController) syncPolicy(task task) {
 	}
 
 	// it is safe to ignore the error
-	namespace, name, _ := ParseNamespaceName(key)
+	namespace, name, _ := nsutils.ParseNamespaceName(key)
```
javorszky: I see that it is safe to ignore the error per the comment. My guess is that L65, the `getNamespacedInformer` will throw an error if there's no namespace anyways. Is that assumption correct?

pkg/apis/configuration/validation/virtualserver.go
```diff
@@ -737,18 +738,18 @@ func validateServiceName(name string, fieldPath *field.Path) field.ErrorList {
 
 // validateVirtualServerServiceName checks if a namespaced service name is valid for VirtualServer upstreams.
 func validateVirtualServerServiceName(name string, fieldPath *field.Path) field.ErrorList {
-	if strings.Contains(name, "/") {
-		parts := strings.Split(name, "/")
-		if len(parts) != 2 {
+	if nsutils.HasNamespace(name) {
+		svcNamespace, svcName, err := nsutils.ParseNamespaceName(name)
```
javorszky: These two lines can also be collapsed to a single function call:

```
ns, name, err := nsutils.ParseNamespaceName(name)
if err != nil {
    validateServiceName(name, fieldPath)
}

// validateDNS1123Label

// validateServiceName
```

pdabelf5:
> the `nsutils.ParseNamespaceName` is [very similar to the `cache.SplitMetaNamespaceKey` function](https://github.com/kubernetes/client-go/blob/v0.35.0/tools/cache/store.go#L174-L191), though the latter has graceful fallback if the argument doesn't have a namespace

Does this mean you want me to refactor all places where `nsutils.ParseNamespaceName` is used and replace it with `cache.SplitMetaNamespaceKey` ?

internal/configs/virtualserver.go
pdabelf5: Reverted ParseNamespaceName() changes to allow more time to consider the correct approach

internal/k8s/controller.go
pdabelf5: Reverted ParseNamespaceName() changes to allow more time to consider the correct approach

internal/k8s/policy.go
pdabelf5: Reverted ParseNamespaceName() changes to allow more time to consider the correct approach

internal/configs/virtualserver.go
pdabelf5: Reverted ParseNamespaceName() changes to allow more time to consider the correct approach

pkg/apis/configuration/validation/virtualserver.go
pdabelf5: Reverted ParseNamespaceName() changes to allow more time to consider the correct approach

javorszky:
> Does this mean you want me to refactor all places where nsutils.ParseNamespaceName is used and replace it with cache.SplitMetaNamespaceKey ?

Not necessarily. If we can reuse the same function, that's awesome. If there's a good reason for keeping a hard fail in the `ParseNamespaceName`, I'm okay with it. What I am looking for is a confirmation of "yeah I've seen it, it's still better to keep this" or "no, that makes sense, we can reuse it". 🙂  Neither is better or worse than the other.

Personally I am leaning towards reuse _when possible_.

PR #8455, opened 2025-10-22
Cleanup stale socket files on startup

AlexFenlon:
### Proposed changes

This change moves clean up logic for the `.sock` files to startup instead of termniation.  Thank you @sigv for the suggestion.
This is allow the users to recover gracefully after an unexpected event that might terminate the process.

Closes https://github.com/nginx/kubernetes-ingress/issues/7622

### Testing
Heavily tested by deploying a pod with memory and cpu limits to 64Mi and 1M respectively, used over 50 ingresses and repeating curl requests to crash the system  to see if it came back.  Tested using the image 5.2.1 -> local image with fix as well as just using the local image. 

### Checklist


cmd/nginx-ingress/main.go
```diff
@@ -844,6 +833,24 @@ func handleTermination(lbc *k8s.LoadBalancerController, nginxManager nginx.Manag
 	os.Exit(0)
 }
 
+func cleanupSocketFiles(l *slog.Logger) {
+	files, readErr := os.ReadDir(fmt.Sprintf("%s/", socketPath))
```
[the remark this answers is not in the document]
AlexFenlon: socketpath does not have the / so this is needed

cmd/nginx-ingress/main.go
```diff
@@ -844,6 +833,24 @@ func handleTermination(lbc *k8s.LoadBalancerController, nginxManager nginx.Manag
...
+	if readErr != nil {
+		nl.Errorf(l, "error trying to read directory %s: %v", socketPath, readErr)
+	} else {
+		for _, f := range files {
+			nl.Debugf(l, "Processing file %s", f.Name())
+			if !f.IsDir() && strings.HasSuffix(f.Name(), ".sock") {
```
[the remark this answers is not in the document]
AlexFenlon: removed debug line as its no longer required.

cmd/nginx-ingress/main.go
```diff
@@ -844,6 +832,25 @@ func handleTermination(lbc *k8s.LoadBalancerController, nginxManager nginx.Manag
 	os.Exit(0)
 }
 
+
```
AlexFenlon:
```suggestion
```

cmd/nginx-ingress/main.go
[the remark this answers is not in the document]
javorszky: os.ReadDir does not need trailing slash to successfully open the path. If the file referenced by the path (trailing slash or not) is a directory, ReadDir will open it as a directory, otherwise it will error out. Internally it uses syscall.Open.

On a mac this means if you have a directory named `bla`, using the command `open bla` (without trailing slash) will open Finder with the directory open. If you try to create a file named `bla`, the same one as the directory, with `touch bla`, that will do nothing.

I also tested this with a tiny Go program with the following directory structure
```
/Users/username/Projects/testproject
├── bla
│   └── .gitkeep
├── go.mod
├── go.sum
└── main.go
```
and the main file:
```go
package main

import (
	"fmt"
	"log"
	"os"
)

func main() {
	files, err := os.ReadDir("/Users/username/Projects/testproject/bla")
	if err != nil {
		log.Fatal(err)
	}

	for _, file := range files {
		fmt.Println(file.Name())
	}
}
```
That will print
```
$ go run ./...
.gitkeep
```

With that, do you have a test case that necessitates the trailing slash that I'm not thinking of? 🤔

cmd/nginx-ingress/main.go
```diff
@@ -936,7 +942,7 @@ func createPlusAndLatencyCollectors(
 			plusCollector = nginxCollector.NewNginxPlusCollector(plusClient, "nginx_ingress_nginxplus", variableLabelNames, constLabels, l)
 			go metrics.RunPrometheusListenerForNginxPlus(ctx, *prometheusMetricsListenPort, plusCollector, registry, prometheusSecret)
 		} else {
-			httpClient := getSocketClient("/var/lib/nginx/nginx-status.sock")
+			httpClient := getSocketClient(fmt.Sprintf("%s/nginx-status.sock", socketPath))
```
javorszky: would it be better to use `filepath.join` here as well?

cmd/nginx-ingress/main.go
```diff
@@ -945,7 +951,7 @@ func createPlusAndLatencyCollectors(
 			if err := lc.Register(registry); err != nil {
 				nl.Errorf(l, "Error registering Latency Prometheus metrics: %v", err)
 			}
-			syslogListener = metrics.NewLatencyMetricsListener(ctx, "/var/lib/nginx/nginx-syslog.sock", lc)
+			syslogListener = metrics.NewLatencyMetricsListener(ctx, fmt.Sprintf("%s/nginx-syslog.sock", socketPath), lc)
```
javorszky: same as above about `filepath.join`

cmd/nginx-ingress/main.go
```diff
@@ -522,7 +525,7 @@ func createPlusClient(ctx context.Context, nginxPlus bool, useFakeNginxManager b
 	var err error
 
 	if nginxPlus && !useFakeNginxManager {
-		httpClient := getSocketClient("/var/lib/nginx/nginx-plus-api.sock")
+		httpClient := getSocketClient(fmt.Sprintf("%s/nginx-plus-api.sock", socketPath))
```
javorszky: this is the third review comment, but same question about using `filepath.join` 🙂

cmd/nginx-ingress/main.go
```diff
@@ -88,6 +89,8 @@ func main() {
 	ctx := initLogger(*logFormat, logLevels[*logLevel], os.Stdout)
 	l := nl.LoggerFromContext(ctx)
 
+	cleanupSocketFiles(l)
```
javorszky, line 92: socket files will be cleaned up as part of ... startup? The previous functionality cleaned up the sockets when nginx was done, no? 🤔

cmd/nginx-ingress/main.go
```diff
@@ -66,6 +66,7 @@ var (
 		"error":   levels.LevelError,
 		"fatal":   levels.LevelFatal,
 	}
+	socketPath = "/var/lib/nginx"
```
javorszky: Out of curiousity, in what instances would this be different? And what is the expectation when passing this in at build time? I assume it's using the linker flag, something like

```
$ go build -ldflags="-X 'github.com/nginx/kubernetes-ingress.socketPath=/some/other/path'"
```
Is that correct? Can that documentation be part of the PR? 🙂

cmd/nginx-ingress/main.go
AlexFenlon: Correct! 

It used to clean up during nginx termination which didn't catch every case, running it on startup ensures that if there are existing .sock files remaining, they are removed before new ones are created during runtime!

cmd/nginx-ingress/main.go
AlexFenlon: I do not think it would ever be different, this was only done to use a variable instead of repeating the same path over and over.

cmd/nginx-ingress/main.go
AlexFenlon: fixed

cmd/nginx-ingress/main.go
AlexFenlon: fixed

cmd/nginx-ingress/main.go
AlexFenlon: fixed

cmd/nginx-ingress/main.go
javorszky: in that case a `const` might be a better fit

AlexFenlon: done

PR #8548, opened 2025-11-19
Add `client-body-buffer-size` directive to Ingress Annotations & Configmap

AlexFenlon:
### Proposed changes

- Add `nginx.org/client-body-buffer-size`  Ingress annotation to change location-specific directive changes
- Add  `client-body-buffer-size` ConfigMap value to change the global directive.
- Default global `client-body-buffer-size` is set to 8k to match ingress-nginx

docs: https://github.com/nginx/documentation/pull/1468

### Checklist

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -17,7 +17,7 @@ ARG PACKAGE_REPO=pkgs.nginx.com
 
 ############################################# Base images containing libs for FIPS #############################################
 FROM ghcr.io/nginx/dependencies/nginx-ubi:ubi8@sha256:cc6e2b5277b4ded251c8055a99be1f930a3aeae150359ff8de820c3e9155b358 AS ubi8-packages
-FROM ghcr.io/nginx/dependencies/nginx-ubi:ubi9@sha256:32cef95db124a0425224b6c4d36ac5ce7fe384e15072492c898d335496399a2b AS ubi9-packages
+FROM ghcr.io/nginx/dependencies/nginx-ubi:ubi9@sha256:cd8693e02cd22ad05d1c5c57f7dd7ccfcf3cca23fac158381a760ff6f4d0f292 AS ubi9-packages
 FROM ghcr.io/nginx/alpine-fips:0.4.0-alpine3.19@sha256:0b400b81b5f403d69535a54839296ae35ced374eb1bb04db5b4282f380fef09a AS alpine-fips-3.19
 FROM ghcr.io/nginx/alpine-fips:0.4.0-alpine3.22@sha256:61ed75f252bde7da1e6db33d2709456e87478280dfae3d11084f94c361e9f329 AS alpine-fips-3.22
 FROM redhat/ubi9-minimal:9.6-1760515502@sha256:34880b64c07f28f64d95737f82f891516de9a3b43583f39970f7bf8e4cfa48b7 AS ubi-minimal


internal/configs/version1/__snapshots__/template_test.snap
```diff
@@ -21,6 +21,7 @@ http {
     map_hash_max_size ;
     map_hash_bucket_size ;
 
+    client_body_buffer_size ;
```
pdabelf5: This looks like invalid nginx config

internal/configs/version1/nginx-plus.tmpl
```diff
@@ -37,6 +37,8 @@ http {
     map_hash_max_size {{.MapHashMaxSize}};
     map_hash_bucket_size {{.MapHashBucketSize}};
 
+    client_body_buffer_size {{.ClientBodyBufferSize}};
```
pdabelf5, line 41: Why not have an if check around this one too?

internal/configs/version1/__snapshots__/template_test.snap
AlexFenlon: that would be but there is no input provided here, in the code it is defaulted to 8k to match ingress-nginx

pdabelf5: the general idea of the snaps is to review the generated nginx config.

internal/configs/version1/nginx-plus.tmpl
AlexFenlon: the global value by default in ingress-nginx was 8k, I am happy not to have the global default set but it is replicating what ingress-nginx does.

The other directives are only added if a value was added to ingress annotations.

internal/configs/config_params.go
```diff
@@ -247,6 +249,7 @@ func NewDefaultConfigParams(ctx context.Context, isPlus bool) *ConfigParams {
 		MainServerNamesHashMaxSize:    "1024",
 		MainMapHashBucketSize:         "256",
 		MainMapHashMaxSize:            "2048",
+		MainClientBodyBufferSize:      "8k",
```
pdabelf5: Just wondering why this is hardcoded to 8k when nginx can figure this out based in the architecture it is running on
> This is 8K on x86, other 32-bit platforms, and x86-64. It is usually 16K on other 64-bit platforms

for example 8k on amd64 & 16k on arm64

AlexFenlon: I tried it in ingress-nginx and I didn't see it change, i don't think it needs to be hard coded to be honest,

internal/configs/version1/nginx-plus.tmpl
AlexFenlon: makes more sense to have the if, since nginx will create the defaults itself. Will fix this.

AlexFenlon: fixed

internal/configs/config_params.go
AlexFenlon: fixed

internal/configs/version1/template_test.go
```diff
@@ -303,6 +303,28 @@ func TestExecuteTemplate_ForMergeableIngressWithOneMinionWithPathRegexAnnotation
...
 }
 
+func TestExecuteTemplate_ForIngressWithClientBodyBufferSize(t *testing.T) {
+	t.Parallel()
+
+	tmpl := newNGINXPlusIngressTmpl(t)
+	buf := &bytes.Buffer{}
+
+	testCfg := ingressCfg
+	testCfg.Servers[0].Locations[0].ClientBodyBufferSize = "16k"
```
pdabelf5: Changing the value here is affecting more than this test.  See output for test `TestExecuteTemplate_ForIngressForNGINXPlus`

AlexFenlon: That is the reason snaps were all over the place, I blame copilot for that. 

Thank you!

AlexFenlon: fixed

PR #8618, opened 2025-12-05
Tweak some maintenance make targets

javorszky:
## Proposed changes

* Update the version of staticcheck being installed from the make target
* Add govulncheck to the list of make targets. v1.1.4 is currently the latest version, see https://go.googlesource.com/vuln
* Adjust the `all-images` target such that when each image gets built, their names are appended to the TAG used, see below

### Reason for the all-images

When you ran the `all-images` make command, each image got built, but it's the same TAG being used, so only the last one was tagged, the rest were overwritten by the next one.

Now the list of images will end up looking like this:

image

This change is not needed for the 5.3.0 release, so no need to cherry pick

## Checklist

### DIFF
--- a/tests/data/modules/data.json
+++ b/tests/data/modules/data.json
@@ -33,7 +33,7 @@
       "packages": [
         {
           "name": "nginx-plus",
-          "version": "36-1"
+          "version": "36-2"
         },
         {
           "name": "nginx-plus-module-njs",
@@ -64,7 +64,7 @@
       "packages": [
         {
           "name": "nginx-plus",
-          "version": "36-1"
+          "version": "36-2"
         },
         {
           "name": "nginx-plus-module-njs",
@@ -110,7 +110,7 @@
       "packages": [
         {
           "name": "nginx-plus",
-          "version": "36-1"
+          "version": "36-2"
         },
         {
           "name": "nginx-plus-module-njs",
@@ -152,7 +152,7 @@
       "packages": [
         {
           "name": "nginx-plus",
-          "version": "36-1"
+          "version": "36-2"
         },
         {
           "name": "nginx-plus-module-njs",
@@ -186,7 +186,7 @@
       "packages": [
         {
           "name": "nginx-plus",
-          "version": "36-1"
+          "version": "36-2"
         },
         {
           "name": "nginx-plus-module-njs",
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -79,9 +79,14 @@ format: ## Run goimports & gofmt
 
 .PHONY: staticcheck
 staticcheck: ## Run staticcheck linter
-	@staticcheck -version >/dev/null 2>&1 || go install honnef.co/go/tools/cmd/staticcheck@2022.1.3;
+	@staticcheck -version >/dev/null 2>&1 || go install honnef.co/go/tools/cmd/staticcheck@2025.1.1;
```
pdabelf5: Lets use renovate if we can to avoid manually needing to manage this

Makefile
```diff
@@ -79,9 +79,14 @@ format: ## Run goimports & gofmt
...
 	staticcheck ./...
 
+.PHONY: govulncheck
+govulncheck:
+	@govulncheck -version >/dev/null 2>&1 || go install golang.org/x/vuln/cmd/govulncheck@v1.1.4;
```
pdabelf5: Lets use renovate if we can to avoid manually needing to manage this

Makefile
```diff
@@ -79,9 +79,14 @@ format: ## Run goimports & gofmt
 
 .PHONY: staticcheck
 staticcheck: ## Run staticcheck linter
-	@staticcheck -version >/dev/null 2>&1 || go install honnef.co/go/tools/cmd/staticcheck@2022.1.3;
+	@staticcheck -version >/dev/null 2>&1 || go install honnef.co/go/tools/cmd/staticcheck@2025.1.1;
 	staticcheck ./...
 
+.PHONY: govulncheck
+govulncheck:
```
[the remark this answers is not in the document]
javorszky: This has been added in a later commit

Makefile
javorszky: I added renovate config which I think resolves this issue, but please double check, and resolve this configuration if it's okay

Makefile
javorszky: I added renovate config which I think resolves this issue, but please double check, and resolve this configuration if it's okay

pdabelf5 (approved):

vepatel (approved):

PR #8692, opened 2025-12-10
support NIC running in proxy mode

haywoodsh:
### Proposed changes

Improves the `-proxy` flag so the NIC can run outside the cluster without crashing.

- Create a synthetic pod object instead of fetching from the API
- Skip license reporting, deployment metadata, pod label updates and agent version operations that require a real pod
- Use default mgmt config params when `-mgmt-configmap` is not specified
- Allow `-nginx-plus` without `-mgmt-configmap` in proxy mode
- Move socket file cleanup into the Manager interface so it is a no-op in proxy mode instead of guarding with a flag check
- Return the correct OSS or Plus version in proxy mode so the binary/flag consistency check passes without a real NGINX binary
- Fix SIGTERM shutdown so the controller exits cleanly in proxy mode
- Add `-oidc-template-path` flag for overriding the OIDC template path

### Checklist

### DIFF
--- a/cmd/nginx-ingress/main.go
+++ b/cmd/nginx-ingress/main.go
@@ -88,11 +88,6 @@ func main() {
 	ctx := initLogger(*logFormat, logLevels[*logLevel], os.Stdout)
 	l := nl.LoggerFromContext(ctx)
 
-	// TODO: Use fake manager
-	if *proxyURL == "" {
-		cleanupSocketFiles(l)
-	}
-
 	initValidate(ctx)
 	parsedFlags := os.Args[1:]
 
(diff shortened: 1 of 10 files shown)


cmd/nginx-ingress/main.go
```diff
@@ -101,10 +99,36 @@ func main() {
 	if err := validateKubernetesVersionInfo(ctx, kubeClient); err != nil {
 		nl.Fatal(l, err)
 	}
-	pod, err := kubeClient.CoreV1().Pods(controllerNamespace).Get(context.TODO(), podName, meta_v1.GetOptions{})
-	if err != nil {
-		nl.Fatalf(l, "Failed to get pod: %v", err)
+
+	var pod *api_v1.Pod
+
+	if *proxyURL != "" {
```
sean-breen, line 105: Can you extract this logic into a function so it can be unit-tested?

PR #8720, opened 2025-12-15
Add `nginx.org/app-root` annotation support

AlexFenlon:
### Proposed changes

- Add `nginx.org/app-root` Ingress annotation
  - allows redirection of root to go to any user-defined path eg coffee
  - Returns a 302 code if curled 
  - validation prevents anything that isn't a path.
- Add telemetry test
- docs https://github.com/nginx/documentation/pull/1578
eg. 

set annotation
```
Annotations:        nginx.org/app-root: /coffee
```

curl root application eg. cafe.example.com/
```
<html>
<head><title>302 Found</title></head>
<body>
<center><h1>302 Found</h1></center>
<hr><center>nginx/1.29.3</center>
</body>
</html>
```

going to the same path in a browser redirects you to the /coffee path

https://github.com/user-attachments/assets/30f9de62-410c-42f4-8a33-343f1d209674

### Checklist

### DIFF
--- a/.github/scripts/requirements.txt
+++ b/.github/scripts/requirements.txt
@@ -417,9 +417,9 @@ typing-extensions==4.15.0 \
     --hash=sha256:0cea48d173cc12fa28ecabc3b837ea3cf6f38c6d1136f85cbaaf598984861466 \
     --hash=sha256:f0fa19c6845758ab08074a0cfa8b7aecb71c999ca73d62883bc25cc018c4e548
     # via pygithub
-urllib3==2.5.0 \
-    --hash=sha256:3fc47733c7e419d4bc3f6b3dc2b4f890bb743906a30d56ba4a5bfa4bbff92760 \
-    --hash=sha256:e6b01673c0fa6a13e374b50871808eb3bf7046c4b125b216f6bf1cc604cff0dc
+urllib3==2.6.2 \
+    --hash=sha256:016f9c98bb7e98085cb2b4b17b87d2c702975664e4f060c6532e64d1c1a5e797 \
+    --hash=sha256:ec21cddfe7724fc7cb4ba4bea7aa8e2ef36f607a4bab81aa6ce42a13dc3f03dd
     # via
     #   pygithub
     #   requests
--- a/internal/healthcheck/healthcheck.go
+++ b/internal/healthcheck/healthcheck.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -19,7 +20,6 @@ import (
 
 	"github.com/nginx/kubernetes-ingress/internal/configs"
 	"github.com/nginx/nginx-plus-go-client/v3/client"
-	"k8s.io/utils/strings/slices"
 )
 
 // RunHealthCheck starts the deep healthcheck service.
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -840,9 +840,9 @@ typing-extensions==4.15.0 \
     # via
     #   -r requirements.in
     #   pyee
-urllib3==2.5.0 \
-    --hash=sha256:3fc47733c7e419d4bc3f6b3dc2b4f890bb743906a30d56ba4a5bfa4bbff92760 \
-    --hash=sha256:e6b01673c0fa6a13e374b50871808eb3bf7046c4b125b216f6bf1cc604cff0dc
+urllib3==2.6.2 \
+    --hash=sha256:016f9c98bb7e98085cb2b4b17b87d2c702975664e4f060c6532e64d1c1a5e797 \
+    --hash=sha256:ec21cddfe7724fc7cb4ba4bea7aa8e2ef36f607a4bab81aa6ce42a13dc3f03dd
     # via
     #   -r requirements.in
     #   docker


internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -172,6 +172,12 @@ server {
 	{{$value}}{{end}}
 	{{- end}}
 
+	{{- if $server.AppRoot }}
+	if ($uri = /) {
```
pdabelf5, line 176: Does the AppRoot annotation only work if the requested path is `/`.  I would think the content of the annotation would be prepended to all paths

AlexFenlon: As far as I can tell when testing this on ingress-nginx, yes, it would only redirect /. Exact functionality here.

The annotation can be whatever eg. `/coffee` so if the user was to go to the root of the website eg. cafe.example.com/ it would redirect to cafe.example.com/coffee.

AlexFenlon: at least from what I can tell, it shouldn't change /tea to /coffee/tea for example

pdabelf5: okay, it's not what I was expecting but if that's the expected behaviour then it's all good

pdabelf5 (dismissed):

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -172,6 +172,12 @@ server {
...
+		return 302 $scheme://$http_host{{ $server.AppRoot }};
```
pdabelf5, line 177: Should this use the redirect code from @haywoodsh PR about redirect codes

AlexFenlon: Had a conversation with Haywood last week based on exactly this, as we want to replicates ingress-nginx's functionality, they have that annotation but no not use it, this is only for return code 302. 

> Haywood
ingress-nginx has the configmap key [http-redirect-code](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#http-redirect-code) (default to 308)
but if it's hardcoded, then AppRoot is ignoring it.
FYI, in my NIC branch, for Ingress, both [nginx.org/redirect-to-https](http://nginx.org/redirect-to-https) and [nginx.org/redirect-to-https](http://nginx.org/redirect-to-https) would be affected by this ConfigMap key or the annotation [nginx.org/http-redirect-code](http://nginx.org/http-redirect-code) if it's set. this configmap key also changes the default redirect code for TLS redirect in VS
I don't know much about when different redirect codes should be used, so I'm not sure if the AppRoot redirect code should always be 302 or if it should be customisable... (edited) 

> Haywood
> but then the AppRoot annoation seems to be pretty straight forward and should be a quick win for us. maybe we dont need to dig this deep...
> 

> Alex
> Yeah I don't think we need to dig deeper, I think I can move forward by hardcoding it as Micheal does say he wants to replicate community ingress, so it should be fine.

pdabelf5: `http-redirect-code` sounds like it affects all redirects, not just https redirect.  I think we should make the annotation name more specific if it only works for one type of redirect.

internal/k8s/validation.go
```diff
@@ -385,6 +389,48 @@ func validateHTTPRedirectCodeAnnotation(context *annotationValidationContext) fi
...
+	if strings.Contains(path, "../") || strings.Contains(path, "..\\") {
+		return field.ErrorList{field.Invalid(context.fieldPath, path, "path traversal patterns not allowed in rewrite target")}
+	}
+
+	// Prevents invalid config characters
+	if !validateRFC1738Path(path) {
+		allErrs = append(allErrs, field.Invalid(context.fieldPath, path, "path must not contain the following characters: whitespace, '{', '}', ';', '$', '|', '^', '<', '>', '\\', '\"', '#', '[', ']'"))
+		return allErrs
+	}
+
```
pdabelf5, line 424: Did we need to remove `~` in a separate check to `validateRFC1738Path()`?

AlexFenlon: yes, disallowing ~ now!

internal/k8s/validation.go
```diff
@@ -385,6 +389,53 @@ func validateHTTPRedirectCodeAnnotation(context *annotationValidationContext) fi
...
+		return field.ErrorList{field.Invalid(context.fieldPath, path, "path traversal patterns not allowed")}
+	}
+
+	// Prevents invalid config characters
+	if !validateRFC1738Path(path) {
+		allErrs = append(allErrs, field.Invalid(context.fieldPath, path, "path must not contain the following characters: whitespace, '{', '}', ';', '$', '|', '^', '<', '>', '\\', '\"', '#', '[', ']'"))
+		return allErrs
+	}
+
+	// Prevent tilda character
```
javorszky, line 425: tiny nit, it's tilde :) https://en.wikipedia.org/wiki/Tilde


```suggestion
	// Prevent tilde character
```

javorszky (approved):
one tiny nit, not a blocker

internal/k8s/validation.go
AlexFenlon: wasn't going to re-request reviews and have the entire pipeline rerun just for this typo so this is fixed here

https://github.com/nginx/kubernetes-ingress/pull/8818

Issue #8735, opened 2025-12-16
[Feature]: Add commonLabels to Helm Chart


nxf5025:
**Is your feature request related to a problem? Please describe.**
Currently, there's no way to add custom labels to all Kubernetes resources managed by the nginx-ingress Helm chart. Users who need to apply organization-wide labels (e.g., for cost tracking, compliance, or resource management) must manually patch each resource type individually, which is error-prone and difficult to maintain across chart upgrades.

**Describe the solution you'd like**
Add a commonLabels configuration option to the Helm chart that allows users to specify custom labels applied to all Kubernetes resource metadata. This follows the same pattern used by the ingress-nginx chart:
```
commonLabels:
  team: platform
  environment: production
  cost-center: "12345"
```
These labels would be automatically applied to all resources (Deployments, Services, ConfigMaps, Roles, etc.) but excluded from pod template labels to avoid selector conflicts.

**Describe alternatives you've considered**

- Using Kustomize overlays to add labels post-rendering
- Manual patching of resources after deployment
- Creating a wrapper chart that adds labels to all resources
- Using kubectl label commands after installation

All alternatives require additional tooling and maintenance overhead compared to native Helm chart support.

**Additional context**
This is a common pattern in enterprise Kubernetes deployments where labels are required for cost allocation, compliance tracking, or automated resource management policies.

PR #8737, opened 2025-12-16
feat: Added commonLabels to helm chart

nxf5025:
### Proposed changes

Adds commonLabels to the Helm chart. Refer to #8735 for more details

### Checklist


nxf5025:

pdabelf5 (dismissed):

AlexFenlon (dismissed):

vepatel (approved):

charts/nginx-ingress/values.schema.json
```diff
@@ -10,6 +10,12 @@
     "serviceInsight"
   ],
   "properties": {
+    "commonLabels": {
+      "type": "object",
```
haywoodsh, line 14: I don't think it should be object without any validation. The schema could be like `extralabels`


https://github.com/nginx/kubernetes-ingress/blob/73ea8d0cda592cd63f5c7e1bba5d56cccca02232/charts/nginx-ingress/values.schema.json#L1852-L1856

there are some other label fields in this schema that use `object` and will need to be fixed, but it would be out of scope of this PR

charts/nginx-ingress/values.schema.json
```diff
@@ -10,6 +10,12 @@
...
+      "default": {},
+      "title": "Common labels to apply to all resources",
+      "additionalProperties": true
```
haywoodsh, line 17: I don't think this is required. `true` seems to be the default anyway.

charts/nginx-ingress/values.yaml
```diff
@@ -1,3 +1,6 @@
+## Labels to apply to all resources.
```
haywoodsh, line 1: should probably clearly state that this field does not apply to the pod label if the following in the [issue](https://github.com/nginx/kubernetes-ingress/issues/8735) is still true. the same in the schema title or description

>  all resources (Deployments, Services, ConfigMaps, Roles, etc.) but excluded from pod template labels to avoid selector conflicts.

pdabelf5:
@nxf5025 can you make the adjustments suggested?  If you no longer have capacity I can update the PR to get it merged.

PR #8869, opened 2026-01-13
Refactor getValidTarget to return a list of items

### LINKED ISSUE #8708: [Bug]: DNSEndpoint recordType is incorrect for dual-stack services
### Version

edge

AKS Azure

### Steps to reproduce

When using a dual-stack LoadBalancer Service, the NGINX Ingress Controller generates an invalid DNSEndpoint.

Instead of creating separate DNS records for IPv4 (A) and IPv6 (AAAA), a single A record is created with both IPv4 and IPv6 addresses as targets.

Actual behavior:
```
apiVersion: externaldns.nginx.org/v1
kind: DNSEndpoint
metadata:
  name: podinfo.example.com
  namespace: podinfo
spec:
  endpoints:
    - dnsName: podinfo.example.com
      recordType: A
      targets:
        - 127.0.0.1
        - ::1
```

Expected behavior:
```
apiVersion: externaldns.nginx.org/v1
kind: DNSEndpoint
metadata:
  name: podinfo.example.com
  namespace: podinfo
spec:
  endpoints:
    - dnsName: podinfo.example.com
      recordType: A
      targets:
        - 127.0.0.1
    - dnsName: podinfo.example.com
      recordType: AAAA
      targets:
        - ::1
```

javorszky:
### Proposed changes

Closes #8708

### Checklist

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -702,9 +702,9 @@ pyopenssl==25.3.0 \
     --hash=sha256:1fda6fc034d5e3d179d39e59c1895c9faeaf40a79de5fc4cbbfbe0d36f4a77b6 \
     --hash=sha256:c981cb0a3fd84e8602d7afc209522773b94c1c2446a3c710a75b06fe1beae329
     # via -r requirements.in
-pyparsing==3.3.1 \
-    --hash=sha256:023b5e7e5520ad96642e2c6db4cb683d3970bd640cdf7115049a6e9c3682df82 \
-    --hash=sha256:47fad0f17ac1e2cad3de3b458570fbc9b03560aa029ed5e16ee5554da9a2251c
+pyparsing==3.3.2 \
+    --hash=sha256:850ba148bd908d7e2411587e247a1e4f0327839c40e2e5e6d05a007ecc69911d \
+    --hash=sha256:c777f4d763f140633dcb6d8a3eda953bf7a214dc4eff598413c070bcdc117cbc
     # via -r requirements.in
 pytest==9.0.2 \
     --hash=sha256:711ffd45bf766d5264d487b917733b453d917afd2b0ad65223959f59089f875b \


internal/externaldns/sync_test.go
```diff
@@ -74,19 +97,57 @@ func TestGetValidTargets(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "from external endpoint with multiple targets",
```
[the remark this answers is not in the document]
javorszky: added that additional info

pkg/apis/externaldns/v1/types.go
```diff
@@ -35,7 +35,7 @@ type Endpoint struct {
 	DNSName string `json:"dnsName,omitempty"`
 
 	// The targets the DNS service points to
-	Targets Targets `json:"targets,omitempty"`
+	Targets []string `json:"targets,omitempty"`
```
pdabelf5: This looks like a breaking change

pkg/apis/externaldns/validation/externaldns.go
```diff
@@ -48,7 +48,7 @@ func validateDNSName(name string) error {
 	return nil
 }
 
-func validateTargets(targets v1.Targets) error {
+func validateTargets(targets []string) error {
```
pdabelf5: Is this change still appropriate as we're no longer updating the CRD

pkg/apis/externaldns/v1/types.go
pdabelf5: Code updated to not need a CRD update

pkg/apis/externaldns/validation/externaldns.go
javorszky: No, good catch, let me fix that!

javorszky: this is fixed in https://github.com/nginx/kubernetes-ingress/pull/8869/commits/7d519a2e22edf5e7df869bd33aed6d7d5013b778

PR #9213, opened 2026-02-19
Fix `unexpected ";"` using zone-sync in the Configmap while disabling IPV6

AlexFenlon:
### Proposed changes

Fix a bug where zone_sync with IPV6 disabled leaves a left over `;`. Updated test to account for this. 

### Checklist

### DIFF
--- a/internal/configs/version1/nginx-plus.tmpl
+++ b/internal/configs/version1/nginx-plus.tmpl
@@ -353,7 +353,9 @@ stream {
     {{- if .ZoneSyncConfig.Enable }}
     server {
         listen {{ .ZoneSyncConfig.Port }};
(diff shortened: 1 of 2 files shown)


internal/configs/version1/nginx-plus.tmpl
```diff
@@ -353,7 +353,7 @@ stream {
     {{- if .ZoneSyncConfig.Enable }}
     server {
         listen {{ .ZoneSyncConfig.Port }};
-        {{ if not .DisableIPV6}}listen [::]:{{ .ZoneSyncConfig.Port }}{{end}};
+        {{- if not .DisableIPV6}}listen [::]:{{ .ZoneSyncConfig.Port }};{{- end}}
```
javorszky: If the leading `-` is removed, does it break the ipv6 into a newline in the snaps? 🤔

PR #9279, opened 2026-03-02
Change resource type to ContainerResource in HPA

BROGOTSOME:
Updated Horizontal Pod Autoscaler (HPA) configuration to use container-level resource metrics instead of overall pod resource metrics for more granular scaling control

### Proposed changes

Updated the HPA metric configuration from resource-level 
(pod-wide CPU/memory usage) to container-level resource metrics.

This ensures scaling decisions are made based on individual 
container utilization, improving precision in multi-container 
pod environments and preventing inaccurate scaling behavior.

### Checklist

### DIFF
--- a/charts/nginx-ingress/templates/controller-hpa.yaml
+++ b/charts/nginx-ingress/templates/controller-hpa.yaml
@@ -23,17 +23,19 @@ spec:
 {{- end }}
   metrics:
     {{- if .Values.controller.autoscaling.targetMemoryUtilizationPercentage }}
-    - type: Resource
-      resource:
+    - type: ContainerResource
+      containerResource:
         name: memory
+        container: {{ include "nginx-ingress.name" . }}
         target:
           type: Utilization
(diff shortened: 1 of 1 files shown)


BROGOTSOME:

BROGOTSOME:
recheck

vepatel:
@BROGOTSOME  thanks for the PR, please take some time to read project contribution guidelines and create an issue outlining the problem this PR solves, ty

haywoodsh:
This seems to be a breaking change in behaviour that could affect existing users. Are you using `extraContainers` to include other containers but you want to exclude the resource usage of those containers to be calculated for autoscaling?

Wouldn't it be better to add a toggle such as `controllerOnly` (name TBD) defaulting to false, so that this change would not affect existing users upgrading?

vepatel:
hey @BROGOTSOME , is this still an issue? are there any custom container you're adding to deployment which needs this?

PR #9280, opened 2026-03-02
Ingress Policy: AccessControl

sean-breen:
### Proposed changes

Recreated PR, old PR with feedback https://github.com/nginx/kubernetes-ingress/pull/9214

This change enables the AccessControl policy type on Ingress objects via the `nginx.org/policies` annotation. This works the same as the AccessControl policy on a VS/VSR.

### Checklist


examples/ingress-resources/access-control/README.md
haywoodsh, line 1: I think the readme should be step-by-step instructions to deploy the working example and testing the response from request showing the effect of the policy. see https://github.com/nginx/kubernetes-ingress/blob/main/examples/ingress-resources/basic-auth/README.md

internal/k8s/controller.go
```diff
@@ -2194,6 +2183,33 @@ func (lbc *LoadBalancerController) createIngressEx(ing *networking.Ingress, vali
 		ingEx.SecretRefs[secretName] = secretRef
 	}
 
+	var policyNames string
+	var policyRefs []conf_v1.PolicyReference
+	if ingEx.Ingress.Annotations[configs.PoliciesAnnotation] != "" {
+		policyNames = ingEx.Ingress.Annotations[configs.PoliciesAnnotation]
+		policyRefs = k8spolicies.GetPolicyRefsFromAnnotation(policyNames, ing.Namespace)
+	}
+	policies, policyErrors := lbc.getPolicies(policyRefs, ing.Namespace)
```
haywoodsh, line 2192: how would it work with policies in a different namepsace when the annotation is like    `nginx.org/policies: "my-policy,mynamespace/another-policy"`?

tests/suite/utils/policy_resources_utils.py
```diff
@@ -60,6 +60,22 @@ def delete_policy(custom_objects: CustomObjectsApi, name, namespace) -> None:
     print(f"Policy was removed with name '{name}'")
 
 
+def apply_and_wait_for_valid_policy(kube_apis, namespace, policy_yaml, retry_count=30, interval=1) -> str:
```
haywoodsh, line 63: Does this policy require a retry or a longer interval? I'm interested to know if there is any reason to create another method rather than apply_and_assert_valid_policy below

internal/k8s/controller.go
pdabelf5: GetPolicyRefsFromAnnotation() will parse the policy, and only add `ing.Namespace` if the policy is not in the `namespace/name` format.

tests/suite/utils/policy_resources_utils.py
pdabelf5: During testing, it was taking some time (1-3 seconds) after the policy apply for the Resource to reflect the new status.  We found that polling is more consistent than just waiting for the changes happen in that time period.

examples/ingress-resources/access-control/README.md
pdabelf5: Readme style has been updated

tests/suite/test_ac_policies_ingress.py
```diff
@@ -0,0 +1,318 @@
...
+
+std_cm_src = f"{DEPLOYMENTS}/common/nginx-config.yaml"
+test_cm_src = f"{TEST_DATA}/access-control/configmap/nginx-config.yaml"
+
+deny_pol_src = f"{TEST_DATA}/access-control/policies/access-control-policy-deny.yaml"
+allow_pol_src = f"{TEST_DATA}/access-control/policies/access-control-policy-allow.yaml"
+invalid_pol_src = f"{TEST_DATA}/access-control/policies/access-control-policy-invalid.yaml"
+
+reload_times = {}
+
```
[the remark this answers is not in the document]
pdabelf5: removed reload_times{}

tests/suite/utils/policy_resources_utils.py
```diff
@@ -60,6 +60,22 @@ def delete_policy(custom_objects: CustomObjectsApi, name, namespace) -> None:
...
+        if (
+            "status" in policy_info
+            and policy_info["status"]["reason"] == "AddedOrUpdated"
+            and policy_info["status"]["state"] == "Valid"
+        ):
+            return pol_name
+        wait_before_test(interval)
+
+    return ""
+
```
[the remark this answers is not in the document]
pdabelf5: refactored to raise an exception on timeout

PR #9316, opened 2026-03-11
Enable Session Persistence for NGINX OSS config

sean-breen:
### Proposed changes

Adds session persistence to NGINX OSS configurations, via the `sticky cookie` method.

Session persistence is no longer a Plus-only feature.

Adds the `nginx.org/sticky-cookie-services` annotation for OSS, acts as an alias for `nginx.com/sticky-cookie-services`. Prioritises `nginx.com/` if both are defined.

- updates nginx OSS + nginx plus templates
- updates Unit tests to move session cookie tests to OSS examples, remove error checks for 'plus-only' feature
- updates the e2e tests

### Checklist

### DIFF
--- a/internal/configs/configurator.go
+++ b/internal/configs/configurator.go
@@ -478,6 +478,25 @@ func (cnf *Configurator) AddOrUpdateMergeableIngress(mergeableIngs *MergeableIng
 	return warnings, nil
 }
 
+// AddOrUpdateMergeableIngresses adds or updates NGINX configuration for the list of mergeable Ingress resources.
+func (cnf *Configurator) AddOrUpdateMergeableIngresses(mergeableIngs []*MergeableIngresses) (Warnings, error) {
+	allWarnings := newWarnings()
+
+	for _, mergeableIng := range mergeableIngs {
+		_, warnings, err := cnf.addOrUpdateMergeableIngress(mergeableIng)
+		if err != nil {
+			return allWarnings, err
+		}
+		allWarnings.Add(warnings)
+	}
+
+	if err := cnf.Reload(nginx.ReloadForOtherUpdate); err != nil {
+		return allWarnings, fmt.Errorf("error when reloading NGINX when updating mergeable Ingresses: %w", err)
+	}
+
+	return allWarnings, nil
+}
+
 func (cnf *Configurator) addOrUpdateMergeableIngress(mergeableIngs *MergeableIngresses) (bool, Warnings, error) {
 	apResources := cnf.updateApResources(mergeableIngs.Master)
 	cnf.updateDosResource(mergeableIngs.Master.DosEx)
--- a/internal/configs/ingress.go
+++ b/internal/configs/ingress.go
@@ -42,6 +42,7 @@ type IngressEx struct {
 	Endpoints        map[string][]string
 	HealthChecks     map[string]*api_v1.Probe
 	Policies         map[string]*conf_v1.Policy
+	PolicyWarnings   []string
 	ExternalNameSvcs map[string]bool
 	PodsByIP         map[string]PodInfo
 	ValidHosts       map[string]bool
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -981,7 +981,8 @@ func (lbc *LoadBalancerController) updateAllConfigs() {
 		lbc.recorder.Eventf(gc, eventType, eventTitle, fmt.Sprintf("GlobalConfiguration %s was updated %s", key, eventWarningMessage))
 	}
 
-	lbc.updateResourcesStatusAndEvents(resources, warnings, updateErr)
+	resourcesWithWarnings := mergeExtendedResourceWarnings(resources, resourceExes)
+	lbc.updateResourcesStatusAndEvents(resourcesWithWarnings, warnings, updateErr)
 }
 
 // preSyncSecrets adds Secret resources to the SecretStore.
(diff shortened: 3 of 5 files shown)


internal/k8s/validation.go
```diff
@@ -360,7 +361,10 @@ var (
 			validateRewriteTargetAnnotation,
 		},
 		stickyCookieServicesAnnotation: {
```
pdabelf5, line 363: This could re-use the const from the `configs` packge

internal/k8s/validation.go
```diff
@@ -360,7 +361,10 @@ var (
...
-			validatePlusOnlyAnnotation,
+			validateRequiredAnnotation,
+			validateStickyServiceListAnnotation,
+		},
+		stickyCookieServicesAnnotationPlus: {
```
pdabelf5, line 367: This could re-use the const from the `configs` package

internal/k8s/validation_test.go
```diff
@@ -3613,6 +3613,115 @@ func TestValidateNginxIngressAnnotations(t *testing.T) {
 			msg: "invalid nginx.org/rewrites annotation",
 		},
 
+		{
```
pdabelf5, line 3616: These could re-use the const from the `configs` package

pkg/apis/configuration/v1/types.go
```diff
@@ -171,7 +171,7 @@ type Upstream struct {
 	SlowStart string `json:"slow-start"`
 	// Configures a queue for an upstream. A client request will be placed into the queue if an upstream server cannot be selected immediately while processing the request. By default, no queue is configured. Note: this feature is supported only in NGINX Plus.
 	Queue *UpstreamQueue `json:"queue"`
-	// The SessionCookie field configures session persistence which allows requests from the same client to be passed to the same upstream server. The information about the designated upstream server is passed in a session cookie generated by NGINX Plus.
+	// The SessionCookie field configures session persistence which allows requests from the same client to be passed to the same upstream server. The information about the designated upstream server is passed in a session cookie generated by NGINX OSS or Plus.
```
pdabelf5:
```suggestion
	// The SessionCookie field configures session persistence which allows requests from the same client to be passed to the same upstream server. The information about the designated upstream server is passed in a session cookie generated by NGINX.
```

internal/k8s/validation.go
```diff
@@ -360,7 +361,10 @@ var (
 			validateRewriteTargetAnnotation,
 		},
 		stickyCookieServicesAnnotation: {
-			validatePlusOnlyAnnotation,
```
haywoodsh, line 363: Even though it has the same functionality, would it cause any confusion if an 'nginx.com' annotation could be used in the OSS image? 
Would it make more sense to maintain the convention that the `nginx.org` annotations are available for OSS and the `nginx.com` annotations are for Plus only, for all Ingress annotations, in order to avoid any confusion regarding the features of the Ingress Controller in different versions?

internal/configs/annotations.go
```diff
@@ -709,14 +716,33 @@ func getGrpcServices(ingEx *IngressEx) map[string]bool {
...
+	valuePlus, plusExists := ingEx.Ingress.Annotations[StickyCookieServicesAnnotationPlus]
+	valueOrg, orgExists := ingEx.Ingress.Annotations[StickyCookieServicesAnnotation]
+
+	if !plusExists && !orgExists {
+		return nil
 	}
-	return nil
+
+	if plusExists && orgExists {
+		nl.Warnf(l, "Ingress %s/%s: both %s and %s annotations are set; using %s",
```
haywoodsh, line 734: If there is no plan to deprecate either of the annotations, is it necessary for the log level to be set to `warn`? A warning-level log message could trigger an alert in some monitoring systems.
Also, the user applying the ingress might not be aware of this immediately, as they may not have access to the logs, and no k8s event is emitted for the ingress in question.
If the log level has to be `warn` here, the ingress might need to equally go into a `Warning` state, but I am not sure whether it should be in this case.

internal/configs/version1/nginx.ingress.tmpl
```diff
@@ -131,11 +139,11 @@ server {
...
-    {{- if $server.AppRoot }}
-    if ($uri = /) {
-        return 302 $scheme://$http_host{{ $server.AppRoot }};
-    }
-    {{- end }}
+	{{- if $server.AppRoot }}
+	if ($uri = /) {
+		return 302 $scheme://$http_host{{ $server.AppRoot }};
+	}
+	{{- end }}
```
haywoodsh: what is the actual change here?

sean-breen: Oh looks like it added tabs instead of spaces, my editor must have done that, I'll revert

internal/configs/annotations.go
sean-breen: Yes good point, thanks Haywood. Do you think it would be better to log the message as `info` to avoid triggering the `Warning`?

sean-breen: Added the message at `Info` level

haywoodsh (approved):

internal/configs/annotations.go
```diff
@@ -76,12 +76,19 @@ const AppProtectDosProtectedAnnotation = "appprotectdos.f5.com/app-protect-dos-r
...
+const StickyCookieServicesAnnotationPlus = "nginx.com/sticky-cookie-services"
+
 var masterDenylist = map[string]bool{
 	"nginx.org/rewrites":                      true,
 	"nginx.org/ssl-services":                  true,
 	"nginx.org/grpc-services":                 true,
 	"nginx.org/websocket-services":            true,
-	"nginx.com/sticky-cookie-services":        true,
+	StickyCookieServicesAnnotation:            true,
+	StickyCookieServicesAnnotationPlus:        true,
```
javorszky, line 91: I really like these being consts!

internal/k8s/validation_test.go
```diff
@@ -3615,21 +3615,130 @@ func TestValidateNginxIngressAnnotations(t *testing.T) {
 
 		{
 			annotations: map[string]string{
-				"nginx.com/sticky-cookie-services": "true",
+				configs.StickyCookieServicesAnnotation: "serviceName=service-1 srv_id expires=1h path=/service-1",
```
javorszky: I think in tests we should use the naked strings, otherwise this becomes circular. If for example the const changes to something different, this test would not catch the issue, but a naked string key would.

internal/configs/annotations_test.go
```diff
@@ -71,6 +71,69 @@ func TestParseStickyServiceInvalidFormat(t *testing.T) {
...
+
+	tests := []struct {
+		name        string
+		annotations map[string]string
+		expected    map[string]string
+	}{
+		{
+			name: "nginx.org annotation only",
+			annotations: map[string]string{
+				StickyCookieServicesAnnotation: "serviceName=service-1 srv_id expires=1h path=/",
```
javorszky: Similarly here, this map key should be the naked string rather than the const. Reasoning is in the other comment

internal/configs/annotations.go
```diff
@@ -718,24 +718,26 @@ func getSessionPersistenceServices(ctx context.Context, ingEx *IngressEx) map[st
...
 		return nil
 	}
 
+	var value string
+	if plusExists {
+		value = valuePlus
+	}
+	if orgExists {
+		value = valueOrg
+	}
```
javorszky, line 731: Super tiny nit, also not a blocker.

This can be simplified. On line 724 we already break when neither of them exists, so that means at least one of them exists.

That means we can do this:

```
value := valuePlus
if orgExists {
    value = valueOrg
}
```

| org | plus | value | expectation met? | notes |
|-|-|-|-|-|
| 🔴  | 🔴 | break | ✅ | returned nil on line 724 |
| 🔴 | ✅  | plus | ✅ | default plus, org does not override |
| ✅ | 🔴 | org | ✅ | default plus, org overrides |
| ✅ | ✅ | org | ✅ | default plus, org overrides |

internal/k8s/validation_test.go
sean-breen: Yes that's a good point, I've reverted to naked strings in the suggested locations, thanks Gabor!

internal/configs/annotations.go
sean-breen: Added!

PR #9331, opened 2026-03-12
fix: dont check for docker on local builds

zamadye:
saw the issue about docker check blocking local builds. moved the docker check into the download/container targets so local builds work without docker installed.

tested locally with `make build TARGET=local` and it builds fine now. let me know if you need any changes!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,15 +137,17 @@ telemetry-schema: ## Generate the telemetry Schema
 
 .PHONY: build
 build: ## Build Ingress Controller binary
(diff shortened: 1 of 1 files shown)


zamadye:

AlexFenlon:
Hey @zamadye, thank you for this change. 

It seems like our F5 CLA check is failing due to a different user in the inital comit, can you please rebase using the @zamadye user so we will be able to continue and review this PR.

zamadye:
thanks for the feedback! 👍

PR #9547, opened 2026-04-07
Support add_header_inherit directive

### LINKED ISSUE #9387: Add support for the add_header_inherit NGINX directive

add_header_inherit directive allows altering inheritance rules for the values specified in the add_header (https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header) directives. By default, the [standard](https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header_default_inherit) inheritance model is used.

For example, as a user I may need to merge headers defined on both ‘location’ and ‘server’.

The ask is to include add_header_inherit as a configurable parameter in NGINX Ingress Controller. 

It should support the three values: on | off | merge;

It's possible with snippets today

**Additional context**
Ideally this is supported in both the Ingress object and VirtualServer, VirtualServerRoute

sean-breen:
### Proposed changes

- [x] Add support for the NGINX `add_header_inherit` directive in Ingress, VS + VSR.
- [ ] Update docs

### Checklist


pkg/apis/configuration/v1/types.go
```diff
@@ -63,6 +63,9 @@ type VirtualServerSpec struct {
 	HTTPSnippets string `json:"http-snippets"`
 	// Sets a custom snippet in server context. Overrides the server-snippets ConfigMap key.
 	ServerSnippets string `json:"server-snippets"`
+	// Controls header inheritance behavior at the server level. Allowed values are: on, off, merge. When set to "merge", headers from this context are merged with headers in child contexts. When set to "on", standard NGINX inheritance applies. When set to "off", no headers are inherited from parent contexts.
+	// +kubebuilder:validation:Pattern=`^[oO][nN]$|^[oO][fF][fF]$|^[mM][eE][rR][gG][eE]$`
```
haywoodsh: I don't see the need to accept strings containing uppercase characters and then convert them to lowercase before validation in go code?
Should we just use enum here since it's very clear that we allow only three values as per the description?

e.g.
```
	// +kubebuilder:validation:Enum=http1;http2;grpc;websocket
```
https://github.com/nginx/kubernetes-ingress/blob/v5.4.1/pkg/apis/dos/v1beta1/types.go#L43

vepatel: @sean-breen  something like `// +kubebuilder:validation:XValidation:rule="self in ['on', 'off', 'merge']",message="allowed values must be one of: on, off, merge"` should work as well.
then we won't need `switch strings.ToLower(value)` in virtualserver.go

pkg/apis/configuration/v1/types.go
```diff
@@ -287,6 +290,9 @@ type Route struct {
 	ErrorPages []ErrorPage `json:"errorPages"`
 	// Sets a custom snippet in the location context. Overrides the location-snippets ConfigMap key.
 	LocationSnippets string `json:"location-snippets"`
+	// Controls header inheritance behavior at the location level. Allowed values are: on, off, merge. When set to "merge", headers from this context are merged with headers in child contexts. When set to "on", standard NGINX inheritance applies. When set to "off", no headers are inherited from parent contexts.
+	// +kubebuilder:validation:Pattern=`^[oO][nN]$|^[oO][fF][fF]$|^[mM][eE][rR][gG][eE]$`
```
vepatel:
```suggestion
	// +kubebuilder:validation:XValidation:rule="self in ['on', 'off', 'merge']",message="allowed values must be one of: on, off, merge"
```

internal/configs/virtualserver.go
```diff
@@ -896,6 +899,7 @@ func (vsc *virtualServerConfigurator) GenerateVirtualServerConfig(
 				)
 				addPoliciesCfgToLocations(routePoliciesCfg, cfg.Locations)
 				addDosConfigToLocations(dosRouteCfg, cfg.Locations)
+				addAddHeaderInheritToLocations(r.AddHeaderInherit, cfg.Locations)
```
haywoodsh: This should fall back to the VS route's value when unset, otherwise the value set on a VS route is never used.

VirtualServer and VirtualServerRoute do not rely on standard NGINX inheritance model, because in the generated NGINX config, both VS routes and VSR subroutes are `location` blocks, so there is no parent-child relationship between them. In NIC's resource hierarchy, the VS route is logically above the VSR subroute, so the fallback chain should be: VSR subroute (`location`) -> VS route (`location`) -> VS spec (`server`)-> ConfigMap (`http`).

internal/configs/configmaps_test.go
```diff
@@ -2990,3 +2990,102 @@ func TestParseConfigMapWithHTTPRedirectCode(t *testing.T) {
...
+			msg:         "invalid value",
+		},
+		{
+			value:       "yes",
+			expect:      "",
+			expectError: true,
+			msg:         "invalid value yes",
+		},
+		{
+			value:       "",
```
pdabelf5, line 3063: ParseConfigMap() calls ParseAddHeaderInherit() which returns an error is the value doesn't match on, off or merge.  Why is an empty value ok?

internal/configs/parsing_helpers.go
```diff
@@ -166,6 +166,25 @@ func ParseBool(s string) (bool, error) {
...
+	AddHeaderInheritOff = "off"
+	// AddHeaderInheritMerge appends inherited add_header values to the current context values.
+	AddHeaderInheritMerge = "merge"
+)
+
+// ParseAddHeaderInherit validates the add_header_inherit directive mode.
+func ParseAddHeaderInherit(value string) (string, error) {
+	switch value {
+	case AddHeaderInheritOn, AddHeaderInheritOff, AddHeaderInheritMerge:
+		return value, nil
```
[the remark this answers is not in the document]
pdabelf5: I agree, we should add a ToLower() here

internal/configs/parsing_helpers_test.go
```diff
@@ -703,6 +703,31 @@ func TestParseBool(t *testing.T) {
 	}
 }
 
+func TestParseAddHeaderInherit(t *testing.T) {
```
pdabelf5, line 706: Add a test case for empty

internal/k8s/validation_test.go
```diff
@@ -963,6 +963,64 @@ func TestValidateNginxIngressAnnotations(t *testing.T) {
 			expectedErrors:        nil,
 			msg:                   "valid nginx.org/location-snippets annotation, multi-value",
 		},
+		{
+			annotations: map[string]string{
+				"nginx.org/add-header-inherit": "on",
```
pdabelf5, line 968: Use the const from the configs package

internal/k8s/validation_test.go
```diff
@@ -963,6 +963,64 @@ func TestValidateNginxIngressAnnotations(t *testing.T) {
...
+			appProtectDosEnabled:  false,
+			internalRoutesEnabled: false,
+			snippetsEnabled:       true,
+			directiveAutoAdjust:   false,
+			expectedErrors:        nil,
+			msg:                   "valid nginx.org/add-header-inherit annotation with on",
+		},
+		{
+			annotations: map[string]string{
+				"nginx.org/add-header-inherit": "off",
```
pdabelf5, line 982: Use the const from the configs package

internal/k8s/validation_test.go
```diff
@@ -963,6 +963,64 @@ func TestValidateNginxIngressAnnotations(t *testing.T) {
...
+			appProtectDosEnabled:  false,
+			internalRoutesEnabled: false,
+			snippetsEnabled:       true,
+			directiveAutoAdjust:   false,
+			expectedErrors:        nil,
+			msg:                   "valid nginx.org/add-header-inherit annotation with off",
+		},
+		{
+			annotations: map[string]string{
+				"nginx.org/add-header-inherit": "merge",
```
pdabelf5, line 996: Use the const from the configs package

internal/k8s/validation_test.go
```diff
@@ -963,6 +963,64 @@ func TestValidateNginxIngressAnnotations(t *testing.T) {
...
+			appProtectDosEnabled:  false,
+			internalRoutesEnabled: false,
+			snippetsEnabled:       true,
+			directiveAutoAdjust:   false,
+			expectedErrors:        nil,
+			msg:                   "valid nginx.org/add-header-inherit annotation with merge",
+		},
+		{
+			annotations: map[string]string{
+				"nginx.org/add-header-inherit": "bogus",
```
pdabelf5, line 1010: Use the const from the configs package

pkg/apis/configuration/validation/virtualserver_test.go
```diff
@@ -4799,3 +4799,46 @@ func TestValidateRouteSelector(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateAddHeaderInherit_ValidValues(t *testing.T) {
+	t.Parallel()
+
+	validValues := []string{
+		"",
```
pdabelf5, line 4807: Is an empty string a valid value?

internal/configs/annotations.go
haywoodsh, line 132: should the annotation be add to this `minionInheritanceList` so that the default value of a `add_header_inherit` of a minion Ingress would follow the Master Ingress at server level instead of the configmap (http level), the same way as how VS and VSR works for this directive?

sean-breen: Setting the inheritance already modifies the behaviour for child contexts, so I don't think we need to explicitly write the values into every minion.

pkg/apis/configuration/v1/types.go
```diff
@@ -63,6 +63,9 @@ type VirtualServerSpec struct {
 	HTTPSnippets string `json:"http-snippets"`
 	// Sets a custom snippet in server context. Overrides the server-snippets ConfigMap key.
 	ServerSnippets string `json:"server-snippets"`
+	// Controls header inheritance behavior at the server level. Allowed values are: on, off, merge. When set to "merge", headers from this context are merged with headers in child contexts. When set to "on", standard NGINX inheritance applies. When set to "off", no headers are inherited from parent contexts.
+	// +kubebuilder:validation:Enum=on;off;merge;
+	AddHeaderInherit string `json:"add-header-inherit"`
```
pdabelf5, line 68: Should we use omitempty here if there is no default

pdabelf5: it seems omitempty would not change if the value in the Spec is an empty string, probably not worth the time to change

internal/configs/ingress.go
```diff
@@ -515,6 +516,9 @@ func generateNginxCfg(ncp NginxCfgParams) (version1.IngressNginxConfig, Warnings
 			}
 
 			if ncp.isMinion {
+				loc.AddHeaderInherit = cfgParams.AddHeaderInherit
+				server.AddHeaderInherit = ""
```
pdabelf5: Adding a comment to highlight why we reset the `server.AddHeaderInherit` would be useful

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -195,6 +195,9 @@ server {
 	add_header {{$h.Name}} {{ printf "%q" $h.Value }}{{if $h.Always}} always{{end}};
 	{{- end }}
 	{{- end }}
+	{{- if $server.AddHeaderInherit}}
+    add_header_inherit {{$server.AddHeaderInherit}};
+    {{- end}}
```
AlexFenlon: minor but there are 4 spaces, it should be consistant with the rest, it should be two tabs. Line 199 and 200

pkg/apis/configuration/v1/types.go
```diff
@@ -63,6 +63,9 @@ type VirtualServerSpec struct {
 	HTTPSnippets string `json:"http-snippets"`
 	// Sets a custom snippet in server context. Overrides the server-snippets ConfigMap key.
 	ServerSnippets string `json:"server-snippets"`
+	// Controls header inheritance behavior at the server level. Allowed values are: on, off, merge. When set to "merge", headers from this context are merged with headers in child contexts. When set to "on", standard NGINX inheritance applies. When set to "off", no headers are inherited from parent contexts.
+	// +kubebuilder:validation:Enum=on;off;merge;
```
AlexFenlon: not sure about this trailing ; either, I think it's best to remove the last ;

internal/configs/version1/nginx-plus.ingress.tmpl
sean-breen: Corrected.

pkg/apis/configuration/v1/types.go
sean-breen: Done.

AlexFenlon (approved):

pdabelf5 (approved):

PR #9590, opened 2026-04-10
Add -oidc-template-path flag

### LINKED ISSUE #9589: Add -oidc-template-path flag for overriding OIDC template path
### Description

NIC supports overriding template paths for main, ingress, virtualserver, and transportserver
configs via CLI flags (`-main-template-path`, `-ingress-template-path`, etc.), but there is no
equivalent flag for the OIDC template.

### Proposed solution

Add an `-oidc-template-path` flag following the same pattern as the other template path flags.

haywoodsh:
### Proposed changes

Add `-oidc-template-path` CLI flag to allow overriding the OIDC template path,
consistent with the existing flags for main, ingress, virtualserver, and
transportserver templates.

- Add flag definition and wiring for template path override
- Add Plus-only validation that warns and clears the flag on OSS

Closes #9589

### Checklist

### DIFF
--- a/cmd/nginx-ingress/flags.go
+++ b/cmd/nginx-ingress/flags.go
@@ -114,6 +114,10 @@ var (
 		`Path to the TransportServer NGINX configuration template for a TransportServer resource.
 	(default for NGINX "nginx.transportserver.tmpl"; default for NGINX Plus "nginx-plus.transportserver.tmpl")`)
 
+	oidcTemplatePath = flag.String("oidc-template-path", "",
+		`Path to the OIDC NGINX configuration template.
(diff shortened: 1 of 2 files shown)


cmd/nginx-ingress/flags.go
```diff
@@ -274,6 +278,11 @@ func initValidate(ctx context.Context) {
 		*mgmtConfigMap = ""
 	}
 
+	if *oidcTemplatePath != "" && !*nginxPlus {
```
AlexFenlon, line 281: Should we also check if enableOIDC is actually ther before allowing the user to set this flag? 

As it is useless without enableOIDC. This is the only policy we have that uses a flag, not sure why but still good to have a check.

PR #9622, opened 2026-04-14
Support WAF IP Intelligence

### LINKED ISSUE #8837: Invetsigate how to enable the IP intelligence feature for F5 NGINX WAF

sean-breen:
### Proposed changes

Adds support F5 WAF IP Intelligence module. 

NIC's WAF integration has two modes, both requiring different sets of changes:

- for WAF "v4" the controller, WAF and its' modules are in a single container image
- for WAF "v5" the modules exist as sidecar containers (total of 4 containers per pod)

Changes to WAF v4 images: 
- install `app-protect-ip-intelligence` package from respective package manager
- allow configuring IPI via the `APPolicy` CRD (should be reviewed by WAF team member)
- adds the `--enable-app-protect-ip-intelligence=true` flag to launch the `iprepd` daemon on pod startup

Changes to v5 images:
- Introduces the `app-protect-ip-intelligence` container image as a sidecar
- Adds a shared volume which is used to store the IpRep database and persist between restarts

### Checklist


charts/nginx-ingress/values.yaml
```diff
@@ -115,6 +115,29 @@ controller:
           drop:
             - all
 
+    ## Configuration for App Protect WAF v5 IP Intelligence
+    ipIntelligence:
+      enable: false
```
vepatel, line 120: should this be enabled by default like v4?

charts/nginx-ingress/values.schema.json
```diff
@@ -2666,15 +2775,33 @@
             "port": 50000,
             "image": {
               "repository": "private-registry.nginx.com/nap/waf-enforcer",
-              "tag": "5.10.0",
+              "tag": "5.12.0",
```
vepatel, line 2778: branch needs rebase as other containers are updated to 5.12.1

charts/tests/__snapshots__/helmunit_test.snap
```diff
@@ -1832,6 +1832,8 @@ spec:
         name: app-protect-config
       - emptyDir: {}
         name: app-protect-bundles
+      - name: app-protect-ipi-db
+        emptyDir: {}
```
javorszky, line 1839: the others are
`- emptyDir: {}` first, and `name` after. Yaml doesn't care about fields within list items, so that should be fine, but would this look better if it was consistent? 🙂

javorszky (dismissed):
approved with a small question

charts/tests/__snapshots__/helmunit_test.snap
sean-breen: Yeah I'll adjust this to be consistent, thanks!

AlexFenlon (dismissed):

pdabelf5 (commented):
For Waf v4, where is the go code that starts the IPI daemon that downloads the database file.  It happens automatically when the V5 image starts

charts/nginx-ingress/values.yaml
sean-breen: Following the same pattern as the `appprotect` module here by defaulting to false. If AppProtect is enabled at all, we will enable IP intelligence also

sean-breen:
> For Waf v4, where is the go code that starts the IPI daemon that downloads the database file. It happens automatically when the V5 image starts

Good catch, I missed adding this to the startup scripts, thanks!

In fact, I've added the `iprepd` daemon to be run in a similar pattern to the agent and other processes we launch.

charts/nginx-ingress/values.schema.json
```diff
@@ -481,6 +481,97 @@
                   "$ref": "https://raw.githubusercontent.com/nginxinc/kubernetes-json-schema/master/v1.35.1/_definitions.json#/definitions/io.k8s.api.core.v1.SecurityContext"
                 }
               }
+            },
+            "ipIntelligence": {
+              "type": "object",
+              "title": "Configuration for App Protect WAF v5 IP Intelligence",
+              "properties": {
```
pdabelf5, line 488: I'm guessing this will get resource requests when the rest of v5 images do?

cmd/nginx-ingress/main.go
```diff
@@ -670,6 +677,8 @@ func startChildProcesses(nginxManager nginx.Manager, appProtectV5 bool) childPro
 		aPDosDone:      aPPDosAgentDone,
 		agentEnable:    *agent,
 		agentDone:      agentDone,
+		ipRepdEnable:   *appProtect && !appProtectV5,
```
pdabelf5, line 682: Do only v5 users get to disable IPI by not deploying the container?

pdabelf5: would a cli flag help here?

internal/nginx/manager.go
```diff
@@ -579,6 +585,39 @@ func (lm *LocalManager) AppProtectPluginQuit() {
...
+// IPRepdStart starts the IP Reputation Daemon (iprepd) for App Protect IP Intelligence.
+func (lm *LocalManager) IPRepdStart(ipRepdDone chan error) {
+	if _, err := os.Stat(ipRepdBinaryPath); os.IsNotExist(err) {
+		nl.Debugf(lm.logger, "iprepd binary not found at %s, skipping IP Intelligence", ipRepdBinaryPath)
+		return
+	}
+
+	nl.Debugf(lm.logger, "Starting IP Reputation Daemon (iprepd)")
+	cmd := exec.CommandContext(context.Background(), ipRepdBinaryPath, ipRepdConfigPath) //nolint:gosec // G204: paths resolve to constants
+	cmd.Stdout = os.Stdout
```
pdabelf5, line 597: This is going to be noisy, does the config file allow us to turn down the logging?

cmd/nginx-ingress/main.go
sean-breen: We could implement a cli flag to allow disabling IP in WAF v4, at the moment it's always on for v4 and optional for v5.

It always gets installed along with the other packages in v4 images

internal/nginx/manager.go
sean-breen: Turn down to less detail is it? It's configurable via the `iprepd.cfg` config file. We could ship a modified version of this file when we build the container image

cmd/nginx-ingress/main.go
danielnginx: User should be able to enable/disable the feature

internal/nginx/manager.go
pdabelf5: we have a cli flag for `appProtectLogLevelDefault` do the log levels of iprepd match up?

deploy/crds-nap-waf.yaml
```diff
@@ -1517,6 +1518,8 @@ spec:
                           type: boolean
                       type: object
                     type: array
+                  ip-intelligence:
```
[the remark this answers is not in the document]
ald8: This is correct, we should add `type: object` @sean-breen

examples/custom-resources/app-protect-waf/ap-dataguard-alarm-policy.yaml
```diff
@@ -30,5 +33,43 @@ spec:
       usSocialSecurityNumbers: true
     enforcementMode: blocking
     name: dataguard-alarm
+    ip-intelligence:
```
ald8: do we need the ip intelligence in that data guard alarm policy example?

sean-breen: I was unsure if adding it as a separate APPolicy would override any values in here so I just combined them. Would it be clearer to break it out into its own policy file?

internal/nginx/manager.go
sean-breen: `appProtectLogLevelDefault` seems to default to `fatal` while the iprepd config defaults to `Info`, so no they don't match up currently

deploy/crds-nap-waf.yaml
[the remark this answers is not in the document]
sean-breen: Added @ald8, please re-review when you can, thanks!

cmd/nginx-ingress/main.go
sean-breen: Added the flag `-enable-app-protect-ip-intelligence`, when true the `iprepd` daemon will launch alongside the ingress controller

examples/custom-resources/app-protect-waf/ap-dataguard-alarm-policy.yaml
ald8: Since the current policy name is “data-guard,” it would be more clear to create another policy example.

deploy/crds-nap-waf.yaml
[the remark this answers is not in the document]
ald8: this is good ("Failed to resolve thread")

examples/custom-resources/app-protect-waf/ap-dataguard-alarm-policy.yaml
sean-breen: Created another APPolicy for IPI in the same example folder @ald8

charts/nginx-ingress/templates/_helpers.tpl
```diff
@@ -529,8 +542,24 @@ volumeMounts:
...
+  securityContext:
+{{ toYaml .Values.controller.appprotect.ipIntelligence.securityContext | nindent 6 }}
+{{- end }}
+  volumeMounts:
+    - name: app-protect-ipi-db
+      mountPath: /var/IpRep
+    - name: app-protect-config
+      mountPath: /opt/app_protect/config
+    - name: app-protect-bundles
+      mountPath: /etc/app_protect/bundles
```
pdabelf5: why are we mounting the config and bundles directory into this container

internal/nginx/manager.go
```diff
@@ -101,6 +104,8 @@ type Manager interface {
 	AgentStart(agentDone chan error, logLevel string)
 	AgentQuit()
 	AgentVersion() string
+	IPRepdStart(ipRepdDone chan error)
+	IPRepdQuit()
```
pdabelf5, line 108: does IPI not work with the rollback manager?

charts/nginx-ingress/templates/_helpers.tpl
sean-breen: > We should only do this when `v5: true`, will update

I misunderstood, I didn't need the volume mounts here so they have been removed

internal/nginx/manager.go
sean-breen: LocalManager is embedded in `RollbackManager` so I expect it to work with both, admittedly I haven't tested it with config safety enabled

pdabelf5 (approved):

AlexFenlon (approved):

PR #9636, opened 2026-04-16
Fix acme challenge

### LINKED ISSUE #9600: [Bug]: ACME challenge fails when ingress uses Basic Auth
### Version

edge

Kind

### Steps to reproduce

When an Ingress is protected by Basic Auth (via nginx.org/basic-auth-secret), cert-manager's HTTP-01 ACME challenge to /.well-known/acme-challenge/... is rejected with 401 Unauthorized. LetsEncrypt cannot issue or renew a certificate on any host that requires Basic Auth.

Today there is no way to exempt specific paths from Basic Auth on an Ingress, so users have to temporarily remove auth, issue the cert, and re-enable it — which breaks automated renewals and is error-prone.

Expected behaviour

The ACME challenge path (/.well-known/acme-challenge/) should be reachable without credentials so LetsEncrypt can complete the HTTP-01 challenge, while the rest of the site remains behind Basic Auth.

Example Ingress (currently broken)

  ```yaml
  kind: Ingress
  metadata:
    name: test-basic-auth
    annotations:
      nginx.org/basic-auth-secret: "basic-auth"
      nginx.org/basic-auth-realm: "Test Realm"
  spec:
    ingressClassName: nginx
    tls:
      - hosts: [basic-auth.test.local]
        secretName: basic-auth-tls
    rules:
      - host: basic-auth.test.local
        http:
          paths:
            - path: /
              pathType: Prefix
              backend:
                service:
                  name: echo-backend
                  port:
                    number: 8080
```

---

Extra info:
- I run Nginx Ingress on my production GKE cluster, but I also reproduced the inability to get LetsEncrypt working correctly in a Kind cluster, hence I filled in Kind.
- For my production cluster I need to use `acme.cert-manager.io/http01-edit-in-place: "true"` because every new ingress gets a default ingress class of `gce` unless I build a template with annotations for the LetsEncrypt resolver. Not sure if relevant; but I cannot resolve the ACME challenge in a separate ingress (without Basic Auth).

### PR DESCRIPTION
**Pull request by danielnginx**
### Proposed changes

Fix https://github.com/nginx/kubernetes-ingress/issues/9600

### Checklist

### CODE REVIEW

File: internal/configs/version1/nginx-plus.ingress.tmpl [hunk 1]
```diff
@@ -252,6 +252,10 @@ server {
 		set $resource_name "{{$location.MinionIngress.Name}}";
 		set $resource_namespace "{{$location.MinionIngress.Namespace}}";
 		{{- end}}
+		{{- if and $server.BasicAuth (hasPrefix $location.Path "/.well-known/acme-challenge/") }}
+		# Disable basic auth for ACME HTTP-01 challenge paths
+		auth_basic off;
+		{{- end }}
```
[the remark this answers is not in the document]
javorszky: later in the same block there's this (lines 320 and 372)

```
		{{- with $location.BasicAuth }}
		auth_basic {{ printf "%q" .Realm }};
		auth_basic_user_file {{ .Secret }};
		{{- end }}
```
This is the section copilot is talking about. It's also true, if the `/.well-known/acme-challenge` has a `location.BasicAuth` set to true, then this `auth_basic off;` is meaningless. These two should be merged together.

internal/configs/version1/nginx.ingress.tmpl
```diff
@@ -166,6 +166,10 @@ server {
 		set $resource_name "{{$location.MinionIngress.Name}}";
 		set $resource_namespace "{{$location.MinionIngress.Namespace}}";
 		{{- end}}
+		{{- if and $server.BasicAuth (hasPrefix $location.Path "/.well-known/acme-challenge/") }}
+		# Disable basic auth for ACME HTTP-01 challenge paths
+		auth_basic off;
+		{{- end }}
```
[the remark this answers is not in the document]
javorszky: Same issue here on line 214 and 257 .

Also it needs to be after the location snippets (line 211), but both basic auth blocks are after that

internal/configs/version1/nginx-plus.ingress.tmpl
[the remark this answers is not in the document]
javorszky: And it needs to be after location snippets, which is 312, so both of the auth blocks will be after.

javorszky (changes requested):
copilot is right, later basic auth directives can override this one

internal/configs/version1/nginx.ingress.tmpl
```diff
@@ -258,6 +258,10 @@ server {
 		auth_basic {{ printf "%q" .Realm }};
 		auth_basic_user_file {{ .Secret }};
 		{{- end }}
+		{{- if and $server.BasicAuth (hasPrefix $location.Path "/.well-known/acme-challenge/") }}
```
javorszky: If you put this after line 305-ish, where the ` {{- if $location.GRPC}} ... {{- else}} ... {{- end}}` ends, you only need one block to negate either cases.

Right now you're adding the same block to both branches of that if statement, and they work the same way.

javorszky (changes requested):
These can be deduplicated

internal/configs/version1/nginx.ingress.tmpl
javorszky: The same logic is in both files

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -369,6 +369,11 @@ server {
 		auth_basic {{ printf "%q" .Realm }};
 		auth_basic_user_file {{ .Secret }};
 		{{- end }}
+		{{- $normalizedLocationPath := trimPrefix $location.Path "= " }}
+		{{- if and (or (eq $normalizedLocationPath "/.well-known/acme-challenge") (hasPrefix $normalizedLocationPath "/.well-known/acme-challenge/")) (or $server.BasicAuth $location.BasicAuth) }}
+		# Disable basic auth for ACME HTTP-01 challenge paths
+		auth_basic off;
```
pdabelf5, line 375: We should only ever have one directive for basic_auth, this one will cause a duplicate directive when `$location.BasicAuth` is set.

internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -369,6 +369,11 @@ server {
...
+		{{- end }}
```
pdabelf5, line 376: I'm not a big fan of having this complex logic in the template.  This should happen in `ingress.go` where the logic would be easier to understand

pdabelf5:
Ideally we should have a NoAuth policy for specific locations.  This way we could use it for Basic, APIKey & ExternalAuth

PR #9679, opened 2026-04-20
improve startup performance

hugolevino:
### Proposed changes

During startup, every ingress event calls rebuildHosts()
This PR introduces a startupComplete flag that skips rebuildHosts() during startup, deferring it to a single call at queue drain.

### Checklist


hugolevino:

PR #9728, opened 2026-04-27
Support empty host ingress

### LINKED ISSUE #209: Support for not setting explicit host
According to the K8S Ingress spec, I should be able to either set a rule with no **host**, or an ingress with no rules, and be able to have a default backend.

In the first case, the controller rejects the ingress because it has a rule with no host, and in the second case, it fails to create a server entry that has a proxy_pass to the default backend of the rule.

Is this a bug, or is this intentional? (if so, why?)

I'd like to see support for both of these cases as they are in the documentation: https://kubernetes.io/docs/concepts/services-networking/ingress/

If there is some reason why they cannot be supported organically, I'd like to see a way to override the server_name directive that gets generated so that I can have a server that listens to all hosts.

haywoodsh:
## What this adds

A new flag `-allow-empty-ingress-host` (default `false`) allows Ingress resources to omit the `host` field. When enabled, the Ingress with no host takes over the NGINX catch-all server — the same server that handles requests that do not match any named virtual host.

Without this flag, an Ingress with a host of "" is rejected. The flag must be explicitly set to opt in.

## How it works

NGINX allows only one `default_server` listener per port. NIC manages this through a single shared config file, conf.d/_default-server.conf:

- On startup, NIC writes a synthetic fallback default server to `_default-server.conf`. This synthetic server mimics the existing behavior by returning the configured `default-server-return` code, uses listener ports with `-default-http-listener-port`/ `-default-https-listener-port`, exposes the health-status endpoint with `-health-satus`, and terminates TLS with `-default-server-tls-secret`.
- When a hostless Ingress is admitted, NIC renders it through the ingress template and writes the output to `_default-server.conf`, replacing the synthetic fallback. The rendered config inherits the same default-server ports, TLS, and health-status settings.
- When the hostless Ingress is deleted, NIC restores the synthetic fallback.

Internally, NIC keeps the ingress map key as the normal namespace/name-derived identifier. Only the config file on disk changes.

## Config rollback behavior

When an Ingress transitions between empty-host and named-host configurations, NIC uses an ordered write strategy to prevent NGINX config conflicts:

**Transitioning an Ingress from empty-host to named-host (adding a host):**
1. Write the new named-host config to its dedicated file and validate with `nginx -t`
2. Only if validation succeeds, delete the old entry from `_default-server.conf` by restoring the synthetic fallback server
3. If validation fails at step 1, rollback discards the write and keeps the Ingress as the active default server

**Transitioning an Ingress from named-host to empty-host (removing a host):**
1. Write the updated Ingress config to `_default-server.conf` and validate with `nginx -t`
2. Only if validation succeeds, delete the old dedicated config file
3. If validation fails at step 1, rollback restores the previous config and the Ingress remains a named-host

This ordered approach ensures NGINX never has a state where both configs could exist simultaneously, avoiding duplicate `default_server` errors.

## Host Conflict resolution

If multiple Ingress resources without a host exist in the same namespace, NIC applies the existing winner-selection algorithm (oldest by creation timestamp). The losing Ingress is marked as invalid in its status.

For hostless Ingresses across namespaces without conflict, use mergeable Ingress with an empty-host master and path-contributing minions.

## Constraints

**Listener ports:** Hostless Ingresses cannot use the `nginx.org/listen-ports` or `nginx.org/listen-ports-ssl` annotations. Their listener ports are fixed by `-default-http-listener-port` and `-default-https-listener-port`.

**TLS:** `spec.tls[*].hosts` cannot contain an explicit empty string. TLS for the default catch-all server is configured via `-default-server-tls-secret`, not via the Ingress `spec.tls` field. TLS entries with no `hosts` slice are silently ignored. Mixed Ingresses (some rules with a named host, others without) are allowed; named-host TLS entries continue to work normally.

**Root location:** If the hostless Ingress defines a `/` path or has `spec.defaultBackend`, NIC uses that as the root location and does not inject the fallback `default-server-return` directive, avoiding a duplicate `location /` in the rendered config.

**Health-status redirect:** When `ssl-redirect` is enabled (either by ConfigMap or annotation) alongside the health-status endpoint, NIC exempts the health-status URI from the HTTP→HTTPS redirect so that liveness probes continue to work.

## What changed architecturally

- **Default server moved from `nginx.conf` to `conf.d/`:** The hardcoded `server { ... default_server ... }` block was removed from `nginx.tmpl` and `nginx-plus.tmpl`. The default server is now written to `conf.d/_default-server.conf` using the ingress template, which allows a hostless Ingress to replace it at runtime. Several `MainConfig` fields that fed the old hardcoded block are gone; their equivalents now live on the `Server` struct.
- **`map` directives at http scope:** The `set $resource_*` variables that were inside the removed server block are now `map` directives in the http block.
- **Bootstrap write:** On startup, the initial write of `_default-server.conf` using the ingress template bypasses `nginx -t` validation (nginx is not yet running). All subsequent writes are validated with rollback protection.
- **`@grpcerror*` locations:** These named locations are now emitted when a server has any gRPC-proxying location (`HasGRPCLocations`), not just when HTTP/2 is globally enabled. The gRPC-only tracking is also refactored from a pre-loop scan to per-location detection, which is more accurate.

## Test coverage

**Unit tests:**
- Validation: empty-host rule accepted/rejected based on flag, duplicate empty-host rule in one Ingress rejected, `tls.hosts` with `""` rejected, `listen-ports` annotation forbidden for hostless.
- Config generation: `IsDefaultServer` rendering, `default_server` in listen directives, root-location suppression of `DefaultServerReturn`, gRPC tracking and `HasGRPCLocations`.
- Configurator: `syncDefaultServerConfig` suppressed while active empty-host Ingress present, error propagation from sync on add/update/delete.
- Template snapshots regenerated.

**Integration tests (Python):**
- `test_empty_host_ingress.py`: basic routing, winner selection, coexistence with named-host Ingress.
- `test_empty_host_ingress_mergeable.py`: master+minion routing.
- `test_empty_host_ingress_reload.py`: lifecycle (create, update, delete), reload and config rollback on bad annotation.
- `test_default_server.py`: updated for the new default-server config structure.

**Helm:** `controller.allowEmptyIngressHost` value, schema, Helm unit test, and snapshot.

**Upgrade testing:**
- In-place upgrade from v5.4.2 to feat/empty-host-ingress validated on k3d: NIC controller reloaded successfully, NGINX configuration remained valid with all default server functionality working correctly.

## Examples

`examples/ingress-resources/empty-host/` and `examples/ingress-resources/empty-host-mergeable/` contain ready-to-run manifests and READMEs.

### Checklist


haywoodsh:
pre-commit.ci run

haywoodsh:
pre-commit.ci run

examples/ingress-resources/empty-host-mergeable/cafe-master.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: cafe-empty-host-master
+  annotations:
+    nginx.org/mergeable-ingress-type: "master"
+spec:
+  ingressClassName: nginx
+  rules:
+  - http: {}
```
AlexFenlon: not sure if this is an error you get but when testing this I get
```
The Ingress "cafe-empty-host-master" is invalid: spec.rules[0].http.paths: Required value 
```

even with the flag enabled, I was getting this.
I fixed it by making this change in the master cafe ingress

```suggestion
  - http: ""
```

AlexFenlon: this was updated. all good now.

AlexFenlon (dismissed):

AlexFenlon (approved):

vepatel (approved):

Issue #9509, opened 2026-03-31
[Bug]: Unguarded conflict between pathType: Exact and nginx.org/path-regex annotation.


javorszky:
### Version

edge

### What Kubernetes platforms are you running on?

Kind

### Steps to reproduce

When pathType is Exact, generateIngressPath() (internal/configs/ingress.go:574) prepends "= " to the path, producing "= /tea".

Then at template rendering time, makePathWithRegex() (internal/configs/version1/template_helper.go:56) sees the path-regex annotation and wraps the already-modified path, producing:

```
~ "^= /tea"
```

which is a malformed NGINX location directive.

There is no validation that detects or prevents this conflict.

See example ingress yaml:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: bug-pathtype-pathregex-conflict
  annotations:
    nginx.org/path-regex: "case_sensitive"
spec:
  ingressClassName: nginx
  rules:
  - host: example.com
    http:
      paths:
      - path: /tea
        pathType: Exact
        backend:
          service:
            name: tea-svc
            port:
              number: 80
```

elliott-logrocket:
@herwinz why did you close the PR fixing this? Would love to get this fixed as it's causing cert-manager HTTP01 challenges to fail for my use case when using [`acme.cert-manager.io/http01-edit-in-place: "true"`](https://cert-manager.io/docs/reference/annotations/#acmecert-manageriohttp01-edit-in-place) and `nginx.org/path-regex: case_sensitive` on the same Ingress since it's trying to match `location ~ "^= /.well-known/acme-challenge/...` which isn't even a valid URL path

RedVortex:
We seem to be impacted by this issue as well since we upgraded to 5.6.0.

We also use acme.cert-manager.io/http01-edit-in-place: "true" and not certificates are able to renew since the upgrade to 5.6.0

Claude seems to tell me this is a new bug introduced in 5.6.0 rather than this bug here but the end result seems the same, the location ends up in quotes and it shouldn't

Apprently this:

⏺ Yes — introduced in 5.6.0, and I can point at the commit and the code

```
  Commit 924ffb19f1e9 — "sync: security hardening and minor bug fixes for release-5.6 (#10775)", 2026-09-02, the day 5.6.0 shipped. It's a config-injection hardening change that added quoting to the location directive, and it swallowed the exact-match modifier.

[...]

PR #9766, opened 2026-05-03
Fix path-regex handling for Exact ingress paths

herwinz:
### Proposed changes

This PR fixes a conflict between `pathType: Exact` and the `nginx.org/path-regex` annotation (issue #9509).

When an Ingress path is exact, the controller generates a location path prefixed with `= ` (for example, `= /tea`).
Previously, regex modifiers from `nginx.org/path-regex` were still applied unconditionally, which could produce invalid/malformed location expressions like `~ "^= /tea"`.

Changes in this PR:
- add `applyPathRegex()` to avoid applying regex modifiers to already-exact location paths
- update `makeLocationPath()` to use this helper for both minion and non-minion ingress paths
- add regression tests to verify exact paths remain unchanged when `nginx.org/path-regex` is set (case_sensitive and case_insensitive)

### Checklist


herwinz:

herwinz:
recheck

PR #9861, opened 2026-05-14
Validate CORS origin port ranges

### LINKED ISSUE #9860: [Bug]: CORS policy allows out-of-range origin ports
### Current Behavior

CORS Policy validation accepts `allowOrigin` values with ports outside the valid TCP port range.

Example:

```yaml
spec:
  cors:
    allowOrigin:
    - https://example.com:99999
```

The CRD only checks that origins are non-empty strings, so this can be submitted and reaches policy validation.

### Expected Behavior

`allowOrigin` should reject ports outside `1-65535` for both exact and wildcard origins.

### Steps To Reproduce

Add a CORS validation test with these origins:

- `https://example.com:99999`
- `https://*.example.com:99999`

Before the fix, both cases pass validation unexpectedly.

### Environment

Current `main` branch.

immanuwell:
### Proposed changes

Fixes #9860

Tiny fix for CORS Policy validation. `allowOrigin` could accept ports outside `1-65535`, like `https://example.com:99999`, because the CRD only requires non-empty strings and the webhook check did not range-check the parsed port.

Repro, before this patch:

1. add a `TestValidateCORS` case with `https://example.com:99999` and `https://*.example.com:99999`
2. run `go test -tags=aws,helmunit ./pkg/apis/configuration/validation -run TestValidateCORS$`
3. both cases pass validation. not great lol

Now exact and wildcard origins share the same port-range check.

### Checklist

- [ ] I have checked that all unit tests pass after adding my changes (`make test` is blocked locally: `helm` is missing from PATH, and `internal/k8s/secrets/validation_test.go` references missing `embeds-ca.crt`)
- [x] I have updated necessary documentation (n/a, validation-only change)

### DIFF
--- a/pkg/apis/configuration/validation/policy.go
+++ b/pkg/apis/configuration/validation/policy.go
@@ -1085,6 +1085,10 @@ func validateOriginFormat(origin string) error {
 		return validateWildcardOriginHost(origin, host)
 	}
 
+	if err := validateOriginPort(parsedOrigin.Port(), origin); err != nil {
+		return err
+	}
+
 	return validateExactOriginHost(host)
 }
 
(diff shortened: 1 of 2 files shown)


immanuwell:

immanuwell:
recheck

pkg/apis/configuration/validation/policy.go
```diff
@@ -1146,6 +1149,22 @@ func validateWildcardOriginHost(origin, host string) error {
...
+func validateOriginPort(port, origin string) error {
+	if port == "" {
+		return nil
+	}
+
+	portNum, err := strconv.Atoi(port)
+	if err != nil {
+		return fmt.Errorf("port must be numeric (invalid: %s)", origin)
+	}
+	if portNum < 1 || portNum > 65535 {
```
javorszky, line 1161: you can reuse `validation.IsValidPortNum` here. It returns a non empty `[]string` if there was an issue which can be used in an error as a message.

The slightly different `validatePortNumber` function in the same file uses `validation.IsValidPortNum` if you need to see patterns. 🙂

javorszky (changes requested):
Tiny change, reuse a function we already use for similar validation elsewhere.

If you could add CRD validation for the option too, that would be excellent! Let me know if you need help with that.

pkg/apis/configuration/validation/policy.go
javorszky: I've created https://github.com/nginx/kubernetes-ingress/pull/10897 as a successor

AlexFenlon:
Closing in favour of https://github.com/nginx/kubernetes-ingress/pull/10897

PR #9875, opened 2026-05-15
Improve TransportServer, `nginx.org/limit-req-key` annotation, and zone size validation

AlexFenlon:
### Proposed changes

* Adds improved validation to the limit-req-key annotation . 
* Fixed typos in pkg/apis/configuration/validation/transportserver.go
* Added normalisation of bare numbers for limit req zone size following CRD docs (bare numbers assume 'k'). Bare numbers above and including 32768 will not get the 'k' suffix, they are already valid as is, and adding the suffix would mean increasing memory need thousand fold, which is not desirable. Numbers 1..31 already get caught with existing validation.

### Checklist

### DIFF
--- a/internal/configs/policy.go
+++ b/internal/configs/policy.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/url"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/nginx/kubernetes-ingress/internal/configs/version2"
@@ -1272,6 +1273,23 @@ func generateLimitReq(zoneName string, rateLimitPol *conf_v1.RateLimit) version2
 	return limitReq
 }
 
+// normalizeRateLimitZoneSize appends "k" to a bare number if its byte value
+// would be below NGINX's minimum shared memory zone size (32k).
+// Per the CRD API contract, bare numbers below 32k are treated as kilobytes.
+// Bare numbers >= 32768 are valid byte values and left unchanged.
+func normalizeRateLimitZoneSize(s string) string {
+	if len(s) == 0 {
+		return s
+	}
+	last := s[len(s)-1]
+	if last >= '0' && last <= '9' {
+		if n, err := strconv.Atoi(s); err == nil && n < 32768 {
+			return s + "k"
+		}
+	}
+	return s
+}
+
 func generateLimitReqZone(zoneName string, policy *conf_v1.Policy, podReplicas int, zoneSync bool) (version2.LimitReqZone, string) {
 	rateLimitPol := policy.Spec.RateLimit
 	rate := rateLimitPol.Rate
@@ -1287,7 +1305,7 @@ func generateLimitReqZone(zoneName string, policy *conf_v1.Policy, podReplicas i
 	return version2.LimitReqZone{
 		ZoneName: zoneName,
 		Key:      rateLimitPol.Key,
-		ZoneSize: rateLimitPol.ZoneSize,
+		ZoneSize: normalizeRateLimitZoneSize(rateLimitPol.ZoneSize),
 		Rate:     rate,
 		Sync:     zoneSync,
 	}, warningText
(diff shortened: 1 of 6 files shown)


internal/configs/policy.go
```diff
@@ -1272,6 +1273,23 @@ func generateLimitReq(zoneName string, rateLimitPol *conf_v1.RateLimit) version2
...
+// Per the CRD API contract, bare numbers below 32k are treated as kilobytes.
+// Bare numbers >= 32768 are valid byte values and left unchanged.
+func normalizeRateLimitZoneSize(s string) string {
+	if len(s) == 0 {
+		return s
+	}
+	last := s[len(s)-1]
+	if last >= '0' && last <= '9' {
+		if n, err := strconv.Atoi(s); err == nil && n < 32768 {
+			return s + "k"
```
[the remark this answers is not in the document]
javorszky: Numbers above and including 32768 are valid for nginx's "must be greater than or equal to 8 x pagesize (32k)"; adding a 'k' suffix would silently increase memory need leading to resource exhaustion. Bare numbers under 32768 led to an nginx crash at the moment, adding 'k' enables those to work.

internal/configs/policy.go
```diff
@@ -1287,7 +1305,7 @@ func generateLimitReqZone(zoneName string, policy *conf_v1.Policy, podReplicas i
 	return version2.LimitReqZone{
 		ZoneName: zoneName,
 		Key:      rateLimitPol.Key,
-		ZoneSize: rateLimitPol.ZoneSize,
+		ZoneSize: normalizeRateLimitZoneSize(rateLimitPol.ZoneSize),
```
pdabelf5, line 1308: This seems unrelated to the annotation

internal/configs/policy.go
```diff
@@ -1316,7 +1334,7 @@ func generateGroupedLimitReqZone(
 	lrz := version2.LimitReqZone{
 		ZoneName: zoneName,
 		Key:      rateLimitPol.Key,
-		ZoneSize: rateLimitPol.ZoneSize,
+		ZoneSize: normalizeRateLimitZoneSize(rateLimitPol.ZoneSize),
```
pdabelf5, line 1337: This seems unrelated to the annotation

AlexFenlon: change required, title updated.

internal/configs/policy.go
AlexFenlon: change required, title updated.

internal/configs/policy.go
```diff
@@ -1272,6 +1273,23 @@ func generateLimitReq(zoneName string, rateLimitPol *conf_v1.RateLimit) version2
 	return limitReq
 }
 
+// normalizeRateLimitZoneSize appends "k" to a bare number if its byte value
+// would be below NGINX's minimum shared memory zone size (32k).
+// Per the CRD API contract, bare numbers below 32k are treated as kilobytes.
+// Bare numbers >= 32768 are valid byte values and left unchanged.
+func normalizeRateLimitZoneSize(s string) string {
```
pdabelf5, line 1280: should this be behind the `-enable-directive-autoadjust` flag?

internal/configs/policy_test.go
```diff
@@ -4854,3 +4854,32 @@ func TestAddWafConfig(t *testing.T) {
...
+		{input: "", expected: ""},
+		{input: "10m", expected: "10m"},
+		{input: "10M", expected: "10M"},
+		{input: "32k", expected: "32k"},
+		{input: "32K", expected: "32K"},
+		{input: "32", expected: "32k"},
+		{input: "100", expected: "100k"},
+		{input: "31", expected: "31k"},
+		{input: "0", expected: "0k"},
+		{input: "32767", expected: "32767k"},
```
pdabelf5, line 4873: if you put in 32767 bytes (32k), you should expect 32Mb to go into the config?

PR #9876, opened 2026-05-15
Improve TransportServer, `nginx.org/limit-req-key` annotation, and zone size validation

AlexFenlon:
### Proposed changes

* Adds improved validation to the limit-req-key annotation . 
* Fixed typos in pkg/apis/configuration/validation/transportserver.go
* Added normalisation of bare numbers for limit req zone size following CRD docs (bare numbers assume 'k'). Bare numbers above and including 32768 will not get the 'k' suffix, they are already valid as is, and adding the suffix would mean increasing memory need thousand fold, which is not desirable. Numbers 1..31 already get caught with existing validation.

### Checklist

### DIFF
--- a/pkg/apis/configuration/validation/transportserver.go
+++ b/pkg/apis/configuration/validation/transportserver.go
@@ -287,7 +287,7 @@ func validateHealthCheckMatch(match *conf_v1.TransportServerMatch, fieldPath *fi
 	}
 
 	allErrs := validateMatchExpect(match.Expect, fieldPath.Child("expect"))
-	allErrs = append(allErrs, validateMatchSend(match.Expect, fieldPath.Child("send"))...)
+	allErrs = append(allErrs, validateMatchSend(match.Send, fieldPath.Child("send"))...)
 	return allErrs
 }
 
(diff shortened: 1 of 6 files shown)


pkg/apis/configuration/validation/transportserver.go
```diff
@@ -180,7 +180,7 @@ func validateTransportServerUpstreams(upstreams []conf_v1.TransportServerUpstrea
 
 		allErrs = append(allErrs, validateServiceName(u.Service, idxPath.Child("service"))...)
 		allErrs = append(allErrs, validatePositiveIntOrZeroFromPointer(u.MaxFails, idxPath.Child("maxFails"))...)
-		allErrs = append(allErrs, validatePositiveIntOrZeroFromPointer(u.MaxFails, idxPath.Child("maxConns"))...)
+		allErrs = append(allErrs, validatePositiveIntOrZeroFromPointer(u.MaxConns, idxPath.Child("maxConns"))...)
```
sean-breen, line 183: Is this change related to the annotation validation problem?

AlexFenlon: different issue, still has to go in, not releated to the annotation, will update PR titles.

PR #9958, opened 2026-05-21
Update workflow trigger to push on release-5.4 branch

pdabelf5:
### Proposed changes

Trigger dependency image build for release-5.4

### Checklist

### DIFF
--- a/.github/workflows/build-ubi-dependency.yml
+++ b/.github/workflows/build-ubi-dependency.yml
@@ -3,11 +3,10 @@ name: Build UBI c-ares Dependency
 on:
   push:
     branches:
-      - main
+      - release-5.4
     paths:
       - build/dependencies/Dockerfile.ubi8
       - build/dependencies/Dockerfile.ubi9
-      - .github/workflows/build-ubi-dependency.yml
   workflow_dispatch:
 
 env:


AlexFenlon (approved):

.github/workflows/build-ubi-dependency.yml
```diff
@@ -3,11 +3,10 @@ name: Build UBI c-ares Dependency
 on:
   push:
     branches:
-      - main
+      - release-5.4
```
javorszky, line 6: does this mean this will **only** trigger for the release-5.4 branch, but not main? (do we need this to trigger on pushes to main?)

javorszky: answered elsewhere:
* main has ubi 10, so this is not needed
* release has ubi 8 and 9, so this is needed
* lts does not have ubi, so this is not needed

javorszky (approved):

.github/workflows/build-ubi-dependency.yml
pdabelf5: only `release-5.4` needs ubi8/9, `main` is moving to ubi10

sean-breen (approved):

PR #10016, opened 2026-05-26
forwarded headers support for reverse proxy deployments

### LINKED ISSUE #10015: Preserve original host in multi-origin / DR environments when X-Forwarded-Host is unavailable
Yes.

In DR and multi-origin Kubernetes environments, requests are routed through multiple proxy layers such as Azure Front Door, F5, CDN, or reverse proxies before reaching the NGINX Ingress Controller.

In our setup:

Client
→ public-domain.com
→ Front Door / F5
→ secondary-origin.internal.com
→ NGINX Ingress
→ backend application

The ingress controller currently determines the upstream Host header primarily using:

$host
or X-Forwarded-Host

When X-Forwarded-Host is absent, stripped, or overwritten by upstream infrastructure, NGINX falls back to the internal ingress/origin hostname instead of the original public hostname.

Example:

Expected:
Host: public-domain.com

Actual:
Host: secondary-origin.internal.com

This causes backend applications to fail when they strictly validate the original host/domain.

Observed impact:

OAuth/OpenID authentication failures
403 Forbidden responses
incorrect redirect generation
cookie domain mismatch
multi-tenant routing issues

Proposed enhancement to the ingress template:

Current implementation:

proxy_set_header Host $host;
proxy_set_header X-Forwarded-Host $host;

Proposed implementation:

set $best_http_host $host;

if ($http_x_forwarded_host != "") {
    set $best_http_host $http_x_forwarded_host;
}

proxy_set_header Host $best_http_host;
proxy_set_header X-Forwarded-Host $best_http_host;

This allows the ingress controller to preserve the original client-facing hostname when X-Forwarded-Host is provided by upstream infrastructure while still falling back safely to $host.

This would improve interoperability for:

DR architectures
multi-origin deployments
Front Door / CDN / WAF environments
reverse proxy chains

We attempted:

custom nginx snippets
manual proxy_set_header overrides
ingress annotations
static Host header configuration
forwarded header configurations

However, these approaches are difficult to maintain consistently across DR failover environments and multiple ingress origins.

However, these approaches become difficult to maintain in DR environments and can introduce inconsistencies between primary and secondary origins depending on upstream proxy behavior.

**Additional context**

Additional context

Current template section:

proxy_set_header Host $host;
proxy_set_header X-Forwarded-Host $host;

Proposed logic improves host preservation behavior by preferring X-Forwarded-Host when available.

This issue is reproducible when:

ingress host differs from public domain
multiple proxy layers are involved
backend applications strictly validate host headers
DR failover routes traffic to alternate origins

Would appreciate feedback on whether this behavior can be considered for enhancement in future releases.

Mohammed-suhail-devops:
Fixes #10015

### Use Case
When the NGINX Ingress Controller is deployed behind reverse proxies (Azure Application Gateway, AWS ALB, etc.), the original Host header is modified to an internal service name, but the original hostname is preserved in the `X-Forwarded-Host` header. Currently, the controller always uses `$host` which contains the modified hostname, causing backend applications to reject requests (403/404) or generate incorrect URLs.

### Changes
This PR adds logic to prioritize the `X-Forwarded-Host` header (when present) over the standard `$host` variable, with a safe fallback.

**Implementation:**
- Introduces `$best_http_host` variable with conditional logic
- Uses `X-Forwarded-Host` when present (original hostname from proxy)
- Falls back to `$host` when `X-Forwarded-Host` is absent or empty
- Modified both `nginx.ingress.tmpl` (OSS) and `nginx-plus.ingress.tmpl` (Plus)

**NGINX Configuration Added:**
```nginx
set $forwarded_host $host;
if ($http_x_forwarded_host != "") {
    set $forwarded_host $http_x_forwarded_host;
}
proxy_set_header Host $forwarded_host;
proxy_set_header X-Forwarded-Host $forwarded_host;

### CODE REVIEW

**Comment by suhail-blueocean** [comment]
Hi @spencerugbo, could you please check this pr we need this change as soon as its blocked the whole migration of us.

**Comment by suhail-blueocean** [comment]
Hi @javorszky @vepatel, please look into this pr we need this change

**Comment by AlexFenlon** [comment]
Hi @suhail-blueocean, we plan on addressing this hopefully later in this year (see https://github.com/nginx/kubernetes-ingress/issues/7555) 

We do like what is here but there is still a lot of discussion internally about related things. We will credit you if we end up taking in some/all of your PR

Thanks

**Comment by vepatel** [comment]
hey @Mohammed-suhail-devops https://github.com/nginx/kubernetes-ingress/pull/10539 might supersede this PR as its covers VirtualServer as well

**Comment by haywoodsh** [comment]
Hi @Mohammed-suhail-devops, thanks for explaining the use case.

We have renamed the ConfigMap key in #10539 to `disable-forwarded-headers` and close #7555. With this you should be able to achieve some of the logic with snippets, without the issue of repeated `proxy_set_header` directives.

That said, you would still not be able to set the `Host` header to another variable without a custom template.

I have reopened #10015 to track this specific gap, and we aim to deliver a complete solution for this with the milestone set to 5.7.0.

On this PR specifically, rather than changing the native behaviour, I think it would be better to have an opt-in ConfigMap key `use-forwarded-headers`, which would also help users migrating from ingress-nginx.

Because of the ongoing work in #10557 and #10539 scheduled for 5.6.0, the template could look like this. It's a rough draft and needs testing:

```
{{- if $location.UseForwardedHeaders}}
set $forwarded_host $host;
if ($http_x_forwarded_host != "") { 
    set $forwarded_host $http_x_forwarded_host; (will need to take first element instead of whole list)
}
set $forwarded_port $server_port;
if ($http_x_forwarded_port != "") {
    set $forwarded_port $http_x_forwarded_port;
}
set $forwarded_proto {{if $server.RedirectToHTTPS}}https{{else}}$scheme{{end}};
if ($http_x_forwarded_proto != "") {
    set $forwarded_proto $http_x_forwarded_proto;
}
{{- else}}
set $forwarded_host $host;
set $forwarded_port $server_port;
set $forwarded_proto {{if $server.RedirectToHTTPS}}https{{else}}$scheme{{end}};
{{- end}}
{{- if $location.UpstreamVhost}}
proxy_set_header Host {{$location.UpstreamVhost}};
{{- else}}
proxy_set_header Host $forwarded_host;
{{- end}}
{{- if not $location.DisableForwardedHeaders}}
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header X-Forwarded-Host $forwarded_host;
proxy_set_header X-Forwarded-Port $forwarded_port;
proxy_set_header X-Forwarded-Proto $forwarded_proto;
{{- end}}
```

Would something like this fix your use case? Let me know what you think. I'll track the details in #10015.

Thanks again for raising this. The use case is valid and we want to support it natively.

PR #10082, opened 2026-06-02
Add nodePort for LoadBalancer only if allocateLoadBalancerNodePorts

Chupaka:
### Proposed changes

In v3.7.0 the chart started to set nodePort field for LoadBalancer service type in addition to NodePort type. But if we disable allocateLoadBalancerNodePorts, we don't need to set nodePorts.

Also, allocateLoadBalancerNodePorts is true by default (in k8s), not false.

### Checklist

### DIFF
--- a/charts/nginx-ingress/values.yaml
+++ b/charts/nginx-ingress/values.yaml
@@ -518,7 +518,7 @@ controller:
     loadBalancerSourceRanges: []
 
     ## Whether to automatically allocate NodePorts (only for LoadBalancers).
-    # allocateLoadBalancerNodePorts: false
+    # allocateLoadBalancerNodePorts: true
 
     ## Dual stack preference.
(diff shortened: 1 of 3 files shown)


Chupaka:

charts/nginx-ingress/values.yaml
```diff
@@ -518,7 +518,7 @@ controller:
     loadBalancerSourceRanges: []
 
     ## Whether to automatically allocate NodePorts (only for LoadBalancers).
-    # allocateLoadBalancerNodePorts: false
+    # allocateLoadBalancerNodePorts: true
```
haywoodsh, line 521: I understand that the Kubernetes spec may have different values, but changing the default value of this Helm chart could result in a breaking change for existing users upgrading.

PR #10111, opened 2026-06-04
Add HSTS Policy to VS

### LINKED ISSUE #9720: VirtualServer/VirtualServerRoute - HSTS Policy
## Overview

The NGINX Ingress Controller's **VirtualServer (VS)** and **VirtualServerRoute (VSR)** resources do not support direct HSTS configuration through ConfigMap keys. The following HSTS-related settings are **not available** for VS/VSR resources:

- `hsts`
- `hsts-max-age`
- `hsts-include-subdomains`
- `hsts-behind-proxy`
## Problem

When using VirtualServer and VirtualServerRoute CRDs, operators cannot configure HSTS headers via the standard ConfigMap interface that is available for traditional Ingress resources. This creates a gap between the global configuration capabilities and what can be implemented at the virtual server level.

## Current Workaround

HSTS must be configured using **snippets** in the VirtualServer resource to inject raw NGINX configuration. VirtualServerRoute snippet support for HSTS may be limited or unavailable depending on the controller version and configuration.

# Solution / Definition of Done

A dedicated **HSTS Policy CRD** would provide a native, Kubernetes-native way to define and manage HSTS settings for VirtualServer and VirtualServerRoute resources. This CRD would allow operators to declare HSTS requirements declaratively through YAML manifests, similar to how other security policies are managed in Kubernetes. The HSTS Policy CRD could specify parameters such as `max-age`, `includeSubDomains`, and `preload` flags, and be referenced or attached to VS/VSR/Ingress resources via selectors, labels, or direct reference fields. This approach eliminates the need for snippets, improves security by avoiding raw configuration injection, and brings HSTS configuration in line with the broader Kubernetes API design patterns for managing policy-based settings across resources.

spencerugbo:
### Proposed changes

Introduces support for configuring HTTP Strict Transport Security in NGINX through the Policy custom resource. It follows the design seen in the Configmap and allows the user to set the `maxAge`, `includeSubDomains` and `behindProxy` fields. HSTS is then set in the server block of the NGINX config but if an `add_header` directive is used in any of the location blocks, HSTS is also set in that block to avoid being ignored. The following is the policy design used for HSTS:
```
apiVersion: k8s.nginx.org/v1
kind: Policy
metadata:
  name: hsts-policy
spec:
  hsts:
    # sets how long in seconds the browser should cache and enforce the HSTS policy
    maxAge: 2592000
    # extends the policy to all subdomains of the host
    includeSubDomains: true
    # sets the HSTS headers based on the X-Forwarded-Proto header
    behindProxy: true
    # when preload is true, the preload directive id added to the header allowing the site to be added to the browser HSTS preload lists
    preload: false
```

This PR also includes a new example, `examples/custom-resources/hsts/README.md`, to show how to setup the policy.

docs: https://github.com/nginx/documentation/pull/2032

### Checklist


examples/ingress-resources/ingress-mtls/cafe-ingress.yaml
```diff
@@ -2,8 +2,6 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: cafe-ingress
```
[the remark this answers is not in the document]
spencerugbo: Unintentional change. Change has been reverted

examples/custom-resources/hsts/virtual-server.yaml
```diff
@@ -0,0 +1,23 @@
...
+      port: 80
+  routes:
+    - path: /
+      action:
+        pass: webapp
+    - path: /test
+      action:
+        pass: webapp
+      policies:
+        - name: cors-policy
```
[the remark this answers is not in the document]
spencerugbo: Forgot to change back after testing the policy with CORS. Has been changed back to normal.

internal/configs/policy.go
```diff
@@ -1132,6 +1133,32 @@ func (p *policiesCfg) addCORSConfig(
...
 }
 
+func (p *policiesCfg) addHSTSConfig(
+	hsts *conf_v1.HSTS,
+	polKey string,
+) *validationResults {
+	res := newValidationResults()
+
+	if p.HSTS != nil {
+		res.addWarningf("Multiple HSTS policies in the same context is not valid. HSTS policy %s will be ignored", polKey)
```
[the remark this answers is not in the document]
spencerugbo: Fixed mistake

pkg/apis/configuration/v1/types.go
```diff
@@ -1324,3 +1326,18 @@ type ExternalAuth struct {
 	// SNIName sets the server name used for SNI and certificate verification when connecting to the external authentication server over TLS. If not specified, defaults to <service-name>.<namespace>.svc derived from authServiceName.
 	SNIName string `json:"sniName,omitempty"`
 }
+
+// HSTS defines a HTTP Strict Transport Security policy for enforcing secure connections to the server.
+type HSTS struct {
+	// +kubebuilder:validation:Required
+	// MaxAge defines how long (in seconds) the browser should cache and enforce the HSTS policy.
+	MaxAge *int `json:"maxAge"`
```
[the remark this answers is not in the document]
spencerugbo: Added in this validation

examples/custom-resources/hsts/hsts-policy.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: hsts-policy
+spec:
+  hsts:
+    # sets how long in seconds the browser should cache and enforce the HSTS policy
+    maxAge: 2592000
+    # extends the policy to all subdomains of the host
+    includeSubDomains: true
+    # sets the HSTS headers based on the X-Forwarded-Proto header
+    behindProxy: false
```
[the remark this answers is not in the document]
spencerugbo: Made comment more descriptive

pkg/apis/configuration/v1/types.go
```diff
@@ -1324,3 +1326,18 @@ type ExternalAuth struct {
 	// SNIName sets the server name used for SNI and certificate verification when connecting to the external authentication server over TLS. If not specified, defaults to <service-name>.<namespace>.svc derived from authServiceName.
 	SNIName string `json:"sniName,omitempty"`
 }
+
+// HSTS defines a HTTP Strict Transport Security policy for enforcing secure connections to the server.
```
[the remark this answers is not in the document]
spencerugbo: Fixed mistake

internal/configs/version2/nginx.virtualserver.tmpl
```diff
@@ -561,6 +575,19 @@ server {
         }
         {{- end }}
 
+        {{- if $l.HSTS}}
```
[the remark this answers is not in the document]
spencerugbo: Decided to move the TLS check to `addHSTSConfig` instead of in the templates and added an error message is that case isn't met.

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -799,6 +813,19 @@ server {
         }
         {{- end }}
 
+        {{- if $l.HSTS}}
```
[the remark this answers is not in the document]
spencerugbo: Same as above

deploy/crds.yaml
```diff
@@ -709,6 +709,29 @@ spec:
...
+                    type: boolean
+                  includeSubDomains:
+                    default: false
+                    description: IncludeSubDomains extends the HSTS policy to all
+                      subdomains of the host.
+                    type: boolean
+                  maxAge:
+                    description: MaxAge defines how long (in seconds) the browser
+                      should cache and enforce the HSTS policy.
+                    minimum: 0
```
vepatel, line 730: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security#expiration
min of 0 disables the hsts, mabe we should enforce positive integers

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -799,6 +813,19 @@ server {
...
+        set $hsts_header_val "";
+        proxy_hide_header Strict-Transport-Security;
+        {{- if $l.HSTS.BehindProxy}}
+        if ($http_x_forwarded_proto = 'https') {
+        {{- else}}
+        if ($https = on) {
+        {{- end}}
+            set $hsts_header_val "max-age={{$l.HSTS.MaxAge}}; {{if $l.HSTS.IncludeSubDomains}}includeSubDomains; {{end}}preload";
```
AlexFenlon: This differs to the server block, in this case, the preload here is always in this string no matter what. There does not seem to be a toggle for preload in location block, is that intentional?

internal/configs/version2/nginx.virtualserver.tmpl
```diff
@@ -561,6 +575,19 @@ server {
...
+        set $hsts_header_val "";
+        proxy_hide_header Strict-Transport-Security;
+        {{- if $l.HSTS.BehindProxy}}
+        if ($http_x_forwarded_proto = 'https') {
+        {{- else}}
+        if ($https = on) {
+        {{- end}}
+            set $hsts_header_val "max-age={{$l.HSTS.MaxAge}}; {{if $l.HSTS.IncludeSubDomains}}includeSubDomains; {{end}}preload";
```
AlexFenlon: same thing here as in the plus template

internal/configs/version2/nginx-plus.virtualserver.tmpl
```diff
@@ -372,6 +372,20 @@ server {
...
 
+    {{- if $s.HSTS }}
+    set $hsts_header_val "";
+    proxy_hide_header Strict-Transport-Security;
+    {{- if $s.HSTS.BehindProxy}}
+    if ($http_x_forwarded_proto = 'https') {
+    {{- else}}
+    if ($https = on) {
+    {{- end}}
+        set $hsts_header_val "max-age={{$s.HSTS.MaxAge}}; {{if $s.HSTS.IncludeSubDomains}}includeSubDomains{{end}}{{if $s.HSTS.Preload}}; preload{{end}}";
```
AlexFenlon: When IncludeSubDomains=false and Preload=false , this produces "max-age=2592000;  "

not sure if the trailing spce will cause an issue.

internal/configs/version2/nginx-plus.virtualserver.tmpl
AlexFenlon: same trailing space issue in server block

internal/configs/version2/nginx.virtualserver.tmpl
```diff
@@ -242,6 +242,20 @@ server {
...
 
+    {{- if $s.HSTS }}
+    set $hsts_header_val "";
+    proxy_hide_header Strict-Transport-Security;
+    {{- if $s.HSTS.BehindProxy}}
+    if ($http_x_forwarded_proto = 'https') {
+    {{- else}}
+    if ($https = on) {
+    {{- end}}
+        set $hsts_header_val "max-age={{$s.HSTS.MaxAge}}; {{if $s.HSTS.IncludeSubDomains}}includeSubDomains{{end}}{{if $s.HSTS.Preload}}; preload{{end}}";
```
AlexFenlon: same issue mentioned in plus for here and locaiton block

tests/suite/test_hsts_policy_vs.py
```diff
@@ -0,0 +1,249 @@
...
+            allow_redirects=False,
+        )
+        print(f"Response headers: {resp.headers}")
+
+        teardown_policy(kube_apis, test_namespace, tls_secret_name, pol_name)
+        delete_and_create_vs_from_yaml(
+            kube_apis.custom_objects, virtual_server_setup.vs_name, std_vs_src, test_namespace
+        )
+
+        assert "Strict-Transport-Security" not in resp.headers
```
AlexFenlon, line 131: I think we should also add an assertion to check the reponse code here eg.


```suggestion
        assert "Strict-Transport-Security" not in resp.headers
        assert resp.status_code == 200
```

test this locally before pushing but might just been good to have this assertion

examples/custom-resources/hsts/README.md
```diff
@@ -0,0 +1,117 @@
...
+Send a plain HTTP request to confirm the header is not sent over unencrypted connections:
+
+```console
+curl --resolve webapp.example.com:$IC_HTTP_PORT:$IC_IP \
+   http://webapp.example.com:$IC_HTTP_PORT/ -v
+```
+
+The `Strict-Transport-Security` header should be absent from the response.
+
+## Step 5 - Removing the HSTS Policy
```
AlexFenlon: Maybe have this as a note instead of a step.

internal/configs/version2/nginx-plus.virtualserver.tmpl
spencerugbo: No, not intentional I must have missed it so I'll fix that. The issue with the extra trailing spaces seems to be coming from the ErrorPageLocation and ReturnLocation blocks just below and I'm unsure if that was done intentionally. The closing tag does not contain tag (eg `{{ - end }}` ) so the whitespace isn't being stripped.

internal/configs/version2/nginx.virtualserver.tmpl
spencerugbo: That is now fixed

internal/configs/version2/nginx-plus.virtualserver.tmpl
spencerugbo: A small problem with the format there, it has been fixed now so the output should be as expected

examples/custom-resources/hsts/README.md
spencerugbo: Sounds good, this step is now turned into a note block

tests/suite/test_hsts_policy_vs.py
spencerugbo: Looked good when tested locally so I've added it into the PR

internal/configs/version2/nginx.virtualserver.tmpl
AlexFenlon: unrelased to current change

deploy/crds.yaml
spencerugbo: We might need to allow an input of 0 in order to safely disable to policy. When the policy is set the browser enforces the policy for the amount of time inputted for maxAge. If the policy is then deleted, the browser will continue to enforce the policy for that stated amount of time and will refuse to connect over HTTP entirely. I have added a note for it in the README and it'll be mentioned in the docs too.

AlexFenlon (dismissed):

pkg/apis/configuration/validation/policy.go
```diff
@@ -1559,6 +1566,34 @@ func validateExternalAuthSSLFields(externalAuth *v1.ExternalAuth, fieldPath *fie
...
+
+	if hsts.Preload {
+		if !hsts.IncludeSubDomains {
+			allErrs = append(allErrs, field.Invalid(
+				fieldPath.Child("preload"), hsts.Preload,
+				"preload requires includeSubDomains to be enabled",
+			))
+		}
+
+		if *hsts.MaxAge < 31536000 {
```
sean-breen: Magic number: could be a const

sean-breen:
Seems we are missing the `hsts` option from the error message in `pkg/apis/configuration/validation/policy.go` at `validatePolicySpec:176`, please add that
```
	if fieldCount != 1 {
		msg := "must specify exactly one of: `accessControl`, `rateLimit`, `ingressMTLS`, `egressMTLS`, `basicAuth`, `apiKey`, `cache`, `cors`, `externalAuth`"
		if cfg.IsPlus {
			msg = fmt.Sprint(msg, ", `jwt`, `oidc`, `waf`")
		}
		allErrs = append(allErrs, field.Invalid(fieldPath, "", msg))
	}
```

pkg/apis/configuration/validation/policy.go
spencerugbo: Turned it into a local variable for more clarity but I don't think it needs to be a const

sean-breen: If it's not going to change during runtime, it should be a const, and it should have a useful name like `hstsPolicyMaxAge`

spencerugbo: Sounds good, I have it set now as a const called `minAgeValueForPreload`

sean-breen (approved):

AlexFenlon (approved):

PR #10137, opened 2026-06-09
fix Unguarded conflict between pathType: Exact and nginx.org/path-regex annotation

elliott-logrocket:
### Proposed changes

- Copied almost entirely from a closed PR (https://github.com/nginx/kubernetes-ingress/pull/9766) 
- Otherwise, if path-regex is set (case_sensitive or case_insensitive) and there is a Ingress path with PathType exact then the generated location block is invalid 
- Previously generated a block like this `location ~ "\^= <url path>" {}` which cannot match any URL (it's matching the literal path `= <url path>` and valid URL paths cannot contain ` ` or `=` characters) 
- If the Ingress rule has PathType = Exact then we should just ignore the regex directive on the parent Ingress and match the path literally

### Checklist


elliott-logrocket:

elliott-logrocket:
fixes #9509

vepatel:
hey @elliott-logrocket, thanks for this we'll be adding this to one of our upcoming releases

PR #10176, opened 2026-06-12
fix illegal keyword path validation rejecting paths containing /var or /root as substring

### LINKED ISSUE #10175: [Bug]: Ingress paths containing /var or /root as substring (e.g. /variables) are rejected by path validation
### Version

5.5.0

EKS Amazon

### Steps to reproduce

1. Create an Ingress with a path that contains `/var` or `/root` as a substring, for example:
 
```yaml
kind: Ingress
metadata:
  name: test-variables
spec:
  ingressClassName: nginx
  rules:
  - host: test.example.com
    http:
      paths:
      - path: /variables
        backend:
          service:
            name: my-service
            port:
              number: 8000
```
 
2. The Ingress is rejected with this event:
 
```
Warning  Rejected  nginx-ingress-controller
spec.rules[0].http.path[0].path: Invalid value: "/variables": must not contain invalid paths
```
 
3. No location is rendered in nginx.conf for this path.
 
---
 
### Why this happens
 
The `illegalKeywordFmt` constant in [internal/k8s/validation.go](https://github.com/nginx/kubernetes-ingress/blob/main/internal/k8s/validation.go#L1401-L1406) is:
 
```go
illegalKeywordFmt = `/etc/|/root|/var|\\n|\\r`
```
 
and it gets compiled as `"^" + illegalKeywordFmt + "$"`. Because `|` has the lowest precedence in regex, the anchors only attach to the first and last alternatives. The effective pattern is:
 
```
^/etc/  |  /root  |  /var  |  \\n  |  \\r$
```
 
So `/root` and `/var` are unanchored and match *anywhere* in the path. Any path that merely contains these substrings is rejected: `/variables`, `/var-config`, `/rootpage`, `/manager/variables`, `/api/variables/list`, etc.
 
I understand the original intent (PR #3094, related to the path injection hardening around CVE-2022-30535) was to block paths pointing at sensitive directories like `/etc/passwd` or `/var/run/secrets` that goal makes sense as a defense-in-depth layer.
 
Note that a path like `/var/run/secrets` is not exploitable by itself. It just becomes
`location /var/run/secrets { proxy_pass ...; }`, a plain URL route that never touches the
filesystem. A real attack would need config injection with `{`, `}`, `;` or whitespace,
and those are already blocked by `pathFmt` and [validateCurlyBraces](https://github.com/nginx/kubernetes-ingress/blob/main/internal/k8s/validation.go#L1373-L1383).
 
So this check is defense-in-depth, and fixing the regex precedence keeps it working as
intended, just without the false positives on paths like `/variables`.
 
A few extra observations:
 
- The same check does not exist for VirtualServer/VirtualServerRoute, so `/variables` is accepted as a VS route but rejected as an Ingress path, even though both render the same way into nginx.conf.
- A user already reported this in a [comment on PR #3094](https://github.com/nginx/kubernetes-ingress/pull/3094#issuecomment-2431590174) back in Oct 2024, but it never became an issue.
- There is no workaround via ConfigMap, annotation or flag. The only options today are renaming application paths or migrating to VirtualServer.
 
### Suggested fix
 
Anchor the keywords to the first path segment:
 
```go
illegalKeywordFmt = `^(?:/etc(?:/|$)|/root(?:/|$)|/var(?:/|$))|\\n|\\r`
```
 
This still blocks everything the original test cases block (`/root`, `/etc/passwd`, `/var/run/secrets`, ...) but stops rejecting legitimate paths like `/variables`.
 
I have a fix with regression tests ready and can open a PR.

hugolevino:
### Proposed changes

Fixes #10175

The `illegalKeywordFmt` regex in [internal/k8s/validation.go](https://github.com/nginx/kubernetes-ingress/blob/main/internal/k8s/validation.go#L1401-L1406) has an alternation precedence problem. The constant

    `/etc/|/root|/var|\\n|\\r`

is compiled as `"^" + fmt + "$"`, but since `|` has the lowest precedence in regular expressions, the `^` only anchors `/etc/` and the `$` only anchors `\\r`. This causes `/root` and `/var` to match as unanchored substrings, rejecting valid Ingress paths like `/variables`, `/var-config` or `/rootpage` with:

    spec.rules[0].http.path[0].path: Invalid value: "/variables": must not contain invalid paths

This change replaces the regex with a self-contained expression that embeds its own anchors, removing the `"^" + fmt + "$"` wrapping:

    `^/etc/|^/root(?:/|$)|^/var(?:/|$)|\\n|\\r$`
 
Each directory keyword is now anchored to match only as a leading path segment, while `\n` and `\r` remain matched anywhere / at end-of-string respectively — preserving existing behaviour.

### Behaviour before / after:

| Path | Before | After |
|------|--------|-------|
| `/root`, `/root/.ssh` | blocked ❌ | blocked ❌ |
| `/etc/passwd`, `/etc/nginx/secrets` | blocked ❌ | blocked ❌ |
| `/var`, `/var/run/secrets` | blocked ❌ | blocked ❌ |
| `/variables`, `/var-config` | blocked ❌ | allowed ✅ |
| `/manager/variables`, `/api/variables/list` | blocked ❌ | allowed ✅ |
| `/rootpage`, `/root-cause`, `/tree/root` | blocked ❌ | allowed ✅ |
| `/etcetera`, `/etcd` | allowed ✅ | allowed ✅ |

Notes:

- All test cases from the original PR #3094 still pass unchanged. The security
  intent (blocking paths that point at sensitive directories, see the hardening
  done around CVE-2022-30535) is preserved — only the unintended substring
  matching is removed.
- Added regression tests for both directions: paths that must stay blocked and paths that must no longer be rejected.

### Checklist

### DIFF
--- a/internal/k8s/validation.go
+++ b/internal/k8s/validation.go
@@ -1399,11 +1399,11 @@ func ValidateEscapedString(body string, examples ...string) error {
 }
 
 const (
-	illegalKeywordFmt    = `/etc/|/root|/var|\\n|\\r`
+	illegalKeywordFmt    = `^/etc/|^/root(?:/|$)|^/var(?:/|$)|\\n|\\r$`
 	illegalKeywordErrMsg = `must not contain invalid paths`
 )
 
-var illegalKeywordFmtRegexp = regexp.MustCompile("^" + illegalKeywordFmt + "$")
+var illegalKeywordFmtRegexp = regexp.MustCompile(illegalKeywordFmt)
 
 func validateIllegalKeywords(path string, fieldPath *field.Path) field.ErrorList {
 	if illegalKeywordFmtRegexp.MatchString(path) {
--- a/internal/k8s/validation_test.go
+++ b/internal/k8s/validation_test.go
@@ -6129,8 +6129,11 @@ func TestValidateIllegalKeywords(t *testing.T) {
 
 	invalidPaths := []string{
 		"/root",
+		"/root/",
+		"/root/.ssh",
 		"/etc/nginx/secrets",
 		"/etc/passwd",
+		"/var",
 		"/var/run/secrets",
 		`\n`,
 		`\r`,
@@ -6139,7 +6142,27 @@ func TestValidateIllegalKeywords(t *testing.T) {
 	for _, path := range invalidPaths {
 		allErrs := validateIllegalKeywords(path, field.NewPath("path"))
 		if len(allErrs) == 0 {
-			t.Errorf("validateCurlyBraces(%q) returned no errors for invalid input", path)
+			t.Errorf("validateIllegalKeywords(%q) returned no errors for invalid input", path)
+		}
+	}
+
+	validPaths := []string{
+		"/variables",
+		"/var-config",
+		"/manager/variables",
+		"/api/variables/list",
+		"/rootpage",
+		"/root-cause",
+		"/tree/root",
+		"/etcetera",
+		"/etcd",
+		"/fetch",
+	}
+
+	for _, path := range validPaths {
+		allErrs := validateIllegalKeywords(path, field.NewPath("path"))
+		if len(allErrs) != 0 {
+			t.Errorf("validateIllegalKeywords(%q) returned errors %v for valid input", path, allErrs)
 		}
 	}
 }


AlexFenlon:
Hi @hugolevino, thank you for this PR and the issue! we really appreciate it, PR looks good as far as I can tell, we will review this in when the team gets some time!

hugolevino:
Thanks @AlexFenlon, really appreciate it!
 
I folded a few extra edge cases into the validation tests (bare `/etc`, an escaped `\r` mid-path, and `/etcd` to confirm `/etc`-prefixed words still pass). Nothing changed in the fix itself, just locking the behaviour down with tests.

internal/k8s/validation.go
```diff
@@ -1399,11 +1399,11 @@ func ValidateEscapedString(body string, examples ...string) error {
 }
 
 const (
-	illegalKeywordFmt    = `/etc/|/root|/var|\\n|\\r`
+	illegalKeywordFmt    = `^(?:/etc(?:/|$)|/root(?:/|$)|/var(?:/|$))|\\n|\\r`
```
vepatel:
```suggestion
	illegalKeywordFmt    = `^/etc/|^/root(?:/|$)|^/var(?:/|$)|\\n|\\r$`
```
~I'd prefer if we keep the earlier allowed path like `/etc` just to avoid breaking changes in case someone has configured it already, otherwise this is good to go~

Updated to match @hugolevino's suggestion

hugolevino: @vepatel
I tested proposed change and it would still block `/etc`

In order to make `/etc` not illegal and keep same behaviour as we currently have, I would suggest:

```suggestion
	illegalKeywordFmt    = `^/etc/|^/root(?:/|$)|^/var(?:/|$)|\\n|\\r`
```

But if we trully want to avoid breaking changes, while only relaxing /var and /root as substrings of path.. we should also anchor `\\r$` to maintain same behaviour.

```suggestion
	illegalKeywordFmt    = `^/etc/|^/root(?:/|$)|^/var(?:/|$)|\\n|\\r$`
```

Let me know what do you think.
We will probably need to change some tests scenarios based on this.

vepatel: @hugolevino Agree, feel free to commit the change `^/etc/|^/root(?:/|$)|^/var(?:/|$)|\\n|\\r$` in your branch and add/update unit tests and I'll get it approved and merged asap

hugolevino: @vepatel done

vepatel (approved):

javorszky (approved):

vepatel:
@hugolevino updated description to match the code update, thanks for the contribution!

pdabelf5 (approved):

hugolevino:
@vepatel I just updated description to match the code update.

PR #10230, opened 2026-06-18
filter by only ingress controller events when updating status

### LINKED ISSUE #10229: VirtualServer status.state set to empty string
### Version

5.5.0

- GKE Google Cloud
- Kops AWS Clusters

### Steps to reproduce

Deploy any controller that might produce an event on a `VirtualServer` or `VirtualServerRoute` with a `Reason` field that's unknonw to the Ingress controller.

 
# Bug: VirtualServer status.state set to empty string

## Summary

When a new leader is elected, VirtualServer and VirtualServerRoute resources can have their `status.state` field overwritten with an empty string, causing the status to become invalid. Additionally, the `status.reason` and `status.message` fields get polluted with messages from other controllers, making it appear as though the ingress controller reported those errors.

## Root Cause

During leader election, the controller calls `updateVirtualServersStatusFromEvents()` to restore status from Kubernetes events. The function finds the latest event for each VirtualServer and derives the status from the event's `Reason` field using `getStatusFromEventTitle()`.

The bug occurs because:

1. The function iterated over **all** events for a VirtualServer, not just events created by the ingress controller
2. Third-party controllers (e.g., Kyverno) create events on VirtualServer resources with their own `Reason` values (e.g., `PolicyViolation`)
3. `getStatusFromEventTitle()` returns an empty string for unrecognized event reasons
4. If a Kyverno event is the most recent event, the status is updated with `state: ""`

## Affected Code

`internal/k8s/controller.go`:
- `updateVirtualServersStatusFromEvents()` (line ~2461)
- `updateVirtualServerRoutesStatusFromEvents()` (line ~2505)
- `getStatusFromEventTitle()` (line ~2448) - returns `""` for unrecognized reasons

## Trigger Scenario

1. VirtualServer exists with valid `status.state` (e.g., "Valid")
2. Kyverno evaluates the VirtualServer and creates a `PolicyViolation` event
3. Leader election occurs (pod restart, scaling, etc.)
4. New leader calls `updateVirtualServersStatusFromEvents()`
5. Kyverno's `PolicyViolation` event is the most recent event
6. `getStatusFromEventTitle("PolicyViolation")` returns `""`
7. Status is updated with empty `state`, overwriting the valid state

## Fix

Filter events to only consider those created by the ingress controller by checking `event.ReportingController == "nginx-ingress-controller"`. Skip the status update entirely if no matching events are found.

## Affected Versions

Unknown - requires investigation of when this code path was introduced.

fcrespofastly:
### Proposed changes
# PR: Filter events by reporting controller when updating VirtualServer/VirtualServerRoute status

## Summary

This PR fixes a bug where VirtualServer and VirtualServerRoute resources could have their `status.state` field overwritten with an empty string during leader election, and their `status.reason` and `status.message` fields polluted with messages from third-party controllers.

## Problem

When a new leader is elected, the controller calls `updateVirtualServersStatusFromEvents()` and `updateVirtualServerRoutesStatusFromEvents()` to restore status from Kubernetes events. The functions find the latest event for each resource and derive the status from the event's `Reason` field using `getStatusFromEventTitle()`.

The bug occurs because:

1. The functions iterated over **all** events for a VirtualServer/VirtualServerRoute, not just events created by the ingress controller
2. Third-party controllers (e.g., Kyverno) create events on these resources with their own `Reason` values (e.g., `PolicyViolation`)
3. `getStatusFromEventTitle()` returns an empty string for unrecognized event reasons
4. If a third-party event is the most recent event, the status is updated with `state: ""`

### Trigger Scenario

1. VirtualServer exists with valid `status.state` (e.g., "Valid")
2. Kyverno evaluates the VirtualServer and creates a `PolicyViolation` event
3. Leader election occurs (pod restart, scaling, etc.)
4. New leader calls `updateVirtualServersStatusFromEvents()`
5. Kyverno's `PolicyViolation` event is the most recent event
6. `getStatusFromEventTitle("PolicyViolation")` returns `""`
7. Status is updated with empty `state`, overwriting the valid state

## Solution

Filter events to only consider those created by the ingress controller by checking `event.ReportingController == EventReporterName`. 

Additionally, fixed a bug where the timestamp variable wasn't being updated in the loop, which could cause incorrect "latest" event selection when multiple NIC events exist.

### `internal/k8s/controller.go`

**`updateVirtualServersStatusFromEvents()`** (line ~2488):
- Added `timestamp = event.CreationTimestamp.Time` to correctly track the latest event timestamp

**`updateVirtualServerRoutesStatusFromEvents()`** (line ~2533):
- Added `&& event.ReportingController == EventReporterName` filter condition
- Added `timestamp = event.CreationTimestamp.Time` to correctly track the latest event timestamp

### `internal/k8s/controller_test.go`

Added two new test functions with 4 test cases each:

- `TestUpdateVirtualServersStatusFromEvents_FiltersEventsByReportingController`
- `TestUpdateVirtualServerRoutesStatusFromEvents_FiltersEventsByReportingController`

Test cases cover:
1. Only NIC event - should use NIC event
2. Third-party event newer than NIC event - should ignore third-party and use NIC event
3. Only third-party events - should not update status (empty state)
4. Multiple NIC events - should use latest NIC event by timestamp

## Testing

```bash
go test -v -run "TestUpdateVirtualServersStatusFromEvents_FiltersEventsByReportingController|TestUpdateVirtualServerRoutesStatusFromEvents_FiltersEventsByReportingController" ./internal/k8s/...
```

All 8 test cases pass.

## Related Issue

Fixes #10229

### Checklist

### DIFF
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -2521,6 +2521,7 @@ func (lbc *LoadBalancerController) updateVirtualServersStatusFromEvents() error 
 			var latestEvent api_v1.Event
 			for _, event := range events.Items {
 				if event.CreationTimestamp.After(timestamp) && event.ReportingController == EventReporterName {
+					timestamp = event.CreationTimestamp.Time
 					latestEvent = event
 				}
 			}
@@ -2564,7 +2565,8 @@ func (lbc *LoadBalancerController) updateVirtualServerRoutesStatusFromEvents() e
 			var timestamp time.Time
 			var latestEvent api_v1.Event
 			for _, event := range events.Items {
-				if event.CreationTimestamp.After(timestamp) {
+				if event.CreationTimestamp.After(timestamp) && event.ReportingController == EventReporterName {
+					timestamp = event.CreationTimestamp.Time
 					latestEvent = event
 				}
 			}
--- a/internal/k8s/controller.go
+++ b/internal/k8s/controller.go
@@ -78,6 +78,8 @@ const (
 	ingressClassKey = "kubernetes.io/ingress.class"
 	// IngressControllerName holds Ingress Controller name
 	IngressControllerName = "nginx.org/ingress-controller"
+	// EventReporterName
+	EventReporterName = "nginx-ingress-controller"
 
 	typeKeyword                                     = "type"
 	helmReleaseType                                 = "helm.sh/release.v1"
@@ -2518,7 +2520,7 @@ func (lbc *LoadBalancerController) updateVirtualServersStatusFromEvents() error 
 			var timestamp time.Time
 			var latestEvent api_v1.Event
 			for _, event := range events.Items {
-				if event.CreationTimestamp.After(timestamp) {
+				if event.CreationTimestamp.After(timestamp) && event.ReportingController == EventReporterName {
 					latestEvent = event
 				}
 			}
--- a/internal/k8s/controller_test.go
+++ b/internal/k8s/controller_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/nginx/kubernetes-ingress/internal/metrics/collectors"
 	"github.com/nginx/kubernetes-ingress/internal/nginx"
 	conf_v1 "github.com/nginx/kubernetes-ingress/pkg/apis/configuration/v1"
+	fake_versioned "github.com/nginx/kubernetes-ingress/pkg/client/clientset/versioned/fake"
 	api_v1 "k8s.io/api/core/v1"
 	networking "k8s.io/api/networking/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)


fcrespofastly:

razaj92:
👋 Any consideration on this?

Thanks

ciarams87 (changes requested):
Thanks for this @fcrespofastly, and sorry for the wait. The diagnosis in #10229 is correct and the two-part fix here (advance the `timestamp` cursor, filter by `ReportingController`) is the right shape. I verified both halves against `main`.

On @x-itm's concern about `ReportingController` being unreliable - you're right that it's
populated here. The field is set client-side by the recorder that emits the event, not by the apiserver:

 ```go
 // k8s.io/client-go@v0.37.0 tools/record/event.go:463-464 (generateEvent)
 event.ReportingInstance   = recorder.source.Host
 event.ReportingController = recorder.source.Component
 ```

and our recorder is built with `EventSource{Component: "nginx-ingress-controller"}` (`cmd/nginx-ingress/main.go:118`), which is exactly the constant this PR introduces. So the value depends on the client-go we vendor, not on the cluster's control-plane version.

The historical caveat is real but doesn't apply to us: the legacy recorder did not set the field at all until [kubernetes/kubernetes#116506](https://github.com/kubernetes/kubernetes/pull/116506)
(v1.28, cherry-picked to release-1.25/1.26/1.27 on 2023-08-28). Our floor is `kubeVersion: "= 1.25.0-0"` (`charts/nginx-ingress/Chart.yaml:5`), and since we emit the events ourselves with a current client-go, every event we write carries the field regardless. Falling back to `Source.Component` isn't needed. Filtering on `Reason` instead would be worse - an unknown `Reason` is the root cause of #10229 in the first place.

One change needed before this can merge. The filter is applied, but the status update still runs when no Ingress Controller event was found. `latestEvent` stays zero-valued, and `getStatusFromEventTitle("")` returns `""`, so `status.state`, `.reason` and `.message` are all overwritten with empty strings - the same symptom #10229 reports, reached by a different route. #10229's own `Fix` section asks for this: "Skip the status update entirely if no matching events are found."

This isn't a corner case. The apiserver expires Events (`--event-ttl`, 1h by default), so a long-lived VirtualServer can have its own events aged out while a third-party controller keeps writing fresh ones. The next leader election then blanks the status.

Reproduced on your branch — a VirtualServer starting at `state: Valid` with a single `kyverno-admission` event ends at `state: "" reason: "" message: ""`:

 ```
 resulting status: state="" reason="" message=""
 status.state was clobbered: want "Valid", got ""
 ```

Worth noting the three new subtests named `"only third-party events - should not update status (empty state)"` currently assert the clobbering as expected behaviour: the fixture starts at `State: "initial"` and the test expects `""`, so it passes *because* the status was overwritten. They should assert the status is left untouched.

The minimal fix is a `found` return on the event selection plus a `continue` at all three call sites. Since the selection loop is duplicated three times (and has already drifted - the operand order differs in `transport_server.go`), it's probably worth extracting. Happy to push the change to your branch if you'd prefer, or you can take it from here - either is fine.

Also needed: the `updateTransportServersStatusFromEvents` path has the fix but no test.

fcrespofastly:
Thanks for the feedback @ciarams87! 

Can you clarify this part please? 

> One change needed before this can merge. The filter is applied, but the status update still runs when no Ingress Controller event was found. latestEvent stays zero-valued, and getStatusFromEventTitle("") returns "", so status.state, .reason and .message are all overwritten with empty strings - the same symptom https://github.com/nginx/kubernetes-ingress/issues/10229 reports, reached by a different route. https://github.com/nginx/kubernetes-ingress/issues/10229's own Fix section asks for this: "Skip the status update entirely if no matching events are found."

How are we going to reach this state? Is there any use-case for the ingress controller to emit an event with no Reason string at all?

> This isn't a corner case. The apiserver expires Events (--event-ttl, 1h by default), so a long-lived VirtualServer can have its own events aged out while a third-party controller keeps writing fresh ones. The next leader election then blanks the status.

But a third-party controller would have a different `ReportingController` ?

> Reproduced on your branch — a VirtualServer starting at state: Valid with a single kyverno-admission event ends at state: "" reason: "" message: "":
resulting status: state="" reason="" message=""
status.state was clobbered: want "Valid", got ""

Can you clarify how did you reproduce it? Because kyverno-admission-controller uses a different ReportingController.

ciarams87:
Good questions @fcrespofastly and you're right on both counts, which is what makes this subtle. Let me be more precise, because I explained it badly.

**Your filter is correct. The bug is what happens *because* it correctly excludes the kyverno event.**

Walking `updateVirtualServersStatusFromEvents` as it stands on the branch:

```go
2516	if len(events.Items) == 0 {
2517		continue
2518	}
2519
2520	var timestamp time.Time
2521	var latestEvent api_v1.Event          // zero value
2522	for _, event := range events.Items {
2523		if event.CreationTimestamp.After(timestamp) && event.ReportingController == EventReporterName {
2524			timestamp = event.CreationTimestamp.Time
2525			latestEvent = event
2526		}
2527	}
2528
2529	err = lbc.statusUpdater.UpdateVirtualServerStatus(vs, getStatusFromEventTitle(latestEvent.Reason), latestEvent.Reason, latestEvent.Message)
```

With one kyverno event and no NIC event:

1. Line 2516 checks the **unfiltered** list. The kyverno event is in it, so `len != 0` and we do **not** `continue`.
2. Your filter at 2523 correctly rejects it, so lines 2524-2525 never run and `latestEvent` is **never assigned**, it stays the zero-valued `api_v1.Event{}` from 2521.
3. Line 2529 then runs **unconditionally**, with `latestEvent.Reason == ""`, not because anything emitted an empty `Reason`, but because that's the zero value of an unassigned struct field. `getStatusFromEventTitle("")` falls through its switch and returns `""`, so we write `state: "", reason: "", message: ""`.

So to your three questions directly:

> Is there any use-case for the ingress controller to emit an event with no Reason string at all?

No, and it doesn't. The empty `Reason` is the zero value of `latestEvent`, which never got assigned.

> But a third-party controller would have a different `ReportingController`?

Yes, exactly, and that's the problem rather than the protection. Because it differs, nothing matches, and line 2529 writes the status from the unassigned `latestEvent`.

> Can you clarify how did you reproduce it? Because kyverno-admission-controller uses a different ReportingController.

Your own test already demonstrates it. `TestUpdateVirtualServersStatusFromEvents_FiltersEventsByReportingController`, the fixture starts at `State: "initial", Reason: "initial", Message: "initial"` (`controller_test.go:4890-4894`), and the subtest *"only third-party events - should not update status (empty state)"* asserts `expectedState: ""` / `expectedReason: ""` (`:4837-4838`). That test passes, which can only happen if the status was overwritten from `"initial"` to `""`. The subtest name says the status should not be updated, but the assertion is pinning the fact that it *is*, to empty.

The pre-existing `len(events.Items) == 0` guard at 2516 was the check for "this VirtualServer has no events". Once a filter is added, a second check is needed for "none of its events are ours":

```go
latestEvent, found := latestEventEmittedByIngressController(events.Items)
if !found {
	continue
}
```

That also matches what #10229 asks for in its own Fix section "Skip the status update entirely if no matching events are found."

On reachability: you're right that it needs a VirtualServer whose NIC events have aged out while a third-party controller keeps writing fresh ones. The apiserver's `--event-ttl` defaults to 1h, so for a stable VirtualServer that has not changed in a while, its own `AddedOrUpdated` event is long gone while kyverno re-evaluates on a schedule. The next leader election then blanks the status. Not the common path, but not hypothetical either, and it's the same symptom #10229 reports, which is why I flagged it.

The fix is just the guard above, the same two lines on the VSR and TransportServer paths, and flipping those three subtests to assert the status is preserved.

fcrespofastly:
Thanks @ciarams87 ! Understood now. I updated the PR, please let me know if there's anything else I should look into, we're kind of in a hurry with this one.

> ## Copilot review overview
> ### 🟡 Changes recommended
> Event ordering can restore stale status, and the duplicated reporter identity risks silently disabling restoration.
> 
> _Get a fresh assessment by requesting another Copilot review._
> 
> **Review effort:** Balanced **Findings:** 1 High severity
> 
> **Open (1)**
> * High severity [Select NIC events by LastTimestamp, not CreationTimestamp](#discussion_r4075060324) · New
> 
> **What changed in this PR**
> Filters leader-election status restoration to NIC-reported events and selects the latest matching event.
> 
> **Changes:**
> 
> * Adds shared event filtering and selection.
> * Applies restoration logic to VirtualServer, VirtualServerRoute, and TransportServer.
> * Adds unit coverage for reporter filtering and event ordering.
> 
> File	Changes and findings
> `internal/​k8s/​transport_server.go`	Applies filtered event restoration to TransportServers. No issues identified.
> `internal/​k8s/​transport_server_test.go`	Adds TransportServer restoration tests. No issues identified.
> `internal/​k8s/​controller.go`	Adds shared filtering and selection. **Critical:** selection must use event occurrence time rather than `CreationTimestamp`. **Moderate:** centralize the reporter identity instead of duplicating the recorder literal.
> `internal/​k8s/​controller_test.go`	Adds VirtualServer and VirtualServerRoute tests; fixtures need aggregated/re-emitted event coverage.
> 💡 [Add a `code-review` agent skill](/nginx/kubernetes-ingress/new/main?filename=.github/skills/code-review/SKILL.md) or configure MCP servers for context-aware, tailored reviews. [Learn more in the docs.](https://docs.github.com/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review?tool=webui#mcp-servers-and-agent-skills)

Also addressed this! Thanks

ciarams87 (commented):
This looks good @fcrespofastly, thanks for turning it round so quickly!

Couple of small things before I approve.

The reporter identity is written out in two places that have to agree but don't reference each other. `cmd/nginx-ingress/main.go` builds the recorder with `api_v1.EventSource{Component: "nginx-ingress-controller"}`, and `EventReporterName` in `internal/k8s/controller.go` is what the filter compares against. They're linked only because client-go copies one into the other (`tools/record/event.go`: `event.ReportingController = recorder.source.Component`) - different field names, no comment tying them together. If either string changes, the filter matches nothing, `found` is false for every resource, and status restoration silently stops working with no test failing. Let's update the string to the constant in main.go too - `main.go` already imports `internal/k8s`, so pointing it at the constant should be a one-liner.

Also, `// EventReporterName` needs a description after the name or `revive` will fail the lint job:

```
internal/k8s/controller.go:80:2: exported: comment on exported const EventReporterName should be of the form "EventReporterName ..." (revive)
```

Could you also rebase against main, it's a bit behind now, and we'll get the CI run done. Thanks again!

fcrespofastly:
Thanks @ciarams87 ! Done !!

ciarams87 (approved):

internal/k8s/controller.go
```diff
@@ -2879,6 +2881,27 @@ func getStatusFromEventTitle(eventTitle string) string {
...
+// CreationTimestamp only tracks the first one.
+func latestEventEmittedByIngressController(events []api_v1.Event) (api_v1.Event, bool) {
+	var latestEvent api_v1.Event
+	var found bool
+
+	for _, event := range events {
+		if event.ReportingController != EventReporterName {
+			continue
+		}
+		if !found || event.LastTimestamp.After(latestEvent.LastTimestamp.Time) {
```
ciarams87:
```suggestion
		if !found || !event.LastTimestamp.Before(&latestEvent.LastTimestamp) {
```

ciarams87 (changes requested):
Sorry @fcrespofastly - I ran the full pipeline and it caught a regression. The VSR smoke tests fail on both alpine and alpine-plus-fips:

```
Failed: virtualservers 'virtual-server-route' did not reach state 'Valid'. Current status:
{'reason': 'AddedOrUpdatedWithWarning', 'state': 'Warning'}
```

The VS gets stuck on a stale Warning. Root cause is a tie-break, and it's genuinely subtle; the code you replaced was buggy in a way that happened to be correct:

```go
var timestamp time.Time            // never assigned inside the loop
if event.CreationTimestamp.After(timestamp) { ... }   // therefore always true
```

so it always ended up with the *last* event in the list. Events come back name-ordered and names are `<obj>.<hex UnixNano>`, so last == newest. It worked by accident.

Your version is the right approach, but `metav1.Time` serialises at RFC3339 **second** precision. When a VS goes Warning → Valid within the same second - which is exactly what the VSR fixture does, creating the VS before its VSRs - both events come back with identical `LastTimestamp`, and strict `After()` keeps the first one, so the stale Warning wins. `updateVirtualServersStatusFromEvents` only runs once on acquiring leadership, so nothing corrects it afterwards.

One line fixes it - prefer the later entry on a tie:

```go
if !found || !event.LastTimestamp.Before(&latestEvent.LastTimestamp) {
```

(`Before` takes a `*metav1.Time`, unlike `After`.)

I verified locally: that makes the smoke-test scenario pass, all your existing cases still pass, and `make test` is green. Worth adding a case to `TestLatestEventEmittedByIngressController` to pin it, something like:

```go
{
    name: "two NIC events in the same second",
    events: []api_v1.Event{
        event("nic-warning", EventReporterName, baseTime, baseTime),
        event("nic-valid", EventReporterName, baseTime, baseTime),
    },
    expectedName:  "nic-valid",
    expectedFound: true,
},
```

I'll re-run the pipeline once you've pushed.

fcrespofastly:
pushed @ciarams87 ! 🚢

ciarams87 (approved):
🚀 Thanks so much @fcrespofastly! I'll reach out to the team to get your second review and we'll try and get this merged in as soon as possible

javorszky (approved):

vepatel (approved):

PR #10282, opened 2026-06-25
Add -latency-metrics-buckets flag to configure upstream latency histogram buckets

### LINKED ISSUE #10281: Make the upstream latency histogram buckets configurable

The `nginx_ingress_controller_upstream_server_response_latency_ms` histogram is created with a hardcoded 26-bucket list in [`internal/metrics/collectors/latency.go`](https://github.com/nginx/kubernetes-ingress/blob/main/internal/metrics/collectors/latency.go). Every histogram bucket is its own time series, so the bucket count multiplies the metric's cardinality across each `upstream` / `server` / `code` (and pod) label combination. On large deployments running with `-enable-latency-metrics`, this fixed bucket count is a meaningful  contributor to active series, remote-write traffic, and storage cost and there is currently no way to trim it.

A command-line argument (and a matching Helm value) to override the latency histogram buckets, defaulting to the current list so existing deployments are unaffected. For example:

```
-latency-metrics-buckets=5,10,25,50,100,250,500,1000
```

This is similar to how`kubernetes/ingress-nginx` already exposes equivalent knobs (`--time-buckets`, `--length-buckets`, `--size-buckets`). https://github.com/kubernetes/ingress-nginx/blob/dbb11b92ddb77bee9c35e462479129354c84f939/docs/user-guide/cli-arguments.md?plain=1#L43-L44

- Dropping or relabeling the metric in the Prometheus scrape config — loses the data entirely instead of letting operators trade resolution for cost.
- Downsampling downstream with recording rules — reduces query cost but not ingestion/active-series cost at the source.

**Additional context**

 I have a change ready https://github.com/nginx/kubernetes-ingress/pull/10282 that adds a `-latency-metrics-buckets` flag (validated as positive and strictly ascending) threaded into `NewLatencyMetricsCollector`, a `controller.latencyMetricsBuckets` Helm value, unit tests, and passing helmunit snapshots. Default behaviour is unchanged.

bm1216:
### Proposed changes

Closes #10281.

The `upstream_server_response_latency_ms` histogram is created with a hardcoded 26-bucket list in `internal/metrics/collectors/latency.go`. Because every histogram bucket is a separate time series, the bucket count multiplies the metric's cardinality across each `upstream` / `server` / `code` (and pod) label combination, and there is no way to trim it. On large deployments running with `-enable-latency-metrics`, this is a meaningful and unavoidable contributor to active series and remote-write/storage cost.

This PR makes the buckets configurable, mirroring how `kubernetes/ingress-nginx` exposes `--time-buckets` / `--length-buckets` / `--size-buckets`:

- New `-latency-metrics-buckets` command-line argument — a comma-separated list of bucket bounds in milliseconds, e.g. `-latency-metrics-buckets=5,10,25,50,100,250,500,1000`. Requires `-enable-latency-metrics`.
- Values are validated as positive and strictly ascending (as required by the Prometheus client). Invalid input logs a warning and falls back to the default list rather than tripping the client's bucket-ordering panic at startup.
- Threaded into `NewLatencyMetricsCollector`, which falls back to the exported `DefaultLatencyBuckets` (the previous hardcoded list) when no buckets are supplied.
- New `controller.latencyMetricsBuckets` Helm value (and `values.schema.json` entry), rendered as the flag only when set — so existing chart output and the helmunit snapshots are unchanged.

Default behaviour is unchanged: with the flag/value unset, the same 26 buckets are used as before.

Tests: added `TestParseLatencyBuckets` (parser validation) and collector tests asserting custom buckets are honoured and that an empty list falls back to the default. `go test ./cmd/nginx-ingress/... ./internal/metrics/collectors/...` and the `helmunit` chart snapshot tests pass.

Docs: the flag's `-help` text and the chart `values.yaml` comment document the option in-repo. The hosted CLI-arguments reference on docs.nginx.com lives in a separate repository and would be a follow-up.

### Checklist


bm1216:

bm1216:
recheck

haywoodsh:
Tested locally with the flag `- -latency-metrics-buckets=5,10,25` and I can get the following metrics with `le="5"`, etc:
```
nginx_ingress_controller_upstream_server_response_latency_ms_bucket{class="nginx",code="200",pod_name="coffee-6db967495b-rt5wh",resource_name="cafe-ingress",resource_namespace="default",resource_type="ingress",server="10.42.0.125:8080",service="coffee-svc",upstream="default-cafe-ingress-cafe.example.com-coffee-svc-80",le="5"} 4
nginx_ingress_controller_upstream_server_response_latency_ms_bucket{class="nginx",code="200",pod_name="coffee-6db967495b-rt5wh",resource_name="cafe-ingress",resource_namespace="default",resource_type="ingress",server="10.42.0.125:8080",service="coffee-svc",upstream="default-cafe-ingress-cafe.example.com-coffee-svc-80",le="10"} 4
nginx_ingress_controller_upstream_server_response_latency_ms_bucket{class="nginx",code="200",pod_name="coffee-6db967495b-rt5wh",resource_name="cafe-ingress",resource_namespace="default",resource_type="ingress",server="10.42.0.125:8080",service="coffee-svc",upstream="default-cafe-ingress-cafe.example.com-coffee-svc-80",le="25"} 5
nginx_ingress_controller_upstream_server_response_latency_ms_bucket{class="nginx",code="200",pod_name="coffee-6db967495b-rt5wh",resource_name="cafe-ingress",resource_namespace="default",resource_type="ingress",server="10.42.0.125:8080",service="coffee-svc",upstream="default-cafe-ingress-cafe.example.com-coffee-svc-80",le="+Inf"} 5
```

- For the Helm chart, I would suggest adding some Helm test cases with a non-empty flag value, as well as possible Helm validations for early rejection, such as a regex in the JSON schema.
- I would also suggest adding Python test cases to tests/suite/test_prometheus_metrics.py showing that it has the exact number of outputs as the number of flag values and contains the correct label values from the flag values.

bm1216:
> Tested locally with the flag `- -latency-metrics-buckets=5,10,25` and I can get the following metrics with `le="5"`, etc:
> 
> ```
> nginx_ingress_controller_upstream_server_response_latency_ms_bucket{class="nginx",code="200",pod_name="coffee-6db967495b-rt5wh",resource_name="cafe-ingress",resource_namespace="default",resource_type="ingress",server="10.42.0.125:8080",service="coffee-svc",upstream="default-cafe-ingress-cafe.example.com-coffee-svc-80",le="5"} 4
> nginx_ingress_controller_upstream_server_response_latency_ms_bucket{class="nginx",code="200",pod_name="coffee-6db967495b-rt5wh",resource_name="cafe-ingress",resource_namespace="default",resource_type="ingress",server="10.42.0.125:8080",service="coffee-svc",upstream="default-cafe-ingress-cafe.example.com-coffee-svc-80",le="10"} 4
> nginx_ingress_controller_upstream_server_response_latency_ms_bucket{class="nginx",code="200",pod_name="coffee-6db967495b-rt5wh",resource_name="cafe-ingress",resource_namespace="default",resource_type="ingress",server="10.42.0.125:8080",service="coffee-svc",upstream="default-cafe-ingress-cafe.example.com-coffee-svc-80",le="25"} 5
> nginx_ingress_controller_upstream_server_response_latency_ms_bucket{class="nginx",code="200",pod_name="coffee-6db967495b-rt5wh",resource_name="cafe-ingress",resource_namespace="default",resource_type="ingress",server="10.42.0.125:8080",service="coffee-svc",upstream="default-cafe-ingress-cafe.example.com-coffee-svc-80",le="+Inf"} 5
> ```
> 
> * For the Helm chart, I would suggest adding some Helm test cases with a non-empty flag value, as well as possible Helm validations for early rejection, such as a regex in the JSON schema.
> * I would also suggest adding Python test cases to tests/suite/test_prometheus_metrics.py showing that it has the exact number of outputs as the number of flag values and contains the correct label values from the flag values.

Thanks for the comment!
Going to add the required tests now

bm1216:
Addressed in 170bfa8..fb10cb1:

- Added a `pattern` to the `controller.latencyMetricsBuckets` schema entry in `values.schema.json` so malformed input (e.g. non-numeric or empty-list-with-comma) is rejected by `helm template`/`install` before the controller starts, rather than only at controller startup. This mirrors the existing duration-string pattern used elsewhere in the schema.
- Added a positive helmunit test case (`latencyMetricsBuckets`) asserting the flag renders correctly with a non-empty value, and a negative case (`latencyMetricsBucketsInvalid`) asserting the new schema pattern rejects a malformed value.
- Added `test_latency_metrics_custom_buckets` to `tests/suite/test_prometheus_metrics.py`, deploying with `-latency-metrics-buckets=5,10,25` and asserting the `/metrics` response contains exactly one bucket series per configured bound plus `+Inf`, and that a default-only bound (`le="1000"`) is absent — confirming both the exact output count and the correct label values you flagged.

All Go unit tests and helmunit tests pass locally (42 + 42).

bm1216:
Hi! Could I get another review on this if possible?

bm1216:
Any luck with pushing this out please?

javorszky:
Hi @bm1216 ,

We talked about this on the community call on 7th september. Would you be able to resolve the conflicts? As we've tested the changeset and it works, we're going to merge this is once all the necessary checks pass.

Thank you for the fix!

bm1216:
Hiya, thanks! Just sorted out the merge conflicts. Sorry about that!

tests/suite/test_prometheus_metrics.py
```diff
@@ -165,6 +165,58 @@ def test_latency_metrics(
...
+        # One series per configured bucket bound, plus the implicit +Inf bucket
+        expected_bounds = custom_buckets + ["+Inf"]
+        assert len(bucket_lines) == len(expected_bounds), (
+            f"Expected {len(expected_bounds)} bucket series for buckets {expected_bounds}, "
+            f"got {len(bucket_lines)}: {bucket_lines}"
+        )
+        for bound in expected_bounds:
+            assert any(
+                f'le="{bound}"' in line for line in bucket_lines
+            ), f'Missing bucket series for le="{bound}" in {bucket_lines}'
```
javorszky: this can be simplified to this

```suggestion
        bounds = {
            line.split('le="', 1)[1].split('"', 1)[0]
            for line in bucket_lines
        }

        assert bucket_lines
        assert bounds == set(custom_buckets + ["+Inf"])
```

reason we can't use the `assert len(bucket_lines)` method is because ingress controller brings up a deployment, a service, and 2 replicas.

`ensure_connection` sends a GET request internally on line 191, then the second `ensure_connection` on line 195 sends a second.

We can't guarantee that both of those connections go to the same pod. The returned bucket lines might be duplicated, because they have `pod_name` label too, which means if the two requests reached two pods, we'd have 8 lines of output.

the suggested code extracts every unique value that the `le` label has, which should equal the 3 set buckets, and the `+Inf`

javorszky (changes requested):
one tiny change so the test won't end up flakey, otherwise excellent work!

tests/suite/test_prometheus_metrics.py
bm1216: Good find!

bm1216: pushed a fix

javorszky (dismissed):

javorszky:
@bm1216 there are some format issues remaining. To solve that we have two options:

* you run `make format` locally, and commit the changes and push it up, or
* with your permission, I can do it myself, and get this merged

Let me know which one sounds better for you 🙂

bm1216:
> @bm1216 there are some format issues remaining. To solve that we have two options:
> 
> 
> 
> * you run `make format` locally, and commit the changes and push it up, or
> 
> * with your permission, I can do it myself, and get this merged
> 
> 
> 
> Let me know which one sounds better for you 🙂 

Feel free to do it yourself and get it merged! Thanks for pushing this one through

bm1216:
@javorszky I pushed a commit https://github.com/nginx/kubernetes-ingress/pull/10282/commits/c0d5013f777d4710d55daf51cec963f80c36b5a0 which regenerates the unit test snapshots as it had older chart versions compared to main. This should fix the failing unit test i hope.

bm1216:
@javorszky sorry, i think you hit your copilot limit 😃

PR #10340, opened 2026-07-01
Fix conditional for nodePort in controller-service.yaml

### LINKED ISSUE #9054: [Bug]: (Helm chart) nodeport setting causes argoCD out-of-sync
### Version

5.3.0

EKS Amazon

### Steps to reproduce

1. Download the helm chart (2.4.2) into a folder
2. Run `helm template -s templates/controller-service.yaml`
3. Observe that with default values the nodePorts will show up as empty
    ```yaml
          - port: 443
            targetPort: 443
            protocol: TCP
            name: https
            nodePort:
   ```
4: If using a tool like ArgoCD to maintain this manifest via GitOps, this causes persistent diffs / out of sync state as it will notice a difference from the empty nodePort to the the ports "injected" by `spec.allocateLoadBalancerNodePorts`

Fix suggestion would be setting a check for emptiness that omits this field, which is what kubernetes/ingress-nginx does:
```go
{{- if and (or (eq .Values.controller.service.type "LoadBalancer") (eq .Values.controller.service.type "NodePort")) (not (empty .Values.controller.service.httpsPort.nodePort)) }}
```

josemaia:
### Proposed changes

Avoid setting an explicit empty value for nodePort when not using nodePorts, which causes issues with systems that compare the manifest with the live values (where a nodePort will be injected). Fixes #9054.

### Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/nginx/kubernetes-ingress/blob/main/CONTRIBUTING.md) doc

### DIFF
--- a/charts/nginx-ingress/templates/controller-service.yaml
+++ b/charts/nginx-ingress/templates/controller-service.yaml
@@ -54,7 +54,7 @@ spec:
     targetPort: {{ .Values.controller.service.httpPort.targetPort }}
     protocol: TCP
     name: {{ .Values.controller.service.httpPort.name }}
-  {{- if or (eq .Values.controller.service.type "LoadBalancer") (eq .Values.controller.service.type "NodePort") }}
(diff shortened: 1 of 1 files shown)


josemaia:

PR #10478, opened 2026-07-16
feat: add additional controller Services to the Helm chart

### LINKED ISSUE #10476: Add support for additional controller Services in the Helm chart
### Is your feature request related to a problem? Please describe

The Helm chart currently creates a single Service for the Ingress Controller pods. `controller.service.customPorts` can expose additional ports, but all ports still share the same Service, LoadBalancer IP, annotations, source ranges, and traffic policy.

A production use case needs one controller to serve both:

- HTTP/HTTPS Ingress resources through one LoadBalancer IP; and
- a Redis TCP listener configured by a TransportServer through a different LoadBalancer IP.

Creating a second controller would duplicate the workload and configuration. The current workaround is to maintain an out-of-chart Service manifest that manually duplicates the controller selector labels. This is also inconvenient for GitOps reconciliation and chart lifecycle management, as discussed in #3662.

### Describe the solution you'd like

Add `controller.service.additionalServices` to the Helm chart. The map key is the exact Service name. Each entry creates another Service that always selects the same controller pods while allowing independent Service configuration, including:

- type;
- annotations and labels;
- ports;
- LoadBalancer IP/class and source ranges;
- external traffic policy;
- external and cluster IPs;
- dual-stack settings;
- session affinity.

Example:

```yaml
controller:
  service:
    additionalServices:
      nginx-ingress-redis:
        type: LoadBalancer
        loadBalancerIP: 10.0.0.11
        externalTrafficPolicy: Local
        ports:
          - name: redis
            port: 6379
            targetPort: 6379
            protocol: TCP
```

A map keeps Helm values overrides composable and makes additional Service names unique. The selector is not configurable and is always rendered from `nginx-ingress.selectorLabels`, preventing drift from the controller pods.

### Alternatives considered

- `customPorts` (introduced from #580): does not provide a separate IP or per-Service annotations/policies.
- An extra manifest outside the chart: works, but duplicates chart-owned selector details and lifecycle management.
- A second controller: provides isolation but duplicates compute and operational overhead when a single NGINX instance is intended to serve both listeners.

### Additional context

This feature only controls Kubernetes Service exposure. The controller currently supports a single `reportIngressStatus.externalService`; selecting different status addresses per resource type is outside the scope of this proposal. TransportServer status does not publish the Service external IP.

GabboPenna:
### Proposed changes

This PR adds `controller.service.additionalServices` to the Helm chart so one NGINX Ingress Controller workload can be exposed through multiple independently configured Kubernetes Services.

The map key is the exact Service name. Each additional Service:

- always selects the controller pods using `nginx-ingress.selectorLabels`;
- can configure its own type, ports, annotations, labels, LoadBalancer settings, traffic policy, IP families, external IPs, and session affinity;
- is rendered independently from `controller.service.create`, allowing installations that only use explicitly defined Services;
- rejects `ExternalName`, missing/empty ports, invalid names, and collisions with the primary controller Service.

The default is an empty map, so existing chart output is unchanged. Status reporting remains intentionally unchanged: `controller.reportIngressStatus.externalService` still identifies a single Service.

Fixes #10476

### Validation

- `make secrets && make test` (Linux, Go 1.26.5, Helm 3.21.2)
- `helm lint charts/nginx-ingress`
- `golangci-lint run --new-from-rev HEAD`
- pre-commit YAML, JSON, JSON Schema, formatting, and security hooks

### Checklist

- [x] I have added tests that prove my feature works

### DIFF
--- a/charts/nginx-ingress/templates/controller-additional-services.yaml
+++ b/charts/nginx-ingress/templates/controller-additional-services.yaml
@@ -1,0 +1,69 @@
+{{- range $name, $service := .Values.controller.service.additionalServices }}
+{{- if eq $name (include "nginx-ingress.controller.service.name" $) }}
+{{- fail (printf "additional Service name %q must not match the primary controller Service name" $name) }}
+{{- end }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ $name }}
+  namespace: {{ $.Release.Namespace }}
+  labels:
+    {{- include "nginx-ingress.labels" $ | nindent 4 }}
+    {{- with $service.extraLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with $service.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- with $service.clusterIP }}
+  clusterIP: {{ . }}
+  {{- end }}
+  {{- if or (eq $service.type "LoadBalancer") (eq $service.type "NodePort") }}
+  {{- with $service.externalTrafficPolicy }}
+  externalTrafficPolicy: {{ . }}
+  {{- end }}
+  {{- end }}
+  {{- if eq $service.type "LoadBalancer" }}
+  {{- if hasKey $service "allocateLoadBalancerNodePorts" }}
+  allocateLoadBalancerNodePorts: {{ $service.allocateLoadBalancerNodePorts }}
+  {{- end }}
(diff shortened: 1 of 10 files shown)


AlexFenlon:
Hi @GabboPenna , thanks for the PR, we will try to review ti as soon as we can!

GabboPenna:
Hi @AlexFenlon, just a quick follow up on this one when you have a chance. Please let me know if the approach makes sense for the chart, or if you'd prefer something different. I'm happy to update/rebase the PR if needed, thanks!

PR #10543, opened 2026-07-28
fix: skip secret lookup when tls[].secretName is empty (fixes spurious warning with wildcard TLS)

### LINKED ISSUE #10218: [Bug]: Spurious warning "Error trying to get the secret" when tls[].secretName is empty and wildcard TLS is configured
### Version

5.5.0

EKS Amazon

## Description

When an `Ingress` resource has a `tls:` block with no `secretName` specified (relying
on the wildcard TLS secret configured via `--wildcard-tls-secret`), the controller
emits a spurious warning on **every sync** of that Ingress:

W controller.go:2691] Error trying to get the secret  for Ingress <name>: secret doesn't exist or of an unsupported type

Note the double space after `secret` — `secretName` is the empty string and is
printed as blank. TLS actually works correctly; the wildcard cert is applied as
intended. The warning is a false positive.

## Steps to Reproduce

1. Configure NIC with `--wildcard-tls-secret=<namespace>/<secret-name>`
2. Create an Ingress with a `tls:` block that has **no `secretName`**:

```yaml
spec:
  ingressClassName: nginx-f5
  rules:
  - host: example.com
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: my-service
            port:
              number: 80
  tls:
  - hosts:
    - example.com
    # secretName intentionally absent — relying on wildcard
```
3. Observe the controller logs — warning fires on every sync of this Ingress.

Root Cause
In internal/k8s/controller.go, createIngressEx() iterates ing.Spec.TLS
unconditionally and attempts a secret store lookup regardless of whether
secretName is empty:
```
for _, tls := range ing.Spec.TLS {
    secretName := tls.SecretName          // = "" when absent
    secretKey  := ing.Namespace + "/" + secretName  // = "namespace/"
    secretRef  := lbc.secretStore.GetSecret(secretKey) // always fails — invalid key
    if secretRef.Error != nil {
        nl.Warnf(lbc.Logger, "Error trying to get the secret %v for Ingress %v: %v",
            secretName, ing.Name, secretRef.Error)
    }
    ingEx.SecretRefs[secretName] = secretRef
}
```

The key "namespace/" (with empty secret name) can never exist in the secret
store. The lookup fails on every sync, forever.
However, the downstream addSSLConfig() in internal/configs/ingress.go
correctly handles this case — when secretName == "" and isWildcardEnabled,
it uses the wildcard cert path:
```
} else if isWildcardEnabled {
    pemFile = pemFileNameForWildcardTLSSecret  // ← correct behaviour
}
```
So TLS works correctly, but the warning fires spuriously.
Expected Behaviour
When tls[].secretName is empty and a wildcard TLS secret is configured
(isWildcardEnabled == true), no warning should be emitted. The controller
already knows at config-generation time that it will use the wildcard cert.
Suggested Fix
Guard the secret store lookup in createIngressEx():
```
for _, tls := range ing.Spec.TLS {
    secretName := tls.SecretName
    if secretName == "" {
        // No secretName specified; wildcard TLS will be used if --wildcard-tls-secret
        // is configured. Skip the lookup to avoid a spurious warning.
        ingEx.SecretRefs[secretName] = &secrets.SecretReference{}
        continue
    }
    secretKey := ing.Namespace + "/" + secretName
    secretRef := lbc.secretStore.GetSecret(secretKey)
    if secretRef.Error != nil {
        nl.Warnf(lbc.Logger, "Error trying to get the secret %v for Ingress %v: %v",
            secretName, ing.Name, secretRef.Error)
    }
    ingEx.SecretRefs[secretName] = secretRef
}
```

Impact
At scale this is not just cosmetic. Each warning indicates a sync that:
1. Iterates the full queue item
2. Calls configurator.AddOrUpdateIngress() which triggers cnf.Reload()
On a cluster with large number of Ingresses all using this pattern (wildcard TLS,
no secretName), every cluster-wide resync event (ConfigMap update, endpoint
change, pod restart) generates spurious warnings and up to 3,163 reload
triggers. Observed reload rate on production: ~1 reload every 10 seconds
on a cluster with minimal traffic.

Environment
 	 
NGINX Ingress Controller version	v5.4.3 (also confirmed present on main/v5.5.0)
NGINX version	                         1.29.1
Kubernetes version	                 1.34.8
Deployment method	                  Helm
--wildcard-tls-secret	                  configured
--nginx-plus	                                  false (OSS)

Verified In
- v5.4.3: internal/k8s/controller.go lines 2285–2296
- main (HEAD @ 6ec9a08): internal/k8s/controller.go lines 2685–2696
— identical code, bug present in both
Related
This pattern (wildcard TLS without per-Ingress secretName) is the intended
design for clusters that serve thousands of tenants behind a shared wildcard
certificate. The current behaviour makes wildcard TLS operationally noisy
at scale and generates misleading log output.

rahulii:
## Problem

Fixes #10218.

When an `Ingress` has a `tls:` block with **no `secretName`** (relying on `--wildcard-tls-secret`), `createIngressEx()` unconditionally attempts:

```go
secretKey := ing.Namespace + "/" + secretName   // = "namespace/" when secretName == ""
secretRef := lbc.secretStore.GetSecret(secretKey) // always fails — key can never exist
```

This fires a spurious warning on **every sync** of the Ingress:

```
W controller.go] Error trying to get the secret  for Ingress <name>: secret doesn't exist or of an unsupported type
```

Note the double space — `secretName` is `""}` and prints as blank. TLS actually works correctly; the downstream `addSSLConfig()` already handles the empty-secretName case by falling through to the wildcard cert path when `isWildcardEnabled == true`.

At scale (clusters serving thousands of tenants with shared wildcard TLS, no per-Ingress `secretName`), every cluster-wide resync generates thousands of spurious warnings and the reload pressure they cause.

## Fix

Guard the store lookup in `createIngressEx()` with a `secretName == ""` check. When empty, store a no-error `&secrets.SecretReference{}` and skip the lookup. The downstream config-generation path is unchanged.

```go
for _, tls := range ing.Spec.TLS {
    secretName := tls.SecretName
    if secretName == "" {
        // No secretName — wildcard TLS will be used if --wildcard-tls-secret is configured.
        // Skip the lookup to avoid a spurious warning on every sync.
        ingEx.SecretRefs[secretName] = &secrets.SecretReference{}
        continue
    }
    secretKey := ing.Namespace + "/" + secretName
    secretRef := lbc.secretStore.GetSecret(secretKey)
    if secretRef.Error != nil {
        nl.Warnf(lbc.Logger, "Error trying to get the secret %v for Ingress %v: %v", secretName, ing.Name, secretRef.Error)
    }
    ingEx.SecretRefs[secretName] = secretRef
}
```

## Testing

Added `TestCreateIngressEx_NoSpuriousWarningWhenTLSSecretNameEmpty` covering both:
- wildcard TLS configured (`--wildcard-tls-secret` set)
- wildcard TLS not configured

Both sub-cases verify that `SecretRefs[""]` is present with no error, matching the existing contract consumed by `addSSLConfig()`.

```
--- PASS: TestCreateIngressEx_NoSpuriousWarningWhenTLSSecretNameEmpty (0.00s)
    --- PASS: TestCreateIngressEx_NoSpuriousWarningWhenTLSSecretNameEmpty/wildcard_TLS_configured (0.00s)
    --- PASS: TestCreateIngressEx_NoSpuriousWarningWhenTLSSecretNameEmpty/wildcard_TLS_not_configured (0.00s)
```

All existing `internal/k8s` tests continue to pass.


rahulii:

vepatel (approved):

haywoodsh (approved):

rahulii:
can we merge this?

pdabelf5 (approved):

PR #10556, opened 2026-07-30
feat: add support for configuring otel-trace-context via ConfigMap

### LINKED ISSUE #9077: OpenTelemetry context propagation

It's currently not possible to configure all directives from the [ngx_otel_module](https://nginx.org/en/docs/ngx_otel_module.html) with NGINX Ingress.
Probably the most important one is [otel_trace_context](https://nginx.org/en/docs/ngx_otel_module.html#otel_trace_context) since the default is **ignore** which might no be desirable for an ingress.

Extend the current NGINX Ingress implementation to support configuring those directives either via global ConfigMap or per location via CRD (e.g. Policy).

If there is another way of setting _otel_trace_context_ to **propagate**, let me know :)

adberger:
### Proposed changes

enables configuring the otel-trace-context via ConfigMap the same as other OTel options are already supported.

### Checklist


adberger:

internal/configs/version1/nginx.tmpl
```diff
@@ -131,19 +131,24 @@ http {
     {{- end}}
 
     {{- if .MainOtelLoadModule }}
+    {{- if .MainOtelExporterEndpoint }}
```
haywoodsh: This toggle shouldn't be needed. the design was that `MainOtelExporterEndpoint` is always needed to set `MainOtelLoadModule` to true.

internal/configs/configmaps.go
```diff
@@ -989,7 +1006,7 @@ func parseConfigMapOpenTelemetry(l *slog.Logger, cfgm *v1.ConfigMap, cfgParams *
 		otelValid = false
 	}
 
-	if cfgParams.MainOtelExporterEndpoint != "" {
+	if cfgParams.MainOtelExporterEndpoint != "" || cfgParams.MainOtelTraceContext != "" {
```
haywoodsh: I don't think this is correct as it is changing the default behaviour

internal/configs/configmaps.go
```diff
@@ -979,6 +979,23 @@ func parseConfigMapOpenTelemetry(l *slog.Logger, cfgm *v1.ConfigMap, cfgParams *
...
 
+	if otelTraceContext, exists := cfgm.Data["otel-trace-context"]; exists {
+		otelTraceContext = strings.TrimSpace(otelTraceContext)
+		switch otelTraceContext {
+		case "extract", "inject", "propagate", "ignore":
+			cfgParams.MainOtelTraceContext = otelTraceContext
+		case "":
+		default:
+			errorText := fmt.Sprintf(
+				"ConfigMap %s/%s: invalid value for 'otel-trace-context': %q, must be one of 'extract', 'inject', 'propagate', 'ignore', ignoring",
```
haywoodsh: `ignoring` was not in the `case` line above

adberger:
@haywoodsh I've made changes you've proposed, shall I rebase?

javorszky:
> @haywoodsh I've made changes you've proposed, shall I rebase?

Hi @adberger ! Apologies for the silence. Yes, please rebase, and we'll take a look at it more thoroughly!

adberger:
@javorszky no worries, want to check again?

internal/configs/version1/template_test.go
```diff
@@ -5497,6 +5524,12 @@ var (
 		MainOtelExporterHeaderName:  "X-Custom-Header",
 		MainOtelExporterHeaderValue: "custom-value",
 		MainOtelServiceName:         "nginx-ingress-controller:nginx",
+		MainOtelTraceContext:        "inject",
+	}
+
+	mainCfgWithOTelTraceContextOnly = MainConfig{
+		MainOtelLoadModule:   true,
```
javorszky, line 5628: this config not only turns on the otel trace context, but also turns on the otel load module.

There's a missing case where the context is present, but the module is not enabled. It should not print the context then.

internal/configs/configmaps.go
```diff
@@ -994,6 +994,23 @@ func parseConfigMapOpenTelemetry(l *slog.Logger, cfgm *v1.ConfigMap, cfgParams *
 		cfgParams.MainOtelTraceInHTTP = otelTraceInHTTP
 	}
 
+	if otelTraceContext, exists := cfgm.Data["otel-trace-context"]; exists {
+		otelTraceContext = strings.TrimSpace(otelTraceContext)
+		switch otelTraceContext {
+		case "extract", "inject", "propagate", "ignore":
```
javorszky, line 1000: elegant solution to go's lack of enums 🙂

internal/configs/configmaps_test.go
```diff
@@ -2023,6 +2024,49 @@ func TestOpenTelemetryConfigurationSuccess(t *testing.T) {
 			expectedTraceInHTTP:         false,
 			msg:                         "no config",
 		},
+		{
+			configMap: &v1.ConfigMap{
+				Data: map[string]string{
+					"otel-trace-context": "propagate",
+				},
+			},
+			expectedLoadModule:       true,
```
javorszky: this should fail the test. Only the endpoint turns on the otel config automatically.

javorszky (changes requested):
Good work! Besides the individual comments, a snapshot test for NGINX OSS is missing.

Let me know if you need help adding that one!

adberger:
@javorszky Thanks for the review, I've tried addressing the things mentioned with the help of Claude Code.  
Let me know what you think.

internal/configs/version1/__snapshots__/template_test.snap
```diff
@@ -7356,6 +7356,420 @@ stream {
...
+        ''      $default_connection_header;
+    }
+    otel_exporter {
+        endpoint ;
+    }
+
+    
+    
+    
+    otel_trace_context propagate;
```
javorszky, line 7730: something is broken here. An empty endpoint should not enable the otel load module.

internal/configs/version1/template_test.go
javorszky: Unresolving.

This is still current. `MainOtelLoadModule` is not something a user can configure directly. The value depends on the endpoint. If the endpoint is empty, then this will be false, if the endpoint is not empty, this will be true.

Having a config where `(the endpoint is empty)` AND `(MainOtelLoadModule is true)` is unreachable.

internal/configs/version1/template_test.go
```diff
@@ -5497,6 +5621,17 @@ var (
...
+		MainOtelTraceContext: "propagate",
+	}
+
+	mainCfgWithOTelTraceContextModuleDisabled = MainConfig{
+		MainOtelLoadModule:   false,
```
javorszky, line 5633: Similarly the zero value of the `MainOtelLoadModule` property is false, this can be omitted from here.

Because the `mainCfgWithOTelTraceContextOnly` is unreachable, and should not be tested against, these two configs collapse into one.

internal/configs/version1/template_test.go
javorszky: See https://github.com/nginx/kubernetes-ingress/pull/10556/changes#diff-386098894c7779bda00021e1c7a4b8a6b4577a80974318f50b63daaee53db702R1024-R1026

javorszky (changes requested):
Thank you for adding the snapshot test for the nginx OSS version.

There are still a few issues with it. If you'd like I can fix up the PR by copying this branch and adding new commits.

You will still be credited for the fix. Let me know if you'd like to keep working on this, or you're happy for us to take over 🙂

adberger:
> Thank you for adding the snapshot test for the nginx OSS version.
> 
> There are still a few issues with it. If you'd like I can fix up the PR by copying this branch and adding new commits.
> 
> You will still be credited for the fix. Let me know if you'd like to keep working on this, or you're happy for us to take over 🙂

I wouldn't mind if you take over from here.
Also I don't need credits, I'm just happy when this is supported :)

javorszky:
> I wouldn't mind if you take over from here.

Will do.

> Also I don't need credits, I'm just happy when this is supported :)

you did a large part of the work, you get credited, unless you specifically don't want to be 🙂 

I'll link the PR once it's up

javorszky:
@adberger PR is up: https://github.com/nginx/kubernetes-ingress/pull/10843

If you're happy with that one, close this PR.

javorszky:
@adberger this has been merged into main, and will form part of the next release 🙂

PR #10571, opened 2026-08-05
fix(templates): omit status_zone for default fallback server

### LINKED ISSUE #10459: [Bug]: wrong number of labels log spam after 5.5.0 upgrade
### Version

edge

Minikube

### Steps to reproduce

1. Deploy NIC with NGINX Plus and Prometheus metrics enabled.
2. Port-forward the Prometheus Exproter IP eg (9113)
3. Check logs after curl

Decision: Do not add the zone when it's not taken by an actual hostless Ingress

`level=WARN source.file=nginx_plus.go source.line=689 msg=wrong number of labels for http zone, empty labels will be used instead zone=_ expected=3 got=0`

javorszky:
### Proposed changes

Closes #10459

When Prometheus metrics are enabled on NGINX Plus (-prometheus-metrics), nginx-prometheus-exporter scrapes /api/10/http/server_zones and attempts to extract 3 custom resource labels (resource_type, resource_name, resource_namespace).

For the default fallback server (default-server.conf), status_zone was set to "_", but no resource labels were defined. This caused nginx-prometheus-exporter to log a warning on every scrape interval: "wrong number of labels for http zone ... zone=_ expected=3 got=0"

This commit updates internal/configs/version1/nginx-plus.ingress.tmpl to omit status_zone _ when rendering default server blocks that are not backed by an actual Ingress resource.

### Checklist

### DIFF
--- a/internal/configs/version1/nginx-plus.ingress.tmpl
+++ b/internal/configs/version1/nginx-plus.ingress.tmpl
@@ -106,13 +106,16 @@ server {
 	{{- end }}
 
 	server_name {{$server.Name}};
+	{{- if $.Ingress.Name}}
 
 	status_zone {{$server.StatusZone}};
-	{{- if $.Ingress.Name}}
 	set $resource_type "ingress";
 	set $resource_name "{{$.Ingress.Name}}";
 	set $resource_namespace "{{$.Ingress.Namespace}}";
 	set $service "-";
+	{{- else if ne $server.StatusZone "_" }}
+
+	status_zone {{$server.StatusZone}};
 	{{- end}}
 
 	{{- with $server.EgressMTLS }}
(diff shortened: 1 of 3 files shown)


internal/configs/version1/nginx-plus.ingress.tmpl
```diff
@@ -106,13 +106,16 @@ server {
...
 	server_name {{$server.Name}};
+	{{- if $.Ingress.Name}}
 
 	status_zone {{$server.StatusZone}};
-	{{- if $.Ingress.Name}}
 	set $resource_type "ingress";
 	set $resource_name "{{$.Ingress.Name}}";
 	set $resource_namespace "{{$.Ingress.Namespace}}";
 	set $service "-";
+	{{- else if ne $server.StatusZone "_" }}
```
haywoodsh, line 116: I dont see a scenario where this ```else if``` block would be reached. If the default server is not claimed by any Ingress, `$server.StatusZone` should also be empty.

javorszky: hm... okay, I'll give this a think, thank you! 🙂

PR #10733, opened 2026-08-27
[main] CI release prep fixes

sean-breen:
### Proposed changes

Fixes for new release process

### Checklist


vepatel (dismissed):

AlexFenlon (dismissed):

.github/scripts/create-release-tarballs.sh
```diff
@@ -40,4 +40,4 @@ for i in ${releases}; do
 done
 
 checksum_file=$(ls "${tarball_dir}"/*_checksums.txt )
-${cosign_binary} sign-blob "${checksum_file}" --output-signature="${checksum_file}.sig" --output-certificate="${checksum_file}.pem" -y
+${cosign_binary} sign-blob "${checksum_file}" --use-signing-config=false --new-bundle-format=false --output-signature="${checksum_file}.sig" --output-certificate="${checksum_file}.pem" -y
```
javorszky, line 43: Opened https://github.com/nginx/kubernetes-ingress/issues/10741 to track figuring out how to re-enable usign signing config and adapt it to new bundle

.github/workflows/version-bump.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       contents: write
       pull-requests: write
       id-token: write
-    if: github.repository == 'nginx/kubernetes-ingress'
+    if: github.repository == 'nginx/kubernetes-ingress-internal'
```
javorszky: this would mean that version bumps on the public repository will not work, is that expected?

javorszky (dismissed):

.github/workflows/version-bump.yml
vepatel: I guess we can remove this change, should be okay in public as its NIC only and can cause sync failures

haywoodsh (dismissed):

.github/workflows/version-bump.yml
sean-breen: With the current gating it's difficult to allow on both repos, so it will be gated to public only

vepatel (approved):

javorszky (approved):

PR #10825, opened 2026-09-09
Fix double-increment of splitClientsIndex for VS route-level splits

### LINKED ISSUE #10826: [Bug]: dynamic weight changes write to the wrong split_clients keyval zone when a VirtualServer has 2+ route-level splits
### Version

edge

Kind

### Steps to reproduce

1. Deploy NIC (Plus) with `-weight-changes-dynamic-reload`.
2. Apply a `VirtualServer` with two route-level 2-way splits, `/a` and `/b`, both 50/50.
3. `kubectl patch` both weights in a single update -`/a` to 70/30 and `/b` to 60/40.

Expected: each path shifts to its own new distribution, no reload.

Actual: `/a` doesn't move, and `/b` gets `/a`'s update. With three changed splits it's worse — the third index lands past the end of the generated zones, so the update is dropped with only a `Failed to add key value pair` warning in the log. No event, no status change, VS still reports Valid.

## Cause

`processVSWeightChangesDynamicReload` (internal/k8s/controller.go:4443) re-derives the `split_clients` index the generator assigned, so it can target the right keyval zone. The generator emits 101 blocks per 2-way split, so the counter must advance by 101 each time. The route-level branch has an increment inside the `if weights changed` block _plus_ the unconditional one below it, so it advances by 202 when a split actually changed.

That's why it's silent until it isn't: the index is correct unless an earlier route-level split changed in the same update, and each changed split adds a surplus 101 to every index after it. Generator ground truth for a 3-split VS is 0, 101, 202; today we produce 0, 202, 404 against only 303 blocks.

The match-level loop in the same function, the route-level loop in the VSR twin, and `getStartingSplitClientsIndex` all use a single unconditional increment, so this branch is the only one of the four that double-counts.

Corruption clears on the next full render, so an operator sees a canary weight change "not take" or the wrong service shift, then sees it behave correctly after any unrelated edit.

pdabelf5:
The route-level branch of `processVSWeightChangesDynamicReload` advanced `splitClientsIndex` by 202 instead of 101 whenever a 2-way split's weights had actually changed: an increment inside the if fired in addition to the unconditional one below it. The generator emits 101 `split_clients` blocks per 2-way split, so the index no longer matched.

Each changed split added a surplus 101 to every index after it, so with two changed route-level splits the second update landed on the third split's keyval zone, and with three the last index ran past the end of the generated zones and was dropped with only a "Failed to add key value pair" warning. The corruption self-heals on the next full render, which makes it intermittent in the field.

Extract the walk as `computeVSWeightUpdates` so it can be tested without a live `Configuration` and `Configurator`, and pin the generator's index sequence from the configs side so the two walks cannot drift silently.

Only affects `-weight-changes-dynamic-reload`.

### Checklist


internal/configs/virtualserver.go
```diff
@@ -22,14 +22,36 @@ import (
...
+	//     blocks are emitted, and therefore how far the running index actually
+	//     moves;
+	//  2. this constant, used only by generateMatchesConfig's main-map walk;
+	//  3. an identical, separate declaration in internal/k8s/controller.go,
+	//     used by computeVSWeightUpdates to re-derive these indices when
+	//     pushing keyval updates without a reload.
+	//
+	// Changing one without the others is not a compile error. The index
+	// sequences are pinned from both sides by
+	// TestGenerateVirtualServerConfigSplitClientsIndexSequence here and
```
vepatel: can we make comments here a bit more compact, no need for test functions

vepatel (dismissed):

internal/configs/virtualserver.go
pdabelf5: Comments have been condenced

AlexFenlon (approved):

vepatel (approved):

PR #10832, opened 2026-09-09
fix(helm): emit allow-empty-ingress-host when custom resources are disabled

### LINKED ISSUE #10831: [Bug]: Helm chart wires `controller.allowEmptyIngressHost` behind `controller.enableCustomResources`, disabling it silently when CRDs are off
### Version

5.5.4

Other

### Version
5.5.4 (also reproduced on current `main`)

Other (platform-independent — this is a Helm templating bug, not runtime-specific)

### Steps to reproduce
1. Install the chart with an internal/CRD-free configuration:
   ```yaml
   controller:
     enableCustomResources: false
     allowEmptyIngressHost: true
   ```
2. Render the manifests: `helm template nginx-ingress charts/nginx-ingress -f values.yaml | grep allow-empty-ingress-host`
3. Observe that the `-allow-empty-ingress-host=true` argument is **absent** from the controller Deployment args, even though `allowEmptyIngressHost: true` was set.
4. Deploy an Ingress resource with no `spec.rules[].host` — it is rejected by admission/validation, because the running controller was never started with `-allow-empty-ingress-host=true`.

### Expected behavior
`controller.allowEmptyIngressHost` should be applied independently of `controller.enableCustomResources`, matching:
- the CLI flag definition, which has no dependency on custom resources (`cmd/nginx-ingress/flags.go:254`, consumed in `cmd/nginx-ingress/main.go:345` and `internal/k8s/controller.go:308`/`512`), and
- the `values.yaml` documentation for the field, which does not mention any dependency on custom resources:
  ```yaml
  ## Allows Ingress resources to omit the host field.
  ## spec.tls is not supported on Ingress resources without a host; use controller.defaultTLS.secret instead.
  ## nginx.org/listen-ports and nginx.org/listen-ports-ssl are not supported; use controller.defaultHTTPListenerPort and controller.defaultHTTPSListenerPort instead.
  allowEmptyIngressHost: false
  ```

### Actual behavior
In `charts/nginx-ingress/templates/_helpers.tpl`, the line building the controller args nests `-allow-empty-ingress-host` inside the `{{- if .Values.controller.enableCustomResources }}` block, alongside genuinely CR-only flags (`-enable-tls-passthrough`, `-enable-cert-manager`, `-enable-oidc`, `-enable-external-dns`, `-global-configuration`, etc.):

```gotemplate
{{- if .Values.controller.enableCustomResources }}
- -enable-tls-passthrough={{ .Values.controller.enableTLSPassthrough }}
{{- if .Values.controller.enableTLSPassthrough }}
- -tls-passthrough-port={{ .Values.controller.tlsPassthroughPort }}
{{- end }}
- -enable-cert-manager={{ .Values.controller.enableCertManager }}
- -enable-oidc={{ .Values.controller.enableOIDC }}
- -enable-external-dns={{ .Values.controller.enableExternalDNS }}
- -default-http-listener-port={{ .Values.controller.defaultHTTPListenerPort}}
- -default-https-listener-port={{ .Values.controller.defaultHTTPSListenerPort}}
- -allow-empty-ingress-host={{ .Values.controller.allowEmptyIngressHost }}   # <-- wrongly gated here
{{- if and .Values.controller.globalConfiguration.create (not .Values.controller.globalConfiguration.customName) }}
- -global-configuration=$(POD_NAMESPACE)/{{ include "nginx-ingress.controller.fullname" . }}
{{- else if .Values.controller.globalConfiguration.customName }}
- -global-configuration={{ .Values.controller.globalConfiguration.customName }}
{{- end }}
{{- end }}
```

As a result, anyone running the controller purely as a classic Ingress controller (no VirtualServer/VirtualServerRoute/Policy CRDs, `enableCustomResources: false` — a supported, documented mode) cannot enable hostless-Ingress support at all, since the flag is silently dropped from the container args regardless of its own value.

### Suggested fix
Move `- -allow-empty-ingress-host={{ .Values.controller.allowEmptyIngressHost }}` (and, arguably, `-default-http-listener-port` / `-default-https-listener-port`, which are also documented as independent of custom resources per their own `values.yaml` comments) out of the `enableCustomResources` conditional block, e.g. next to the other unconditional flags:

```gotemplate
- -enable-snippets={{ .Values.controller.enableSnippets }}
- -disable-ipv6={{ .Values.controller.disableIPV6 }}
- -allow-empty-ingress-host={{ .Values.controller.allowEmptyIngressHost }}
{{- if .Values.controller.directiveAutoAdjust }}
- -enable-directive-autoadjust={{ .Values.controller.directiveAutoAdjust }}
{{- end }}
{{- if .Values.controller.enableCustomResources }}
...
{{- end }}
```

Happy to open a PR with this fix (and accompanying `helm unittest` coverage) if that's welcome — let me know if you'd like it scoped differently (e.g. also relocating the listener-port flags).

cestercian:
### Proposed changes

Helm chart `_helpers.tpl` nested `-allow-empty-ingress-host` inside `controller.enableCustomResources`. When CRDs are off (`enableCustomResources: false`) and a user sets `allowEmptyIngressHost: true`, the Deployment arg was silently omitted.

That CLI flag applies to Ingress host validation, not VirtualServer CRDs:

- `values.yaml` documents `allowEmptyIngressHost` with no CR requirement (unlike TLS passthrough, cert-manager, OIDC, external-dns, and GlobalConfiguration).
- The controller flag in `cmd/nginx-ingress/flags.go` has the same independent semantics.

This PR moves **only** `-allow-empty-ingress-host` out of the CR block so it is always emitted. Default HTTP/HTTPS listener-port flags stay gated on custom resources here and are split into a follow-up PR.

The existing CR-enabled `allowEmptyIngressHost` helm unit case is unchanged; a new regression covers `enableCustomResources: false` + `allowEmptyIngressHost: true`.

Fixes https://github.com/nginx/kubernetes-ingress/issues/10831

### Checklist

### DIFF
--- a/charts/nginx-ingress/templates/_helpers.tpl
+++ b/charts/nginx-ingress/templates/_helpers.tpl
@@ -383,15 +383,15 @@ Build the args for the service binary.
 - -enable-cert-manager={{ .Values.controller.enableCertManager }}
 - -enable-oidc={{ .Values.controller.enableOIDC }}
 - -enable-external-dns={{ .Values.controller.enableExternalDNS }}
-- -default-http-listener-port={{ .Values.controller.defaultHTTPListenerPort}}
-- -default-https-listener-port={{ .Values.controller.defaultHTTPSListenerPort}}
-- -allow-empty-ingress-host={{ .Values.controller.allowEmptyIngressHost }}
 {{- if and .Values.controller.globalConfiguration.create (not .Values.controller.globalConfiguration.customName) }}
 - -global-configuration=$(POD_NAMESPACE)/{{ include "nginx-ingress.controller.fullname" . }}
 {{- else if .Values.controller.globalConfiguration.customName }}
 - -global-configuration={{ .Values.controller.globalConfiguration.customName }}
 {{- end }}
 {{- end }}
+- -default-http-listener-port={{ .Values.controller.defaultHTTPListenerPort}}
+- -default-https-listener-port={{ .Values.controller.defaultHTTPSListenerPort}}
+- -allow-empty-ingress-host={{ .Values.controller.allowEmptyIngressHost }}
 - -ready-status={{ .Values.controller.readyStatus.enable }}
 - -ready-status-port={{ .Values.controller.readyStatus.port }}
 - -enable-latency-metrics={{ .Values.controller.enableLatencyMetrics }}
--- a/charts/tests/__snapshots__/helmunit_test.snap
+++ b/charts/tests/__snapshots__/helmunit_test.snap
@@ -448,6 +448,428 @@ metadata:
     app.kubernetes.io/managed-by: Helm
 ---
 
+[TestHelmNICTemplate/allowEmptyIngressHostWithoutCRs - 1]
+/-/-/-/
+# Source: nginx-ingress/templates/controller-serviceaccount.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: allow-empty-ingress-host-no-crs-nginx-ingress
+  namespace: default
+  labels:
+    helm.sh/chart: nginx-ingress-2.8.0
+    app.kubernetes.io/name: nginx-ingress
+    app.kubernetes.io/instance: allow-empty-ingress-host-no-crs
+    app.kubernetes.io/version: "5.7.0"
+    app.kubernetes.io/managed-by: Helm
+/-/-/-/
+# Source: nginx-ingress/templates/controller-configmap.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: allow-empty-ingress-host-no-crs-nginx-ingress
+  namespace: default
+  labels:
+    helm.sh/chart: nginx-ingress-2.8.0
+    app.kubernetes.io/name: nginx-ingress
+    app.kubernetes.io/instance: allow-empty-ingress-host-no-crs
+    app.kubernetes.io/version: "5.7.0"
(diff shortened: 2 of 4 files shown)


cestercian:

cestercian:
recheck

haywoodsh:
This PR appears to also modify the `defaultHTTPListenerPort` and `defaultHTTPSListenerPort`. Could you edit the issue and PR title (I saw it in the description) to reflect the actual code changes? Additional Helm snap tests for these two arguments would be great.

Alternatively, you could split the listener port changes into another PR, leaving this one to contain only the changes described, which would make it easier to review. I'm happy to approve this one first and then review the other changes separately.

cestercian:
@haywoodsh Split as suggested.

- This PR now only moves `-allow-empty-ingress-host` out of the `enableCustomResources` block (listener ports stay gated on CRs here). Snapshots refreshed for the no-crs case.
- Follow-up for the default HTTP/HTTPS listener ports (with helm snap coverage): https://github.com/nginx/kubernetes-ingress/pull/10874

cestercian:
@haywoodsh Could you run `/approve-pipeline-run` again when you have a moment?

The last pipeline failed two smoke jobs that look unrelated to this Helm change (Helm + unit were green):

- `ingresses` alpine-plus: controller stayed `Pending` after a missing GCR image manifest
- `ingresses` debian: custom error-pages expected body, got nginx `502 Bad Gateway`

Happy to dig further if you think either is real.

PR #10897, opened 2026-09-17
Fix/cors origin port validation redo 9861

### LINKED ISSUE #9860: [Bug]: CORS policy allows out-of-range origin ports

javorszky:
### Proposed changes

Fixes https://github.com/nginx/kubernetes-ingress/issues/9860

Internal rework of #9861 to address feedback.

### Checklist

### DIFF
--- a/pkg/apis/configuration/validation/policy.go
+++ b/pkg/apis/configuration/validation/policy.go
@@ -1085,6 +1085,10 @@ func validateOriginFormat(origin string) error {
 		return validateWildcardOriginHost(origin, host)
 	}
 
+	if err := validateOriginPort(parsedOrigin.Port(), origin); err != nil {
+		return err
+	}
+
 	return validateExactOriginHost(host)
 }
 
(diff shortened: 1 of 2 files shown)


promptless-for-oss:

- [Document CORS allowOrigin port validation in the NIC Policy reference](https://app.gopromptless.ai/suggestions/f9ce3c33-9386-4daf-8eeb-c9663d7cdca4) adds a note to the Policy CORS `allowOrigin` field reference stating that an origin's port must be a valid TCP port (1–65535), or the Policy is rejected at admission.

pkg/apis/configuration/validation/policy.go
```diff
@@ -1734,6 +1737,24 @@ func validateWildcardOriginHost(origin, host string) error {
 	return nil
 }
 
+func validateOriginPort(port, origin string) error {
+	if port == "" {
+		return nil
+	}
```
sean-breen, line 1743: Are we allowing an empty port field here?

sean-breen: Nvm, the guard clauses at the call site look to have that covered already for this case.

sean-breen (dismissed):

haywoodsh:
Sholud we also try adding CRD validation in types.go so that wrong configuration can be rejected early?

javorszky:
> Sholud we also try adding CRD validation in types.go so that wrong configuration can be rejected early?

Added in 6973abf, reverted in 772c764

The CEL validation needed was too expensive, and the kubernetes engine kicked it back.

Cheaper CEL validation options exist, but they are syntactic, and aren't worth doing given we already handle validation on the Go side anyways.

As an example, we could validate that the ports are numbers at most 5 long, but that wouldn't catch the `99999` case, and it would have to duplicate the origin syntax in CEL that we already have Go validation for.

haywoodsh (approved):

vepatel (approved):

PR #10905, opened 2026-09-18
Enable http2 for upstream proxying

### LINKED ISSUE #9656: Support configurable proxy_http_version directive
Allow users to configure the HTTP version used for upstream connections via Service appProtocol, Ingress annotation, or VirtualServer field. Ingress/VirtualServer configuration takes precedence over Service appProtocol.

Acceptance Criteria
- Service objects with `appProtocol: kubernetes.io/h2c` automatically infer HTTP/2 for upstream connections
- Ingress resources support a new `nginx.org/proxy-http-version` annotation with valid values: "1.0", "1.1", "2"
- VirtualServer resources support a new [Upstream](https://docs.nginx.com/nginx-ingress-controller/configuration/virtualserver-and-virtualserverroute-resources/#upstream) `proxy-http-version` field with valid values: "1.0", "1.1", "2"
- Precedence: Ingress/VirtualServer annotation/field › Service appProtocol › default (HTТP/1.1)
- When configured, the generated NGINX config includes proxy_http_version in location blocks
- Invalid values are rejected with clear validation errors
- Documentation includes YAML examples for Service appProtocol, Ingress, and VirtualServer configuration

Dev notes:
- Note that today `proxy_http_version 1.1;` is hardcoded into the templates.
- NGINX now defaults to `proxy_http_version 1.1;` so we can omit the directive unless it is explicitly configured by the user

NGF Reference: https://github.com/nginx/nginx-gateway-fabric/blob/main/internal/controller/state/dataplane/types.go#L713-L721

sean-breen:
### Proposed changes

Add support for configuring the HTTP protocol version (proxy_http_version) used when proxying requests to upstream servers.

Summary of changes:
- VirtualServer & VirtualServerRoute: Added proxy-http-version upstream field supporting "1.0", "1.1", and "2".
Ingress resources: Added nginx.org/proxy-http-version annotation (supported on standard and mergeable minion Ingresses).
- Automatic inference via appProtocol: If proxy-http-version is not explicitly set, NIC infers HTTP/2 ("2") when the upstream Kubernetes Service port defines appProtocol: kubernetes.io/h2c.
Precedence: Explicit CRD field / Ingress annotation > Service appProtocol > unset (NGINX default).
HTTP/2 hop-by-hop header handling: Omit Connection and Upgrade headers when proxying over HTTP/2 per RFC 9113.

Validation & warnings:
- Validates accepted values ("1.0", "1.1", "2").
- Emits warnings if configured on grpc upstreams or paired with WebSockets on HTTP/2.
- Triggers reload when a Service port's appProtocol changes.

Docs & examples: 
- Updated CRD docs and added examples for both Ingress and VirtualServer under examples/.

Tests: Added unit tests, snapshot updates, and integration tests.

### Checklist


promptless-for-oss:

- [Document NIC upstream proxy-http-version (VirtualServer/VirtualServerRoute field and nginx.org/proxy-http-version Ingress annotation)](https://app.gopromptless.ai/suggestions/e84fd0ce-0ead-40df-b162-d1a11b3b9b38) adds the new `proxy-http-version` upstream option — the VirtualServer/VirtualServerRoute field and the `nginx.org/proxy-http-version` Ingress annotation — to the two NIC reference pages that cover it, including accepted values, `appProtocol: kubernetes.io/h2c` HTTP/2 inference and precedence, and the gRPC behavior.

AlexFenlon (dismissed):

pkg/apis/configuration/v1/types.go
```diff
@@ -136,6 +136,16 @@ type Upstream struct {
...
 	Keepalive *int `json:"keepalive"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum="1.0";"1.1";"2"
+	// Sets the HTTP protocol version used for connections to the upstream servers.
+	// Supported values are "1.0", "1.1" and "2". If unset, the appProtocol of the backing
+	// Service port is used ("kubernetes.io/h2c" implies "2"); otherwise the directive is
+	// omitted and NGINX applies its own default, which is 1.1 as of NGINX 1.29.7.
+	// Note: this field is ignored for upstreams with type "grpc", where grpc_pass always
+	// uses HTTP/2, and it cannot be used with WebSocket, which requires HTTP/1.1.
+	// The value "2" requires NGINX 1.29.4 or later.
```
haywoodsh: I don't think we need to mention the NGINX versions. We can safely assume that NIC edge and future releases that include this PR won't use older NGINX versions. Mentioning the NGINX versions could cause confusion and mislead users into checking the NGINX version unnecessarily.

internal/k8s/validation.go
```diff
@@ -808,6 +812,14 @@ func validateProxyNextUpstreamAnnotation(context *annotationValidationContext) f
 	return allErrs
 }
 
+// validateProxyHTTPVersionAnnotation validates the nginx.org/proxy-http-version annotation.
+// It shares the accepted value set with the VirtualServer upstream proxy-http-version field so
+// that the two surfaces cannot drift apart. Unlike the CRD field, an empty value is rejected
+// here: an annotation that is present but blank is a mistake, not "unset".
```
haywoodsh: >Unlike the CRD field, an empty value is rejected here: an annotation that is present but blank is a mistake, not "unset".

This does not seem correct to me. The empty string is actually rejected by `validateRequiredAnnotation` which runs before this validator.

examples/custom-resources/proxy-http-version/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+```console
+curl --resolve cafe.example.com:$IC_HTTP_PORT:$IC_IP http://cafe.example.com:$IC_HTTP_PORT/coffee
+curl --resolve cafe.example.com:$IC_HTTP_PORT:$IC_IP http://cafe.example.com:$IC_HTTP_PORT/tea
+```
+
+> [!NOTE]
+> `nginxdemos/nginx-hello` does not serve cleartext HTTP/2. The `/tea` request is expected to
+> fail against it; replace `tea` with an h2c-capable backend to exercise the HTTP/2 upstream
+> path end to end.
```
haywoodsh: Since this is a "Test", I think we should use a working upstream that allows a user following this guide to run curl commands and see a real, working 200 OK response showing the correct HTTP protocol, probably by using an `nginx:alpine` upstream to set NGINX conf with ConfigMap, e.g. https://github.com/nginx/kubernetes-ingress/blob/main/examples/ingress-resources/proxy-redirect/redirect-backend.yaml#L6-L21

same as the ingress example.

haywoodsh (dismissed):

AlexFenlon (approved):

pkg/apis/configuration/v1/types.go
sean-breen: Removed the references to nginx version

examples/custom-resources/proxy-http-version/README.md
sean-breen: Updated the example as suggested

internal/k8s/validation.go
sean-breen: Updated the comment

haywoodsh (approved):

PR #10913, opened 2026-09-19
Warn when routeSelector matches no VirtualServerRoutes

### LINKED ISSUE #10899: [Bug]: A VirtualServer attached via `routeSelector` silently loses routes and stays `Valid`
### Version

edge

Kind

## Summary

When a VirtualServerRoute leaves the Ingress Controller's internal store, every VirtualServer that attached it loses
the corresponding routes. A VirtualServer that attached it **by name** (`route: default/coffee`) reports a warning and
goes to `State: Warning`. A VirtualServer that attached it **by `routeSelector`** reports **nothing** and stays
`State: Valid`, while serving no locations for the affected paths.

In the worst cases (VirtualServerRoute deleted, or its `ingressClassName` changed to another controller) there is no
signal anywhere: not on the VirtualServer, and not on the VirtualServerRoute either, because NIC writes no status for
an object that is gone or not owned by it.

## Reproduction

Reproduces on `main`. No backends needed — the subroutes use `action.return`.

<details>
<summary><code>vs.yaml</code> — one VirtualServer, one route attached by name, one by selector</summary>

```yaml
apiVersion: k8s.nginx.org/v1
kind: VirtualServer
metadata:
  name: cafe
spec:
  host: cafe.example.com
  routes:
  # attached BY NAME -> route loss is reported
  - path: /coffee
    route: default/coffee-named
  # attached BY SELECTOR -> route loss is silent
  - path: /tea
    routeSelector:
      matchLabels:
        route-group: cafe
```

</details>

<details>
<summary><code>vsr-named.yaml</code> and <code>vsr-selected.yaml</code></summary>

```yaml
apiVersion: k8s.nginx.org/v1
kind: VirtualServerRoute
metadata:
  name: coffee-named
spec:
  host: cafe.example.com
  subroutes:
  - path: /coffee
    action:
      return:
        body: "served by coffee-named"
---
apiVersion: k8s.nginx.org/v1
kind: VirtualServerRoute
metadata:
  name: tea-selected
  labels:
    route-group: cafe
spec:
  host: cafe.example.com
  subroutes:
  - path: /tea
    action:
      return:
        body: "served by tea-selected"
```

</details>

Using one VirtualServer for both attachment methods keeps every other variable fixed — same host, same reconcile, same
rebuild. Only the attachment method differs.

```console
kubectl apply -f vs.yaml -f vsr-named.yaml -f vsr-selected.yaml
# both paths serve, VirtualServer is Valid

# 1. Delete the SELECTOR-attached VirtualServerRoute
kubectl delete virtualserverroute tea-selected
kubectl get virtualserver cafe
```

```text
NAME   STATE   HOST               IP             PORTS      AGE
cafe   Valid   cafe.example.com   XX.YY.ZZ.III   [80,443]   1m
```

`/tea` now has no location and returns `404`, but the VirtualServer is `Valid` with no warning and no event.
`kubectl describe virtualserver cafe` shows nothing about the lost route.

```console
# 2. Now delete the NAME-attached VirtualServerRoute
kubectl delete virtualserverroute coffee-named
kubectl get virtualserver cafe
```

```text
NAME   STATE     HOST               IP             PORTS      AGE
cafe   Warning   cafe.example.com   XX.YY.ZZ.III   [80,443]   2m
```

```text
Warning  AddedOrUpdatedWithWarning  ...  Configuration for default/cafe was added or updated with warning(s):
                                         VirtualServerRoute default/coffee-named doesn't exist or invalid
```

## Expected

Both attachment methods express the same intent ("serve this path from this VirtualServerRoute"), so losing the route
should be equally visible. Step 1 should also produce a warning and `State: Warning`.

## Actual

Only the named form is visible. A `routeSelector` route that resolves to zero VirtualServerRoutes is
indistinguishable, to the reporting code, from a selector that was never expected to match.

## Root cause

Two sibling functions resolve routes, and only one reports a failure to resolve.

`validateVSRs` (by name) warns when the key is not in the store — `internal/k8s/configuration.go:1967`:

```go
vsr, exists := c.virtualServerRoutes[vsrKey]
if !exists {
    warning := fmt.Sprintf("VirtualServerRoute %s doesn't exist or invalid", vsrKey)
    warnings = append(warnings, warning)
    return vsrs, warnings
}
```

`validateVSRSelectors` (by selector) has no equivalent — `internal/k8s/configuration.go:1983`. It iterates the store,
collects matches, and returns an empty slice with no warning when nothing matches. VirtualServer state follows directly
from the warning list (`internal/k8s/controller.go:2127`), so no warnings means `StateValid`.

Note the per-VirtualServer validation failure *inside* that loop **is** reported; only the "matched nothing" outcome is
silent.

## Suggested fix

Give `validateVSRSelectors` the warning its named counterpart already has.

**Important:** it must fire only when the selector matched *no VirtualServerRoutes at all*, not merely when the
validated result set is empty. A naive `if len(matched) == 0` is wrong — when a selector matches a VirtualServerRoute
that is then rejected by `ValidateVirtualServerRouteForVirtualServer`, `matched` is empty but an accurate per-VS
warning was already emitted, so the VirtualServer would get a spurious second warning. Count label matches separately:

```go
var matched []matchedVSR
labelMatches := 0

for vsrKey, vsr := range c.virtualServerRoutes {
    if sel.Matches(labels.Set(vsr.Labels)) {
        labelMatches++
        // ... existing per-VS validation, unchanged
    }
}

if labelMatches == 0 {
    warnings = append(warnings, fmt.Sprintf(
        "VirtualServerRoute routeSelector %s matched no VirtualServerRoutes", selectorStr))
}
```

## Things to decide before fixing

1. **Is zero-match a warning or benign?** This is the central decision. The fix treats a route path that resolves to
   nothing as a warning regardless of cause, which also makes *removing a label* from a VirtualServerRoute warn
   (arguably a deliberate detach, and it already signals on the VirtualServerRoute side).
2. **Startup/reconcile ordering.** A VirtualServer processed before the VirtualServerRoutes it selects will warn
   transiently until they appear. This is consistent with the named path, which already behaves that way, and the
   warning clears on the next rebuild — but it is a real increase in warning-event volume during startup and rollout.
3. **Release scoping.** `routeSelector` already ships, so this changes the reported state of existing users'
   VirtualServers: some that report `Valid` today will begin reporting `Warning`. It also cannot be narrowed to a
   subset of users — when a selector matches zero objects there is no VirtualServerRoute left to inspect, so the
   warning has to be emitted for every `routeSelector` route that resolves to nothing. Recommend landing it as its own
   change with a release note.

## Tests that codify today's behaviour and will need updating

| Test | Location |
| --- | --- |
| `TestValidateVSRSelectors/VSR does not exist` | `configuration_test.go:6038` |
| `TestAddVirtualServerWithVirtualServerRoutesVSR` | `configuration_test.go:5764`, `:5788` |
| `TestAddVirtualServerWithExistingVirtualServerRoute` | `configuration_test.go:1406` |
| `TestMatchVSwithVSRusingSelector` | `configuration_test.go:5638` |

Full analysis: [`docs/developer/issue-routeselector-silent-route-loss.md`](https://github.com/nginx/kubernetes-ingress/blob/fix/route-selector-error-propogation/docs/developer/issue-routeselector-silent-route-loss.md)

cestercian:
## Summary

A VirtualServer that attaches routes via \`routeSelector\` stayed \`Valid\` with no event when the selected VirtualServerRoutes left the store (deleted, \`ingressClassName\` changed, or labels removed). Named \`route:\` attachments already reported a warning and set \`State: Warning\`.

\`validateVSRSelectors\` now warns when a selector matches no VirtualServerRoutes. Label matches are counted separately from the validated result set so a selector that matches an invalid VSR keeps its existing per-VSR warning and does not get a spurious \"matched nothing\" warning.

Fixes #10899

## Test plan

- [x] \`go test -tags=aws,helmunit ./internal/k8s\` — pass (includes \`TestValidateVSRSelectors\`, \`TestLostVirtualServerRouteWarnsForNamedAndSelectorAttachment\`)
- [x] \`go test -tags=aws -shuffle=on ./...\` — pass (unit tests except Helm)


promptless-for-oss:

- [Document VirtualServer routeSelector no-match warning (NIC)](https://app.gopromptless.ai/suggestions/b5f95958-6d4e-4f37-9c8a-bda3d885b2ec) documents that a VirtualServer whose `routeSelector` matches no VirtualServerRoutes now reports `State: Warning` and emits a warning event instead of silently staying `Valid`. [Docs PR](https://github.com/nginx/documentation/pull/2301)

